PR #1, opened 2017-12-31
Minor changes in README

Sajjon:
It is always difficult to decide format on names/terms, but I felt that we might want to emphasise names such as _Durian_, _Linux_ etc...

**Not sure if we want to do the same with the name _Zilliqa_?**

 Since the command
`sudo apt-get install libboost-all-dev libssl-dev libleveldb-dev libjsoncpp-dev cmake`
was a bit longer I thought it would be clearer putting it in ``` 

**I added link to the Telegram group as well, but maybe that is unwanted?**

Please feel free to suggest improvements to this PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Zilliqa (codename Durian)
+# Zilliqa (codename _Durian_)
 ## Overview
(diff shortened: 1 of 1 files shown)


bb111189:
* For now, we will put slack only as we are considering opening a dedicated dev channel on Slack. Telegram will be catered more towards general chat. 
* I like ``` for the command. Thank for the suggestion. 
* I think Zilliqa name can remain the same. 

Approved the PR. Thank you

Sajjon:
@bb111189 Sorry! I just added telegram before you merged!

PR #14, opened 2018-01-06
Add initial .gitattributes

### LINKED ISSUE #13: Add initial .gitattributes
I'm working on adding an initial **.gitattributes** file. It's generally straightforward¹, but I have a few questions:

1. Was the file [src/depends/Makefile](https://github.com/Zilliqa/Zilliqa/blob/21cb318b0923b7714088729e084bd48a5033ffa6/src/depends/Makefile) intended to be removed as part of 2c459abe314b5f40658121a72c07741baeefee45? I see that the file is excluded per **.gitignore** starting in bb9e4e6571f511bcdf186892e6f0906e465f5091, but I wanted to verify before submitting a pull request to remove all these files.
2. While scripts should have permissions 100755 in Git, I see that other files like [CMakeLists.txt](https://github.com/Zilliqa/Zilliqa/blob/21cb318b0923b7714088729e084bd48a5033ffa6/CMakeLists.txt) and [tests/Data/Test_Block.cpp](https://github.com/Zilliqa/Zilliqa/blob/21cb318b0923b7714088729e084bd48a5033ffa6/tests/Data/Test_Block.cpp) are also added like this. Is it a problem to submit a pull request that changes these files from 100755&rarr;100644?
  

¹ I always make sure **.gitattributes** and **.gitignore** are consistent during this change, since it can otherwise lead to weird merge conflicts down the road. It's much easier to get it sorted out up front, hence the questions before sending the PR. :smile:

sharwell:
Closes #13 

This pull request limits changes to only the following:

1. Add the minimal **.gitattributes** file that will work with the files already in source control
2. Normalize line endings for the two source files directly impacted by this change

:memo: Please use the Merge strategy for merging this pull request (as opposed to the Rebase and/or Squash strategy).


sharwell:
@sidutta I just verified that this pull request is still up-to-date. Keep in mind that delays on this increase the chances that one or more additional files in the repository will need to be normalized for consistent use across different platforms.

sidutta:
Hi @sharwell, @bb111189 manages github-related administrative  issues. Feel free to tag me on core code related reviews.

PR #16, opened 2018-01-06
Remove execute permissions bit from source files

sharwell:
⚠️ While this change seems logical to me, I am not certain that it is desired. Please see #13 for my question(s) on the subject. The impacted file extensions are:

* cpp
* h
* ~~py~~ (edit: no longer impacted)
* txt

:memo: Please use the Merge strategy for merging this pull request (as opposed to the Rebase and/or Squash strategy).


sharwell:
> Can't python scripts remain 755?

It was not clear to me what the preference here was, which is a substantial reason why I listed the file extensions explicitly. Looking at the source, it does appear that this was intended to be a script, and not just a source file in the context of a larger program. I'll make the change and update this pull request.

sharwell:
@sidutta Restored the Python script

Issue #7, opened 2018-01-03
Support for Mac OS X


sidutta:
Compilation(warnings and enhancements)
=
- syscall is deprecated
- pthread unused argument during compilation
- remove need of -l<LIBRARY_NAME> in root CMakeLists.txt and replace them by cmake FindModules and library variables in submodule CMakeLists.txt
- remove need of hardcoded paths like -L/usr/local/Cellar/openssl/1.0.2/lib from CMakeLists.txt if possible
- replace the sysctl commands used to enable core dumps
- fuser unknown option k from the test script

Runtime
=
- thread id incorrectly shown as -1 always in the logs

Sajjon:
Did anyone get it building on mac? (high sierra)

I installed all dependencies except `build-essential`, using `brew`. 

However, when I try runnning `./build.sh` I get `fatal error 'boost/algorithm/hex.hpp'` file not found.

This is using Boost 1.66 (installed via brew).

Sajjon:
@sharwell @sidutta did you get it to work on mac? Did you get the same errors I got?

ckyang:
Dear Sajjon,

Here is a temporary solution for "boost/algorithm/hex.hpp file not found."

Modify CMakeCache.txt with
CMAKE_CXX_FLAGS:STRING=-I/usr/local/include

Please have a try.

Sajjon:
@ckyang Thanks for your suggestion and sorry for my late response, I tried your solution but I got new errors.

<details><summary>I got this without your suggestion</summary>
<p>

```
-- We are on a Darwin system
-- Boost version: 1.66.0
CMake Warning at /usr/local/Cellar/cmake/3.10.1/share/cmake/Modules/FindBoost.cmake:801 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.10.1/share/cmake/Modules/FindBoost.cmake:907 (_Boost_COMPONENT_DEPENDENCIES)
  /usr/local/Cellar/cmake/3.10.1/share/cmake/Modules/FindBoost.cmake:1542 (_Boost_MISSING_DEPENDENCIES)
  tests/Crypto/CMakeLists.txt:28 (find_package)


CMake Warning at /usr/local/Cellar/cmake/3.10.1/share/cmake/Modules/FindBoost.cmake:801 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
[...]

PR #22, opened 2018-01-08
[wip] build : flags for mac

bguiz:
## openssl

- assumes installation via homebrew, and avoids using
  a version specific path (see output below)
- related: https://github.com/Zilliqa/Zilliqa/issues/7

```
$ brew info openssl
...
For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/openssl/lib
    CPPFLAGS: -I/usr/local/opt/openssl/include
```


sidutta:
@bguiz There is branch called macosx-support. Can you please use that branch instead?

sidutta:
I am closing this PR. The plan is to have sufficient number of people test the code in macosx-support on macs first(both compilation as well as tests). Then we will check whether none of changes break anything on ubuntu. If there are no issues, we could merge with master.

Issue #19, opened 2018-01-08
Support for building on Windows 


sharwell:
I'm in the process of updating the code to build on Windows. The work is not yet complete, but I wanted to post an issue early to communicate that it's at least being investigated.

digiront:
Hello, are you still working on this?

sharwell:
@digiront It was not clear how committed the core team was to preserving this functionality in a working state if I got it there. I'm worried the proposed change would go out of date and require a large amount of work to update later (which has already happened to me once).

whalehub:
@sharwell The main net will be launching later this month. Are you going to resume work on this issue at that point? It would be really awesome because most people don't run Linux on the home computers that have an Nvidia GPU in them, so it restricts those people from being able to mine Zilliqa with their GPU. Of course, it also restricts most people in general from being able to mine Zilliqa with their CPU.

If required as an incentive, I'd be willing to put up a bounty of $500 for you to implement Windows support to Zilliqa.

Skytokenn:
Kindly open this live chat link to talk to the customer service directly on your issue  https://direct.lc.chat/14334663/

PR #23, opened 2018-01-08
Use inet_pton instead of inet_aton for portability

sharwell:
This change appears to be neutral for the Ubuntu builds, but `inet_aton` is not defined on Windows platforms. I saw two straightforward resolutions:

1. Use `inet_pton` instead, which is defined on all platforms
2. Define `inet_aton` within a shared header for Windows platforms

I felt that the first approach is likely to be the simplest in the long run.

:link: Related to #19

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -71,7 +71,7 @@ void Lookup::SetLookupNodes()
         if (v.first == "peer")
         {
             struct in_addr ip_addr;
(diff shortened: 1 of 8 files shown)


ansnunez:
Now integrated in #987.  Thank you for the support @sharwell

bb111189:
Closing this old PR as it was implemented in https://github.com/Zilliqa/Zilliqa/pull/987

PR #25, opened 2018-01-08
Add explicit casts where intended to avoid compiler warnings

sharwell:
:link: Related to #19

### DIFF
--- a/src/depends/Sha3/Sha3.c
+++ b/src/depends/Sha3/Sha3.c
@@ -156,7 +156,7 @@ static void store64(UINT8 *x, UINT64 u)
     unsigned int i;
 
     for(i=0; i<8; ++i) {
-        x[i] = u;
+        x[i] = (UINT8)u;
         u >>= 8;
     }
 }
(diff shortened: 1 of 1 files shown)


src/depends/Sha3/Sha3.c
```diff
@@ -298,7 +298,7 @@ void Keccak(unsigned int rate, unsigned int capacity, const unsigned char *input
 
     /* === Absorb all the input blocks === */
     while(inputByteLen > 0) {
-        blockSize = MIN(inputByteLen, rateInBytes);
+        blockSize = (unsigned)MIN(inputByteLen, rateInBytes);
```
sharwell: :memo: Let me know if the following alternative form is preferred:

```c
blockSize = MIN((unsigned)inputByteLen, rateInBytes);
```

The alternative form has the advantage that it guarantees (in the form of a compiler warning on some platforms) that `rateInBytes` will not be out of range, but it may be less readable for some people.

sharwell: ➡️ Ah, the downside of the alternative form is it could cause an incorrect result if `inputByteLen` overflows.

ansnunez: Not totally sure at the moment, but for our purposes I do not think this method is expected to be used with large values of inputByteLen

sharwell: I agree, but given the unknown I prefer the approach that is known to give a correct answer. If this was a C++ file we could use `numeric_cast`, but the approach in this PR should work well in context.

ansnunez: Agree. In that case, we can stick to your current change.

PR #26, opened 2018-01-08
Use std::chrono instead of clock_gettime for portability

sharwell:
This change simplifies the code using new C++ standard library features, and addresses a future issue where `clock_gettime` is not defined on Windows platforms.
  
:link: Related to #19

### DIFF
--- a/src/libUtils/TimeUtils.h
+++ b/src/libUtils/TimeUtils.h
@@ -18,12 +18,11 @@
 #ifndef __TIMEUTILS_H__
 #define __TIMEUTILS_H__
 
-#include <time.h>
-
+#include <chrono>
 #include <boost/multiprecision/cpp_int.hpp>
 
-struct timespec r_timer_start();
-double r_timer_end(struct timespec start_time);
+std::chrono::system_clock::time_point r_timer_start();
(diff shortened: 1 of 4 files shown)


src/libUtils/TimeUtils.h
```diff
@@ -18,12 +18,11 @@
...
 
-#include <time.h>
-
+#include <chrono>
 #include <boost/multiprecision/cpp_int.hpp>
 
-struct timespec r_timer_start();
-double r_timer_end(struct timespec start_time);
+std::chrono::system_clock::time_point r_timer_start();
+double r_timer_end(std::chrono::system_clock::time_point start_time);
```
sharwell: 💭 It may be desirable to return `std::microseconds` instead of just `double`, but I left that for a future change.

ansnunez:
It seems timespec is also used in other files like PoW1Processing.cpp.  Would you be able to update those files as well, to avoid compile errors?

sharwell:
@ansnunez I'll look into the `STAT_TEST` preprocessor define, thanks 😄

sharwell:
@ansnunez Done. The only required change occurred in **DirectoryService.h**.

PR #28, opened 2018-01-08
Use boost::stacktrace for portability

sharwell:
This change simplifies the code and addresses a future compatibility concern when Windows platforms are supported.

⚠️ This pull request changes the minimum version of Boost to 1.65.1. The Travis build is updated to account for this.
  
:link: Related to #19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ matrix:
         apt:
           packages:
             - g++-5
-            - libboost-all-dev
             - libleveldb-dev
             - libjsoncpp-dev
             - libsnappy-dev
@@ -25,7 +24,6 @@ matrix:
         apt:
           packages:
             - g++-5
-            - libboost-all-dev
             - libleveldb-dev
             - libjsoncpp-dev
             - libsnappy-dev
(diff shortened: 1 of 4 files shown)


bb111189:
There is error in compilation. 
/home/junhao/Desktop/octcoin/other_br/codereview/stacktrace/Zilliqa/tests/Zilliqa/main.cpp:22:32: fatal error: boost/stacktrace.hpp: No such file or directory

I will look more into it in the near future. 

print_trace() is currently not used in the Zilliqa codebase and I was considering to remove it previously. As such, I will put this PR on low priority.

sharwell:
@bb111189 This change may unintentionally depend on #30. Also, it appears stacktrace was added in [Boost 1.65.0](http://www.boost.org/users/history/version_1_65_0.html), so if we want to use it then we'd need to update the minimum boost library requirement to match.

sharwell:
@bb111189 Closing and reopening this PR triggered a build (necessary since it was created prior to the Travis integration). You can see that the build failed, and the Boost version is ~1.54. I'll leave this PR for a bit and follow up with more details/suggestions later.

sharwell:
Looks like after about a dozen tries I managed to get this PR to work (separate from the discussion of whether or not we want it).

bb111189:
PR closed. Function no longer used. Refer to PR #102

PR #29, opened 2018-01-08
Use this_thread::get_id instead of SYS_gettid for portability

sharwell:
This change simplifies the code and addresses a future compatibility concern when Windows platforms are supported.
  
:link: Related to #19

### DIFF
--- a/src/libUtils/Logger.cpp
+++ b/src/libUtils/Logger.cpp
@@ -20,9 +20,7 @@
 #include <cstring>
 #include <iomanip>
 #include <iostream>
-#include <pthread.h>
-#include <sys/syscall.h>
-#include <unistd.h>
+#include <thread>
 
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


ansnunez:
As SYS_gettid and this_thread::get_id don't exactly provide the same info, I'm wondering if this will have some impact on testing or whatever (e.g., maybe for some reason we'll be selecting and terminating threads in Linux, so we'll need the actual OS thread IDs).

How do you feel about putting your changes inside #ifdefs for now so that Linux is unaffected?

sharwell:
@ansnunez The main problem with an approach like that (testing or otherwise) is it leads to an application that is very difficult to maintain across platforms. I would prefer to find a common approach that meets the needs with as few conditional blocks as possible.

I'd like to see if any of the current or actively planned test sequences rely on the previous behavior before making the new behavior conditional. If I end up blocked on this after everything else is done we can put the conditions in place to move forward, but I would view that as acquired technical debt.

ansnunez:
Pulling in another reviewer who might be better aware if this change has any impact on the testing activities. @sidutta

Gnnng:
#129 introduces a conditional fix 93bfe6b to make Linux unaffected like @ansnunez said. It was inspired by https://github.com/PowerShell/PowerShell/pull/2675/files. I'm not quite confident on this fix as it was intended for removing the deprecated syscalls on MacOS.

bb111189:
Refer to 93bfe6b75e05a13d7f61102f47085a155aee3d27
Windows support can be easily added as a elif condition. 

Closing this pr

PR #30, opened 2018-01-08
[WIP] Simplify dependencies

sharwell:
⚠️ See comments in the diff for questions.

:link: Related to #19


CMakeLists.txt
```diff
@@ -31,5 +31,20 @@ if(NOT IS_LOOKUP_NODE)
...
+    SET( LevelDB_LIBRARIES leveldb )
+endif()
+if(WIN32)
+    FIND_LIBRARY( LevelDB_LIBRARY libleveldb )
+    SET( LevelDB_LIBRARIES ${LevelDB_LIBRARY} shlwapi )
+endif()
+
+# Dependencies
+FIND_PACKAGE(Boost 1.33 REQUIRED COMPONENTS filesystem system unit_test_framework)
+INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIRS})
```
sharwell: ❓ Do we still need to specify `-lboost_system -lboost_filesystem` in CMAKE_EXE_LINKER_FLAGS above? It seems like this is all covered in the individual projects.

src/depends/json_spirit/CMakeLists.txt
```diff
@@ -3,8 +3,5 @@ json_spirit_reader.cpp
 json_spirit_value.cpp
 json_spirit_writer.cpp)
 
-FIND_PACKAGE(Boost 1.34 REQUIRED)
```
sharwell: ❓ This one says version 1.34, but others said 1.33. I used 1.33 in the updated CMakeLists.txt file. Should it be changed to something else?

CMakeLists.txt
```diff
@@ -31,5 +31,20 @@ if(NOT IS_LOOKUP_NODE)
...
+
+FIND_PACKAGE(OpenSSL REQUIRED)
```
sharwell: ⚠️ I tested this on Windows, but have not tested it on Ubuntu. Since we don't already have a CI in place for pull requests (e.g. travis-ci), someone should manually verify this.

CMakeLists.txt
sharwell: ➡️ Now that Travis-CI is set up, I was able to remove these lines and see that the build still works without them.

CMakeLists.txt
sharwell: :memo: This appears to be working on Travis CI.

src/depends/json_spirit/CMakeLists.txt
sharwell: :memo: I'm trying to figure out how to more easily install specific Boost versions in Travis CI for build validation.

sharwell:
Marking as a WIP. This PR was ignored and eventually conflicted with recent merges.

PR #52, opened 2018-01-14
Remove the references to unused library libjsoncpp

sharwell:
I was unable to find any indication that this library is used. This pull request removes the linker references in an attempt to simplify the cross-platform build.


bb111189:
It was used in a set of the test cases. We removed that set of test cases as we need to do some re-organizing of the test cases. However, we will be putting that test cases back again. Issue #60 
 
Hence, we will need libjsoncpp. PR closed as libjsoncpp will be necessary.

sharwell:
@bb111189 Thanks, I will update the dependent PRs to remove these commits

Issue #32, opened 2018-01-09
Add state root to final block composed at the last mini-epoch of every major epoch


sidutta:
The DS committee doesn't have the txn bodies when composing final blocks usually. Since txn body sharing is asynchronous, there are no guarantees as to when the entire network can agree on a common account state. The current plan is to make the network synchronous every 2 hrs or so(at the last min epoch of every major epoch) at the cost of extra delays and refetches for missing txn bodies and add the state root to the final block composed then. After this epoch, the sharding structure will reshuffle.

bb111189:
Closed. Implemented at PR #56

PR #56, opened 2018-01-15
added state root to txBlockHeader

sidutta:


build_lookup.sh
```diff
@@ -19,5 +19,9 @@ rm -rf ./txblocks.db
 rm -rf ./test.db
 rm -rf ./txbodies.db
 
+find . -name *.cmake | xargs rm -rf
```
sharwell: 💡 Another option is to not have this in the script, but instead run the following to clean *everything* in the working directory:

```
git clean -dxf
```

💡 It's good to make changes like this in their own commits, since it's not directly related to the rest of this pull request.

src/libData/BlockData/Block/TxBlock.cpp
```diff
@@ -28,7 +28,8 @@ unsigned int TxBlock::Serialize(vector<unsigned char> & dst, unsigned int offset
 
     unsigned int header_size_needed = sizeof(uint8_t) + sizeof(uint32_t) + UINT256_SIZE + UINT256_SIZE + 
                                       BLOCK_HASH_SIZE + UINT256_SIZE + UINT256_SIZE + TRAN_HASH_SIZE + 
-                                      sizeof(uint32_t) + sizeof(uint32_t) + PUB_KEY_SIZE + UINT256_SIZE + BLOCK_HASH_SIZE;
+                                      TRAN_HASH_SIZE + sizeof(uint32_t) + sizeof(uint32_t) + PUB_KEY_SIZE + 
+                                      UINT256_SIZE + BLOCK_HASH_SIZE;
```
sharwell: :bulb: This expression appears several times. Consider extracting a constant.

src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
```diff
@@ -169,6 +174,11 @@ const TxnHash & TxBlockHeader::GetTxRootHash() const
     return m_txRootHash;
 }
 
+const StateHash & TxBlockHeader::GetStateRootHash() const
```
sharwell: :bulb: Consider defining this method where it is declared (improves efficiency with no apparent downside).

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -888,7 +888,7 @@ bool Node::ProcessFinalBlock(const vector<unsigned char> & message, unsigned int
     if(m_state == MICROBLOCK_CONSENSUS)
     {
         unsigned int time_pass = 0;
-        while(m_state != PROCESS_FINALBLOCK)
+        while(m_state != WAITING_FINALBLOCK)
```
sharwell: ❔ Why did this change? It seems unrelated to the rest of this PR.

sharwell:
@sidutta Can you briefly describe the relation between this pull request and #32?

build_lookup.sh
```diff
@@ -19,5 +19,9 @@ rm -rf ./txblocks.db
...
+find . -name CMakeLists.txt | xargs rm -rf
```
sharwell: ❗️ This line breaks the build by deleting the CMake input.

src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
sidutta: Sure.

src/libNode/FinalBlockProcessing.cpp
sidutta: This was a bug that had to be corrected.

sidutta:
@sharwell The commits 32d80a0, e3035c8, b390ebd nearly complete #32. The last mini epoch of every major epoch has been made vacuous(the microblock in that epoch doesn't have any transactions). This should give (hopefully) sufficient time for the DS nodes to receive transaction bodies from all the prior mini epochs. In the commit 32d80a0 I added a new field in the final block header to store the state root. In e3035c8, I added some synchronization barriers so that DS node leader has all transaction bodies before it attempts to create the final block. In b390ebd, I compute the state root and store it in the field created as part of 32d80a0.

build_lookup.sh
sidutta: I corrected it. It should have been CMakeCache. Thanks.

src/libData/BlockData/Block/TxBlock.cpp
sidutta: Yeah, even I was wondering the same when making the change.

src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
sidutta: For such changes though could you rather make generalized github issues. May be as we grow someone from the community can pick them up. We could tag such enhancements something so that they can be found.

PR #55, opened 2018-01-14
Enable unit tests on travis-ci builds

sharwell:
This pull request enables unit testing during Travis CI builds.

PR #58, opened 2018-01-21
[WIP] Enable code coverage

sharwell:
⚠️ Blocked on #55


ckyang:
@sharwell Could you please fix conflict files?
Thanks!

sharwell:
@ckyang Sure I'll look into updating it later today. I wasn't sure if there was interest previously so I was waiting for a comment like that before spending more time on it. 😄

sharwell:
@ckyang It's going to take me a few tries to get this updated. I'm going to keep amending the merge commit until it's the _correct_ merge commit given the changes that happened in two different branches. Once it's working I'll reply with a comment that it's ready for review.

sharwell:
It's late here so I'm pausing for the evening but I'll give a status update.

* Items to fix that I will fix
    * Script execution in the docker container doesn't inherit environment variables from the host, so the codecov.io upload script can't automatically determine parameters like the pull request and commit number. I need to propagate the correct variables when executing **ci_report_coverage.sh**.
    * Failure to in **ci_report_coverage.sh** does not get reported as a build failure, but it should.
* Items to fix that I won't fix in this pull request
    * The build uses optimized binaries, which could negatively impact the code coverage report clarity. The reports should be reviewed later to determine if the development experience would benefit from gathering and reporting coverage data from non-optimized binaries.
    * Coverage is only gathered and reported for the Linux build. The OSX build could be added for improved coverage information.

sharwell:
@ckyang This is ready now. Please be aware of the issues from https://github.com/Zilliqa/Zilliqa/pull/58#issuecomment-391600328. In addition, keep in mind that codecov.io relies on differential coverage data, so in the initial days of adopting code coverage functionality some pull requests will not have complete information (including this one). In Git terms, the condition for it working properly is the [merge-base](https://git-scm.com/docs/git-merge-base) needs to have coverage information available. Over time this will happen automatically.

ckyang:
Let me verify in my site first

ckyang:
@sharwell It works in my site.
Please update README.md dependency part for lcov & curl
Thanks for your effort!

sharwell:
> Some folders/files are missing and should be included, ex: src/libDirectoryService and src/libNode

I'm not 100% clear what this is referring to. If I broke one or more tests (were running before but no longer are with this pull request), then I certainly need to fix that. However, if this is just a situation where one or more source files are not appearing in the coverage reports, I would prefer to leave this as a follow-up item.

> For lookup mode, ci_build.sh should mkdir/cd "build_lookup" instead of "build"

I don't believe I changed anything related to this request in this pull request. If this is not correct, can you point out the item that I need to update in the diff?

> Update README.md dependency part for lcov & curl

**curl** is only needed on Travis (local coverage collection won't get uploaded to codecov), so it shouldn't need to be updated. **lcov** and **gcov** are currently unintentional dependencies. I will try to restrict their requirement to cases where users are specifically attempting to collect code coverage information, and update this pull request accordingly.

sharwell:
:memo: Build failure in 398cabb was due to #308.

sharwell:
@ckyang Please let me know if my replies+updates addressed the outstanding concerns.

ckyang:
@sharwell Nice modification, thanks for your effort!

Regarding to the LCOV report, supposedly it should show every files inside src/ folder, including src/libDirectoryService/ and src/libNode/; however it doesn't.
Is it possible to adjust gcov/lcov setting for showing overall files inside Zilliqa project?
We think it's more reasonable to calculate the coverage % under whole project scope (under src/).
Thanks again for your great help!

sharwell:
@ckyang The files are not intentionally excluded. If you expected the code to be covered by one or more specific tests, the first step is making sure that test is running during the Travis builds. If the files are not covered by current tests, the first step is making sure one or more test classes exists which at least links against the libraries containing the source files. If there are no tests running which reference the library code, then the tools have no way of knowing the files exist for inclusion.

This process can involve investigation and sometimes trial and error, which is why I typically suggest to leave it as a follow-up item after the initial support for code coverage is operational.

ckyang:
@sharwell Fine, understand the situation and it seems the original design of gcov/lcov: only include the files under test.
Could you please raise an issue for the follow-up and let's make this PR on-line first.

sharwell:
Filed #313

ckyang:
@Gnnng This PR will automatically generate .gcda files for lcov report usage after building procedure. Please help to check if it will impact current Travis CI flow, especially the docker deploying process. Thank you very much!

.travis.yml
```diff
@@ -35,6 +35,12 @@ script:
...
 
+# Code coverage is currently only implemented for GCC builds, so OSX is currently excluded from reporting
+after_success:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+      docker exec -e CI="true" -e TRAVIS="$TRAVIS" -e TRAVIS_PULL_REQUEST_SHA="$TRAVIS_PULL_REQUEST_SHA" -e TRAVIS_COMMIT="$TRAVIS_COMMIT" -e TRAVIS_JOB_NUMBER="$TRAVIS_JOB_NUMBER" -e TRAVIS_PULL_REQUEST="$TRAVIS_PULL_REQUEST" -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" -e TRAVIS_REPO_SLUG="$TRAVIS_REPO_SLUG" -e TRAVIS_OS_NAME="$TRAVIS_OS_NAME" -e TRAVIS_TAG="$TRAVIS_TAG" -e TRAVIS_BRANCH="$TRAVIS_BRANCH" -t ci /bin/sh -c "./scripts/ci_report_coverage.sh";
+    fi
+
 jobs:
   include:
   - stage: deploy
```
Gnnng: I suggest adding a line `after_success: # left blank` to disable it just like those for `before_install`, `install` and `cache`. They are not intended for deploy stage so should be disabled.

Gnnng:
For some reason I couldn't check the travis running [status](https://travis-ci.org/Zilliqa/Zilliqa/builds/385495844?utm_source=github_status&utm_medium=notification) for a91af27. So I read the change on `.travis.yaml` and suggested a modification there. Hope it helps.

sharwell:
@Gnnng Fixed and amended to the merge commit (I consider the change to be part of resolving the merge between branches.

README.md
```diff
@@ -1,6 +1,9 @@
 # Zilliqa (codename _Durian_)
+
 [![Build Status](https://travis-ci.org/Zilliqa/Zilliqa.svg?branch=master)](https://travis-ci.org/Zilliqa/Zilliqa)
 
+[![codecov](https://codecov.io/gh/Zilliqa/Zilliqa/branch/master/graph/badge.svg)](https://codecov.io/gh/Zilliqa/Zilliqa)
```
Gnnng: Can we align them into one line?

sharwell:
> Just one last comment about the CMake flag `ENABLE_COVERAGE`: can we simply set it to `ON` and evaluate it as boolean type in `CMakeLists.txt` just like the flag `TESTS`?

Yep, I didn't realize this was a thing. 😄 Fixed in 3548987e9ea3121c1d17abeb3ac1b281c730d75e.

README.md
sharwell: Also fixed in 3548987e9ea3121c1d17abeb3ac1b281c730d75e.

CMakeLists.txt
```diff
@@ -65,6 +65,13 @@ add_compile_options(-Wall)
 add_compile_options(-Werror)
 # add_compile_options(-Wextra) #TODO: enable it
 
+if (ENABLE_COVERAGE AND CMAKE_COMPILER_IS_GNUCXX)
+    include(CodeCoverage)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GCC_COVERAGE_COMPILE_FLAGS}")
```
Gnnng: how about using `add_compile_option` here?

sharwell:
> ... and you perhaps also need to merge the latest master once.

Pull request builds always run on the result of merging the pull request into the target branch (in this case **master**). As long as there are no merge conflicts, the pull request *should* be safe to merge. If the latest pull request build is old on Travis (it won't be in this case due to the changes this morning), you can use Travis to rebuild the PR. I don't have permissions to retry builds for Zilliqa, but here's what the button should look like for y'all:

image

If GitHub is still saying the branch needs to be updated, it almost certainly means the following option is checked. Considering Travis and other CI services for public GitHub projects build the merged result and not just the pull request commit, I typically recommend projects disable this option when using those services to avoid unnecessary overhead of updating every pull request every time something is merged.

⚠️ This change would make it more likely that a breaking change is merged into master. However, of all the build breaks I've seen over time with public projects, this one is exceedingly rare (I can remember only two cases in all my time on GitHub, and both were caught in short order because one of the merge commits failed).

image

In the meantime I'll update the branch so it's not blocking this PR.

CMakeLists.txt
sharwell: Love the idea in theory, but it treated the whole expression as a single command line argument and the build failed.

Gnnng:
I just had some experiments in branch https://github.com/Zilliqa/Zilliqa/commits/fix/pr-58. 

To resolve the warning in `CodeCoverage.cmake`, I think we are good to use `Debug` build for code coverage build  (although I haven't checked what is the gain here).

The use of `--converage` in the compiler and linker flags comes from https://github.com/simsong/tcpflow/blob/master/cmake/coverage.cmake#L19

sharwell:
> To resolve the warning in CodeCoverage.cmake, I think we are good to use Debug build for code coverage build (although I haven't checked what is the gain here).

I'm not sure this is a necessary change at this point. From what I can tell, **CodeCoverage.cmake** sets up the ability to do `-DCMAKE_BUILD_TYPE=Coverage` with optimum configuration, but I wasn't ready to force that change on the CI process yet.

> The use of --converage in the compiler and linker flags comes from ...

Nice! Updated the PR with this (and credited you for the idea).

Gnnng:
Sure, I agree that the build type can be decided later. Now the pr looks good, will get it merged soon.

sharwell:
🎉

PR #65, opened 2018-01-26
Added feedback mechanism for consensus, implemented for microblock consensus missing txns

sidutta:
TODO: restart announcement from leader with customized proposal


sidutta:
replaced by #66, merging with master leading to weird behaviour

PR #78, opened 2018-02-21
Simplify the travis config

Gnnng:
- Remove the redundant lines
- Show two different builds with "BUILD_COMMAND" variable on build result page
- Allow osx failures (will fix soon)


sharwell:
:memo: It's going to be much easier to fix #55 and #58 first rather than trying to deal with those after all platform builds are enabled.

PR #79, opened 2018-02-21
Add libServer and json-rpc API for lookup nodes

sidthekidder:


CMakeLists.txt
```diff
@@ -81,11 +81,11 @@ endif()
 if(IS_LOOKUP_NODE)
     if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
         SET( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GCC_COVERAGE_COMPILE_FLAGS} -pthread -std=c++14 -Wall -DSTAT_TEST -ggdb -DIS_LOOKUP_NODE" )
-        SET( CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${GCC_COVERAGE_LINK_FLAGS} -pthread -ljsoncpp -lboost_system -lboost_filesystem -std=c++14 -Wall -DSTAT_TEST -DIS_LOOKUP_NODE" )
+        SET( CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${GCC_COVERAGE_LINK_FLAGS} -pthread -ljsoncpp -ljsonrpccpp-common -ljsonrpccpp-server -lboost_system -lboost_filesystem -std=c++14 -Wall -DSTAT_TEST -DIS_LOOKUP_NODE" )
```
sharwell: ❔ These are increasingly difficult to maintain. Is there a way to place each of these included libraries on a separate line of this file?

src/libZilliqa/Zilliqa.cpp
```diff
@@ -66,6 +67,19 @@ Zilliqa::Zilliqa(const std::pair<PrivKey, PubKey> & key, const Peer & peer, bool
 #ifndef IS_LOOKUP_NODE
     m_n.StartSynchronization();
 #endif // IS_LOOKUP_NODE
+
+#ifdef IS_LOOKUP_NODE
```
sharwell: 💡 `#else`

src/libZilliqa/Zilliqa.h
```diff
@@ -28,6 +28,7 @@
 #include "libNetwork/PeerStore.h"
 #include "libNetwork/PeerManager.h"
 #include "libNode/Node.h"
+#include "libServer/Server.h"
```
sharwell: :bulb: It appears this can be surrounded by `#ifdef IS_LOOKUP_NODE`

.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: generic
-sudo: false
+sudo: true
```
sharwell: :memo: I think you can leave this set to `false`. See #28 for example.

sidthekidder: Thanks! I'll try that

sidthekidder: Libjson-rpc-cpp required a slight modification to run on ubuntu 14 (https://gist.github.com/stefanofiorentino/594858159c23841ec4eec56184852d51) so I pointed it to my own fork, and am building it from source. Is this sustainable or should we look for some other solution

PR #87, opened 2018-02-23
Fix testing errors on Mac

Gnnng:
The ctest is having errors on osx platform.

```
The following tests FAILED:
	  3 - Test_Account (Failed)
	  4 - Test_AccountStore (Failed)
	  6 - Test_Transaction (Failed)
	  7 - Test_LevelDB (Failed)
	  8 - Test_LookupNodeForDSBlock (Failed)
	  9 - Test_LookupNodeForTxBlock (Failed)
	 11 - Test_DSPersistence (Failed)
	 12 - Test_TxPersistence (Failed)
	 13 - Test_TxBody (Failed)
	 14 - ReadBlock (Failed)
```

found errors both on local Mac machines and travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,3 @@ script:
     else
       $BUILD_COMMAND && ctest;
     fi
-
-# TODO: remove when osx is fully supported
-matrix:
-  include:
-  allow_failures:
-    - os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
       ";
     else
       brew update;
-      brew install pkg-config jsoncpp leveldb;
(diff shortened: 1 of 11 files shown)


Gnnng:
The tests pass now, though the root cause of this has not been identified. The solution actually comes from [here](https://stackoverflow.com/questions/39171467/there-is-no-argument-provided-for-parameter-color-output-with-boost-test-and-cte) and works well. However, the reason why changing it to a dynamic link solves the problem is not clear. The [doc](http://www.boost.org/doc/libs/1_66_0/libs/test/doc/html/boost_test/usage_variants.html) about the choice of these usage variants provides no other information except the use case.

PR #89, opened 2018-02-24
 fix : osx build issues for openssl and boost

bguiz:
- removed hardcoded version in boost path
- removed trailing `/openssl` in linker flags
  - this allows us to remove the workaround in `src/libCrypto`
  - needed because: e.g. `#include <openssl/ec.h>` in
    `src/libCrypto/*.h` would wind up looking in path
    `.../openssl/openssl/ec.h` instead of `.../openssl/ec.h`
- auto-detect openssl include and lib paths

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,32 @@ link_libraries(${SNAPPY_LIBRARIES})
 
 ###################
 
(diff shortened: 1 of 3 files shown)


bguiz:
Sweet!

PR #93, opened 2018-02-26
Support for nodes to connect to multiple lookup nodes

bb111189:
- State sync-ing
- Better resync when not during ds-epoch
- Various concurrency issues fixes
- Lookup node script fix
- Sharding structure fix 
- "Bootstrap DS" information sync to lookup node 
- Accountstore serialization and deserialization
- Logmarker to various function in lookup class


jiayaoqijia:
State sync-ing
Better resync when not during ds-epoch
Various concurrency issues fixes
Lookup node script fix
Sharding structure fix
"Bootstrap DS" information sync to lookup node
Accountstore serialization and deserialization
Logmarker to various function in lookup class

PR #101, opened 2018-03-03
p2pcomm concurrency issue fix 

bb111189:
p2pcomm concurrency issue fix by adding mutex
threadpool number to be obtain from config file


src/libNetwork/P2PComm.cpp
```diff
@@ -259,7 +259,7 @@ void P2PComm::SendBroadcastMessageCore(const vector<Peer> & peers,
         indexes.at(i) = i;
     }
     random_shuffle(indexes.begin(), indexes.end());
-
+
```
Gnnng: seems an extra tab is added here.

src/libNetwork/P2PComm.cpp
```diff
@@ -609,7 +609,7 @@ void P2PComm::SendBroadcastMessage(const vector<Peer> & peers,
                                    const vector<unsigned char> & message)
 {
     LOG_MARKER();
-
+
```
Gnnng: extra tab

PR #103, opened 2018-03-03
Add getBalance API

sidthekidder:


src/libServer/Server.cpp
```diff
@@ -92,26 +91,16 @@ Json::Value Server::getLatestTxBlock()
...
-	// Address addr = Account::GetAddressFromPublicKey(pub);
-	// Account* account = AccountStore::GetInstance().GetAccount(addr);
-
-	// if(account != nullptr)
- //    {
- //    	Json::Value ret;
- //        ret["balance"] = 24;//account->GetBalance();
- //        ret["nonce"] = 0;//account->GetNonce();
+	vector<unsigned char> input_address(address.begin(), address.end());
+	const PrivKey priv(input_address, 0);
```
bb111189: Don't use magic number. Consider the following 
unsigned int offset = 0; 
const PrivKey priv(input_address, offset);

PR #104, opened 2018-03-03
Added APIs for getLatestTxBlock and getLatestDsBlock

KaustubhShamshery:


bb111189:
Add LOG_MARKER(); and start logging information

src/libServer/JSONConversion.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "libData/BlockData/Block.h"
+#include "JSONConversion.h"
+
+
+
+using namespace std;
+using namespace boost::multiprecision;
+
+
+const Json::Value JSONConversion::convertBoolArraytoJson(vector<bool> v)
```
Gnnng: It's better to use const reference here to pass the parameter (e.g., `const vector<bool> &v`)

src/libServer/JSONConversion.cpp
```diff
@@ -0,0 +1,129 @@
...
+	Json::Value jsonBool;
+	for(auto i:v)
+	{
+		jsonBool.append(i?1:0);
+	}
+	return jsonBool;
+}
+
+
+const Json::Value JSONConversion::convertTxnHashArraytoJson(vector<TxnHash> v)
```
Gnnng: similar situation here, use reference to avoid unnecessary copy

src/libServer/JSONConversion.cpp
KaustubhShamshery: Okay I will do that

Gnnng:
@KaustubhShamshery, glad to see these improvements. I can see other places having the same issue as this, can you fix them as well? In general, we could follow the principle listed [here](https://isocpp.org/wiki/faq/references#call-by-reference)

src/libServer/JSONConversion.cpp
jiayaoqijia: How about txblock and dsblock? @KaustubhShamshery

PR #112, opened 2018-03-08
Support Mac OS

ckyang:
I've run Zilliqa on my MacOS environment, and got build error when building.
There seems to be lack of jsoncpp & json-rpc-cpp library path in some CMakeLists.txt.
Please have a review to see if it's necessary to add them.

Maybe it's not the best solution because I used absolute path inside...
link_directories(${JSONCPP_LIBDIR} /usr/local/Cellar/libjson-rpc-cpp/1.1.0/lib)

### DIFF
--- a/src/depends/common/CMakeLists.txt
+++ b/src/depends/common/CMakeLists.txt
@@ -1,3 +1,4 @@
+link_directories(${JSONCPP_LIBDIR})
 add_library (Common SHARED Common.cpp CommonData.cpp CommonIO.cpp FileSystem.cpp FixedHash.cpp RLP.cpp SHA3.cpp)
 include_directories(${Boost_INCLUDE_DIRS})
 target_include_directories (Common PUBLIC ${PROJECT_SOURCE_DIR}/src)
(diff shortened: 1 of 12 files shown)


bb111189:
Will need to remove the absolute path and version specific information

ckyang:
Sure, removed in 5b409bbd453e0b248ef1dafc11305e06171733ff

Gnnng:
The Mac build on travis was okay, so the reason of this could be different OSX versions or different brew versions. Can you refer to the travis build log [here](https://travis-ci.org/Zilliqa/Zilliqa/jobs/350833221) to check the building environment difference?

ckyang:
After comparing travis & my environment, I found the root cause.
In my environment, following environment variable should be set:

export LIBRARY_PATH="/usr/lib:/usr/local/lib"

Then the Zilliqa compile procedure could be completed smoothly.

"/usr/local/lib" is the default library path for Mac OS, so the environment variable (LIBRARY_PATH) should always be setup as well.
Thus, I think no modification is needed.

Issue #96, opened 2018-02-28
Smart pointer for HttpServer


sidthekidder:
Improve the JsonRpc server with `unique_ptr` for better security

bb111189:
Closed

PR #114, opened 2018-03-09
[WIP] Added Server to Zilliqa constructor

KaustubhShamshery:
A solution to Issue #96


Gnnng:
It looks good, though I would suggest reducing the places of using `#ifdef IS_LOOKUP_NODE` macros, especially introducing it into the definition of class `Zilliqa`.

One possible way is making the instance of `HttpServer` a data member of `Server` and letting `Server`'s constructor and deconstructor handle it.  As for the port number, we can simply use a local const inside `Server'`s constructor for now. Later, we can load it from the xml config file.

PR #116, opened 2018-03-13
Fix potential deadlock when acquiring two locks

Gnnng:


Gnnng:
See the equivalent code here: http://en.cppreference.com/w/cpp/thread/scoped_lock

Gnnng:
These bugs were introduced in #56. Should we review it once again?

Issue #51, opened 2018-01-14
build a docker container


bb111189:

Tails:
+1.

The difficulty is (I believe) that there is not a single endpoint to run Zilliqa right now, since a lot of test and internal functions are now still being used for the testnet. It would either need an entrypoint with multiple functions that can be called from a compose file, or multiple Dockerfiles.

Gnnng:
It may also benefit the continuous integration. The project is mainly developed under Ubuntu 16.04 while the Travis CI only has native Ubuntu build environments for 12.04 and 14.04, which cause troubles when updating the dependencies. The developers need to maintain a separate set of dependency installation instructions in `.travis.yml` to pass the travis check. 

Since Travis CI support docker and we are rapidly developing things on Ubuntu 16.04, it would great if we could refrain from supporting previous Ubuntu versions shortly and simplify the developing, CI and testing process by only maintaining the 16.04 version.

Here are [examples](https://github.com/search?utf8=%E2%9C%93&q=docker%20pull%2016.04+filename%3A.travis.yml) of using Ubuntu 16.04 docker on Traivs CI.

Gnnng:
The support for Ubuntu 16.04 travis build is added in #100.

bb111189:
Closed. Community contributions has created the Dockerfile and it has been merged into the master branch.

PR #120, opened 2018-03-15
Docker image for Zilliqa

nickvanbaelen:
Added a Dockerfile that allows building a Docker image of Zilliqa. From this image, you can spawn Docker containers (running Ubuntu 16.04) which contain the entire Zilliqa source tree and it's built binaries.

Currently by spawning a container from this image, you can enter an interactive shell. This image can however be used to flexibly define several Docker services running a specific Zilliqa node. For example:

- one Docker service for running a Zilliqa _testnet_ node listening on _localhost_ on port _20_
- one Docker service for running a Zilliqa _mainnet_ node listening on _another IP_ on port _40_
- ...

These services would all use the same (prebuilt) Docker image to spawn their own container instance.

Let me know what you think or if you have any questions.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,44 @@
+# escape=\
+FROM ubuntu:16.04
+
+ARG REPO=https://github.com/Zilliqa/Zilliqa.git
+ARG BRANCH=master
+ARG CONFIG=Debug
+
+#--------------------------------------------------------------------
+# PACKAGES - UPDATE DATABASE
+#--------------------------------------------------------------------
+RUN apt-get -qq update
+
+#--------------------------------------------------------------------
+# PACKAGES - BUILDING
+#--------------------------------------------------------------------
+RUN apt-get install -y pkg-config
+RUN apt-get install -y build-essential
+RUN apt-get install -y cmake
+RUN apt-get install -y git
+
+#--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


bb111189:
This pr is related to Issue #51

Gnnng:
Seems that the image built from this Dockerfile is a bit too large, ~1.3GB. The multiple `RUN` lines could be the cause and the solution is to merge them into few lines as much as possible.

nickvanbaelen:
Combining multiple RUN stages into one will have a negligible effect on image size. I agree that for example the apt-get RUN's could be merged into one (or two) RUN commands to reduce some package manager overhead. Other than that, there's no reason real reason to mash several commands into one RUN command.

The majority of the current image size is represented by the ubuntu image. So if your goal is to have an image that is as small as possible, a better route to go down is to use alpine (https://store.docker.com/images/alpine). But you could potentially open yourself up to a myriad of issues since this isn't the platform the team is currently using to develop Zilliqa on.

kishankumar258:
@nickvanbaelen  "one Docker service for running a Zilliqa testnet node listening on localhost on port 20"
Can you tell how would I proceed for doing this I am able to just run the zilliqa shell from the installed docker image

kishankumar258:
I also want to connect multiple Docker containers locally to form a network of nodes. So that I would be able to make transactions between them. Can any one help how would I be able to do that?

PR #124, opened 2018-03-18
Microblock cosig verification

ansnunez:
Hi Yaoqi, this is the updated version of "antonio" branch which you requested last week.

**What this branch mainly contains:**

1. Microblock now has cosig and cosig bitmap members (old microblock had signature but didn't use Signature data type)
2. DirectoryService now includes the checking for this cosig when it processes each microblock.

**Other necessary updates I made to make it work:**

1. I moved the BitVector methods from ConsensusCommon into libUtils so that MicroBlock can also access them
2. I created NumForConsensus in ConsensusCommon so that both ConsensusLeader and DirectoryService can access it
3. A bit of other cleanup done to MicroBlock.h and cpp

**Note:** I'm unable to test this on my end so please see if someone there can do a quick local_run of this branch (I guess even one shard should be OK, just need to make sure microblock creation and verification works fine).


Gnnng:
I can take a look at this as I'm developing on the `feature/coinbase` branch and working on microblock too.

src/libConsensus/ConsensusCommon.h
```diff
@@ -80,7 +72,7 @@ class ConsensusCommon
     State m_state;
 
     /// The minimum fraction of peers necessary to achieve consensus.
-    const double TOLERANCE_FRACTION;
+    static constexpr double TOLERANCE_FRACTION = (double) 0.667;
```
Gnnng: 0.667 is already double literal, probably we can drop the conversion. See http://en.cppreference.com/w/cpp/language/floating_literal

src/libConsensus/ConsensusBackup.cpp
```diff
@@ -629,7 +630,7 @@ bool ConsensusBackup::ProcessMessageCollectiveSigCore(const vector<unsigned char
     // Note on N-byte bitmap: N = number of bytes needed to represent all nodes (1 bit = 1 node) + 2 (length indicator)
 
     const unsigned int length_available = collectivesig.size() - offset;
-    const unsigned int length_needed = sizeof(uint32_t) + BLOCK_HASH_SIZE + sizeof(uint16_t) + SIGNATURE_CHALLENGE_SIZE + SIGNATURE_RESPONSE_SIZE + GetBitVectorLengthInBytes(m_pubKeys.size()) + 2 + SIGNATURE_CHALLENGE_SIZE + SIGNATURE_RESPONSE_SIZE;
+    const unsigned int length_needed = sizeof(uint32_t) + BLOCK_HASH_SIZE + sizeof(uint16_t) + SIGNATURE_CHALLENGE_SIZE + SIGNATURE_RESPONSE_SIZE + BitVector::GetBitVectorLengthInBytes(m_pubKeys.size()) + 2 + SIGNATURE_CHALLENGE_SIZE + SIGNATURE_RESPONSE_SIZE;
```
Gnnng: Maybe we could take this chance to refactor `BitVector::GetBitVectorLengthInBytes(m_pubKeys.size()) + 2` into `BitVector::GetBitVectorSerializedSize(m_pubKeys.size())` to hide the magic number.

src/libConsensus/ConsensusBackup.cpp
```diff
@@ -672,8 +673,8 @@ bool ConsensusBackup::ProcessMessageCollectiveSigCore(const vector<unsigned char
     }
 
     // N-byte bitmap
-    m_responseMap = GetBitVector(collectivesig, curr_offset, GetBitVectorLengthInBytes(m_pubKeys.size()));
-    curr_offset += GetBitVectorLengthInBytes(m_pubKeys.size()) + 2;
+    m_responseMap = BitVector::GetBitVector(collectivesig, curr_offset, BitVector::GetBitVectorLengthInBytes(m_pubKeys.size()));
+    curr_offset += BitVector::GetBitVectorLengthInBytes(m_pubKeys.size()) + 2;
```
Gnnng: The same refactoring could happen here (e.g., `BitVector::GetBitVectorSerializedSize(m_pubKeys.size())`).

src/libConsensus/ConsensusCommon.h
ansnunez: Agree. Pushed update.

src/libConsensus/ConsensusBackup.cpp
ansnunez: Agree. Pushed update.

src/libConsensus/ConsensusBackup.cpp
ansnunez: Agree. Pushed update.

PR #142, opened 2018-03-22
More shard nodes to send microblock to ds committee,  better concurrency handling and updating of welcome message

bb111189:
- Shard leader and 4 other shard node to send Microblock to DS committee
- Better concurrency handling in finalblock and schnorr 
- Add Trie Performance unit test
- Broadcast to support deque 
- Added log marker to various function and clean up unnecessary logging 
- Update Zilliqa welcome message


src/libCrypto/Schnorr.cpp
```diff
@@ -120,6 +125,8 @@ void BIGNUMSerialize::SetNumber(vector<unsigned char> & dst, unsigned int offset
 
 shared_ptr<EC_POINT> ECPOINTSerialize::GetNumber(const vector<unsigned char> & src, unsigned int offset, unsigned int size)
 {
+    lock_guard<mutex> g(m_mutexECPOINT);
+
     shared_ptr<BIGNUM> bnvalue = BIGNUMSerialize::GetNumber(src, offset, size);
```
KaustubhShamshery: Is this function locking two locks? because it calls GetNumber which also aquires a lock?

KaustubhShamshery: Is this function locking two locks? because it calls GetNumber which also aquires a lock?
Maybe lock it after the call to GetNumber.

PR #143, opened 2018-03-22
Revert the deleting of the sleeping

kaikawaliu:


src/libNode/Node.cpp
```diff
@@ -80,8 +80,8 @@ void Node::StartSynchronization()
             m_synchronizer.FetchLatestTxBlocks(m_mediator.m_lookup, m_mediator.m_txBlockChain.GetBlockCount());
             m_synchronizer.FetchLatestState(m_mediator.m_lookup);
             m_synchronizer.AttemptPoW(m_mediator.m_lookup);
-
-			this_thread::sleep_for(chrono::seconds(30));
+		
+	    this_thread::sleep_for(chrono::seconds(30));
```
bb111189: Can set this as a constant value. For local, 30. For non-local, 300

PR #192, opened 2018-04-01
Microblock Cosignature

ansnunez:
Hi Guys, here’s an updated version of the microblock cosignature (tested with test_node_simple).

### But first, a few comments:

  1. Our 2-round consensus currently generates co-sigs this way:
       First round  => co-sig over the original message (e.g., DSBlock, Microblock)
       Second round => co-sig over the co-sig from the first round

     Therefore, what we actually want to retrieve after consensus is the signature from the first round, not the second.
     And currently it’s the second that we retain in the consensus object.

  2. Our consensus doesn’t actually care what’s inside the message (that is the job of the Validator function).
     The consensus just generates signature over the ENTIRE message.

     This becomes a problem with our message types - the data blocks.
     Recall that our Microblock contains a header, a co-sig over the header, and the txn bodies.
     First, the co-sig “over the header” is actually wrong because what we are signing is the ENTIRE Microblock.
     Second, it seems weird that the co-sig field itself is part of the message to be co-signed.
     
     The same is true of DSBlock, and I think of Tx Block.

### What are my changes in this PR:

  1. I’ve updated the consensus to address comment 1 above.

  2. For comment 2, I’ve taken the co-sig out of the Microblock.
     When the primary node submits the Microblock to DS, it just appends the co-sig (and bitmap) to the Microblock.
     Then, the DS does the actual co-sig verification.
     Finally, the DS discards the co-sig.

### So what I think needs to be done after:

  1. Need to decide if it’s OK for the Microblock’s co-sig to be discarded after DS has verified it
  2. Need to decide how to handle DSBlock and TxBlock


Gnnng:
@ansnunez We've recently enabled `.clang-format` checks and fixes, which causes the Travis build failures. Can you refer to https://github.com/Zilliqa/Zilliqa/wiki/Developer-Tools for the instructions and fix the format issues?

ansnunez:
Thanks @Gnnng for the info. Formatting fixed.

bb111189:
With the recent PR, is this PR obsolete?

ansnunez:
Yes, this is PR is now obsolete. Closing PR and deleting branch.

PR #248, opened 2018-04-21
Removed redundant mutex m_mutexProducerConsumer

jiayaoqijia:


bb111189:
An interim fix was proposed in this PR. We leave the full fix to issue #252. PR approved.

PR #251, opened 2018-04-22
Fix Zilliqa/Issues#19

Gnnng:
🔗Zilliqa/Issues#19


bb111189:
To copy .sh files if exists.

Gnnng:
I've added `*.sh`

PR #258, opened 2018-04-24
Use fork g3log

ckyang:


Gnnng:
@ckyang Can you confirm that the log filename is the same as before in the 20-nodes local run?

PR #259, opened 2018-04-24
NAT using miniupnpc

KaustubhShamshery:
* Tries to map a given port to a port in router
     * First tries to map the same port number in router as in machine
     * If failed, then gets any random port and returns it
* Destroys mapping when Object is out of scope

Credit to @digiront for helping in this PR


bb111189:
Modify readme to include libminiupnpc-dev dependency

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,161 @@
+#include <miniupnpc/miniupnpc.h>
```
bb111189: alphabetic order

src/depends/NAT/nat.h
```diff
@@ -0,0 +1,30 @@
...
+class NAT
+{
+public:
+	NAT();
+	~NAT();
+
+	std::string externalIP();
+	int addRedirect(int port);
+	void removeRedirect(int port);
+
```
bb111189: remove extra line

src/depends/NAT/nat.cpp
KaustubhShamshery: resolved

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+    m_urls = make_shared<UPNPUrls>();
+    m_data = make_shared<IGDdatas>();
+
+    memset(m_urls.get(), 0, sizeof(struct UPNPUrls));
+    memset(m_data.get(), 0, sizeof(struct IGDdatas));
+
+    m_initialized = false;
+
+    struct UPNPDev* devlist;
```
bb111189: avoid raw ptr

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+    int error = 0;
+
+#if MINIUPNPC_API_VERSION >= 14
+    devlist = upnpDiscover(2000, NULL, NULL, 0, 0, 2, &error);
```
bb111189: What are all these magic numbers? Can we refactor it.. Can be just a variable in this class

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+void NAT::removeRedirect(int _port)
+{
+    if (!m_initialized)
+    {
+        LOG_GENERAL(WARNING, "Unitialized");
+        return;
+    }
+    char port_str[16];
+    sprintf(port_str, "%d", _port);
```
bb111189: sprintf is a dangerous function and can leads to buffer overflow. In this case, if my _port is more than the size of char array, buffer overflow can happen.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+    UPNP_DeletePortMapping(m_urls->controlURL, m_data->first.servicetype,
+                           port_str, "TCP", NULL);
+    m_reg.erase(_port);
+}
+
+NAT::~NAT()
+{
+    auto r = m_reg;
+    for (auto i : r)
```
bb111189: enclose in bracket

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    m_lanAddress = lan_address;
+    m_initialized = true;
+    freeUPNPDevlist(devlist);
+}
+
+string NAT::externalIP()
+{
+    if (!m_initialized)
```
bb111189: enclosed in bracket

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+    else
+    {
+        LOG_GENERAL(WARNING, "Unable to get external IP");
+        return "0.0.0.0";
+    }
+}
+
+int NAT::addRedirect(int _port)
+{
+    if (!m_initialized)
```
bb111189: enclosed in bracket

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+        return -1;
+
+    char port_str[16];
+    sprintf(port_str, "%d", _port);
```
bb111189: This is dangerous function and can lead to buffer overflow

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+        m_lanAddress.c_str(), "zilliqa", "TCP", NULL, NULL);
+
+    if (!error)
+    {
+        m_reg.insert(_port);
+        return _port;
+    }
+    else
+    {
+        LOG_GENERAL(INFO, "Failed to map same port in router");
```
bb111189: we should return error here

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    if (UPNP_AddPortMapping(m_urls->controlURL, m_data->first.servicetype,
+                            port_str, NULL, m_lanAddress.c_str(), "zilliqa",
+                            "TCP", NULL, NULL))
+    {
+        LOG_GENERAL(WARNING, "Failed to map Port");
+        return 0;
+    }
+
+    unsigned num = 0;
```
bb111189: Is that any special reason for using "unsigned" type?
http://soundsoftware.ac.uk/c-pitfall-unsigned.html

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+    UPNP_GetPortMappingNumberOfEntries(m_urls->controlURL,
+                                       m_data->first.servicetype, &num);
+    for (unsigned i = 0; i < num; ++i)
```
bb111189: unsigned without type

src/depends/NAT/nat.h
```diff
@@ -0,0 +1,28 @@
...
+	NAT();
+	~NAT();
+
+	std::string externalIP();
+	int addRedirect(int port);
+	void removeRedirect(int port);
+	bool isIntialized() const {return m_initialized;}
+
+private:
+	std::set<int> m_reg;
```
bb111189: should it be set of unsigned int?

src/depends/NAT/nat.h
```diff
@@ -0,0 +1,28 @@
...
+	bool m_initialized;
+	std::string m_lanAddress;
+	std::shared_ptr<UPNPUrls> m_urls;
+	std::shared_ptr<IGDdatas> m_data;
+
+
```
bb111189: extra line

src/cmd/main.cpp
```diff
@@ -22,6 +22,7 @@
 #include <arpa/inet.h>
```
bb111189: is this our code? if no, we may have to credit or check the licence
if yes, add our copyright comments

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
+#include "nat.h"
```
bb111189: is this our code? if no, we may have to credit or check the licence
if yes, add our copyright comments

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+        return;
+    }
+    char lan_address[INET6_ADDRSTRLEN];
+
+    int status = UPNP_GetValidIGD(devlist, m_urls.get(), m_data.get(),
+                                  lan_address, sizeof(lan_address));
+
+    if (status != 1)
+    {
+        LOG_GENERAL(WARNING, "Unable to get Valid IGD");
```
bb111189: what is IGD?

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+        char protocol[4];
+        char desc[80];
+        char enabled[4];
+        char rHost[64];
+        char duration[16];
+        UPNP_GetGenericPortMappingEntry(
+            m_urls->controlURL, m_data->first.servicetype, to_string(i).c_str(),
+            extPort, intClient, intPort, protocol, desc, enabled, rHost,
+            duration);
+        if (string("zilliqa") == desc)
```
bb111189: do we need to cast to string here?

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+        {
+            m_reg.insert(atoi(extPort));
+            return atoi(extPort);
+        }
+    }
+
+    LOG_GENERAL(WARNING, "Failed to find port");
```
bb111189: warning should be more verbose such as printing the port

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    return -1;
+}
+
+void NAT::removeRedirect(int _port)
+{
+    if (!m_initialized)
```
bb111189: log marker

src/depends/NAT/nat.cpp
bb111189: log marker

src/depends/NAT/nat.cpp
digiront: A good practice is to declare these as const variables in an anonymous namespace inside the .cpp e.g.

```
// foo.cpp
namepsace {
    const int MY_CONSTANT = 1;
} // close anonymous namespace
```

> If you plan to expose them to other classes then they have to be static members.

src/depends/NAT/nat.cpp
digiront:
```
auto portStr = std::to_string(42);
port_str.c_str(); // returns null terminated C string
```

src/depends/NAT/nat.cpp
bb111189: change logging to warning

src/depends/NAT/nat.cpp
KaustubhShamshery: Internet Gateway Device (eg. router)

src/depends/NAT/nat.cpp
KaustubhShamshery: yes , otherwise we would have to use strcmp

src/depends/NAT/nat.cpp
bb111189: Maybe to make to friendly to other developer, we can change it to 

`LOG_GENERAL(WARNING, "Unable to get Valid Internet Gateway Device (IGD)");`

src/cmd/main.cpp
Gnnng: It has copyright information already.

bb111189: ok

src/depends/NAT/nat.cpp
bb111189: ok

src/depends/NAT/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_library (NAT nat.cpp)
```
bb111189: add_library (NAT SHARED nat.cpp)

bb111189:
Outstanding task: nat in depends change to dynamically linking.

src/cmd/main.cpp
```diff
@@ -44,58 +52,77 @@ int main(int argc, const char* argv[])
...
-        INIT_EPOCHINFO_LOGGER("epochinfo");
-
-        vector<unsigned char> tmpprivkey
-            = DataConversion::HexStrToUint8Vec(argv[1]);
-        vector<unsigned char> tmppubkey
-            = DataConversion::HexStrToUint8Vec(argv[2]);
-        // PrivKey privkey(tmpprivkey, 0);
-        PrivKey privkey;
-        if (privkey.Deserialize(tmpprivkey, 0) != 0)
+        NAT nt;
```
satoshi101: `nt` is local within the `if` which means that by the end of the `if` scope it will be destructed and `removeRedirect()` will be called. To make this work, probably use a `std::unique_ptr` to keep it alive till end of `main()` if the pointer is initialized.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+        return;
+    }
+
+    m_lanAddress = lan_address;
+    m_initialized = true;
+}
+
+NAT::~NAT()
+{
+    auto r = m_reg;
```
satoshi101: So this is tricky. Clever that you do a copy, because you iterate `m_reg`, while removing from it using `removeRedirect`. As a naive maintainer, I would be tempted to remove it, optimize it though. So, I think I would prefer it to have a more intention seeking name e.g. `r` -> `regCopySafeToIterateDuringDelete` or something like that.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+        m_urls->controlURL, m_data->first.servicetype, port_str.c_str(), port_str.c_str(),
+        m_lanAddress.c_str(), "zilliqa", "TCP", NULL, NULL);
+
+    if (!error)
+    {
+        m_reg.insert(_port);
+        return _port;
+    }
+
+    // 2) Failed - now try (random external, port internal) and cycle up to 10 times.
```
satoshi101: Almost certainly this needs a little bit of fix for the indentation level

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+		_port = rand() % (32768 - 1024) + 1024;
+		string ext_port_str = to_string(_port);
+		if (!UPNP_AddPortMapping(m_urls->controlURL, m_data->first.servicetype, ext_port_str.c_str(), port_str.c_str(), m_lanAddress.c_str(), "ethereum", "TCP", NULL, NULL))
+        {
+			return _port;
+        }
+	}
+
+	// 3) Failed. Try asking the router to give us a free external port.
+    // This may not works on some routers.
```
satoshi101: `works` -> `work`

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+    char reservedPort[6];
+    int result = UPNP_AddAnyPortMapping(m_urls->controlURL, m_data->first.servicetype,
+                            port_str.c_str(), port_str.c_str(), m_lanAddress.c_str(), "zilliqa",
+                            "TCP", NULL, NULL, reservedPort);
+
+    if (result)
+    {
+        return 0;
```
satoshi101: Probably would still return `-1`. This feels a bit like success.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    if (_port < 0 && _port > 65535)
+    {
+        return -1;
+    }
+
+    string port_str = to_string(_port);
+
+    // Remove any dangling mapping
+    UPNP_DeletePortMapping(m_urls->controlURL, m_data->first.servicetype,
```
satoshi101: Hmmm... Since we do this here, should we do it when we try random ports a bit below, as well?

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    // 2) Failed - now try (random external, port internal) and cycle up to 10 times.
+	srand(static_cast<unsigned int>(time(nullptr)));
+	for (unsigned i = 0; i < 10; ++i)
+	{
+		_port = rand() % (32768 - 1024) + 1024;
+		string ext_port_str = to_string(_port);
+		if (!UPNP_AddPortMapping(m_urls->controlURL, m_data->first.servicetype, ext_port_str.c_str(), port_str.c_str(), m_lanAddress.c_str(), "ethereum", "TCP", NULL, NULL))
+        {
+			return _port;
```
satoshi101: We likely need `m_reg.insert(_port);` here as well.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+                            port_str.c_str(), port_str.c_str(), m_lanAddress.c_str(), "zilliqa",
+                            "TCP", NULL, NULL, reservedPort);
+
+    if (result)
+    {
+        return 0;
+    }
+    else
+    {
+        return atoi(reservedPort);
```
satoshi101: We likely need `m_reg.insert(atoi(reservedPort));` here as well.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,176 @@
...
+NAT::NAT()
+{
+    m_urls = make_shared<UPNPUrls>();
+    m_data = make_shared<IGDdatas>();
+
+    memset(m_urls.get(), 0, sizeof(struct UPNPUrls));
+    memset(m_data.get(), 0, sizeof(struct IGDdatas));
+
+    m_initialized = false;
+
```
satoshi101: At this point, I would "stop" the constructor and start an `init()` method. `upnpDiscover` and `UPNP_GetValidIGD` are both very heavy and I think it would be better if this was returning error codes. Overall people recommend constructors to be relatively "light" and if they end up "heavy" turn them into `init()` methods. By using the `m_initialized` member, you're anyway halfway there.

tests/Zilliqa/test_zilliqa_NAT.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python
```
satoshi101: Apart from the name of the file and the mention of `NAT` in one of the arguments, I can't find any other NAT reference. This likely means that it's a slight modification of another script. Maybe the other script should be (slightly) extended instead of a duplication here. Otherwise, let's do the NAT-related testing functionality more explicit.

src/depends/NAT/nat.cpp
```diff
@@ -0,0 +1,177 @@
...
+        return _port;
+    }
+
+    // 2) Failed - now try (random external, port internal) and cycle up to 10 times.
+	srand(static_cast<unsigned int>(time(nullptr)));
+	for (unsigned i = 0; i < 10; ++i)
+	{
+		_port = rand() % (32768 - 1024) + 1024;
+		string ext_port_str = to_string(_port);
+		if (!UPNP_AddPortMapping(m_urls->controlURL, m_data->first.servicetype, ext_port_str.c_str(), port_str.c_str(), m_lanAddress.c_str(), "ethereum", "TCP", NULL, NULL))
```
KaustubhShamshery: Also here

src/depends/NAT/nat.cpp
bb111189: Done

tests/Zilliqa/test_zilliqa_NAT.py
bb111189: Created issue #352 to track as it will impact other internal scripts.

PR #263, opened 2018-04-26
Use libevent as network receiving framework

ckyang:


src/libNetwork/P2PComm.h
```diff
@@ -76,20 +76,23 @@ class P2PComm
     Peer m_selfPeer;
 #endif // STAT_TEST
 
-    ThreadPool m_SendPool{MAXMESSAGE, "SendPool"};
-    ThreadPool m_RecvPool{MAXMESSAGE, "RecvPool"};
+    ThreadPool m_SendPool{MAXMESSAGE / 2, "SendPool"};
```
bb111189: Any rationale to divide by 2? Or maybe can we just edit the constant file

tests/Network/Test_P2PComm.cpp
```diff
@@ -22,22 +22,42 @@
...
+    }
 }
 
 int main()
 {
     INIT_STDOUT_LOGGER();
 
     auto func = []() mutable -> void {
-        P2PComm::GetInstance().StartMessagePump(30303, process_message);
+        P2PComm::GetInstance().StartMessagePump(30303, process_message,
```
bb111189: Maybe we can remove the magic number 30303 here

src/libNetwork/P2PComm.h
KaustubhShamshery: I did that. Maybe it should be reverted to the constant file. I did this because always the number of threads were twice of maxmessages which made it easier to know how many threads are being formed

KaustubhShamshery: I did that. Maybe it should be reverted to the constant file. I did this because always the number of threads were twice of maxmessages which made it easier to know how many threads are being formed

ckyang: OK, I will change constant_local.xml instead of modifying P2PComm.h

PR #272, opened 2018-05-07
Remove inefficient sha2 update and use openssl update function

ckyang:
https://github.com/Zilliqa/Issues/issues/27


bb111189:
Verified that it must existing unit test for Sha2

PR #281, opened 2018-05-11
Importing json string for Interfacing interpreter

kaikawaliu:


src/libData/AccountData/Account.cpp
```diff
@@ -146,16 +232,65 @@ vector<string> Account::GetKeys()
...
+    {
+        vector<string> v = GetStorage(k);
+        if (v[0] == "False")
+        {
+            continue;
+        }
+        Json::Value item;
+        item["vname"] = k;
+        item["type"] = v[1];
+        if (v[1] == "Map")
```
AmritKumar: I think we may another special type ADT. We are trying to finalize that. Once it is done, we may need to update this part of the code.

src/libData/AccountData/Account.h
```diff
@@ -43,29 +44,39 @@ class Account : public Serializable
 {
     boost::multiprecision::uint256_t m_balance;
     boost::multiprecision::uint256_t m_nonce;
-    h256 m_storageRoot = h256();
-    h256 m_prevRoot;
+    h256 m_storageRoot, m_prevRoot;
```
AmritKumar: Why do we need a m_prevRoot exactly?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -179,19 +180,194 @@ void AccountStore::AddAccount(const PubKey& pubKey, const Account& account)
...
+            SysCommand::ExecuteCmdWithOutput(GetContractCmdStr());
+            ParseContractOutput();
+        }
+    }
+
     TransferBalance(fromAddr, toAddr, amount);
     IncreaseNonce(fromAddr);
+}
+
+Json::Value AccountStore::GetBlockStateJson(const uint64_t& BlockNum) const
```
AmritKumar: I guess this function can easily be modified in case  blockstatejson has other values such as `TIMESTAMP` and `BLOCKHASH`?

kaikawaliu: Yes, it's easy to modify

src/libData/AccountData/Account.h
kaikawaliu: That's necessary for persistence recovery purpose at current design, it won't be written into disk. Let's keep it first.

src/libData/AccountData/Account.cpp
kaikawaliu: Okay

src/libData/AccountData/Account.h
AmritKumar: ok

PR #282, opened 2018-05-12
Consensus Subsets

ansnunez:
I've modified the consensus process as follows:

1. Leader sends announcement to all backups.
2. Leader receives as many commits as it can (before, we used to stop at numForConsensus).
3. After COMMIT_WINDOW_IN_SECONDS, leader generates 50 subsets each with exactly numForConsensus backups assigned at random.
4. Leader continues consensus (starting at challenge) within each subset.  Now all messages at this point have an extra subsetID value.
5. To avoid flooding the network, leader delays starting each subset by a few seconds.
6. As soon as one subset reaches the end of consensus, leader sends finalcollectivesig to everyone and stops processing other subsets.
7. Backups behave just as before, except now they also keep track of whichever subset a leader is sending a message for. When a backup receives finalcollectivesig, it also stops processing other subsets.

I've only tested so far in test_node_simple.sh.
I propose to integrate this slowly into your larger-scale tests with maybe just a few subsets at first.
Timing and subset parameters are defined as consts inside ConsensusLeader.h


ansnunez:
The work in this PR has now been implemented in #742

PR #285, opened 2018-05-14
Adds libevent-dev dependency

satoshi101:
Adds libevent-dev dependency. Fixes compilation error:

```
P2PComm.cpp:19:19: fatal error: event.h: No such file or directory
compilation terminated.
```


bb111189:
This pr will be obsoleted due to PR #295

PR #287, opened 2018-05-14
Use condition variable instead of busyloop on ProcessFinalBlockConsensus

ckyang:
https://github.com/Zilliqa/Issues/issues/25


bb111189:
Documenting a discussion down. 

+                cv_lkObject, std::chrono::seconds(CONSENSUS_OBJECT_TIMEOUT))
For this timeout, we should set it to longer. This is because, we will wait if the state is at microblock submission or final block consensus prep state. 

The timeout is ok for final block consensus prep state but maybe me too short for microblock submission state. 

To enter this function, the ds leader should be receive all microblock. However, the ds backup may not have received it. A timeout of 10s may be too short. Hence, we will be giving it a longer timeout in this case.

PR #293, opened 2018-05-16
Add POW unit test

ckyang:
Enable POW unit test case and make travis check will run it.


bb111189:
The reason for the unit test cases to failed is due to the current test case expected result is for sha3-fips not sha3-keccak. In the early days, we changed the pow(ethash) from sha3-keccak to sha3-fips. However, a later decision reverted the changes, but the unit test has not been changed to reflect that. 

The expected output can be found at https://github.com/ethereum/ethash/blob/master/test/c/test.cpp

PR #297, opened 2018-05-17
[Testing] Feature/viewchange v2

bb111189:
In this new version of view change, view change will happen when there is a timeout (only). When there is a timeout, it means that during this period of time, consensus (among the DS committee) did not happen or the DS committee failed to reach a consensus. When there is a timeout, the DS committee will run view change consensus protocol, which will piggyback on the existing PBFT protocol, to agree on the the incoming new DS leader. 

When view change occurs, outgoing leader will get push to the back of the deque, ready to be kick out at the next ds-epoch. 

Recursive view change is supported in this


bb111189:
This PR is way larger than I first envisioned it will be. Will try to keep to a smaller PR in the future. 

This PR is ready for code review.

src/libDirectoryService/ViewChangePostProcessing.cpp
```diff
@@ -0,0 +1,413 @@
...
+            m_mode = PRIMARY_DS;
+        }
+        else
+        {
+            LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
+                      "After view change, I am ds backup");
+            m_mode = BACKUP_DS;
+        }
+
+        // Kick ousted leader to the back of the queue, waiting to be eject at
```
jiayaoqijia: Discuss with @ckyang , once he merges pow1/pow2.

src/libData/BlockData/BlockHeader/VCBlockHeader.cpp
```diff
@@ -0,0 +1,228 @@
...
+    unsigned int curOffset = offset;
+    SetNumber<uint64_t>(dst, curOffset, m_VieWChangeDSEpochNo,
+                        sizeof(uint64_t));
+    curOffset += sizeof(uint64_t);
+    SetNumber<uint64_t>(dst, curOffset, m_VieWChangeEpochNo, sizeof(uint64_t));
+    curOffset += sizeof(uint64_t);
+    SetNumber<unsigned char>(dst, curOffset, m_ViewChangeState,
+                             sizeof(unsigned char));
+    curOffset += sizeof(unsigned char);
+    SetNumber<unsigned int>(dst, curOffset, m_CandidateLeaderIndex,
```
jiayaoqijia: Change sizeof(unsigned char) to sizeof(uint8_t). We should keep the length of a field to be fixed, as we cannot predict which os is used by another node.

constants_local.xml
```diff
@@ -16,15 +16,17 @@
...
 		<VIEWCHANGE_TIME>180</VIEWCHANGE_TIME>
+        <VIEWCHANGE_EXTRA_TIME>10</VIEWCHANGE_EXTRA_TIME>
 		<SHARDING_TIMEOUT>180</SHARDING_TIMEOUT>
 		<CONSENSUS_OBJECT_TIMEOUT>10</CONSENSUS_OBJECT_TIMEOUT>
 		<FINALBLOCK_CONSENSUS_OBJECT_TIMEOUT>30</FINALBLOCK_CONSENSUS_OBJECT_TIMEOUT>
 		<NUM_FINAL_BLOCK_PER_POW>5</NUM_FINAL_BLOCK_PER_POW>
 		<NUM_DS_KEEP_TX_BODY>5</NUM_DS_KEEP_TX_BODY>
-		<MAXMESSAGE>64</MAXMESSAGE>
+		<MAXMESSAGE>32</MAXMESSAGE>
 		<MAXSUBMITTXNPERNODE>10</MAXSUBMITTXNPERNODE>
```
ckyang: Why need to reduce MAXMESSAGE? If needed to do so, make sure to sync to testnet.

src/common/Constants.h
```diff
@@ -30,6 +30,7 @@ const unsigned int BLOCK_HASH_SIZE = 32;
 const unsigned int BLOCK_SIG_SIZE = 64;
 
 // Numeric types sizes
+const unsigned int UNSIGNEDCHAR_SIZE = 1;
```
ckyang: Why need to define this instead of using sizeof(unsigned char)?

src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
```diff
@@ -23,7 +23,6 @@ using namespace boost::multiprecision;
 TxBlockHeader::TxBlockHeader()
 {
     m_blockNum = (boost::multiprecision::uint256_t)-1;
```
ckyang: Coding style of constructor initialization should be consistent.
At least, we should choose one style and follow it inside same file.
For example, replace this line to
TxBlockHeader::TxBlockHeader()
: m_blockNum((boost::multiprecision::uint256_t)-1)
{}

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -105,12 +102,14 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary()
     m_consensusBlockHash.resize(BLOCK_HASH_SIZE);
     fill(m_consensusBlockHash.begin(), m_consensusBlockHash.end(), 0x77);
 
-    // kill first ds leader
-    // if (m_consensusMyID == 0 && temp_todie)
-    // {
-    //    LOG_GENERAL(INFO, "I am killing myself to test view change");
-    //    throw exception();
-    // }
+    // kill first ds leader (used for view change testing)
```
ckyang: Why need to preserve this?

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -449,13 +449,15 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSPrimary()
...
     vector<unsigned char> finalBlockMessage = ComposeFinalBlockMessage();
 
-    // kill first ds leader
-    //if (m_consensusMyID == 0 && temp_todie)
-    //{
-    //    LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-    //                 "I am killing myself to test view change");
-    //    throw exception();
-    //}
+    // kill first ds leader (used for view change testing)
```
ckyang: Why need to preserve this?

src/libNode/ViewChangeBlockProcessing.cpp
```diff
@@ -0,0 +1,276 @@
...
+        dsblock.GetHeader().GetBlockNum());
+#endif
+}
+**/
+
+void Node::UpdateDSCommiteeComposition()
+{
+    LOG_MARKER();
+
+    {
```
ckyang: Needless to use one more level

constants_local.xml
bb111189: I noticed some of the user observed out of resource issue on slower machine. Hence, I have advise them to change to a lower number. There is no noticeable impact. 

Good point though. I need to check other scripts too.

src/common/Constants.h
bb111189: @jiayaoqijia pointed out that sizeof(unsigned char) or sizeof(unsigned int) may different on different OS. Hence, it is good that we fixed the length.

src/libDirectoryService/DSBlockPreProcessing.cpp
bb111189: Although ugly, we got to preserve it for future view change testing. I can't think of a simple unit test to remove this part.

src/libDirectoryService/FinalBlockPreProcessing.cpp
bb111189: Although ugly, we got to preserve it for future view change testing. I can't think of a simple unit test to remove this part.

src/libDirectoryService/ViewChangePostProcessing.cpp
bb111189: Will be discussed in his PR once it is ready

src/common/Constants.h
ckyang: https://www.tutorialspoint.com/cprogramming/c_data_types.htm

This page shows only sizeof(int) and sizeof(unsigned int) will differ on different OS.

Gnnng: Good catch, although I would refer to the c++ [standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf) **5.3.3 Sizeof**

Gnnng: We probably want to fix this

src/libData/BlockData/Block/VCBlock.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const VCBlockHeader& VCBlock::GetHeader() const { return m_header; }
+
+bool VCBlock::operator==(const VCBlock& block) const
+{
+    return (m_header == block.m_header);
+}
+
+bool VCBlock::operator<(const VCBlock& block) const
```
Gnnng: How about this
```cpp
bool VCBlock::operator<(const VCBlock& block) const { return m_header < block.m_header; }  
```

src/libDirectoryService/ViewChangePostProcessing.cpp
```diff
@@ -0,0 +1,413 @@
...
+    {
+        LOG_GENERAL(WARNING, "Aggregated key generation failed");
+    }
+
+    vector<unsigned char> message;
+    m_pendingVCBlock->GetHeader().Serialize(message, 0);
+    m_pendingVCBlock->GetCS1().Serialize(message, VCBlockHeader::SIZE);
+    BitVector::SetBitVector(message, VCBlockHeader::SIZE + BLOCK_SIG_SIZE,
+                            m_pendingVCBlock->GetB1());
+    if (Schnorr::GetInstance().Verify(message, 0, message.size(),
```
Gnnng: Understand that `== false` makes this explict, but how about using `not` keyword http://en.cppreference.com/w/cpp/language/operator_alternative

src/libNode/DSBlockProcessing.cpp
```diff
@@ -193,6 +180,10 @@ bool Node::VerifyDSBlockCoSignature(const DSBlock& dsblock)
         == false)
     {
         LOG_GENERAL(WARNING, "Cosig verification failed");
+        for (auto& kv : keys)
+        {
+            LOG_GENERAL(WARNING, "" << kv);
```
Gnnng: `"" << kv` looks unnecessary, any specific reason for doing this?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -250,6 +235,10 @@ bool Node::VerifyFinalBlockCoSignature(const TxBlock& txblock)
         == false)
     {
         LOG_GENERAL(WARNING, "Cosig verification failed");
+        for (auto& kv : keys)
+        {
+            LOG_GENERAL(WARNING, "" << kv);
```
Gnnng: same question

src/libNode/DSBlockProcessing.cpp
bb111189: This is to identify who cosign the message. This is important for debugging and such information should be captured when cosig verification fail.

src/libNode/FinalBlockProcessing.cpp
bb111189: This is to identify who cosign the message. This is important for debugging and such information should be captured when cosig verification fail.

src/libData/BlockData/Block/VCBlock.cpp
bb111189: Fised

src/libNode/DSBlockProcessing.cpp
Gnnng: Sorry for not making the question clear, what I mean is can we just use `kv` instead of `"" << kv`?

PR #298, opened 2018-05-17
Remove persistence storage on each new local test run

bb111189:


KaustubhShamshery:
test_node_late.sh also needs the same update.

Issue #304, opened 2018-05-23
Replace asserts with appropriate error handling code


ansnunez:
Need to make sure we aren’t skipping some necessary checks because of assert code, which only works in debug mode

ckyang:
Currently g3log will automatically crash the program for FATAL message.
Thus the proposal of this issue should be:

assert(condition);
->
LOG_GENERAL(FATAL, "assertion failed (" << __FILE__ << ":" << __LINE__ << ": " << __FUNCTION__ << ")" << IfThereIsCustomMsg);

ansnunez:
Interesting, didn't know.  Yes, we can use FATAL, and on the other hand I think some existing asserts may not necessary need to be at FATAL level (maybe just WARNING).

ckyang:
Done in https://github.com/Zilliqa/Zilliqa/pull/314

PR #314, opened 2018-05-30
Replace assert to LOG_GENERAL with FATAL

ckyang:
https://github.com/Zilliqa/Zilliqa/issues/304


ckyang:
@ansnunez LOG_GENERAL shows function name only, however for the assert fail maybe file name & line number are also important.

Issue #308, opened 2018-05-28
Travis: Test_POW taking too much time (6 min)


Gnnng:
See https://travis-ci.org/Zilliqa/Zilliqa/jobs/381962476#L2042-L2043

Just found that the test case `Test_POW` lasts for nearly 6 minutes, which is close to 50% of the travis running time.

Haven't checked the source code of the test case, perhaps we can tune the parameters and reduce the testing time a bit.

KaustubhShamshery:
I have also experienced the same maybe reducing this https://github.com/Zilliqa/Zilliqa/blob/8c6f49c1f696a35be2c79b6723a1237847d4fbaf/tests/POW/test_POW.cpp#L706

will help.

Gnnng:
A temporary hotfix is coming in #309, which temporarily disables the time-consuming part in Test_POW. Let's keep this issue open until the root cause is found.

Another possible workaround is increasing the timeout on travis: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received. However, this is just another workaround.

ckyang:
Done in https://github.com/Zilliqa/Zilliqa/pull/315

Gnnng:
This issue is reoccurring lately, see 
- https://travis-ci.com/Zilliqa/Zilliqa/jobs/135498870#L2233 (nearly 10 minutes)
- https://travis-ci.com/Zilliqa/Zilliqa/jobs/135520896#L2234 (over 10 minutes, killed by Travis)
- https://travis-ci.org/Zilliqa/Zilliqa/jobs/405471024#L682 (8 minutes)

I guess it is related to the change brought by #422. @ShengguangXiao can you take a look when you are available?

ShengguangXiao:
The problem is because in the Test_POW, it is using full mining now, and the full mining need to generate the dag data which need about 5 minutes to complete on normal CPU machine. I can change the test to light mining for now.

PR #315, opened 2018-05-31
Modify Test_POW test case in order to shorten testing time.

ckyang:
https://github.com/Zilliqa/Zilliqa/issues/308


ckyang:
Current Test_POW takes about 5 seconds with removing fullmine testing.
In Zilliqa code we never used fullmine so it should be enough to verify lightmine only.

PR #320, opened 2018-06-02
Add leader to consensus cosig

ansnunez:
Modified the consensus to include the leader in the cosig


src/libConsensus/ConsensusCommon.cpp
```diff
@@ -207,5 +207,5 @@ const vector<bool>& ConsensusCommon::GetB2() const
 
 unsigned int ConsensusCommon::NumForConsensus(unsigned int shardSize)
 {
-    return ceil(shardSize * TOLERANCE_FRACTION) - 1;
+    return ceil(shardSize * TOLERANCE_FRACTION);
```
bb111189: @jiayaoqijia  Please go through the calculation again

PR #322, opened 2018-06-04
Removes the option of null pointer in validator

satoshi101:
It's very weird to log and continue on a null pointer. 2 lines later, on dereference, a seg fault will happen.

```
    if (m_mediator == nullptr)
    {
        LOG_GENERAL(WARNING, "Error: pointer to Mediator is nullptr.");
    }
```

The easiest way to fix this is to get rid of the pointer altogether, I think.

### DIFF
--- a/src/libValidator/Validator.cpp
+++ b/src/libValidator/Validator.cpp
@@ -24,7 +24,7 @@ using namespace std;
 using namespace boost::multiprecision;
 
 Validator::Validator(Mediator& mediator)
-    : m_mediator(&mediator)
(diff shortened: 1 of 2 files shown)


satoshi101:
```
    Mediator* m = nullptr;
    unique_ptr<ValidatorBase> m_validator = make_unique<Validator>(*m);
```

https://github.com/Zilliqa/Zilliqa/blob/master/tests/Data/Test_Transaction.cpp?utf8=%E2%9C%93#L46

PR #323, opened 2018-06-04
[WIP] Clean validator code

nataouze:
Const-correctness.
Avoid multiple account lookups. Even if the lookup itself has a constant complexity, code is cleaner like that.


src/libValidator/Validator.cpp
```diff
@@ -74,7 +74,8 @@ bool Validator::CheckCreatedTransaction(const Transaction& tx) const
     }
 
     // Check if from account exists in local storage
-    if (!AccountStore::GetInstance().DoesAccountExist(fromAddr))
+    const Account* fromAccnt = AccountStore::GetInstance().GetAccount(fromAddr);
```
bb111189: Do not use raw ptr

src/libValidator/Validator.cpp
```diff
@@ -85,20 +86,21 @@ bool Validator::CheckCreatedTransaction(const Transaction& tx) const
 
     // Check if to account exists in local storage
     const Address& toAddr = tx.GetToAddr();
-    if (!AccountStore::GetInstance().DoesAccountExist(toAddr))
+    const Account* toAccnt = AccountStore::GetInstance().GetAccount(toAddr);
```
bb111189: Do not use raw ptr

src/libValidator/Validator.cpp
```diff
@@ -132,7 +134,8 @@ bool Validator::CheckCreatedTransactionFromLookup(const Transaction& tx)
     }
 
     // Check if from account exists in local storage
-    if (!AccountStore::GetInstance().DoesAccountExist(fromAddr))
+    const Account* fromAccnt = AccountStore::GetInstance().GetAccount(fromAddr);
```
bb111189: Do not use raw ptr

satoshi101:
Whenever one sees `*` in the code (where it didn't exist before), the code is definitely _NOT_ cleaner. "constant complexity" or any other complexity is not an issue, unless there's a performance problem, proved with a profiler.

src/libValidator/Validator.cpp
```diff
@@ -184,20 +184,21 @@ bool Validator::CheckCreatedTransactionFromLookup(const Transaction& tx)
 
     // Check if to account exists in local storage
     const Address& toAddr = tx.GetToAddr();
-    if (!AccountStore::GetInstance().DoesAccountExist(toAddr))
+    const Account* toAccnt = AccountStore::GetInstance().GetAccount(toAddr);
```
bb111189: Do not use raw ptr

satoshi101:
Nice @nataouze - I can see a monad.

`AccountStore` is not thread safe and all sorts of bad things can happen as far as I can see when people remove or there's a rehash on that `unordered_map`. One relatively thread safe to fix this problem is to use `shared_ptr`s. Another, potentially more clean is to use callbacks, which will be invoked while holding a lock (omited below):

```
void AccountStore::GetAccount(const Address& address,
                              const std::function<void(const Account&)>& callback)
{
    //LOG_MARKER();

    auto it = m_addressToAccount.find(address);
    // LOG_GENERAL(INFO, (it != m_addressToAccount.end()));
    if (it != m_addressToAccount.end())
    {
        callback(&it->second);
    }

    string accountDataString = m_state.at(address);
    if (accountDataString.empty())
    {
        return;
    }

    dev::RLP accountDataRLP(accountDataString);
    auto it2 = m_addressToAccount.emplace(
        std::piecewise_construct, std::forward_as_tuple(address),
        std::forward_as_tuple(
            accountDataRLP[0].toInt<boost::multiprecision::uint256_t>(),
            accountDataRLP[1].toInt<boost::multiprecision::uint256_t>(),
            accountDataRLP[2].toHash<dev::h256>(),
            accountDataRLP[3].toHash<dev::h256>()));

    callback(&it2.first->second);
}
```

by using this one could reform `CheckCreatedTransaction*()` to use:

```
AccountStore::GetInstance().GetAccount(toAddr, [this](const Account& toAccount){
    ... toAccount.GetBalance() ...
    ... etc.
});
```

PR #325, opened 2018-06-05
Tiny refactoring: Delete copy ctors and assignment operators for

digiront:
singletons


digiront:
@ckyang it's already fixed here https://github.com/Zilliqa/Zilliqa/pull/321/files#diff-68c3c9beb564258bfda9b2662a7d5d59R76

PR #326, opened 2018-06-06
simplifies Node::CheckState()

satoshi101:
Simplifies `Node::CheckState()`.


src/libNode/Node.cpp
```diff
@@ -246,265 +286,33 @@ bool Node::CheckState(Action action)
...
+              << " but receiving ERROR message";
+        LOG_GENERAL(WARNING, error.str());
+        return false;
+    }
+
+    error << "Doing " << ActionString(action) << " but already in "
+          << NodeStateString(m_state);
+    LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
+              error.str());
+
```
ckyang: LOG_XXX already support operator<< inside.
It's needless to use another ostringstream to do it and that is, we can use following:
LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(), "Doing " << ActionString(action) << " but already in " << NodeStateString(m_state));

satoshi101: Great tip, thanks!

src/libNode/Node.cpp
```diff
@@ -246,265 +286,31 @@ bool Node::CheckState(Action action)
...
-    default:
         LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
                   "Unrecognized action");
-        result = false;
-        break;
+        return false;
     }
 
-    return result;
+    if (m_state == ERROR)
```
digiront: I would move this to the top, as a first check

satoshi101: :D I've validated this and it's 100% same error messages with the original. If I move it it will print different error messages in some cases. I don't know if this change is important for people!

digiront: Your changes made this extremely hard to read code much clearer :). The error checking is still weird, I guess you did it to preserve the existing order of checks.

IMO nobody should depend on error logs to make any decision and this makes more sense:

```
bool Node::CheckState(Action action)
{
    if (m_mediator.m_ds->m_mode != DirectoryService::Mode::IDLE)
    {
        LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
                                               string("I am a DS node.") +
                                               string(" Why am I getting this message? Action: ") << ActionString(action));
        return false;
    }

    if (m_state == ERROR)
    {
        LOG_GENERAL(WARNING, "Doing " << ActionString(action) << " but received ERROR message");
        return false;
    }

    if ((action == PROCESS_TXNBODY) || (action >= NUM_ACTIONS))
    {
        LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(), "Unrecognized action");
        return false;
    }

    if (action == m_state)
    {
        LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
                                                "Doing " << ActionString(action) << " but already in "
                                                << NodeStateString(m_state));
        return false;
    }

    // might want to log if it returns false to type, assuming not
    return compatibleState(m_state, action);
}
```

satoshi101: Maybe you could confirm that what you posted has the same behavior as the original code. Specially the `action == m_state`.

> nobody should depend on error logs to make any decision
famous last words

In this commit, I do this transparent, zero-risk improvement. If people believe this isn't readable, they can do further commits in coordination with whoever operates the testnet. I don't want to change the logging. :)

digiront: The thing is the original code should have never been written like and since it's not a real production system, now it's the time to do such changes. 

Maybe somebody who actually reads the logs can comment.

digiront:
Comment out of the scope of this PR:

Looking at [this](https://github.com/satoshi101/Zilliqa/blob/3096f4dbb8dbce354e620ee2cefe3c0802a30183/src/libNode/Node.cpp#L738):

```
void Node::SetState(NodeState state)
{
    m_state = state;
              "Node State is now " << m_state << " at epoch "
                                   << m_mediator.m_currentEpochNum);
}
```

Setting a new state always succeeds. This means, it's the responsibility of the caller to verify the state transition. 

It's worth implementing a state machine that abstracts away the state transition complexity from the caller of `setState`. Then the caller just needs to know if the state was changed successfully.

src/libNode/Node.cpp
digiront: OK, my comments is out of the scope of this pr.

Raised this https://github.com/Zilliqa/Zilliqa/issues/340

Gnnng:
A note here: the same improvement can be applied in other `CheckState`

https://github.com/Zilliqa/Zilliqa/blob/dc6389e473862a14da3933d60efcc6d388c54971/src/libDirectoryService/DirectoryService.cpp#L100

https://github.com/Zilliqa/Zilliqa/blob/dc6389e473862a14da3933d60efcc6d388c54971/src/libConsensus/ConsensusLeader.cpp#L29

https://github.com/Zilliqa/Zilliqa/blob/dc6389e473862a14da3933d60efcc6d388c54971/src/libConsensus/ConsensusBackup.cpp#L28

bb111189:
With refactoring of checking state transition for node done,  we will also need to refactor it for directory service and consensus.

PR #330, opened 2018-06-11
getSmartContractCode API

sidthekidder:


src/libServer/Server.cpp
```diff
@@ -360,6 +360,42 @@ Json::Value Server::GetSmartContractState(const string& address)
...
+
+    try
+    {
+        Json::Value _json;
+        if (address.size() != ACC_ADDR_SIZE * 2)
+        {
+            _json["Error"] = "Address size inappropriate";
+            return _json;
+        }
+        vector<unsigned char> tmpaddr = DataConversion::HexStrToUint8Vec(address);
```
KaustubhShamshery: Can add a check for a contract account and return the error json.

PR #334, opened 2018-06-12
getSmartContractCode API

sidthekidder:
Previous PR got closed so opening a new one, added check for contract account

### DIFF
--- a/src/libServer/Server.cpp
+++ b/src/libServer/Server.cpp
@@ -360,6 +360,49 @@ Json::Value Server::GetSmartContractState(const string& address)
         return _json;
     }
 }
+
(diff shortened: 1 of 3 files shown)


src/libServer/Server.cpp
```diff
@@ -360,6 +360,49 @@ Json::Value Server::GetSmartContractState(const string& address)
         return _json;
     }
 }
+
+Json::Value Server::GetSmartContractCode(const string& address)
+{
+    LOG_MARKER();
+
+    try
```
digiront: Why is whole block in a `try` block?

sidthekidder: In case of any unforeseen error...similar to other APIs

PR #318, opened 2018-06-01
Smart-Contract Update v1

kaikawaliu:
Includes the updates for backend support of smart contract.

1. Added ContractStorage storage support for contract.
2. Major changes on Account and AccountStore.
3. API functions for RPC service
4. Configurations on Scilla

PR #343, opened 2018-06-18
Account delta

kaikawaliu:
1. Refactored AccountStore for introducing AccountStoreTemp which is to store the temporary account state for generating state delta.
2. Modified Microblock and TxBlock for validating state delta hash.
3. Add checkers for account state delta in both MicroBlock and FinalBlock consensus.
4. Modified the logic of checking whether all microblocks are received in the FinalBlock stage.
5. Temporarily added a sleep before Beginning next round transaction submission to wait for the whole network reaching to a synchronized state, TO FIX it by replacing with condition variable in another PR.


Gnnng:
🔗 #318 is required by this PR.

bb111189:
Depreciated

PR #345, opened 2018-06-19
Gossip refactoring

satoshi101:
Refactoring for gossip work

### DIFF
--- a/src/libNetwork/P2PComm.cpp
+++ b/src/libNetwork/P2PComm.cpp
@@ -42,6 +42,8 @@ const unsigned int HDR_LEN = 6;
 const unsigned int HASH_LEN = 32;
 const unsigned int BROADCAST_EXPIRY_SECONDS = 600;
 
+static const bool USE_GOSSIP = false;
+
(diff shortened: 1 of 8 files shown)


src/libNetwork/P2PComm.cpp
```diff
@@ -376,133 +525,97 @@ void P2PComm::HandleAcceptedConnection(
...
 
-        // Check if this message has been received before
-        bool found = false;
-        {
-            lock_guard<mutex> guard(
-                P2PComm::GetInstance().m_broadcastHashesMutex);
-            vector<unsigned char> msg_hash(hash_buf, hash_buf + HASH_LEN);
-            found = (P2PComm::GetInstance().m_broadcastHashes.find(msg_hash)
-                     != P2PComm::GetInstance().m_broadcastHashes.end());
-            // While we have the lock, we should quickly add the hash
```
satoshi101: This isn't that "quickly", I think. We hold the global lock - `m_broadcastHashesMutex` while we do blocking calls to read from the socket (`read()`).

Issue #347, opened 2018-06-19
Refactoring of state transition for libDirectoryService and libConsensus


bb111189:
PR #326 refactors the checking of the state transition for libNode. A further todo will be require for libDirectoryService and libConsensus,

bb111189:
This was recently implemented.

PR #349, opened 2018-06-21
Refactoring of CheckState for libDirectoryService and libConsensus

satoshi101:
Treats #347. 

`DirectoryService` equivalence proof: http://cpp.sh/8c55w
`ConsensusLeader` equivalence proof: http://cpp.sh/9xnxx


bb111189:
This was recently implemented in a seperate pr. Thanks.

PR #356, opened 2018-06-23
Support version information inside network message header

ckyang:
For each received network message, we should check version first to avoid potential attack.
Now we can define version number in the constans.xml and constants_local.xml, to sepcify the version information inside the transmitted message.


bb111189:
is this ready to review or still wip?

ckyang:
@bb111189 Ready to review

satoshi101:
> avoid potential attack

What type of attack? Whoever could form malicious messages before, should still be able to fake the message version number as far as I can see.

bb111189:
I think this PR does not address attack issue, but rather it allows us to easily drop network messages from older version. 

Spoofing of version number is still an issue, but that we should address it together with spoofing of any network messages in the network.

PR #357, opened 2018-06-23
Added and fixed issues for fetching SCs

jiayaoqijia:
Functions to fetch all SCs for an address.

### DIFF
--- a/src/libData/AccountData/Account.h
+++ b/src/libData/AccountData/Account.h
@@ -57,6 +57,8 @@ class Account : public Serializable
 public:
     Account();
 
+    //tells if the account is a contract accoun
+    bool isContract() const { return m_codeHash != h256(); }
     /// Constructor for loading account information from a byte stream.
     Account(const vector<unsigned char>& src, unsigned int offset);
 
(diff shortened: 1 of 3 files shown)


src/libData/AccountData/Account.h
```diff
@@ -57,6 +57,8 @@ class Account : public Serializable
 public:
     Account();
 
+    //tells if the account is a contract accoun
+    bool isContract() const { return m_codeHash != h256(); }
```
sidthekidder: Can remove this function as its already merged

jiayaoqijia: Fixed.

src/libServer/Server.cpp
```diff
@@ -408,9 +408,63 @@ string Server::GetStorageAt(const string& address, const string& position)
...
+            {
+                if (!contractAccount->isContract())
+                {
+                    continue;
+                }
+                Json::Value tmpJson;
+                tmpJson["address"] = contractAddr.hex();
+                tmpJson["state"] = contractAccount->GetStorageJson();
+                _json.append(tmpJson);
+            }
```
Gnnng: Probably we could try this
```
if (contractAccount == nullptr || !contractAccount->isContract())
{
    continue;
}

Json::Value tmpJson;
tmpJson["address"] = contractAddr.hex();
tmpJson["state"] = contractAccount->GetStorageJson();

_json.append(tmpJson);
```

PR #319, opened 2018-06-01
Merge PoW1 and PoW2 into one PoW

ckyang:
In this PR, PoW1 and PoW2 are merged into one PoW to simplify code flow.

* [ ] ~~Merged m_DSCommitteePubKeys and m_DSCommitteeNetworkInfo into one data structure, in order to avoid potential misused of these 2 variables. (https://github.com/Zilliqa/Issues/issues/48)~~ (#359 reopened to handle this separately)
* [ ] ~~Remove DSBlockChain/TxBlockChain/VCBlockChain and use one template.~~ (#435 created to handle this separately)
* [x] Remove all POW2 related file/function/state.
* [x] After DS consensus, don't call POW2 and directly start sharding process.
* [x] For new-joined node, it will call POW1.

Related bugs:

https://github.com/Zilliqa/Issues/issues/70
https://github.com/Zilliqa/Issues/issues/72

Follow-ups:

1. Create sharding block
https://github.com/Zilliqa/Issues/issues/33
1. DS and sharding consensus should be merged into one consensus
https://github.com/Zilliqa/Zilliqa/issues/350

PR #359, opened 2018-06-25
Merge m_DSCommitteePubKeys and  m_DSCommitteeNetworkInfo

ckyang:
As https://github.com/Zilliqa/Issues/issues/48 described, this PR merged m_DSCommitteePubKeys and m_DSCommitteeNetworkInfo into one data structure, in order to avoid potential misused of these 2 variables.

Ex. Update m_DSCommitteePubKeys but forget to update m_DSCommitteeNetworkInfo, lead to un-synchronized data structure and very hard to debug. After this PR, compiler can help to guard this kind of bug.


ckyang:
This PR already merged into https://github.com/Zilliqa/Zilliqa/pull/319, so I close this one.

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -1265,7 +1279,13 @@ bool ConsensusLeader::StartConsensus(const vector<unsigned char>& message,
     // Multicast to all nodes in the committee
     // =======================================
 
-    P2PComm::GetInstance().SendMessage(m_peerInfo, announcement);
+    deque<Peer> peer;
+
+    for (auto const& i : m_committee)
+        peer.push_back(i.second);
```
bb111189: should enclosed in {}

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -231,8 +230,7 @@ void DirectoryService::UpdateMyDSModeAndConsensusId()
                              << "] DSBK");
     }
     // Check if I am the oldest backup DS (I will no longer be part of the DS committee)
-    else if ((uint32_t)(m_consensusMyID + 1)
-             == m_mediator.m_DSCommitteeNetworkInfo.size())
+    else if ((uint32_t)(m_consensusMyID + 1) == m_mediator.m_DSCommittee.size())
```
bb111189: Not sure why the casting to uint32 is required here.

src/libDirectoryService/ViewChangePostProcessing.cpp
```diff
@@ -173,16 +172,15 @@ void DirectoryService::ProcessViewChangeConsensusWhenDone()
     unsigned int offsetToCandidateLeader = 1;
 
     Peer expectedLeader;
-    if (m_mediator.m_DSCommitteeNetworkInfo.at(offsetToCandidateLeader)
-        == Peer())
+    if (m_mediator.m_DSCommittee.at(offsetToCandidateLeader).second == Peer())
```
bb111189: while this does not relate directly to this pr,  maybe we can add an extra check before this. Check whether offsetToCandidateLeader < m_DSCommittee,size() to prevent out of bound issue. This will become an issue once we randomly pick a leader.

ansnunez: Since we haven't yet determined what the DS should do if the offset is somehow beyond the committee size, I propose we don't add a check yet if we can't add any corrective action along with it.

In any case, I have added a note about this in the random selection issue (https://github.com/Zilliqa/Issues/issues/103) so we don't forget this comment.

src/libDirectoryService/DSBlockPostProcessing.cpp
ansnunez: Without the cast, we get -Werror=sign-compare (I guess the arithmetic on LHS results in a signed value by default)

PR #361, opened 2018-06-25
Testnet mode

bb111189:
This PR support whitelisting of DS Nodes when testnet mode is enabled. Whitelisting is currently supported for PoW1. 

Todo:

- [x] To adjust internal scripts to support TEST_NET_MODE constants and whitelist.xml  
- [x] Shard node whitelisting (PoW2)
- [x] Testing


constants.xml
```diff
@@ -29,6 +29,7 @@
         <N_PREFILLED_PER_ACCOUNT>1000000</N_PREFILLED_PER_ACCOUNT>
         <NEW_NODE_POW_DELAY>10</NEW_NODE_POW_DELAY>
         <POST_VIEWCHANGE_BUFFER>10</POST_VIEWCHANGE_BUFFER>
+        <TEST_NET_MODE>false</TEST_NET_MODE>
```
Gnnng: I suggest putting this constant value into a different xml node (like `smart_contract`) as it's not numeric value.

src/libNetwork/Whitelist.cpp
```diff
@@ -0,0 +1,116 @@
...
+void Whitelist::AddToDSWhitelist()
+{
+    LOG_MARKER();
+
+    ifstream config("whitelist.xml");
+
+    if (!TEST_NET_MODE)
+    {
+        LOG_GENERAL(WARNING, "Not in testnet mode. Whitelisting not allowed");
+        return;
```
ckyang: config.close() should be called here because ifstream constructor would open file.
http://www.cplusplus.com/reference/fstream/ifstream/ifstream/

src/libNetwork/Whitelist.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    if (config.fail())
+    {
+        LOG_GENERAL(WARNING, "No whitelist xml present");
+        return;
```
ckyang: config.close() should be called here because ifstream constructor would open file.
http://www.cplusplus.com/reference/fstream/ifstream/ifstream/

src/libNetwork/Whitelist.cpp
bb111189: Updated. ifstream will not be called if not in TEST_NET_MODE

src/libNetwork/Whitelist.cpp
bb111189: On a deeper look, close() is not necessary here as it failed to open file. Calling close() will result in exception here.

ckyang: ok

src/libNetwork/Whitelist.cpp
Gnnng: Some different opinion here: according to C++ RAII, `.close()` is called when `ifstream` destructor is called. See the doc: http://en.cppreference.com/w/cpp/io/basic_ifstream/close. So it's redundant to make an explict call to `.close()`

src/libNetwork/Whitelist.cpp
Gnnng: It's not very necessary to call `close()` according to RAII.

src/libNetwork/Whitelist.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    using boost::property_tree::ptree;
+    ptree pt;
+    read_xml(config, pt);
+
+    lock_guard<mutex> g(m_mutexShardWhiteList);
+    m_ShardWhiteList.clear();
+
+    BOOST_FOREACH (ptree::value_type const& v, pt.get_child("address"))
```
Gnnng: Seems a simple C++11 auto-for will do. See https://github.com/Zilliqa/Zilliqa/blob/6973099893e2759beb46655996a35cf89c4424df/src/common/Constants.cpp#L53

src/libNetwork/Whitelist.cpp
```diff
@@ -0,0 +1,167 @@
...
+    {
+        return true;
+    }
+    else
+    {
+        LOG_GENERAL(INFO,
+                    "Not inside whitelist " << nodeNetworkInfo << " "
+                                            << nodePubKey);
+        return false;
+    }
```
Gnnng: perhaps put the `else` statement in the top-level?

```
    if (m_DSWhiteList.at(nodeNetworkInfo) == nodePubKey)
    {
        return true;
    } 
    LOG_GENERAL(INFO, ...);
    return false;
```

src/libNetwork/Whitelist.cpp
bb111189: good point @Gnnng

ckyang: A good coding style should always trying to keep open/close and constructor/destructor in pairs. In this case, since the constructor already implicit do "open", we only need to make sure destructor be called, thus I agree not calling any "close." Another approach is, we can choose default constructor (without any input argument) and make a explicit "open" call, then make some explicit "close" calls. In my personal opinion, I prefer to choose first one because it's more simple.

bb111189:
@ckyang  @Gnnng  PR is ready for one final pass.

src/libDirectoryService/PoW1Processing.cpp
```diff
@@ -211,7 +212,18 @@ bool DirectoryService::ParseMessageAndVerifyPOW1(
                 return false;
             }
 
-            m_allPoW1s.push_back(make_pair(key, nonce));
+            if (TEST_NET_MODE
```
KaustubhShamshery: Don't you think we should check if the DS belongs to whitelist before verifying the PoW submission by it ?

PR #364, opened 2018-06-26
Fix submit transaction

kaikawaliu:
1. Fixed the problem when a transaction comes from lookup will call UpdateAccountTemp twice.
2. Combine all the transaction submitted in one node into one message


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1459,6 +1459,7 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
 
     unsigned int cur_offset = offset;
 
+    // Initialize it with 255
     uint8_t shard_id = (uint8_t)-1;
```
Gnnng: Why not use 255 instead?

src/libValidator/Validator.cpp
```diff
@@ -205,7 +205,8 @@ bool Validator::CheckCreatedTransactionFromLookup(const Transaction& tx)
         return false;
     }
 
-    return AccountStore::GetInstance().UpdateAccountsTemp(
-        m_mediator.m_currentEpochNum, tx);
+    // return AccountStore::GetInstance().UpdateAccountsTemp(
+    //     m_mediator.m_currentEpochNum, tx);
+    return true;
```
Gnnng: Reason for such changes? Could have explained a bit in the comment

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: I have no idea...lol

src/libValidator/Validator.cpp
kaikawaliu: The transaction comes from lookup will also be processed in SubmitTransaction, where all the transaction will do UpdateAccountsTemp, remove this line to avoid it being processed twice.

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: Maybe the purpose is to initialize with the largest number in the valid range, despite from the size of the unsigned int

PR #366, opened 2018-06-26
 PoW submission signature verification and PoW mixhash verification

KaustubhShamshery:
* Added signature to PoW messages
* Signature verified by DS nodes
* Mixhash Verification
* [WIP] IP filter


bb111189:
@ckyang Check the mixhash verification part. Does it resolve your previous issue found during your investigation.

PR #368, opened 2018-06-27
Include Gas Price / Limit

KaustubhShamshery:
Include Gas Price / Limit in the parsing of txn


jiayaoqijia:
The commit has been included by PR #364

PR #373, opened 2018-06-27
Added IP Filter

KaustubhShamshery:
* 0.0.0.0 and 255.255.255 are considered invalid
* Added Support for disabling private IPs
* PoW1 and PoW2 submissions are checked to have valid IPs

Based on #361


src/libNetwork/Whitelist.cpp
```diff
@@ -165,4 +166,86 @@ bool Whitelist::IsPubkeyInShardWhiteList(const PubKey& nodePubKey)
...
+    }
+
+    if (!EXCLUDE_PRIV_IP)
+    {
+        // No filtering enable. Hence, IP (other than 0.0.0.0 and 255.255.255.255) is allowed.
+        return true;
+    }
+
+    lock_guard<mutex> g(m_mutexIPexclusion);
+    for (auto& ip_pair : m_IPexclusionRange)
```
ckyang: auto const &ip_pair

src/libZilliqa/Zilliqa.cpp
```diff
@@ -89,7 +89,7 @@ Zilliqa::Zilliqa(const std::pair<PrivKey, PubKey>& key, const Peer& peer,
     {
     case SyncType::NO_SYNC:
         LOG_GENERAL(INFO, "No Sync Needed");
-        Whitelist::GetInstance().UpdateDSWhitelist();
+        Whitelist::GetInstance().Init();
```
ckyang: How about calling Init() inside Whitelist constructor as default private IP exclusion list?
Thus the user do not need to call Init() by themselves.

src/libNetwork/Whitelist.cpp
KaustubhShamshery: noted

src/libZilliqa/Zilliqa.cpp
bb111189: This implementation is fine for now

PR #378, opened 2018-06-28
Heartbeat recovery for dead node

ckyang:
Construct heartbeat for each node.
For the dead DS/normal node (no heartbeat > specific time), rejoin it to network.


src/libMediator/Mediator.h
```diff
@@ -99,6 +99,16 @@ class Mediator
...
+
+    /// Initialize heart beat checking
+    void HeartBeat_Init();
+
+    /// Trigger heart beat
+    void HeartBeat();
+
+private:
+    std::mutex m_heartBeatMutex;
+    uint64_t m_heartBeatTime;
```
KaustubhShamshery: why is this a uint64_t int? I don't see this going beyond 3600

ckyang: ok, I will change to unsigned int

src/libMediator/Mediator.cpp
```diff
@@ -118,3 +119,44 @@ std::string Mediator::GetNodeMode(const Peer& peer)
...
+            if (m_heartBeatTime < heartBeatTimeout)
+            {
+                LOG_GENERAL(INFO, "Still alive...");
+                continue;
+            }
+
+            LOG_GENERAL(WARNING, "I am DEAD, rejoin to network");
+            (DirectoryService::Mode::IDLE == m_ds->m_mode)
+                ? m_node->RejoinAsNormal()
+                : m_ds->RejoinAsDS();
```
kaikawaliu: we cannot let a DS to rejoin as DS under this heartBeatTimeout, consider when this DS is going to be kicked out in the next DS epoch.

kaikawaliu: So we may let the DS rejoin as a normal node, and inform everyone he's no longer a DS when he tries to rejoin

PR #380, opened 2018-06-29
Don't wait for CommitTemp if my shard microblock not in finalblock

kaikawaliu:
1. To fix the bug that the shard without microblock in finalblock will be waiting for m_tempStateDeltaCommitted becomes true until forever if there are any microblock from other shard.
2. Also add a clean up for m_createdTransactions if a shard node becomes a DS Leader.

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -790,7 +790,6 @@ bool Node::IsMyShardsMicroBlockInFinalBlock(const uint256_t& blocknum)
         return false;
     }
 
-    lock_guard<mutex> g(m_mutexUnavailableMicroBlocks);
     auto it = m_unavailableMicroBlocks.find(blocknum);
     if (it == m_unavailableMicroBlocks.end())
     {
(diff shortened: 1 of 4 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -782,6 +783,37 @@ bool Node::IsMyShardsMicroBlockStateDeltaHashInFinalBlock(
                isEveryMicroBlockAvailable);
 }
 
+bool Node::IsMyShardsMicroBlockInFinalBlock(const uint256_t& blocknum)
+{
+    if (m_microblock == nullptr)
```
KaustubhShamshery: Should not you return true here ??

kaikawaliu: Why return true here? if no m_microblock ofcourse my shards not in finalblock

KaustubhShamshery: if no microblock, that means I did not submit a microblock. Just make sure the implementation works in case of no micro block consensus. Otherwise it is ok.

kaikawaliu: Yup, the purpose is just to make sure there is no microblock in finalblock is related to this node

PR #395, opened 2018-07-02
Use emplace & emplace_back as more as possible

ckyang:
## Description

Use emplace() instead of push_back()/insert() can improve efficiency and reduce memory stack usage, because this API would not create a temporary object when copying object into container.

Following syntax needs to be refined:

unurdered_map::insert() -> unurdered_map::emplace()
vector::push_back() -> vector::emplace_back()
vector::insert() -> vector::emplace()
deque::push_back() -> deque::emplace_back()
deque::push_front() -> deque::emplace_front()
deque::insert() -> deque::emplace()

https://github.com/Zilliqa/Issues/issues/87

### Implementation

- [ ] implement ...
- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test


bb111189:
Type casting can be remove here. 
https://github.com/Zilliqa/Zilliqa/blob/4510e5f69291b961f616320f24f1f7a274751642/src/libDirectoryService/FinalBlockPostProcessing.cpp#L340

bb111189:
https://github.com/Zilliqa/Zilliqa/blob/4510e5f69291b961f616320f24f1f7a274751642/src/libNode/Node.cpp#L146

https://github.com/Zilliqa/Zilliqa/blob/4510e5f69291b961f616320f24f1f7a274751642/src/libLookup/Lookup.cpp#L1669

ckyang:
This PR already merged into https://github.com/Zilliqa/Zilliqa/pull/378/, so I close this one.

ansnunez:
Reopened branch and PR so we can review and commit this particular update separately from #378

src/libData/BlockData/Block/TxBlock.cpp
```diff
@@ -116,7 +116,7 @@ int TxBlock::Deserialize(const vector<unsigned char>& src, unsigned int offset)
                 = GetNumber<uint32_t>(src, curOffset, sizeof(uint32_t));
             curOffset += sizeof(uint32_t);
 
-            m_shardIDs.emplace_back(shardID);
+            m_shardIDs.push_back(shardID);
```
ansnunez: Any reason why this can't be emplace_back?

ckyang: Please refer to this article: https://abseil.io/tips/112
There is always trade-off between performance & maintenance.
That is, for native data type, it doesn't matter to use push_back/emplace_back and we prefer to use push_back for good maintenance.
For user-defined data type, of course we prefer emplace_back for better performance.

PR #399, opened 2018-07-04
Combine response of MissingTxn in single message & Fix double lock & Clean Logs

kaikawaliu:
1. OnCommitFailure excessive logging 
2. Log shard leader 
3. Remove re-run of microblock consensus 
4. 02040000001 ← change to 1 byte
5. UpdateCoinBaseTemp excessive logging 
6. FindTxnInReceivedTxn - excessive logging 
7. Endless view change after waiting microblock times out
8. Use unsigned int instead of unsigned char for storing offsets for serialization and deserialization 
9. Logging refactoring


src/libNode/MicroBlockProcessing.cpp
```diff
@@ -402,6 +407,9 @@ bool Node::RunConsensusOnMicroBlockWhenShardLeader()
                                     << " m_consensusMyID: " << m_consensusMyID);
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
```
KaustubhShamshery: Can be one LOG_EPOCH statement ??

kaikawaliu: Noted

PR #402, opened 2018-07-05
Fix Shard Leader Ignoring CommitFailure

kaikawaliu:
1. Fix Ignore CommitFailure 
2. Wait all microblock received with timeout
3. Remove double lock for Rejoin by DetachFunction 
4. Typo 
5. Clean logs
6. Reduce mutex lock scope inside P2PComm::HandleAcceptedConnection, and make sure read-loop be executed at the same time


src/libNetwork/P2PComm.cpp
```diff
@@ -387,49 +391,66 @@ void P2PComm::HandleAcceptedConnection(
...
         bool found = false;
         {
-            lock_guard<mutex> guard(
-                P2PComm::GetInstance().m_broadcastHashesMutex);
             vector<unsigned char> msg_hash(hash_buf, hash_buf + HASH_LEN);
-            found = (P2PComm::GetInstance().m_broadcastHashes.find(msg_hash)
-                     != P2PComm::GetInstance().m_broadcastHashes.end());
+
+            {
+                lock_guard<mutex> guard(
```
lookfwd: Better make methods in `P2PComm` instead of using locks outside. (the original code also had the same issue)

PR #403, opened 2018-07-06
Make system more stable

ckyang:
1. Roll-back (enlarge) p2pcomm mutex lock scope
2. Replace FATAL log to WARNING log, in order to avoid node crash


Gnnng:
Also, may I know how do you roll it back? Normally when you do `git revert xxxxxx`, the new commit will automatically reference the old one in the commit message.

ckyang:
The reverted commit ID:
73a32d3fd66116232c1e67412ca5eb2e1c8e192b
da950ab519c324de7116ca7c0ecd85cbf540533e

PR #411, opened 2018-07-09
Use only 1 thread to handle to-be-removed broadcast messages

ckyang:
## Description
For broadcast messages which will be removed soon, currently we create threads for each message.
This would result in the overwhelmed thread usage.
In this PR, we use only 1 thread to handle this.

### Implementation

- [x] **ready for review**

- [x] add unit test
- [x] local machine test
- [x] small-scale cloud test


src/libNetwork/P2PComm.cpp
```diff
@@ -68,7 +67,33 @@ static void close_socket(int* cli_sock)
...
+            this_thread::sleep_for(chrono::seconds(BROADCAST_INTERVAL));
+            lock_guard<mutex> guard(m_broadcastToRemovedMutex);
+
+            for (unsigned int i = 0; i < BROADCAST_INTERVAL; ++i)
+            {
+                auto it = m_broadcastToRemoved.find(time(nullptr)
+                                                    - BROADCAST_EXPIRY - i);
+
+                if (m_broadcastToRemoved.end() == it)
+                    continue;
```
bb111189: Enclosed with {}

src/libNetwork/P2PComm.cpp
```diff
@@ -68,7 +67,36 @@ static void close_socket(int* cli_sock)
...
+                auto it = m_broadcastToRemoved.find(time(nullptr)
+                                                    - BROADCAST_EXPIRY - i);
+
+                if (m_broadcastToRemoved.end() == it)
+                {
+                    continue;
+                }
+
+                auto it2 = m_broadcastHashes.find(it->second);
+                m_broadcastHashes.erase(m_broadcastHashes.begin(), it2);
```
Gnnng: Although we already had a discussion, I'd like to put the comment here as a reminder: `m_broadcastHashes` stores the hashes by their alphabetical order other than insertion time, so the erase operation here may not work as expected (which should remove the hash entries since the beginning until `it2`)

src/libNetwork/P2PComm.cpp
```diff
@@ -68,7 +67,36 @@ static void close_socket(int* cli_sock)
...
+        while (true)
+        {
+            this_thread::sleep_for(chrono::seconds(BROADCAST_INTERVAL));
+            lock(m_broadcastToRemovedMutex, m_broadcastHashesMutex);
+            lock_guard<mutex> g(m_broadcastToRemovedMutex, adopt_lock);
+            lock_guard<mutex> g2(m_broadcastHashesMutex, adopt_lock);
+
+            for (unsigned int i = 0; i < BROADCAST_INTERVAL; ++i)
+            {
+                auto it = m_broadcastToRemoved.find(time(nullptr)
```
KaustubhShamshery: I am a little concerned about the typecasting here. For eg. what would happen if BROADCAST_EXPIRTY + i > time(nullptr) ?

src/libNetwork/P2PComm.cpp
```diff
@@ -68,7 +67,46 @@ static void close_socket(int* cli_sock)
     }
 }
 
-P2PComm::P2PComm() {}
+static bool mycomp(const std::pair<std::vector<unsigned char>, time_t>& a,
```
bb111189: what is the meaning of mypool?

src/libNetwork/P2PComm.cpp
```diff
@@ -68,7 +67,46 @@ static void close_socket(int* cli_sock)
...
+            lock_guard<mutex> g2(m_broadcastHashesMutex, adopt_lock);
+
+            if (m_broadcastToRemoved.empty()
+                || m_broadcastToRemoved[0].second
+                    > time(nullptr) - BROADCAST_EXPIRY)
+            {
+                continue;
+            }
+
+            std::vector<unsigned char> tmp;
```
bb111189: good to give a better name for this vector

src/libNetwork/P2PComm.cpp
```diff
@@ -68,7 +67,46 @@ static void close_socket(int* cli_sock)
...
-P2PComm::P2PComm() {}
+static bool mycomp(const std::pair<std::vector<unsigned char>, time_t>& a,
+                   const std::pair<std::vector<unsigned char>, time_t>& b)
+{
+    return a.second < b.second;
+}
+
+P2PComm::P2PComm()
+{
+
```
bb111189: Remove the extra line

Issue #414, opened 2018-07-10
Finalblock processed earlier than microblock announcement


kaikawaliu:
Due to network condition, this may happen and cause the finalblock being dropped.
Proper logic should be implemented to handle this situation.
Thus we need to solidify how finalblock should be sync with the microblock consensus.

bb111189:
Implemented.

PR #415, opened 2018-07-10
Solidating Synchronization between Microblock and Finalblock

kaikawaliu:
## Description

Implement the logic for handling the out of order issue between finalblock and microblock

Context: current master branch

Related issue: https://github.com/Zilliqa/Zilliqa/issues/414
## Review Suggestion

Look at the implementation of condition variable first.
Review the handling of each situations for the shard backup node.

### Implementation

TODOs
- [x] Add condition variable in ProcessFinalBlock for waiting Microblock Consensus done
- [x] Handle logic for shard leader when the cv times out (do nothing)
- [x] Handle logic for shard backup when the cv times out if it found its microblock in finalblock (rejoin)
- [x] Handle logic for shard backup when cv times out, microblock not found in finalblock, but shard ID in finalblock (Rejoin)
- [x] Handle logic for shard backup when cv timesout, microblock not found in finalblock, neither shard ID is in finalblock (do Nothing)
- [X] Consolidate shard_id to uint32_t
- [x] Add shard ID in microblock
- [x] Add checker of shard leader pubkey when DS received shard ID in microblock
- [x] Add list of shard IDs in finalblock
- [x] Load shard ID in ProcessFinalBlock
- [x] Add searching function for shard ID

- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1518,7 +1572,7 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
     unsigned int cur_offset = offset;
 
     // Initialize it with 255
```
Gnnng: The comment is out of sync with the code. Better remove it or reflect the change.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1518,7 +1572,7 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
...
-    uint8_t shard_id = (uint8_t)-1;
+    uint32_t shard_id = (uint32_t)-1;
```
Gnnng: I did some search over the codebase and it turns out that this the only place we assign a maximum value like this. Can we have a try of [numeric_limits](http://en.cppreference.com/w/cpp/types/numeric_limits) and replace the use of this?

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -86,10 +88,10 @@ void DirectoryService::ExtractDataFromMicroblocks(
         if (!isVacuousEpoch && !isEmpty)
         {
             m_mediator.m_node->m_unavailableMicroBlocks[blockNum].insert(
-                {{microBlock.GetHeader().GetTxRootHash(),
-                  microBlock.GetHeader().GetStateDeltaHash()},
+                {{{microBlock.GetHeader().GetTxRootHash(),
+                   microBlock.GetHeader().GetStateDeltaHash()}},
                  // {!isEmptyTxn, true}});
-                 {false, true}});
+                 vector<bool>{false, true}});
```
ckyang: The casting might be redundant because there is no casting in line 981. It's better to make the code be consistent.

src/libLookup/Synchronizer.cpp
```diff
@@ -86,11 +86,12 @@ TxBlock Synchronizer::ConstructGenesisTxBlock()
...
-    return TxBlock(
-        TxBlockHeader(TXBLOCKTYPE::FINAL, BLOCKVERSION::VERSION1, 1, 1,
-                      BlockHash(), 0, 151384616955606, TxnHash(), StateHash(),
-                      StateHash(), 0, 5, keypair.second, 0, BlockHash()),
-        vector<bool>(1), vector<MicroBlockHashSet>(5), CoSignatures());
+    return TxBlock(TxBlockHeader(TXBLOCKTYPE::FINAL, BLOCKVERSION::VERSION1, 1,
+                                 1, BlockHash(), 0, 151384616955606, TxnHash(),
+                                 StateHash(), StateHash(), 0, 5, keypair.second,
+                                 0, BlockHash()),
+                   vector<bool>(1), vector<MicroBlockHashSet>(5),
```
ckyang: Should use "true" instead of "1" into bool container to avoid ambiguous though they're identical in almost all compiler.

tests/Lookup/Test_LookupNodeForTxBlock.cpp
```diff
@@ -122,11 +122,12 @@ BOOST_AUTO_TEST_CASE(testTxBlockStoring)
...
-    TxBlock txblock(
-        TxBlockHeader(TXBLOCKTYPE::FINAL, BLOCKVERSION::VERSION1, 1, 1,
-                      BlockHash(), 0, get_time_as_int(), TxnHash(), StateHash(),
-                      StateHash(), 0, 5, pubKey1.second, 0, BlockHash()),
-        vector<bool>(1), vector<MicroBlockHashSet>(5), CoSignatures());
+    TxBlock txblock(TxBlockHeader(TXBLOCKTYPE::FINAL, BLOCKVERSION::VERSION1, 1,
+                                  1, BlockHash(), 0, get_time_as_int(),
+                                  TxnHash(), StateHash(), StateHash(), 0, 5,
+                                  pubKey1.second, 0, BlockHash()),
+                    vector<bool>(1), vector<MicroBlockHashSet>(5),
```
ckyang: Should use "true" instead of "1" into bool container to avoid ambiguous though they're identical in almost all compiler.

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: Noted

Issue #417, opened 2018-07-11
Remove Transaction Hashes in Microblock when sending to DS Committee


kaikawaliu:
These hashes is never used in DS but takes a lot of resource in transmission.

Gnnng:
test comment

ansnunez:
Fixed in #418

PR #418, opened 2018-07-11
Remove TxHashes in Microblock sending to DS

kaikawaliu:
## Description

goal: Remove TxHashes in Microblock sending to DS to decrease the overall network consumption

context: https://github.com/Zilliqa/Zilliqa/pull/415

issue: https://github.com/Zilliqa/Zilliqa/issues/417

### Implementation

- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test

### DIFF
--- a/src/libData/BlockData/Block/MicroBlock.cpp
+++ b/src/libData/BlockData/Block/MicroBlock.cpp
@@ -26,14 +26,9 @@ using namespace boost::multiprecision;
 unsigned int MicroBlock::Serialize(vector<unsigned char>& dst,
                                    unsigned int offset) const
 {
-    if (m_header.GetNumTxs() != m_tranHashes.size())
-    {
-        LOG_GENERAL(WARNING,
-                    "assertion failed (" << __FILE__ << ":" << __LINE__ << ": "
-                                         << __FUNCTION__ << ")");
-    }
+    offset += SerializeCore(dst, offset);
 
-    unsigned int size_needed = GetSerializedSize();
+    unsigned int size_needed = GetSerializedTxnHashesSize();
 
     unsigned int size_remaining = dst.size() - offset;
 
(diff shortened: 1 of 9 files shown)


src/libData/BlockData/Block/MicroBlock.cpp
```diff
@@ -109,6 +109,61 @@ unsigned int MicroBlock::GetSerializedSize() const
...
+        BlockBase::Deserialize(src, curOffset);
+    }
+    catch (const std::exception& e)
+    {
+        LOG_GENERAL(WARNING,
+                    "Error with MicroBlock::Deserialize." << ' ' << e.what());
+        return -1;
+    }
+    return 0;
+}
```
ckyang: SerialzeMin/DeserialzeMin is used to parsing header, and should be re-used by Serialze/Deserialze, in order to reduce duplicated code.

src/libData/BlockData/Block/MicroBlock.cpp
```diff
@@ -26,14 +26,11 @@ using namespace boost::multiprecision;
...
-        LOG_GENERAL(WARNING,
-                    "assertion failed (" << __FILE__ << ":" << __LINE__ << ": "
-                                         << __FUNCTION__ << ")");
-    }
+    unsigned int size_core = SerializeCore(dst, offset);
+
+    offset += size_core;
 
-    unsigned int size_needed = GetSerializedSize();
+    unsigned int size_needed = GetSerializedTxnHashesSize();
```
ansnunez: Even though the tx hashes are no longer added to the serialized message, it looks like we're still resizing the destination buffer to include the size of the hashes, and then also iterating through the hashes without actually doing anything to them.  Perhaps we can get rid of this part already?

src/libNode/Node.h
```diff
@@ -418,8 +425,9 @@ class Node : public Executable, public Broadcastable
 
     // Transaction body sharing variables
     std::mutex m_mutexUnavailableMicroBlocks;
-    std::unordered_map<boost::multiprecision::uint256_t,
-                       std::unordered_map<MicroBlockHashSet, std::vector<bool>>>
+    std::unordered_map<
```
ansnunez: I easily forget what is the meaning of the two bools. Can you add some explanation here?

kaikawaliu: This is used to mark down what are the forwarding data this node is supposed to receive for this epoch. Once it receives all the forwarding data, this map will be cleaned, thus it can proceed to the next round of consensus.

PR #419, opened 2018-07-11
Resolve microblock consensus message being processed in an out of order manner

bb111189:
## Description

This PR resolve an issue where consensus message can arrive in an out of order fashion. An previous fix (#410) was found to cause temporary deadlock issue. In this PR, we preprocess the consensus and block (using conditional variable) it if it is out of order.

### Implementation

- [ ] Add consensus preprocess to other consensus (ds block, sharding structure, final block and view change block). This will be implement in a seperate PR. 
- [x] **ready for review**

- [x] local machine test
- [x] medium-scale cloud test


src/libConsensus/ConsensusUser.cpp
```diff
@@ -165,6 +165,22 @@ bool ConsensusUser::ProcessConsensusMessage(
 {
     LOG_MARKER();
 
+    std::shared_lock<shared_timed_mutex> cv_lk(m_mutexProcessConsensusMessage);
+    if (cv_processConsensusMessage.wait_for(
+            cv_lk, std::chrono::seconds(10), [this, message, offset]() -> bool {
```
ckyang: Please move the timeout "10" second to constant file

src/libNode/MicroBlockProcessing.cpp
```diff
@@ -89,6 +89,22 @@ bool Node::ProcessMicroblockConsensus(const vector<unsigned char>& message,
 #ifndef IS_LOOKUP_NODE
     LOG_MARKER();
 
+    std::shared_lock<shared_timed_mutex> cv_lk(m_mutexProcessConsensusMessage);
+    if (cv_processConsensusMessage.wait_for(
+            cv_lk, std::chrono::seconds(10), [this, message, offset]() -> bool {
```
ckyang: Please move the timeout "10" second to constant file

src/libConsensus/ConsensusUser.h
```diff
@@ -38,6 +39,9 @@ class ConsensusUser : public Executable, public Broadcastable
     bool m_leaderOrBackup; // false = leader, true = backup
     std::shared_ptr<ConsensusCommon> m_consensus;
 
+    std::shared_timed_mutex m_mutexProcessConsensusMessage;
```
ckyang: Why use shared_timed_mutex to only limit writer without reader?

bb111189: To allow other threads to use m_consensusObject for read purposes (in this case, using CanProcessMessage() ).  But it is still vital to ensure m_consensusObject data is not overwritten (eg reset).

ckyang: noted.

bb111189:
This PR will require fixes from PR #415 which fixes the issue of processing final block after receiving microblock cosig2

PR #422, opened 2018-07-13
Feature/gpu mining - Add constant value to control full dataset mining

ShengguangXiao:
## Description

Expose the full dataset mining for user.

Add a const value FULL_DATASET_MINE so user can enable the full dataset in mining.

PoW mining shouldn't use light client. Should use full #92 
https://github.com/Zilliqa/Issues/issues/92

### Implementation

- [ ] implement ...
- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test
Tested


bb111189:
Can you fix the merge conflicts?

Gnnng:
@ShengguangXiao The commits you made seem to have some issue with the commit author information. This [link](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/) might be helpful to you.

constants.xml
```diff
@@ -37,6 +37,7 @@
         <TXN_SUBMISSION>5</TXN_SUBMISSION>
         <TXN_BROADCAST>10</TXN_BROADCAST>
         <DEBUG_LEVEL>3</DEBUG_LEVEL>
+        <FULL_DATASET_MINE>0</FULL_DATASET_MINE>
```
ckyang: This option should belongs to "switch", so I suggest to move to <options> part.

constants_local.xml
```diff
@@ -37,6 +37,7 @@
         <TXN_SUBMISSION>5</TXN_SUBMISSION>
         <TXN_BROADCAST>10</TXN_BROADCAST>
         <DEBUG_LEVEL>3</DEBUG_LEVEL>
+        <FULL_DATASET_MINE>0</FULL_DATASET_MINE>
```
ckyang: This option should belongs to "switch", so I suggest to move to part.

ShengguangXiao: In the future, there should be only full dataset mining, it is not an option for user.

tests/POW/test_POW.cpp
```diff
@@ -706,7 +706,7 @@ BOOST_AUTO_TEST_CASE(mining_and_verification)
     uint8_t difficultyToUse = 10;
     uint8_t blockToUse = 0;
     ethash_mining_result_t winning_result = POWClient.PoWMine(
-        blockToUse, difficultyToUse, rand1, rand2, ipAddr, pubKey, false);
+        blockToUse, difficultyToUse, rand1, rand2, ipAddr, pubKey, true);
```
KaustubhShamshery: This is mining full but verifying light . Do we need to enable some more test ? @ckyang what are your thoughts ?

ShengguangXiao: The mining should always be full, otherwise it is too easy to get the result. The verify should use light because every client can verify the result quickly.
I agree we need to add more auto test case for it, but it will be after the GPU mining is ready.

KaustubhShamshery: The issue is that if you verifyLight, it does not enforce full mining i.e. verifyLight would be true even if you mined light or mined full.

ShengguangXiao: Hi Kaustubh, sorry for delayed reply. This is a good question, i also don't have the answer yet. Need to discuss with the team how to enforce full mining in the future. @Gnnng @bb111189 @jiayaoqijia @Gnnng

KaustubhShamshery: Yes, just trying to bring it to your attention

ShengguangXiao: @KaustubhShamshery @bb111189 @Gnnng @jiayaoqijia After i do some investigation and test, i find actually light dataset mine is much slower than full dataset mine. Because light dataset mine use less memory, every ethash access need to regenerate dag data, it takes a lot of time. I tested with PoW difficulty 15, light mine takes 19 seconds, full dataset mine only 9 seconds. So in the long run, people will not consider using light dataset mine.

PR #424, opened 2018-07-13
Fix circular array

kaikawaliu:
## Description

Recover the m_size and implement m_index in CircularArray to keep track of the size of blocks in chain at meanwhile guarantees the correctness of using back() when it have inconsistency in the array.

based-on pr: https://github.com/Zilliqa/Zilliqa/pull/420

### Implementation

- [X] **ready for review**

- [X] local machine test
- [ ] small-scale cloud test

### DIFF
--- a/src/libConsensus/ConsensusCommon.cpp
+++ b/src/libConsensus/ConsensusCommon.cpp
@@ -23,7 +23,38 @@
 #include "libUtils/DataConversion.h"
 #include "libUtils/Logger.h"
 
+#define LITERAL(s) #s
+
 using namespace std;
+
+map<ConsensusCommon::ConsensusErrorCode, std::string>
+    ConsensusCommon::CONSENSUSERRORMSG
+    = {{NO_ERROR, LITERAL(NO_ERROR)},
+       {GENERIC_ERROR, LITERAL(GENERIC_ERROR)},
+       {INVALID_DSBLOCK, LITERAL(INVALID_DSBLOCK)},
+       {INVALID_MICROBLOCK, LITERAL(INVALID_MICROBLOCK)},
+       {INVALID_FINALBLOCK, LITERAL(INVALID_FINALBLOCK)},
+       {INVALID_VIEWCHANGEBLOCK, LITERAL(INVALID_VIEWCHANGEBLOCK)},
+       {INVALID_DSBLOCK_VERSION, LITERAL(INVALID_DSBLOCK_VERSION)},
+       {INVALID_MICROBLOCK_VERSION, LITERAL(INVALID_MICROBLOCK_VERSION)},
+       {INVALID_FINALBLOCK_VERSION, LITERAL(INVALID_FINALBLOCK_VERSION)},
+       {INVALID_FINALBLOCK_NUMBER, LITERAL(INVALID_FINALBLOCK_NUMBER)},
+       {INVALID_PREV_FINALBLOCK_HASH, LITERAL(INVALID_PREV_FINALBLOCK_HASH)},
+       {INVALID_VIEWCHANGEBLOCK_VERSION,
+        LITERAL(INVALID_VIEWCHANGEBLOCK_VERSION)},
+       {INVALID_TIMESTAMP, LITERAL(INVALID_TIMESTAMP)},
+       {INVALID_BLOCK_HASH, LITERAL(INVALID_BLOCK_HASH)},
+       {INVALID_MICROBLOCK_ROOT_HASH, LITERAL(INVALID_MICROBLOCK_ROOT_HASH)},
+       {MISSING_TXN, LITERAL(MISSING_TXN)},
+       {FINALBLOCK_MISSING_HASH, LITERAL(FINALBLOCK_MISSING_HASH)},
+       {FINALBLOCK_INVALID_MICROBLOCK_ROOT_HASH,
+        LITERAL(FINALBLOCK_INVALID_MICROBLOCK_ROOT_HASH)},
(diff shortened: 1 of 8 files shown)


src/libData/DataStructures/CircularArray.h
```diff
@@ -85,7 +92,7 @@ template<class T> class CircularArray
             LOG_GENERAL(WARNING, "m_array is empty")
             throw;
         }
-        return m_array[(int)((m_size - 1) % m_capacity)];
+        return m_array[(int)((m_index - 1) % m_capacity)];
```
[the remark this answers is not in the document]
kaikawaliu: Noted

src/libData/DataStructures/CircularArray.h
```diff
@@ -28,8 +28,13 @@ template<class T> class CircularArray
     std::vector<T> m_array;
 
     int m_capacity;
+
+    /// return the actual size of how many blocks being stored in the array
     boost::multiprecision::uint256_t m_size;
 
+    /// return the index of the latest block inserted
+    boost::multiprecision::uint256_t m_index;
```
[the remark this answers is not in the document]
kaikawaliu: Noted

src/libData/DataStructures/CircularArray.h
```diff
@@ -73,7 +81,8 @@ template<class T> class CircularArray
             LOG_GENERAL(WARNING, "m_array is empty")
             throw;
         }
-        m_array[(int)(index % m_capacity)] = element;
+        m_index = (int)index % m_capacity;
+        m_array[m_index] = element;
         m_size++;
```
digiront: Although this is not exposed, the size can't keep increasing. Should be `std::max(m_size++, m_capacity)`

KaustubhShamshery: The usage of this variable is such that it tells the total size of the elements in the array, it would cause some internal issues if we do that.

satoshi101: > would cause some internal issues if we do that

So true! I don't know if what's there right now works. Highly doubt it...  But @digiront - did you see [this](https://github.com/Zilliqa/Zilliqa/blob/f0d8d388fe907dd73b85a7b8e329cc7872da92db/src/libData/BlockChainData/DSBlockChain.cpp#L49), [this](https://github.com/Zilliqa/Zilliqa/blob/f0d8d388fe907dd73b85a7b8e329cc7872da92db/src/libData/BlockChainData/TxBlockChain.cpp#L49) and [this](https://github.com/Zilliqa/Zilliqa/blob/dc6389e473862a14da3933d60efcc6d388c54971/src/libData/BlockChainData/VCBlockChain.cpp#L43) (copies)? What about [this](https://github.com/Zilliqa/Zilliqa/blob/9f2aef617ca8ba0cfd8bd361eed63cfd268d122b/src/libServer/Server.cpp#L832) and [this](https://github.com/Zilliqa/Zilliqa/blob/9f2aef617ca8ba0cfd8bd361eed63cfd268d122b/src/libServer/Server.cpp#L939) or [this](https://github.com/Zilliqa/Zilliqa/blob/9f2aef617ca8ba0cfd8bd361eed63cfd268d122b/src/libServer/Server.cpp#L1003)?! It's so funny!! So effectively they have your 'max' suggestion - but they just have it outside the class (in the latter case) and not in the first one. Likely they use this as a circular buffer and something fancy - different - like a thing where they put known blocks and if they've seen more than expected (even if some of them are duplicates overriding the old values - a bug) something happens.

kaikawaliu: This circular array here is not just a circular buffer. It's designed to be able to receive element with inconsistent index, thus insert at the responding position, and be able to returns the element in a desired index. At meanwhile, it will recycle the space for the latest `m_capacity` interested insertion.

satoshi101: I'm sorry. This implementation you have is not correct, though, even with what you just said. :( It's better than it used to be but it still doesn't work.

satoshi101: Those are simple stuff - we shouldn't talk about them, compromise etc. They can easily be 100% correct.

src/libData/DataStructures/CircularArray.h
```diff
@@ -73,7 +81,8 @@ template<class T> class CircularArray
             LOG_GENERAL(WARNING, "m_array is empty")
```
satoshi101: I would add the question marks after all those `LOG_GENERAL`.

tests/Data/Test_CircularArray.cpp
```diff
@@ -48,7 +48,7 @@ BOOST_AUTO_TEST_CASE(CircularArray_test)
 
     for (int i = 0; i < 100; i++)
     {
-        arr.push_back(11);
+        arr.insert_new(arr.size(), 11);
     }
 
     BOOST_CHECK_MESSAGE(arr[101] == 11, "arr[101] != 2!");
```
satoshi101: Should be: `"arr[101] == 11"`

src/libData/DataStructures/CircularArray.h
digiront: > So effectively they have your 'max' suggestion - but they just have it outside the class

I guess we can all agree this should be inside the class.

> Likely they use this as a circular buffer and something fancy

That begs for bugs to appear, I think we can also agree to refactor a little bit to make the special usage obvious.

KaustubhShamshery:
@kaikawaliu it is fine to close this one right?? I think already merged some part of this one.

PR #426, opened 2018-07-13
Fetch missing txns and re-run consensus (reprocess announcement) 

bb111189:
## Description

When a shard backup receive announcement from shard leader, a shard backup may find that it is missing some txns in the microblock (announcement). The previously implementation will fetch all the missing txns from the leader and set consensus to error state. 

In this implementation, we extends the implementation. Recall that the backup has request missing txn from the shard leader and consensus is at error state. Upon this state, the backup will block (with timeout of 10s) to wait for the missing txns. If it received the missing transactions, it will set it consensus state back to INITIAL and rerun processMessageAnnouncement again. Since it have all the txns, it should be able to process Announcement, Cosig1 and Cosig2. The later 2, is currently being blocked by other CV as describe in PR #419 

## Review Suggestion

This PR is build on top of PR #425  which implements consensus error number. Reviewer is suggested to review and approve #425 before this PR. 

## Status
Testing, in progress. 

### Implementation

- [ ] implement ...
- [ ] **ready for review**

- [ ] local machine test
- [ ] small-scale cloud test


src/libNode/MicroBlockProcessing.cpp
```diff
@@ -195,11 +195,58 @@ bool Node::ProcessMicroblockConsensus(const vector<unsigned char>& message,
...
+            LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
+                      "Oops, no consensus reached - consensus error. "
+                      "error number: "
+                          << (m_consensusObject->GetConsensusErrorCode())
+                          << " error message: "
+                          << (m_consensusObject->GetConsensusErrorMsg()));
+
+            // Block till txn is fetched
+            unique_lock<mutex> lock(m_mutexCVMicroBlockMissingTxn);
+            if (cv_MicroBlockMissingTxn.wait_for(
```
ansnunez: Should notify be called on cv_MicroBlockMissingTxn somewhere?

bb111189: yep. I also notice a notify is missing

PR #432, opened 2018-07-18
While loop keep searching for prefill txn

kaikawaliu:
## Description

Avoid `SubmitTransaction` keep looping for valid transaction to submit endlessly.

### Implementation

- [X] implement ...
- [X] **ready for review**

- [x] local machine test
- [ ] small-scale cloud test

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -48,9 +48,11 @@
         <FULL_DATASET_MINE>false</FULL_DATASET_MINE>
     </options>
     <smart_contract>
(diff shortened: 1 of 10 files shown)


Gnnng:
Won't this break the loop?https://github.com/Zilliqa/Zilliqa/blob/24c8b97bf56c9ca38cea2c08e7eac749a481cd5d/src/libNode/Node.cpp#L838-L841

PR #433, opened 2018-07-18
Fix race condition when using m_consensusObject->CanProcessMessage(message, offset) and add atomic to m_state in consensusCommon

bb111189:
## Description

In PR #419 and #428, `m_consensusObject->CanProcessMessage(message, offset)` was added before each consensus message to check whether the ordering of the consensus message is correct or wrong. However, race condition (when using `m_consensusObject`) may occurs. 

One example will be for `libNode`. When the conditional variable `cv_processConsensusMessage` timeout, depending on the situation, a node may rejoin as a new node. During this period of time, `m_consensusObject` will be reset. However, there may be more threads (in the threadpool) that will use `m_consensusObject` in the near future. There is currently no such issue for `libDirectoryService`.

This PR fixes the race condition by introducing a mutex to protect `m_consensusObject`. In addtion, for `libNode`, an additional check was added to check whether is the node in the process of rejoining. If true, it will ignore the rest of the consensus message. 

## Review Suggestion

This PR does not catered for the case when m_consensusObject is reset to prepare for the next round of consensus. 

## Status
Ready for review 

### Implementation

- [x] use atomic for m_state in consensusCommon
- [x] Fix race condition when using m_consensusObject
- [x] **ready for review**

- [x] local machine test
- [x] medium-scale cloud test

### DIFF
--- a/src/libNode/MicroBlockProcessing.cpp
+++ b/src/libNode/MicroBlockProcessing.cpp
@@ -581,25 +581,29 @@ bool Node::RunConsensusOnMicroBlock()
         lock_guard<mutex> g2(m_mutexNewRoundStarted);
         m_newRoundStarted = false;
     }
-
-    if (m_isPrimary == true)
     {
-        if (!RunConsensusOnMicroBlockWhenShardLeader())
+        lock_guard<mutex> lock(m_mutexConsensus);
+        if (m_isPrimary == true)
         {
-            LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
-                      "Error at RunConsensusOnMicroBlockWhenShardLeader");
-            // throw exception();
-            return false;
+            if (!RunConsensusOnMicroBlockWhenShardLeader())
+            {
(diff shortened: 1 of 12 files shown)


src/libNode/MicroBlockProcessing.cpp
```diff
@@ -93,6 +93,15 @@ bool Node::ProcessMicroblockConsensus(const vector<unsigned char>& message,
...
             [this, message, offset]() -> bool {
+                lock_guard<mutex> g(m_mutexConsensus);
+                if (m_mediator.m_lookup->m_syncType != SyncType::NO_SYNC)
+                {
+                    LOG_GENERAL(WARNING,
+                                "The node started the process of rejoining, "
+                                "Ignore rest of "
+                                "consensus msg.")
+                    return false;
+                }
```
kaikawaliu: I think we should add this to every cases of consensus. In case rejoin as DS was called in the future

bb111189: Implemented

PR #435, opened 2018-07-18
Create single template class for all blockchain types

ansnunez:
## Description

Replaced DSBlockChain/TxBlockChain/VCBlockChain with template class BlockChain

This change was initially part of #319 but moved here for faster review and merging.

## Review Suggestion

This is basically a refactoring of the code inside DSBlockChain.cpp, TxBlockChain.cpp, and VCBlockChain.cpp, so you can compare the contents of those against what's inside BlockChain.h.

Some review has also been done on these changes inside #319.

### Implementation

- [x] implement ...
- [x] **ready for review**

- [x] local machine test


bb111189:
While it is a template class, it looks awkward to have 
GetTxBlock()
GetDSBlock()
GetVCBlock()

src/libData/BlockChainData/BlockChain.h
```diff
@@ -0,0 +1,176 @@
...
+        return m_blocks[blockNum];
+    }
+
+    DSBlock GetDSBlock(const boost::multiprecision::uint256_t& blockNum)
+    {
+        lock_guard<mutex> g(m_mutexBlocks);
+
+        if (blockNum >= m_blocks.size())
+        {
+            throw "Blocknumber Absent";
```
bb111189: I think throwing a string is not an ideal way to handle this. Maybe we can throw nullptr or log warning or log fatal in this case

src/libData/BlockChainData/BlockChain.h
```diff
@@ -0,0 +1,176 @@
...
+        return m_blocks[blockNum];
+    }
+
+    VCBlock GetVCBlock(const boost::multiprecision::uint256_t& blockNum)
+    {
+        lock_guard<mutex> g(m_mutexBlocks);
+
+        if (blockNum >= m_blocks.size())
+        {
+            throw "Blocknumber Absent";
```
bb111189: I think throwing a string is not an ideal way to handle this. Maybe we can throw nullptr or log warning or log fatal in this case

src/libData/BlockChainData/BlockChain.h
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /// Returns the block at the specified block number.
+    TxBlock GetTxBlock(const boost::multiprecision::uint256_t& blockNum)
+    {
+        lock_guard<mutex> g(m_mutexBlocks);
+
+        if (blockNum >= m_blocks.size())
+        {
+            throw "Blocknumber Absent";
```
bb111189: I think throwing a string is not an ideal way to handle this. Maybe we can throw nullptr or log warning or log fatal in this case

src/libData/BlockChainData/BlockChain.h
```diff
@@ -0,0 +1,176 @@
...
+    {
+        lock_guard<mutex> g(m_mutexBlocks);
+        return m_blocks.size();
+    }
+
+    /// Returns the last stored block.
+    T GetLastBlock()
+    {
+        lock_guard<mutex> g(m_mutexBlocks);
+        return m_blocks.back();
```
bb111189: What is there is no element in m_blacks? It will result in the circular array throwing an exception which we did not handle.

ansnunez:
@bb111189 addressed your changes:

- Changed GetXXXBlock to just GetBlock
- If some checks fail, log a warning or catch the exception from CircularArray, and return an empty block

src/libData/BlockChainData/BlockChain.h
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Returns the last stored block.
+    T GetLastBlock()
+    {
+        lock_guard<mutex> g(m_mutexBlocks);
+        try
+        {
+            return m_blocks.back();
+        }
+        catch (...)
```
Gnnng: I didn't find this try-catch in the original code but it's a good fix since `CircularArray::back()` is throwing [something](https://github.com/Zilliqa/Zilliqa/blob/c2b49315a37f36b035d1576ced796b728cc6ca26/src/libData/DataStructures/CircularArray.h#L90-L98). However, from the aspect of interface design I feel that `CircularArray::back` is doing too much (detecting empty-array error), since the caller could just call something like `.empty()` beforehand instead of going through the exception stuff.

In the standard library, for example, [`vector::back`](http://en.cppreference.com/w/cpp/container/vector/back) is not handling the empty container error and also not throwing any exception. (It actually leaves it as a UB)

src/libData/BlockChainData/BlockChain.h
```diff
@@ -0,0 +1,160 @@
...
+
+class DSBlockChain : public BlockChain<DSBlock>
+{
+public:
+    DSBlock GetBlockFromPersistentStorage(
+        const boost::multiprecision::uint256_t& blockNum)
+    {
+        DSBlockSharedPtr block;
+        BlockStorage::GetBlockStorage().GetDSBlock(blockNum, block);
+        return *block;
```
Gnnng: Seems we are doing stuff unnecessary here and it's an existing issue in the code. The [`GetDSBlock()`](https://github.com/Zilliqa/Zilliqa/blob/c2b49315a37f36b035d1576ced796b728cc6ca26/src/libPersistence/BlockStorage.cpp#L152) function probably needs some re-design where the return value can be a shared_ptr or unique_ptr. As for the case of not able to get an block, a nullptr wrapped by either share_ptr or unique_ptr can be returned.

Gnnng:
Just made some comments there. They are not critical issues and probably are out of the scope of this PR. It would be nice if they got attention from others and would be improved in the future.

PR #438, opened 2018-07-19
Add mutex to protect AccountStoreTemp Init

kaikawaliu:
## Description

During test we found that some node crashed due to heap-use-after-free issue in AccountStoreTemp, which is actually because of the Inititialization function in AccountStoreTemp is not protected by mutex, which lead to usage after the element being deleted in the later invocation in SubmitTransaction.

### Implementation

- [X] implement ...
- [X] **ready for review**

- [x] local machine test
- [ ] small-scale cloud test


ansnunez:
Closed, already part of merge in #432

PR #439, opened 2018-07-19
Merge PoW1 and PoW2 into one PoW (re-branched)

ansnunez:
## Description

This is the same change as in branch mergepow in PR #319. That branch was merged with two other completely unrelated large changes which were not easy to decouple from. Please refer to PR #319 for more information on this change and some initial review done on it in branch mergepow.

### Implementation

- [x] implement ...
- [ ] **ready for review**

- [x] local machine test
- [ ] small-scale cloud test


src/libDirectoryService/DirectoryService.cpp
```diff
@@ -1111,6 +942,118 @@ bool DirectoryService::ProcessAllPoWConnResponse(
...
+    LOG_MARKER();
+    LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
+              "I am sending AllPow2 to requester");
+
+    lock(m_mutexAllPOW1, m_mutexAllPOW2, m_mutexAllPoWConns);
+    lock_guard<mutex> g(m_mutexAllPOW1, adopt_lock);
+    lock_guard<mutex> g2(m_mutexAllPOW2, adopt_lock);
+    lock_guard<mutex> g3(m_mutexAllPoWConns, adopt_lock);
+
+    if (!m_allPoW1s.empty())
```
ansnunez: This message is sent by the new DS leader to the old DS leader.
In DirectoryService::ProcessDSBlockConsensusWhenDone(), the old DS leader already cleared its m_allPoW1 (so it is empty).
Does that mean this condition will never evaluate to true?

bb111189: In my opinion, after the merging, we should not have old leader sending new leader the info. An old leader at this point of time, can choose to conduct censorship. 

I have told clark previously that once the sharding structure and dsblock is merge, this function should be remove.

PR #440, opened 2018-07-19
Put transactions processing into MicroBlockConsensus with ordering

kaikawaliu:
## Goals:
This pr is trying to put the transaction submission inside Microblock Consensus.
The shard leader in the microblock consensus will generate an order for processing the transactions and broadcast to shard nodes via microblock message, all the shard nodes will follow the order to process transaction.

## Context:
This pr should be merged after the txn dispatcher being finished.

## Related PR:
- PR for txn dispatcher (TODO)
- PR for a new data structure for created transaction (https://github.com/Zilliqa/Zilliqa/pull/442)
- PR for clean-up, including removing the sleeping for TXN_SUBMISSION and TXN_BROADCAST and removing prefilled txns (https://github.com/Zilliqa/Zilliqa/pull/444)
- PR for txn ordering mechanism (https://github.com/Zilliqa/Zilliqa/pull/451)

### Implementation

Implemented:
- [X] Add transaction processing function for shard leader in Microblock consensus.
- [X] Add transaction processing function for shard backup in Microblock consensus.
- [X] Handle missing transactions

- [ ] **ready for review**

### Integration Test (Core Team)

- [ ] local machine test
- [ ] small-scale cloud test

PR #442, opened 2018-07-20
Use boost_multi_index for createdtransaction

kaikawaliu:
## Description

With the new implementation of transaction sharing.
We need an efficient data structure which can either sequentially access the elements by FIFO (like queue, list ...) and randomly accessed by hash value (like unordered_map).
In such case, `boost_multi_index` was chosen as the container for the transaction comes from the transaction dispatcher.

# Context
PR: https://github.com/Zilliqa/Zilliqa/pull/440

## Review Suggestion

Review this PR after https://github.com/Zilliqa/Zilliqa/pull/440 being merged.

### Implementation

- [X] implement ...
- [ ] **ready for review**

- [ ] local machine test
- [ ] small-scale cloud test


src/libNode/MicroBlockProcessing.cpp
```diff
@@ -475,6 +466,103 @@ bool Node::OnCommitFailure(
...
+        {
+            break;
+        }
+        txn_sent_count++;
+    }
+
+    OrderingTxns(curTxns);
+
+    boost::multiprecision::uint256_t blockNum
+        = (uint256_t)m_mediator.m_currentEpochNum;
```
bb111189: why is there a cast to uint256?

kaikawaliu: It's updated in later pr

ansnunez:
Work done in this PR has been integrated in #551

PR #214, opened 2018-04-09
[WIP] Fixed state change for pow1

bb111189:

PR #460, opened 2018-07-25
Move SetState for POW_SUBMISSION to new epoch

kaikawaliu:
## Description

Fix issue:
https://github.com/Zilliqa/Issues/issues/126

Simply let the DS call `SetState(POW_SUBMISSION)` only when a new DS epoch is about to begin.

### Implementation

- [X] **ready for review**

- [x] local machine test
- [ ] small-scale cloud test


bb111189:
Related obsolete PR is #214

Issue #215, opened 2018-04-09
review ProcessFinalBlockConsensusWhenDone


bb111189:
We may need to add state check and mutex at the start of the func

bb111189:
Fixed in PR #462

PR #462, opened 2018-07-25
Fix a bug where "consensus done" will be processed more than once

bb111189:
## Description

When consensus have been reached, leader (and also the rest of the backup), will process to `ProcessXXXXXConsensusWhenDone`. For the leader, after it send out the final co-signature, it may still received consensus `commit` from nodes which are slower. This is where the bug may occurs. Upon receiving the late `commit`, it will be process by `ProcessXXXXXConsensus`. Recall that the consensus state is now `DONE`, hence, it will again goes into `ProcessXXXXXConsensusWhenDone` (for the second time)

In this situation, segmentation fault may occurs as certain object such as `m_finalBlockMessage` are already cleared in the first `ProcessXXXXXConsensusWhenDone`, resulting in unexpected behavior. 

**Additional mitigation (TODO):**
`BitVector::SetBitVector`: check for underflow at  `const unsigned int length_available = dst.size() - offset;` https://sourcegraph.com/github.com/Zilliqa/Zilliqa/-/blob/src/libUtils/BitVector.cpp#L90:24

## Review Suggestion

Related issue are
https://github.com/Zilliqa/Issues/issues/127
https://github.com/Zilliqa/Zilliqa/issues/215

## Status
Ready. TODO will be implemented in separate pr.

### Implementation

- [ ] Check for int underflow at `const unsigned int length_available = dst.size() - offset;` at `BitVector::SetBitVector` (Separate PR) https://github.com/Zilliqa/Issues/issues/130
- [x] **ready for review**

- [X] local machine test
- [x] medium-scale cloud test


src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -502,6 +502,12 @@ bool DirectoryService::ProcessDSBlockConsensus(
     lock_guard<mutex> g(m_mutexConsensus);
 
     bool result = m_consensusObject->ProcessMessage(message, offset, from);
+
+    if (!result)
+    {
+        return result;
+    }
```
ckyang: Since when ProcessMessage return false we will directly return it, we don't need an extra variable to remember the result.
Suggest to rewrite as following:

if(!m_consensusObject->ProcessMessage(message, offset, from))
{
    return false;
}

And in the end of this function (line 536), we can always return true

PR #464, opened 2018-07-25
Move sharding serialization code into one place

ansnunez:
## Description

Another small update to prepare for PoW merging.  This time, I have moved the sharding structure serialization code into one new file called DSBlockMessage.h.

## Review Suggestion

Before this change, we used sharding structure serialization in two places:

- DirectoryService::SendEntireShardingStructureToLookupNodes
- DirectoryService::SerializeShardingStructure

The only difference between the two is that the first one has the IP info for all nodes, while the second one does not.

I have now made just one function now called ShardingStructure::Serialize that has the IP info.  The part of the code that does not need the IP info (DirectoryService::ShardingValidator) can just skip over it for now, to maintain the current behavior.

Regarding the file name DSBlockMessage.h:
When PoW merging is completed, the DS Block message will actually contain three things:
- DS Block
- Sharding structure
- Txn sharing assignments
So, the naming is in preparation for the merging.

### Implementation

- [x] **ready for review**

- [x] local machine test

### DIFF
--- a/src/libDirectoryService/DSBlockMessage.h
+++ b/src/libDirectoryService/DSBlockMessage.h
@@ -1,0 +1,93 @@
+/**
+* Copyright (c) 2018 Zilliqa 
+* This source code is being disclosed to you solely for the purpose of your participation in 
+* testing Zilliqa. You may view, compile and run the code for that purpose and pursuant to 
+* the protocols and algorithms that are programmed into, and intended by, the code. You may 
+* not do anything else with the code without express permission from Zilliqa Research Pte. Ltd., 
+* including modifying or publishing the code (or any part of it), and developing or forming 
+* another public or private blockchain network. This source code is provided ‘as is’ and no 
+* warranties are given as to title or non-infringement, merchantability or fitness for purpose 
+* and, to the extent permitted by law, all liability for your use of the code is disclaimed. 
+* Some programs in this code are governed by the GNU General Public License v3.0 (available at 
+* https://www.gnu.org/licenses/gpl-3.0.en.html) (‘GPLv3’). The programs that are governed by 
+* GPLv3.0 are those programs that are located in the folders src/depends and tests/depends 
+* and which include a reference to GPLv3 in their program files.
+**/
+
+#ifndef __DSBLOCKMESSAGE_H__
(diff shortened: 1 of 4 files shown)


src/libDirectoryService/ShardingPreProcessing.cpp
```diff
@@ -345,9 +287,9 @@ bool DirectoryService::RunConsensusOnShardingWhenDSPrimary()
     vector<unsigned char> sharding_structure;
 
     ComputeSharding();
-    SerializeShardingStructure(sharding_structure);
 
-    unsigned int txn_sharing_offset = sharding_structure.size();
+    unsigned int txn_sharing_offset
+        = ShardingStructure::Serialize(m_shards, sharding_structure, 0);
```
ckyang: It seems logic change here because ShardingStructure::Serialize() will return offset instead of sharding_structure.size().
Is it bug fix?

src/libDirectoryService/DSBlockMessage.h
```diff
@@ -0,0 +1,93 @@
...
+                        << " at " << kv.second.GetPrintableIPAddress()
+                        << " Port: " << kv.second.m_listenPortHost);
+            }
+        }
+
+        return cur_offset;
+    }
+};
+
+#endif // __DSBLOCKMESSAGE_H__
```
ckyang: Why file named "DSBlockMessage.h" contain the only one class named "ShardingStructure"? The naming seems no obvious relationship.

src/libDirectoryService/ShardingPreProcessing.cpp
ansnunez: In original code, offset was also expected (hence the name txn_sharing_offset), not size.
It just so happened that the offset was the same as the size.

src/libDirectoryService/DSBlockMessage.h
ansnunez: When PoW merging is completed, the DS Block message will actually contain three things:
DS Block
Sharding structure
Txn sharing assignments

So, the naming is in preparation for the merging.
I'll put a comment in the Description section above to make this clear.

PR #468, opened 2018-07-26
Unifying BlockChain.GetLastBlock.GetHeader.GetBlockNum

kaikawaliu:
## Description

After the circulararray has being modified. There are still many places using the old way of getting the latest Block number by using `GetBlockCount()`. This PR is to unify the usage of `GetLastBlock().GetHeader().GetBlockNum()`

### Implementation

- [X] **ready for review**

- [X] local machine test (**including new node joining**)
- [ ] small-scale cloud test


kaikawaliu:
Those left several GetBlockCount() are indeed trying to get the count of blocks in the blockchain. So they should remain unchanged.

PR #474, opened 2018-07-26
[PoW Merge] Move txn sharing assignments serialization code

ansnunez:
## Description

Another PR in the PoW merging work.

This is a similar effort to #464 where sharding info serialization was moved to DSBlockMessage.h.
This time, serialization of txn sharing assignments has been moved to that same file.

## Review Suggestion

What I basically did was split `AppendSharingSetupToShardingStructure()` into two functions:
- `DirectoryService::ComputeTxnSharingAssignments()` handles the computation part
- `TxnSharingAssignments::Serialize()` handles the serialization part

The assignment algorithm remains the same, except now we have to store the result of the computation part so that we can pass it to the serialization part.  Hence, 3 new member variables `m_DSReceivers`, `m_shardReceivers`, and `m_shardSenders`.

### Implementation

- [x] **ready for review**

- [x] local machine test

### DIFF
--- a/src/libDirectoryService/ShardingPreProcessing.cpp
+++ b/src/libDirectoryService/ShardingPreProcessing.cpp
@@ -88,104 +88,64 @@ void DirectoryService::ComputeSharding()
     }
 }
 
-void DirectoryService::AppendSharingSetupToShardingStructure(
-    vector<unsigned char>& sharding_structure, unsigned int curr_offset)
+void DirectoryService::ComputeTxnSharingAssignments()
 {
-    // Transaction body sharing setup
-    // Everyone (DS and non-DS) needs to remember their sharing assignments for this particular block
-
-    // Transaction body sharing assignments:
-    // PART 1. Select X random nodes from DS committee for receiving Tx bodies and broadcasting to other DS nodes
-    // PART 2. Select X random nodes per shard for receiving Tx bodies and broadcasting to other nodes in the shard
-    // PART 3. Select X random nodes per shard for sending Tx bodies to the receiving nodes in other committees (DS and shards)
-
-    // Message format:
-    // [4-byte num of DS nodes]
-    //   [16-byte IP] [4-byte port]
-    //   [16-byte IP] [4-byte port]
-    //   ...
-    // [4-byte num of committees]
-    // [4-byte num of committee receiving nodes]
-    //   [16-byte IP] [4-byte port]
-    //   [16-byte IP] [4-byte port]
-    //   ...
-    // [4-byte num of committee sending nodes]
-    //   [16-byte IP] [4-byte port]
-    //   [16-byte IP] [4-byte port]
-    //   ...
-    // [4-byte num of committee receiving nodes]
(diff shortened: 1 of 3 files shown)


src/libDirectoryService/ShardingPreProcessing.cpp
```diff
@@ -88,131 +88,82 @@ void DirectoryService::ComputeSharding()
...
     for (unsigned int i = 0; i < m_shards.size(); i++)
     {
         const map<PubKey, Peer>& shard = m_shards.at(i);
 
-        LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-                  "Shard " << i << " forwarders:");
-
         // PART 2
+
+        m_shardReceivers.emplace_back(vector<Peer>());
```
ckyang: The reason we recommend emplace_back is to reduce the object construction overhead. Thus it's better to just write m_shardReceivers.emplace_back() here.

src/libDirectoryService/ShardingPreProcessing.cpp
```diff
@@ -88,131 +88,82 @@ void DirectoryService::ComputeSharding()
...
+            m_shardReceivers.back().emplace_back(node_peer->second);
             node_peer++;
         }
 
-        LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-                  "Shard " << i << " senders:");
-
         // PART 3
+
+        m_shardSenders.emplace_back(vector<Peer>());
```
ckyang: The reason we recommend emplace_back is to reduce the object construction overhead. Thus it's better to just write m_shardReceivers.emplace_back() here.

src/libDirectoryService/ShardingPreProcessing.cpp
```diff
@@ -88,131 +88,82 @@ void DirectoryService::ComputeSharding()
...
             nodes_recv_hi = shard.size() - 1;
         }
 
         unsigned int num_nodes = nodes_recv_hi - nodes_recv_lo + 1;
 
-        Serializable::SetNumber<uint32_t>(sharding_structure, curr_offset,
-                                          num_nodes, sizeof(uint32_t));
-        curr_offset += sizeof(uint32_t);
-
         map<PubKey, Peer>::const_iterator node_peer = shard.begin();
```
ckyang: Maybe we can use "auto node_peer" to make code simpler?

src/libDirectoryService/ShardingPreProcessing.cpp
ansnunez: Done

src/libDirectoryService/ShardingPreProcessing.cpp
ansnunez: Done

src/libDirectoryService/ShardingPreProcessing.cpp
ansnunez: Done

Issue #469, opened 2018-07-26
Data race when calling gmtime


Gnnng:
`gmtime()` is not a thread-safe function, so the caller should properly use the lock to avoid data race.

For example, in `LogGeneral()`, the lock at line 168 is acquired after one of the invocation of `gmtime` ath line 163.

https://github.com/Zilliqa/Zilliqa/blob/3c47506ca8302adcad8da7829c981557c60f6b17/src/libUtils/Logger.cpp#L155-L186

This is caught by ThreadSanitizer 

```
WARNING: ThreadSanitizer: data race (pid=123)
  Write of size 8 at 0x7f52e85354a0 by thread T36:
    #0 gmtime <null> (libtsan.so.0+0x00000002e81f)
    #1 P2PComm::HandleAcceptedConnection(int, Peer, std::function<void (std::vector<unsigned char, std::allocator<unsigned char> > const&, Peer const&)>, std::function<std::vector<Peer, std::allocator<Peer> > (unsigned char, unsigned char, Peer const&)>) /zilliqa/src/libNetwork/P2PComm.cpp:445 (zilliqa+0x00000044c808)
    #2 operator() /zilliqa/src/libNetwork/P2PComm.cpp:638 (zilliqa+0x00000044e053)
    #3 _M_invoke /usr/include/c++/5/functional:1871 (zilliqa+0x00000044e053)
    #4 std::function<void ()>::operator()() const <null> (zilliqa+0x000000450a0c)
    #5 ThreadPool::Task() /zilliqa/src/libUtils/ThreadPool.h:189 (zilliqa+0x000000452311)
    #6 ThreadPool::ThreadPool(unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)::{lambda()#1}::operator()() const /zilliqa/src/libUtils/ThreadPool.h:65 (zilliqa+0x000000452311)
    #7 void std::_Bind_simple<ThreadPool::ThreadPool(unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)::{lambda()#1} ()>::_M_invoke<>(std::_Index_tuple<>) /usr/include/c++/5/functional:1531 (zilliqa+0x000000452311)
[...]

PR #477, opened 2018-07-27
Adjust log mutex range to avoid potential race-condition from gmtime

ckyang:
## Description

https://github.com/Zilliqa/Zilliqa/issues/469
By ThreadSanitizer, there is potential race-condition coming from gmtime. Now we adjust the logger mutex to avoid this potential problem.

### Implementation

- [ ] **ready for review**

- [ ] local machine test
- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/Logger.cpp
+++ b/src/libUtils/Logger.cpp
@@ -21,7 +21,6 @@
 #include <pthread.h>
 #include <sys/syscall.h>
 #include <unistd.h>
-
 using namespace std;
 using namespace g3;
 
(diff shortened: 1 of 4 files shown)


src/libUtils/Logger.cpp
```diff
@@ -277,7 +278,7 @@ void Logger::LogEpochInfo(const char* msg, const char* function,
     std::time_t curTime = std::chrono::system_clock::to_time_t(
         std::chrono::system_clock::now());
 
-    lock_guard<mutex> guard(m);
+    lock_guard<mutex> guard(m_logMutex);
```
Gnnng: It's better to move this lock acquiring before calling `system_clock::now()` because by the time you acquire the lock, `curTime` may already be a bit late (although it's minor difference)

ckyang: ok

src/libUtils/TimeUtils.cpp
```diff
@@ -33,3 +34,16 @@ uint256_t get_time_as_int()
...
     return static_cast<uint256_t>(microsecs.count());
 }
+
+struct tm* gmtime_safe(const time_t* timer)
+{
+    std::lock_guard<std::mutex> guard(gmtimeMutex);
+    return gmtime(timer);
+}
+
+long int get_ms(const time_point<high_resolution_clock> time)
```
ansnunez: Can we pass `time` by reference here, to improve performance?

ckyang: ok

ckyang:
The latest version already support millisecond precision in log file, however we still need to run ThreadSanitizer to make sure gmtime() is thread-safe or not.

PR #480, opened 2018-07-28
Travis build speed-up and improvement

Gnnng:
## Description

The main contribution of this PR is to fix Zilliqa/Issues#124. The previous way of doing code coverage has several issues:
- The target `Zilliqa_coverage` already includes the `lcov` commands listed in `scripts/ci_report_coverage.sh`, 
https://github.com/Zilliqa/Zilliqa/blob/14be39234c99d772affff9bd6c71ff931a8c5053/cmake/CodeCoverage.cmake#L157-L160. So the duplicated command in `scripts/ci_report_coverage.sh` is removed.
- The code coverage prefer `Debug` build to have more accurate coverage report. There was some [discussion](https://github.com/Zilliqa/Zilliqa/pull/58#issuecomment-393599408) about this already. This may increase the `ctest` time a bit but in it's not a majority of the Travis build time.
- In `scripts/ci_build.sh`, we are running `ctest` twice, one via explicitly callinig `ctest`, and the other one via implicitly calling `make Zilliqa_coverage`. This is unnecessary and has been corrected.

The `ccache` configuration also have one issue:

- The setting `max_size` is reported to be ~500MB previously on Linux build jobs while Mac build is using 5GB. Since mac build works fine with 5GB, now explicitly set it to `5GB` via `ccache -M`

## Future Work

The `deploy` stage is not using any cache mechanism. It's now becoming the majority of the build time.

## Review Suggestion

See the description above and the [build log](https://travis-ci.com/Zilliqa/Zilliqa/builds/80195020?utm_source=github_status&utm_medium=notification) to verify the speed-up.

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ jobs:
     install: # left blank
     - pip install --user awscli
     - export PATH=$PATH:$HOME/.local/bin
-    cache: # left blank
+    cache: false
     script:
     - $BUILD_COMMAND
     after_success: # left blank
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,11 +65,15 @@ add_compile_options(-Werror)
 add_compile_options(-Wextra)
 
 if (ENABLE_COVERAGE AND CMAKE_COMPILER_IS_GNUCXX)
+    if (NOT TESTS)
+        message(FATAL_ERROR "TESTS is not ON")
+    endif()
     include(CodeCoverage)
     add_compile_options(--coverage)
     link_libraries(--coverage)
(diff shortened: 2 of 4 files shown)


sharwell:
@Gnnng I noticed that the codecov comment did not get added in this pull request. Do you know if this was this an intentional change, an unintentional configuration change, or a transient behavior that will likely not affect other pull requests?

Gnnng:
@sharwell we didn't directly change any settings in codecov. However, recently we disabled pull-request builds on Travis temporarily. It may cause the codecov commenting being stopped.

PR #488, opened 2018-07-30
Buffer for forwarding transaction & state delta

kaikawaliu:
## Description

Instead of using cv to block the forwarding transaction. Here simply a buffer can hold the forwarded data received earlier than finishing the processing of finalblock.

### Implementation

- [X] **ready for review**

- [X] local machine test
- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -1997,6 +1997,10 @@ void Node::CommitForwardedMsgBuffer()
                 m_forwardedTxnBuffer.erase(it);
                 break;
             }
+            else
+            {
+                it++;
+            }
         }
     }
 
@@ -2023,6 +2027,10 @@ void Node::CommitForwardedMsgBuffer()
                 }
                 m_forwardedDeltaBuffer.erase(it);
                 break;
+            }
+            else
+            {
+                it++;
             }
         }
     }
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -1491,6 +1491,8 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
                              << "] LAST");
     }
 
+    CommitForwardedMsgBuffer();
+
     // Assumption: New PoW done after every block committed
     // If I am not a DS committee member (and since I got this FinalBlock message,
     // then I know I'm not), I can start doing PoW again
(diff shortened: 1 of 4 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1927,3 +1970,60 @@ bool Node::ProcessForwardStateDelta(const vector<unsigned char>& message,
...
+
+void Node::CommitForwardedMsgBuffer()
+{
+    {
+        lock_guard<mutex> g(m_mutexForwardedTxnBuffer);
+
+        for (auto it = m_forwardedTxnBuffer.begin();
+             it != m_forwardedTxnBuffer.end();)
+        {
+            if (it->first < m_mediator.m_txBlockChain.GetLastBlock()
```
ansnunez: Can this condition actually be hit?

kaikawaliu: Yes, for the forwarding message too late to arrive

ansnunez: I think the check for blocknum in `ProcessForwardTransaction` already takes care of this, but anyway no harm in keeping this code.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1927,3 +1970,60 @@ bool Node::ProcessForwardStateDelta(const vector<unsigned char>& message,
...
     return true;
 }
+
+void Node::CommitForwardedMsgBuffer()
+{
+    {
+        lock_guard<mutex> g(m_mutexForwardedTxnBuffer);
+
+        for (auto it = m_forwardedTxnBuffer.begin();
+             it != m_forwardedTxnBuffer.end();)
```
ckyang: Forget to add ++it?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1927,3 +1970,60 @@ bool Node::ProcessForwardStateDelta(const vector<unsigned char>& message,
...
+                break;
+            }
+        }
+    }
+
+    {
+        lock_guard<mutex> g(m_mutexForwardedDeltaBuffer);
+
+        for (auto it = m_forwardedDeltaBuffer.begin();
+             it != m_forwardedDeltaBuffer.end();)
```
ckyang: Forget to add ++it?

kaikawaliu: noted

ansnunez:
Won't be merged. Changes carried forward to #485

PR #491, opened 2018-07-31
Use evpp as network protocol

ckyang:
## Description

Currently, we use libevent as network server protocol.
However, evpp performed much better than libevent and similar to asio, with thread-safe design.
https://github.com/Qihoo360/evpp/blob/master/docs/benchmark_throughput_vs_libevent.md
https://github.com/Qihoo360/evpp/blob/master/docs/benchmark_throughput_vs_asio.md

This PR will apply evpp as network server/client protocols.

### Implementation

- [ ] **ready for review**

- [ ] local machine test
- [ ] small-scale cloud test


Gnnng:
test comment

ckyang:
After implementation, it seems evpp not really thread-safe as it claimed.
When creating tcpserver with thread_num > 1, plenty of exceptions are thrown as following:

terminate called after throwing an instance of 'std::logic_error' what(): basic_string::_S_construct null not valid

However if we create tcpserver with thread_num = 1, no this kind of error occurred although this setting is obviously cannot be applied in real Zilliqa code base.

Thus we stop apply evpp and will look for another solution for network enhancement.

PR #446, opened 2018-07-20
Origin/feature/include protobuf

HuKuToH:
## Description
Investigation into the possibility to include protobuf. Replaced one type of messages (DSInstructionType::POW1SUBMISSION) for serialization with the help of the protobuf, to assess the possibility of further implementation.

Thanks to satoshi101 for supporting this investigation.

PR #502, opened 2018-08-05
Add sample and test code for protobuf messaging

ansnunez:
## Description

Provided a cleaned-up version of sample protobuf usage originally submitted in #446.
A test message has been defined in src/libMessage.
A test routine has been created in tests/Message to demonstrate setting, getting, serializing, and deserializing the test message.

### Implementation

- [x] **ready for review**

- [x] local machine test
- [ ] small-scale cloud test


satoshi101:
Looks great to me! Overall the idea would be to have a very Top-Level message type which is the only thing that goes on the wire. Everything that can't load as a top-level message, gets discarded:

```
message PoW1Message {
	bytes block_number	 	= 1;
	uint32 listening_port 	= 2;
	bytes pub_key 			= 3;
	uint64 nonce 			= 4;
	bytes result_hash 		= 5;
	bytes mixhash 			= 6;
	bytes signature 		= 7;
}

message PoW2Message {
	bytes block_number	 	= 1;
	uint32 listening_port 	= 2;
	bytes pub_key 			= 3;
	uint64 nonce 			= 4;
	bytes result_hash 		= 5;
	bytes mixhash 			= 6;
	bytes signature 		= 7;
}

message ConsensusMessage {
  repeated CommitConsensusMessage commit = 1;
  repeated CommitFailureConsensusMessage commitFailure = 2;
  repeated ResponseConsensusMessage response = 3;
  repeated FinalCommitConsensusMessage finalCommit = 4;
  repeated FinalResponseConsensusMessage finalResponse = 5;
}

message ZilliqaMessage {
  repeated PoW1Message pow1 = 1;
  repeated PoW2Message pow2 = 2;
  repeated ConsensusMessage consensus = 3;
}
```

This way you will completely get rid of all the if/then/else decoding code and you can pack as many messages as you need. It's worth trying a little experiment, but very highly likely leaving `repeated` (and `optional` fields empty) won't cost you anything in terms of encoding (and even if it costs a little bit - who cares?) I would prefer `repeated` to `optional` because they give opportunity of extra packing and they're better algebraically i.e. easier to handle. For example, to process `ZilliqaMessage`, as expressed above, all you have to do is:

```
ZilliqaMessage zil;
∀x∈zil.pow1 → process(x),
∀x∈zil.pow2 → process(x),
∀x∈zil.consensus → process(x),
...
```

that's just `N`' for loops. It handles automatically missing values, duplicates and empty messages (which are all valid'ish messages).

In contrast, an optional would be like this:

```
ZilliqaMessage zil;
if zil.pow1: 
  process(zil.pow1)
if zil.pow2: 
  process(zil.pow2)
if zil.consensus: 
  process(zil.consensus)
...
```

I don't know if you can see this is way more ugly. It legitimizes a message where `pow1` and `consensus` are both set, whereas what was likely meant to be expressed with the `optional` is that only one of those messages should be set. I think it's better to allow many of them and allow full expression of all those options of none, one or many, as expressed by the `repeated` keyword.

ansnunez:
If we don't go the multiple packed messages route, there is also the option of using `oneof` which acts like a `union` type.

satoshi101:
Still that's a bit uglier. :) And actually it's good one to make a small-scale check those alternative encodings and their costs in terms of space. I've seen in other systems union being bad i.e. taking as much space as the largest message regardless on if you have small messages. But still, for me the `∀ x ∈ zil.*` is the strong argument. I prefer to deal with `for`s rather than `if`s. But's a rather philosophical argument based on "linearities" and the size of the validation (testing) space.

PR #503, opened 2018-08-06
Feature/gpu mining

ShengguangXiao:
## Description

Support Nvidia GPU mining using CUDA library.

Add the libethash-cuda in src/depends. And Updated the libPOW.

Support GPU mining #94
https://github.com/Zilliqa/Issues/issues/94

## Review Suggestion

Check the src/depends/libethash-cuda and src/libPOW

1. Enable the CUDA_GPU_MINE option in constants.xml, build the Zilliqa application, run the ./build/tests/POW/Test_POW to run unit tests.
2. If you have CUDA enabled GPU, you can run 4 tests nodes locally enable the FULL_DATASET_MINE and CUDA_GPU_MINE option.

### Implementation

- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test


src/libPOW/CMakeLists.txt
```diff
@@ -5,3 +5,7 @@ target_include_directories (POW PUBLIC ${PROJECT_SOURCE_DIR}/src ${G3LOG_INCLUDE
 
 find_library(OPENCL_LIBRARIES OpenCL ENV LD_LIBRARY_PATH)
 target_link_libraries (POW PRIVATE ${OPENCL_LIBRARIES} ethash ethash-cl Constants)
+
+if(CUDA_FOUND)
+    target_link_libraries(POW PRIVATE ethash-cuda)
```
bb111189: can you change to dynamic linking? using the SHARED instead of PRIVATE?

src/libPOW/pow.cpp
```diff
@@ -47,7 +50,29 @@ POW::POW()
...
+        LOG_GENERAL(INFO, "OpenCL GPU initialized in POW");
+    }
+    else if (CUDA_GPU_MINE)
+    {
+#ifdef ETHASH_CUDA
+        using namespace dev::eth;
+        CUDAMiner::setNumInstances(UINT_MAX);
+        if (!CUDAMiner::configureGPU(
+                CUDAMiner::c_defaultBlockSize, CUDAMiner::c_defaultGridSize,
+                CUDAMiner::c_defaultNumStreams, 4, 0, 0, false, false))
```
bb111189: Are they values that are configurable (like the number of streams)? If so, can we provide a config file for users to change the setting easily?

src/depends/libethash-cl/CLMiner.cpp
```diff
@@ -295,7 +293,8 @@ bool CLMiner::mine(const WorkPackage &w, Solution &solution)
                     ++s_dagLoadIndex;
                 }
 
-                init(w.blockNumber);
+                if(!init(w.blockNumber))
+                    return false;
```
bb111189: since you are changing it, enclosed it with {}

scripts/install-cuda-ubuntu1604.sh
```diff
@@ -0,0 +1,23 @@
...
+# 'source' this script, rather than executing it in a sub-process.
+#
+# Taken from https://github.com/tmcdonell/travis-scripts.
+
+set -e
+
+export CUDA_VER=9.2.148-1
+apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/7fa2af80.pub
+wget http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/cuda-repo-ubuntu1604_${CUDA_VER}_amd64.deb
+dpkg -i cuda-repo-ubuntu1604_${CUDA_VER}_amd64.deb
```
bb111189: can the script also remove the .deb file after installation?

bb111189:
I install the dep using the script and change constant files. This is the error I encountered.
```
Running 23 test cases...
unknown location(0): fatal error in "mining_and_verification": std::runtime_error: The software is not build with CUDA. Please install CUDA and build software again
<refacted>/Zilliqa/tests/POW/test_POW.cpp(696): last checkpoint
CUDA_GPU_MINE enabled, test with CUDA GPU
unknown location(0): fatal error in "gpu_mining_and_verification_1": std::runtime_error: The software is not build with CUDA. Please install CUDA and build software again
<refacted>/Zilliqa/tests/POW/test_POW.cpp(696): last checkpoint
CUDA_GPU_MINE enabled, test with CUDA GPU
unknown location(0): fatal error in "gpu_mining_and_verification_2": std::runtime_error: The software is not build with CUDA. Please install CUDA and build software again
<refacted>/Zilliqa/tests/POW/test_POW.cpp(696): last checkpoint

*** 3 failures detected in test suite "powtest"
```

src/libPOW/pow.cpp
ShengguangXiao: Yes, i think in next PR i need to create a GPU configuration file to let user set the GPU parameters.

tests/POW/test_POW.cpp
```diff
@@ -792,17 +803,28 @@ BOOST_AUTO_TEST_CASE(opencl_gpu_mining_and_verification_1)
...
+BOOST_AUTO_TEST_CASE(gpu_mining_and_verification_2)
 {
-    if (!OPENCL_GPU_MINE)
+    if (!OPENCL_GPU_MINE && !CUDA_GPU_MINE)
     {
-        std::cout << "OPENCL_GPU_MINE option is not enabled, skip test case "
-                     "opencl_gpu_mining_and_verification_2"
+        std::cout << "OPENCL_GPU_MINE and CUDA_GPU_MINE option are not "
+                     "enabled, skip test case "
+                     "gpu_mining_and_verification_1"
```
Gnnng: I guess this should be 2 instead of 1, right?

src/libPOW/CMakeLists.txt
ShengguangXiao: The libethash-cuda is created as a SHARED library already. 
cuda_add_library(ethash-cuda SHARED ${sources} ${headers})
The PRIVATE in target_link_libraries means the POW depends on ethash, but ethash not depends on POW.

tests/POW/test_POW.cpp
ShengguangXiao: Yes, you are correct.

README.md
```diff
@@ -67,6 +67,7 @@ To compile and run the Zilliqa codebase, you will need the following dependencie
 * `libevent`
 * `miniupnpc`
 * `ocl-icd-opencl-dev`
+* `cuda9.2`
```
Gnnng: Let's only put the minimal dependencies here. I can help to update the README if you want.

README.md
```diff
@@ -83,6 +84,8 @@ sudo apt-get update
 sudo apt-get install git clang-format-5.0 clang-tidy-5.0 clang-5.0 libboost-system-dev libboost-filesystem-dev libboost-test-dev libssl-dev libleveldb-dev libjsoncpp-dev libsnappy-dev cmake libmicrohttpd-dev libjsonrpccpp-dev build-essential pkg-config libevent-dev libminiupnpc-dev ocl-icd-opencl-dev
 ```
 
+If you want to use CUDA GPU, please use `sudo ./scripts/install-cuda-ubuntu1604.sh` to install CUDA library first, you may need to reboot your PC to make the CUDA installation take effect.
+
```
Gnnng: It's better to leave CUDA installation for users because it requires some familiarity and experiences with CUDA. It's okay we provide some quick installation (like the script `install-cuda-ubuntu1604.sh`) but I suggest moving it to our wiki from the codebase as (1) GPU mining (as well as CUDA) is not a compulsory requirement; (2) we may edit the installation instructions frequently when we are adding support for more hardware; (3) previously, the intention of having the script is to run it on Travis but we don't need it anymore.

bb111189:
Do move `GPU support for PoW` to available feature in `README.md`

README.md
ShengguangXiao: Ok, understand, i will provide the nvidia website link instead.

README.md
ShengguangXiao: Thanks, i will handle it.

ShengguangXiao:
Yes, we can have a cmake flag, but then should we need change the build.sh? What will be the default setting?

Gnnng:
> Yes, we can have a cmake flag, but then should we need change the build.sh? What will be the default setting?

Default would still be the current one (without CUDA). We can mention in the README that the cmake flag is needed for CUDA build.

bb111189:
```
Running 23 test cases...
unknown location(0): fatal error in "mining_and_verification": std::runtime_error: The software is not build with CUDA. Please install CUDA and build software again
<refacted>/Zilliqa/tests/POW/test_POW.cpp(696): last checkpoint
CUDA_GPU_MINE enabled, test with CUDA GPU
unknown location(0): fatal error in "gpu_mining_and_verification_1": std::runtime_error: The software is not build with CUDA. Please install CUDA and build software again
<refacted>/Zilliqa/tests/POW/test_POW.cpp(696): last checkpoint
CUDA_GPU_MINE enabled, test with CUDA GPU
unknown location(0): fatal error in "gpu_mining_and_verification_2": std::runtime_error: The software is not build with CUDA. Please install CUDA and build software again
<refacted>/Zilliqa/tests/POW/test_POW.cpp(696): last checkpoint
```

This is due to insufficient available gpu ram. Running the testcase in single user mode (where there is no display manager) works. 

Tested on Nvidia 930M

ShengguangXiao:
@Gnnng @bb111189 I have added a cmake option CUDA_MINE, default it is OFF to control if build the libethash-cuda. Please review again.

bb111189:
```If you want to use CUDA GPU, please download CUDA package from https://developer.nvidia.com/cuda-downloads and follow the instruction to install. And add -DCUDA_MINE=ON in ./build.sh as the cmake option to enable the CUDA mine library.```


I think asking the user to modify build.sh can be user unfriendly. A better option is to pass the option into build.sh and build.sh will pass it cmake.

**Comment by Gnnng** [comment]
> I think asking the user to modify build.sh can be user unfriendly. A better option is to pass the option into build.sh and build.sh will pass it cmake.

I have some thoughts here. We can create a top-level Makefile and call different targets via `make XXX`. This will reduce the number of `build_xxx.sh` scripts in the root.

```
normal:
    cmake ...

lookup:
    cmake ...

cuda-normal:
    cmake ...

cuda-lookup:
    cmake ...
```

PR #505, opened 2018-08-06
Use standard libevent procedure

ckyang:
## Description

Since P2PComm implementation is based on libevent, some standard procedure should be followed, in order to improve maintenance and readability.

1. Re-factor reading, apply event_base & evconnlistener to handle socket bind()/listen()/accept()
2. For writing part, it's unnecessary to re-factor to libevent-style because there is no obviously performance gain however will cause SIGPIPE problem for socket closing
3. Wrap socket read() & write() function to be more modularize
4. For EAGAIN error raised in non-blocking read(), allow to retry
5. Make Dispatcher and Broadcast_list_func as P2PComm member variables

Follow-up PR: provide blacklist API in P2PComm

### Implementation

- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test
- [x] large-scale cloud test

### DIFF
--- a/src/libNetwork/P2PComm.cpp
+++ b/src/libNetwork/P2PComm.cpp
@@ -16,6 +16,7 @@
 
 #include <cstring>
 #include <errno.h>
+#include <event2/bufferevent.h>
 #include <event2/event.h>
 #include <event2/listener.h>
 #include <memory>
@@ -54,15 +55,6 @@ struct hash_compare
         return equal(l.begin(), l.end(), r.begin());
     }
 };
-
-static void close_socket(int* cli_sock)
-{
-    if (cli_sock != NULL)
-    {
-        shutdown(*cli_sock, SHUT_RDWR);
-        close(*cli_sock);
-    }
-}
 
 static bool comparePairSecond(
     const pair<vector<unsigned char>, chrono::time_point<chrono::system_clock>>&
@@ -241,7 +233,6 @@ namespace
     {
         return (buf[2] << 24) + (buf[3] << 16) + (buf[4] << 8) + buf[5];
     }
-
 } // anonymous namespace
 
 bool P2PComm::SendMessageSocketCore(const Peer& peer,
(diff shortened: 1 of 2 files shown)


src/libNetwork/P2PComm.h
```diff
@@ -98,31 +97,35 @@ class P2PComm
     using Dispatcher
         = std::function<void(const std::vector<unsigned char>&, const Peer&)>;
 
+    using Broadcast_list_func = std::function<std::vector<Peer>(
```
Gnnng: Not sure if we are enforcing the naming convention or not, but `BraodcastListFunc` is preferred in google's coding [guide](https://google.github.io/styleguide/cppguide.html#Type_Names). @jiayaoqijia, any idea?

ckyang: It's good guideline and I will modify this

ansnunez: Agree

src/libNetwork/P2PComm.cpp
```diff
@@ -148,6 +146,104 @@ void P2PComm::SendMessageCore(const Peer& peer,
...
+    bool readMsg(vector<unsigned char>* buf, int cli_sock, const Peer& from,
+                 const uint32_t message_length)
+    {
+        // Read the rest of the message
+        assert(buf);
+        buf->resize(message_length);
+        uint32_t read_length = 0;
+        uint32_t retryDuration = 1;
+
+        while (read_length != message_length)
```
ansnunez: In the future I think maybe we should use a stronger check here (read_length < message_length)

src/libNetwork/P2PComm.h
jiayaoqijia: Since we use clang-format, we can choose a conversion it support. You can take a look at https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options.

src/libNetwork/P2PComm.cpp
ckyang: OK, I'm good to make modification in this PR

PR #509, opened 2018-08-08
Implementation of DS Microblock

kaikawaliu:
## Description

**This PR is for implement the microblock consensus for DS committee**
- It happens after the DS received all the microblocks from shard, or triggered by timeout for waiting microblock submission.
- DS will process txns based on the temp account states generated by state delta received from shard microblocks
- If microblock consensus failed or time-out. DS will disgard the changes in state delta resulted from DS microblock and proceed to finalblock consensus.

### Implementation

- [X] **ready for review**

- [X] local machine test
- [ ] small-scale cloud test


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -351,7 +351,8 @@ bool Node::CheckMicroBlockRootHash(const TxBlock& finalBlock,
 }
 
 #ifndef IS_LOOKUP_NODE
-void Node::BroadcastTransactionsToLookup()
+void Node::BroadcastTransactionsToLookup(
```
ansnunez: This doesn't seem to be called by the DS node, only by the shard nodes after final block is received

PR #512, opened 2018-08-08
Feature/gpu mining - Read gpu config parameters from gpu_config.xml

ShengguangXiao:
## Description

Let user change the config parameters of GPU

Add the gpu_config.xml, read the parameters before initilizing the GPU, and use the parameters to config GPU

Support GPU mining #94
https://github.com/Zilliqa/Issues/issues/94

## Review Suggestion

Check the gpu_config.xml and the constructor function of POW.

Enable the OPENCL_GPU_MINE or CUDA_GPU_MINE and run ./build/tests/POW/Test_POW to check the result.

### Implementation

- [ ] **ready for review**

- [ ] local machine test
- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -39,6 +39,7 @@ POW::POW()
     ethash_light_client = EthashLightNew(
         0); // TODO: Do we still need this? Can we call it at mediator?
 
+#ifndef IS_LOOKUP_NODE
     if (OPENCL_GPU_MINE)
     {
         InitOpenCL();
@@ -47,6 +48,7 @@ POW::POW()
     {
         InitCUDA();
     }
+#endif // IS_LOOKUP_NODE
 }
 
 POW::~POW() { EthashLightDelete(ethash_light_client); }
@@ -490,9 +492,8 @@ void POW::InitCUDA()
     if (!CUDAMiner::configureGPU(blockSize, gridSize, streamNum, scheduleFlag,
                                  0, 0, false, false))
     {
-        LOG_GENERAL(FATAL,
-                    "Failed to configure CUDA GPU, please check hardware");
-        exit(1);
+        throw std::runtime_error(
+            "Failed to configure CUDA GPU, please check hardware");
     }
 
     CUDAMiner::setNumInstances(UINT_MAX);
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -15,6 +15,8 @@
 **/
 
 #include <boost/algorithm/string/predicate.hpp>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/xml_parser.hpp>
 #include <ctime>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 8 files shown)


jiayaoqijia:
Does it require some changes to Kubernete scripts? @Gnnng @ShengguangXiao

ShengguangXiao:
Actually originally i want to put the parameters into constants.xml, but Jun Hao suggest to use another standalone file......

bb111189:
I leave the constant file issue open for discussion. 

On other matter, @ShengguangXiao do add the appropriate label (eg. Ready) and set project to `core` in github. It is easier for other to track the PRs.

ShengguangXiao:
I moved already. Previously i thought it will automatically go to the core project. 

I think we can move the gpu parameters to constants.xml, then we only need to maintain one constants file, it that OK?

Gnnng:
Thanks @ShengguangXiao and @bb111189. I was thinking there will be no changes needed from the testnet side if no new config files are added. Also, the same reason applies in CMakeLists.txt.

ShengguangXiao:
@Gnnng @bb111189 I have moved the parameters from gpu_config.xml to constants.xml, and delete the gpu_config.xml file, please review again.

src/libPOW/pow.cpp
```diff
@@ -474,4 +449,59 @@ ethash_return_value_t POW::LightHash(uint64_t blockNum,
...
+    using namespace dev::eth;
+    CLMiner::setCLKernel(CLKernelName::Stable);
+
+    ptree pt;
+    read_xml("constants.xml", pt);
+    std::string parentNode = "node.gpu.opencl.";
+    auto localWorkSize = pt.get<unsigned int>(parentNode + "LOCAL_WORK_SIZE");
+    auto globalWorkSizeMultiplier
+        = pt.get<unsigned int>(parentNode + "GLOBAL_WORK_SIZE_MULTIPLIER");
+    auto startEpoch = pt.get<unsigned int>(parentNode + "START_EPOCH");
```
Gnnng: Since we are adding the parameters to `constants.xml`, can we move these lines of reading configuration to `constants.h/cpp`? It will reduce the difficulty of refactoring the configuration loading in the future.

src/libPOW/pow.cpp
```diff
@@ -474,4 +449,59 @@ ethash_return_value_t POW::LightHash(uint64_t blockNum,
...
+    LOG_GENERAL(INFO, "OpenCL GPU initialized in POW");
+}
+
+void POW::InitCUDA()
+{
+#ifdef CUDA_MINE
+    using namespace dev::eth;
+
+    ptree pt;
+    read_xml("constants.xml", pt);
```
Gnnng: Since we are adding the parameters to `constants.xml`, can we move these lines of reading configuration to `constants.h/cpp`? It will reduce the difficulty of refactoring the configuration loading in the future.

src/libPOW/pow.cpp
ShengguangXiao: Ok, already moved to constants.h/cpp.

ShengguangXiao:
@Gnnng @bb111189 Now the reading gpu parameters are moved to constants.h/cpp, please review again...

src/libPOW/pow.cpp
```diff
@@ -34,46 +34,17 @@ POW::POW()
     currentBlockNum = 0;
     ethash_light_client = EthashLightNew(
         0); // TODO: Do we still need this? Can we call it at mediator?
+
+#ifndef IS_LOOKUP_NODE
```
Gnnng: I'm not creating troubles here but I don't really like using `IS_LOOKUP_NODE` macro here. Maybe the discussion was not clear and transparent before. Let's revisit this issue again. My suggestion is gradually removing the uses of this macro (or at least not adding any new ones 'cause it's adding up the difficulty). We don't actually enjoy too many benefits from it but suffer a lot in sorting out the build system (CMake) and CI/CD workflow.

ShengguangXiao: If don't use this MACRO, is there any other variable can differentiate the normal node or look_up node?

src/libPOW/pow.cpp
```diff
@@ -457,8 +457,10 @@ void POW::InitOpenCL()
                                OPENCL_GLOBAL_WORK_SIZE_MULTIPLIER, 0,
                                OPENCL_START_EPOCH, 0, 0, false, false))
     {
-        throw std::runtime_error(
+        std::string msg(
             "Failed to configure OpenCL GPU, please check hardware");
+        LOG_GENERAL(FATAL, msg);
+        throw std::runtime_error(msg);
```
jiayaoqijia: Once you have LOG FATAL, you don't need throw.

src/libPOW/pow.cpp
jiayaoqijia: I think we don't need to fix this issue here. I'll create a separate issue and you can fix all the MACRO in another PR. @ShengguangXiao

src/libPOW/pow.cpp
ShengguangXiao: If don't throw, later the software will crash, and from the terminal user need to think about why the software is crashed.

jiayaoqijia: The LOG  FATAL will throw the error message.

jiayaoqijia: Hi @ShengguangXiao , if it cannot be resolved now, please create an issue for it. Let's finish this PR by today.

jiayaoqijia:
If possible, let's also make OpenCL as an optional feature, same as CUDA. Let users or devs decide whether they want to include it or not.

ckyang:
The MacOS Travis failed should belongs to this: https://github.com/KjellKod/g3log/issues/89
I've just modify Logger and hopefully it works.

ShengguangXiao:
@Gnnng @bb111189 This PR should be ready now. Please review again...

PR #516, opened 2018-08-13
Modify log message and avoid terminate program when lookup cannot rec…

ckyang:
## Description

As https://github.com/Zilliqa/Issues/issues/148 and https://github.com/Zilliqa/Issues/issues/149 discussed, sometimes lookup node cannot received final block on time.
In this PR, we remove the FATAL message that will cause program termination and allow user to handle this situation by themselves.

### Implementation

- [ ] **ready for review**

- [ ] local machine test
- [ ] small-scale cloud test

### DIFF
--- a/src/libData/BlockChainData/BlockChain.h
+++ b/src/libData/BlockChainData/BlockChain.h
@@ -88,9 +88,11 @@ public:
 
         if (m_blocks[blockNum].GetHeader().GetBlockNum() != blockNum)
         {
-            LOG_GENERAL(FATAL,
-                        "assertion failed (" << __FILE__ << ":" << __LINE__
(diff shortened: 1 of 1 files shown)


bb111189:
@ckyang  An extra check to be done to avoid regression. 
- [ ] Check through codes that used GetBlock(). Can these codes process a block with no content?

ckyang:
@bb111189 I just checked all GetBlock() callers, they will neither terminate nor throw exception when receiving dummy (empty) block. However, of course some abnormal behavior will happened when using dummy (empty) block but at least we already log a WARNING message. I think it's better to leave the error handling to caller because it's case-by-case. BTW, I'll modify the WARNING message to be more noticeable.

PR #520, opened 2018-08-14
Blacklist support

ckyang:
## Description

With the need of DDoS reputation manager, blacklist is a fundamental feature and should be provided first.
There is always a blacklist existed (singleton) in each process, and P2PComm will internally use blacklist to block all sending/receiving message with the blacklist nodes.

After this PR, the reputation manager can adjust blacklist by other functions: Add/Remove.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNetwork/Blacklist.cpp
+++ b/src/libNetwork/Blacklist.cpp
@@ -1,0 +1,61 @@
+/**
+* Copyright (c) 2018 Zilliqa
+* This source code is being disclosed to you solely for the purpose of your participation in
+* testing Zilliqa. You may view, compile and run the code for that purpose and pursuant to
+* the protocols and algorithms that are programmed into, and intended by, the code. You may
+* not do anything else with the code without express permission from Zilliqa Research Pte. Ltd.,
+* including modifying or publishing the code (or any part of it), and developing or forming
+* another public or private blockchain network. This source code is provided ‘as is’ and no
+* warranties are given as to title or non-infringement, merchantability or fitness for purpose
+* and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+* Some programs in this code are governed by the GNU General Public License v3.0 (available at
+* https://www.gnu.org/licenses/gpl-3.0.en.html) (‘GPLv3’). The programs that are governed by
+* GPLv3.0 are those programs that are located in the folders src/depends and tests/depends
+* and which include a reference to GPLv3 in their program files.
+*
+* This should only be used in testnet release  only. This is to ensure the stability of testnet.
+* Mainnet will not require this function and nodes will be incentivise to perform the role
+* as member of DS committee.
+**/
+
+#include "Blacklist.h"
+
+using namespace std;
+
+Blacklist::Blacklist() {}
+
+Blacklist::~Blacklist() {}
+
+Blacklist& Blacklist::GetInstance()
+{
+    static Blacklist blacklist;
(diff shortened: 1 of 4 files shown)


src/libNetwork/Blacklist.cpp
```diff
@@ -0,0 +1,61 @@
...
+* including modifying or publishing the code (or any part of it), and developing or forming
+* another public or private blockchain network. This source code is provided ‘as is’ and no
+* warranties are given as to title or non-infringement, merchantability or fitness for purpose
+* and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+* Some programs in this code are governed by the GNU General Public License v3.0 (available at
+* https://www.gnu.org/licenses/gpl-3.0.en.html) (‘GPLv3’). The programs that are governed by
+* GPLv3.0 are those programs that are located in the folders src/depends and tests/depends
+* and which include a reference to GPLv3 in their program files.
+*
+* This should only be used in testnet release  only. This is to ensure the stability of testnet.
```
bb111189: Remove this para

src/libNetwork/Blacklist.h
```diff
@@ -0,0 +1,63 @@
...
+* including modifying or publishing the code (or any part of it), and developing or forming
+* another public or private blockchain network. This source code is provided ‘as is’ and no
+* warranties are given as to title or non-infringement, merchantability or fitness for purpose
+* and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+* Some programs in this code are governed by the GNU General Public License v3.0 (available at
+* https://www.gnu.org/licenses/gpl-3.0.en.html) (‘GPLv3’). The programs that are governed by
+* GPLv3.0 are those programs that are located in the folders src/depends and tests/depends
+* and which include a reference to GPLv3 in their program files.
+*
+* This should only be used in testnet release  only. This is to ensure the stability of testnet.
```
bb111189: Remove this para

src/libNetwork/Blacklist.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "Blacklist.h"
+
+using namespace std;
+
+Blacklist::Blacklist() {}
+
+Blacklist::~Blacklist() {}
+
+Blacklist& Blacklist::GetInstance()
```
Gnnng: Any reason to use the singleton pattern here? It will be easier to do unit test if we can avoid it.

ckyang: The reason using singleton here is because singleton can avoid user creating more than 2 instances unintentionally. For blacklist/whitelist, it would be unreasonable if there exist more than 2 blacklist/whitelist. I agree singleton would make unit test look a little bit weird, however in this case, singleton may avoid potential misused.

Gnnng: How about adding this Blacklist class into other existing singleton classes and get the reference to the only copy there?

bb111189: Maybe we can add to mediator?

ckyang: OK, I will put one blacklist instance inside mediator.

ckyang: So far mediator is not singleton, thus if blacklist is inside mediator, it cannot be called from P2PComm static function. Hence, we just skip this issue now.

PR #523, opened 2018-08-16
Make SafeMath as template

ckyang:
## Description

As https://github.com/Zilliqa/Issues/issues/152 described, SafeMath should be a type-independent class. Hence we make it as template, and at the same time, add some unit test cases for it.

Known issue: (fixed)
Overflow-checking of add() only works on uint8_t and uint16_t, not work on bigger data size.

### Implementation

- [x] **ready for review**

- [x] unit test
- [x] local machine test

- [x] large-scale cloud test (commit: e71c2ec)

### DIFF
--- a/src/libData/AccountData/Account.cpp
+++ b/src/libData/AccountData/Account.cpp
@@ -601,7 +601,7 @@ int Account::DeserializeDelta(const vector<unsigned char>& src,
 
 bool Account::IncreaseBalance(const uint256_t& delta)
 {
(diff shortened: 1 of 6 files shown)


jiayaoqijia:
We need 200-node AWS test for every PR, to ensure there's no stability issue.

ckyang:
I just done the large-scale aws of 190 nodes with about 40 ds epoch, it seems good

PR #526, opened 2018-08-16
Reputation Manager

bb111189:
## Description

**Reputation Manager** 
This is part of phase 1 of DDoS and DoS defense. 

The idea is to penalize the misbehaving node’s reputation. Reputation score will also be added to good behaving nodes at the end of each DS Epoch. The reputation score is localized to each node and thereby no consensus is required in the process. In addition, a node cannot influence other nodes on the reputation score of certain nodes(s). 

When misbehaving node exceeds a certain reputation score, the node will be added to blacklist and the bad behaving node will not be able to connect to the particular node that blacklist it. Blacklist will be lifted after a period of time. 

**Rationale of reputation score localization**
When a node start behaving badly, it may behave badly to a particular node only. It may behave as a good node to the rest of the network. As such, reputation score should be localized to each node. As such, each node will have a different view of a certain node’s reputation score. 

Concretely, in a targeted attack, a malicious may choose to target the shard leader only. In this case, shard leader will reduce the reputation of the malicious node. Another shard member who is not aware of the attack, will retain the same reputation as the malicious node. 

**Defense surface**
The reputation manager is a mechanism for self defence only. Treat it as a personal network filtering mechanism using naive heuristics and scoring mechanism. 

The reputation manager is not meant for defense of the Zilliqa network as a whole. 

## Review Suggestion

Reputation manager is not connected to core protocol at the moment. It is pending P2Pcomm refactoring and profobuf serialization and deserialization. Hence, no large scale testing will be done for this PR. 

Core team members are encourage to refer to internal documentation of Reputation Manager.

### Implementation

- [x] **ready for review**

- [x] Unit tests
- [ ] ~local machine test~


ansnunez:
I think most of the public API can be converted to const& type.
Also, minor correction: THRESHHOLD -> THRESHOLD

bb111189:
@ansnunez Have updated and fix the typo.

src/libNetwork/ReputationManager.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+    int32_t NewRep = GetReputation(IPAddress);
+
+    // Update result with score delta
+    bool UpdateResult
+        = SafeMath<int32_t>::add(NewRep, ReputationScoreDelta, NewRep);
+    if (!UpdateResult)
+    {
+        LOG_GENERAL(WARNING, "Underflow/overflow detected.");
+    }
```
ckyang: Needless to declare a new variable the function return value here

src/libNetwork/ReputationManager.cpp
```diff
@@ -0,0 +1,164 @@
...
+    if (NewRep <= REPTHRESHOLD && !IsNodeBanned(IPAddress))
+    {
+        UpdateResult = SafeMath<int32_t>::sub(
+            NewRep, ScoreType::BAN_MULTIPLIER * ScoreType::AWARD_FOR_GOOD_NODES,
+            NewRep);
+
+        if (!UpdateResult)
+        {
+            LOG_GENERAL(WARNING, "Underflow detected.");
+        }
```
ckyang: Needless to declare a new variable the function return value here

src/libNetwork/ReputationManager.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+    return AllKnownIPs;
+}
+
+void ReputationManager::AwardNode(
+    const boost::multiprecision::uint128_t& IPAddress)
+{
+    UpdateReputation(IPAddress, ScoreType::AWARD_FOR_GOOD_NODES);
+
+    if (Blacklist::GetInstance().Exist(IPAddress) and !IsNodeBanned(IPAddress))
```
digiront: I would replace `and` with `&&`

tests/Network/Test_ReputationManager.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <arpa/inet.h>
+#include <boost/multiprecision/cpp_int.hpp>
+#include <boost/test/unit_test.hpp>
+#include <limits.h>
+
+using namespace std;
+
+BOOST_AUTO_TEST_SUITE(reputationManager)
+
+BOOST_AUTO_TEST_CASE(test_fundamental)
```
digiront: It would be nice to split this into multiple test cases (for readability and maintainability)

src/libNetwork/ReputationManager.cpp
```diff
@@ -0,0 +1,164 @@
...
+    for (const auto& ip : AllKnownIPs)
+    {
+        AwardNode(ip);
+    }
+}
+
+void ReputationManager::AddNodeIfNotKnown(
+    const boost::multiprecision::uint128_t& IPAddress)
+{
+    std::lock_guard<std::mutex> lock(m_mutexReputations);
```
digiront: Might be worth having two versions of this method. One public that acquires the `m_mutexReputations` and one used privately that doesn't. Then you can use it in other methods like `GetReputation`, `SetReputation` *after* locking this avoid locking twice.

PR #529, opened 2018-08-17
[Refactoring] Redesign P2PComm around lock-free message queues

ansnunez:
## Description

Message receipt now uses a message queue, but processing has been preserved as multi-threaded for now until the control thread is put in place.

* Previous way:
Connection established -> recv_pool thread reads in message -> dispatch and execute

* New way:
Connection established -> recv_pool thread reads in message -> push to queue
Dedicated thread -> pop from queue -> spawn new thread -> dispatch and execute

* Future way:
Dedicated thread -> pop from queue -> dispatch and execute

Message sending now uses a message queue to serially prepare messages before putting into the send pool.

* Previous way:
Send message request -> send_pool thread prepares message -> send message

* New way:
Send message request -> prepare message -> push to queue
Dedicated thread -> pop from queue -> send_pool thread -> send message

Additional update:
Reading now uses bufferevent. This means we don't manually read in the bytes anymore.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNetwork/P2PComm.cpp
```diff
@@ -555,18 +587,32 @@ void P2PComm::ConnectionAccept([[gnu::unused]] evconnlistener* listener,
...
                                BroadcastListFunc broadcast_list_retriever)
 {
     LOG_MARKER();
 
+    // Launch the thread that reads messages from the send queue
+    auto funcCheckSendQueue = [this]() mutable -> void {
+        SendJob* job = NULL;
+        while (true)
+        {
+            while (m_sendQueue.pop(job))
```
ckyang: We could change while() to if(), it's equivalent

src/libZilliqa/Zilliqa.cpp
```diff
@@ -73,6 +113,19 @@ Zilliqa::Zilliqa(const std::pair<PrivKey, PubKey>& key, const Peer& peer,
 {
     LOG_MARKER();
 
+    // Launch the thread that reads messages from the queue
+    auto funcCheckMsgQueue = [this]() mutable -> void {
+        pair<vector<unsigned char>, Peer>* message = NULL;
+        while (true)
+        {
+            while (m_msgQueue.pop(message))
```
ckyang: We could change while() to if(), it's equivalent

src/libNetwork/P2PComm.cpp
```diff
@@ -360,193 +288,297 @@ bool P2PComm::SendMessageSocketCore(const Peer& peer,
...
+                      << "] RECV");
 
-    LOG_STATE(
-        "[BROAD][" << std::setw(15) << std::left
-                   << P2PComm::GetInstance().m_selfPeer << "]["
-                   << DataConversion::Uint8VecToHexStr(msg_hash).substr(0, 6)
-                   << "] RECV");
+            // Move the shared_ptr message to raw pointer type
+            pair<vector<unsigned char>, Peer>* raw_message
+                = new pair<vector<unsigned char>, Peer>(
```
ckyang: Is it possible to use make_pair?

src/libNetwork/P2PComm.cpp
```diff
@@ -360,193 +288,297 @@ bool P2PComm::SendMessageSocketCore(const Peer& peer,
...
-    // Dispatch message normally
-    m_dispatcher(message, from);
+            // Queue the message
+            m_dispatcher(raw_message);
+        }
+        else if (startByte == START_BYTE_NORMAL)
+        {
+            // Move the shared_ptr message to raw pointer type
+            pair<vector<unsigned char>, Peer>* raw_message
+                = new pair<vector<unsigned char>, Peer>(
```
ckyang: Is it possible to use make_pair?

src/libNetwork/P2PComm.cpp
```diff
@@ -360,193 +288,297 @@ bool P2PComm::SendMessageSocketCore(const Peer& peer,
...
-        return;
+        LOG_GENERAL(WARNING, "Error from bufferevent");
     }
 
-    // Check if this message has been received before
-    vector<unsigned char> message;
-    bool found = false;
+    // All bytes read out
+    if (events & (BEV_EVENT_EOF | BEV_EVENT_ERROR))
     {
```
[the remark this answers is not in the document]
ansnunez: Will do

src/libNetwork/P2PComm.cpp
```diff
@@ -360,193 +288,297 @@ bool P2PComm::SendMessageSocketCore(const Peer& peer,
...
+                                       evutil_socket_t cli_sock,
+                                       struct sockaddr* cli_addr,
+                                       [[gnu::unused]] int socklen,
+                                       [[gnu::unused]] void* arg)
 {
-    Peer from(uint128_t(((struct sockaddr_in*)cli_addr)->sin_addr.s_addr),
-              ((struct sockaddr_in*)cli_addr)->sin_port);
+    Peer* from
+        = new Peer(uint128_t(((struct sockaddr_in*)cli_addr)->sin_addr.s_addr),
+                   ((struct sockaddr_in*)cli_addr)->sin_port);
```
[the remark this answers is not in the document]
ansnunez: Will do

src/libNetwork/P2PComm.cpp
ansnunez: Since they're equivalent, would prefer to keep it this way, because the inner while makes it clear that what we want is to keep popping while there is something left to pop, and the outer while makes it clear that after everything has been popped we want to try again. Using if instead of while doesn't make it as clear I think.

src/libNetwork/P2PComm.cpp
ansnunez: Can't use make_pair here since what we want to construct is a heap object, not a local.

src/libNetwork/P2PComm.cpp
```diff
@@ -422,130 +451,136 @@ void P2PComm::HandleAcceptedConnection(int cli_sock, Peer from)
...
-
-    m_dispatcher(message, from);
-}
-
-void P2PComm::HandleAcceptedConnectionBroadcast(int cli_sock, Peer from,
-                                                uint32_t message_length,
-                                                SocketCloser cli_sock_closer)
-{
-    vector<unsigned char> msg_hash;
+    int fd = bufferevent_getfd(bev);
```
Gnnng: `bev` is already being freed

src/libNetwork/P2PComm.cpp
```diff
@@ -378,13 +391,29 @@ void P2PComm::ClearBroadcastHashAsync(const vector<unsigned char>& message_hash)
...
+        return;
+    }
 
-    SocketCloser cli_sock_closer(&cli_sock, close_socket);
+    // Get the data stored in buffer
+    struct evbuffer* input = bufferevent_get_input(bev);
+    size_t len = evbuffer_get_length(input);
+    vector<unsigned char> message(len);
+    evbuffer_copyout(input, message.data(), len);
+    evbuffer_drain(input, len);
```
Gnnng: error handling for the return value of `evbuffer_drain`

src/libNetwork/P2PComm.cpp
```diff
@@ -422,130 +451,136 @@ void P2PComm::HandleAcceptedConnection(int cli_sock, Peer from)
...
-            }
-
-            SHA2<HASH_TYPE::HASH_VARIANT_256> sha256;
-            sha256.Update(message);
-            vector<unsigned char> this_msg_hash = sha256.Finalize();
+            lock_guard<mutex> guard(
+                P2PComm::GetInstance().m_broadcastHashesMutex);
 
-            if (this_msg_hash == msg_hash)
+            found = (P2PComm::GetInstance().m_broadcastHashes.find(msg_hash)
```
Gnnng: If `P2PComm::GetInstance()` is called multiple times in this function, we can simply declare a variable with a short name to hold the reference.

src/libNetwork/P2PComm.cpp
```diff
@@ -378,13 +391,29 @@ void P2PComm::ClearBroadcastHashAsync(const vector<unsigned char>& message_hash)
...
+    if (events & BEV_EVENT_ERROR)
+    {
+        LOG_GENERAL(WARNING, "Error from bufferevent");
+    }
 
-    LOG_GENERAL(INFO, "Incoming message from " << from);
+    // Not all bytes read out
+    if (!(events & (BEV_EVENT_EOF | BEV_EVENT_ERROR)))
+    {
+        return;
```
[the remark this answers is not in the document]
ansnunez: Thanks, missed that after doing the last revision

src/libNetwork/P2PComm.cpp
```diff
@@ -555,18 +590,32 @@ void P2PComm::ConnectionAccept([[gnu::unused]] evconnlistener* listener,
...
     }
 
-    auto func = [cli_sock, from]() -> void {
-        HandleAcceptedConnection(cli_sock, from);
-    };
-
-    GetInstance().m_RecvPool.AddJob(func);
+    // Set up buffer event for this new connection
+    struct event_base* base = evconnlistener_get_base(listener);
+    struct bufferevent* bev = bufferevent_socket_new(
```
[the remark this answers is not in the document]
ansnunez: Got it

src/libNetwork/P2PComm.cpp
ansnunez: Thanks, missed that after doing the last revision

src/libNetwork/P2PComm.cpp
ansnunez: Got it

src/libNetwork/P2PComm.cpp
```diff
@@ -555,18 +590,32 @@ void P2PComm::ConnectionAccept([[gnu::unused]] evconnlistener* listener,
...
+        base, cli_sock, BEV_OPT_CLOSE_ON_FREE | BEV_OPT_DEFER_CALLBACKS);
+    bufferevent_setcb(bev, NULL, NULL, EventCallback, NULL);
```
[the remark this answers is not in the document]
ansnunez: Yep, that makes sense. I agree it can be done in next optimization work.

ansnunez:
Hi @Gnnng thanks for the thorough review. The reason for the raw pointers is that Boost's lock-free queue requires that the contained class have a trivial destructor. Smart pointer types don't qualify in this respect. We'll have to search for an alternative library if we want to be able to use smart pointers. For now, I think as long as we have the code to delete the entries (after popping and processing, and during P2PComm and Zilliqa instance destruction) we'll be ok.

src/libNetwork/P2PComm.cpp
ansnunez: Got it. Will add more error handling for the rest too.

ckyang:
Suggest to do large-scale testnet before merging

src/libNetwork/P2PComm.cpp
```diff
@@ -394,15 +394,8 @@ void P2PComm::ClearBroadcastHashAsync(const vector<unsigned char>& message_hash)
...
-    // Get the IP info
-    int fd = bufferevent_getfd(bev);
-    struct sockaddr_in cli_addr;
-    socklen_t addr_size = sizeof(struct sockaddr_in);
-    getpeername(fd, (struct sockaddr*)&cli_addr, &addr_size);
-    Peer from(cli_addr.sin_addr.s_addr, cli_addr.sin_port);
-
-    // Close the socket
-    bufferevent_free(bev);
+    unique_ptr<struct bufferevent, void (*)(struct bufferevent*)> socket_closer(
```
Gnnng: for simplicity and better connection with the free function, we could use `decltype(bufferevent_free)` here

ansnunez:
decltype done. large-scale testnet working fine.

PR #537, opened 2018-08-23
[Graceful Termination] Implement signal handling using libevent

Gnnng:
## Description

This PR adds support for signal handling to terminate the process using <kbd>Ctrl</kbd>-<kbd>C</kbd> or other signals. It's prepared for the graceful termination.

The implementation is based on libevent's signal handling. The signal to be used for exiting is first registered using `RegisterExitSignal` which adds the signal event to the event base (`m_eventBase`) and binds the callback function `OnExitingSignal` to the event. 

A small refactoring is also included in this PR to adopt RAII in libevent-related resource management, including the event base (`m_eventBase`) and signal events (`m_signalEvents`).

The test case `Test_Signal` shows a minimal working example. Due to an unknown issue in CTest on macOS, the test case itself only works on Linux build so a branch condition is added in 
`tests/Network/CMakeLists.txt`. Despite of this issue, the signal handling still works fine when not running in CTest.

## Review Suggestion

Start from the test case `Test_Signal` to have a brief understanding of the idea. Then move to the newly added data member and functions for `P2PComm`. 

If you prefer having a try on the termination for a single node, follow these steps:

1. `./build.sh`
2. `cd build`
3. `./bin/zilliqa $(./bin/genkeypair | awk '{ printf $2 " " $1 }') 127.0.0.1 30303 0 0 0`
4. <kbd>Ctrl</kbd>-<kbd>C</kbd>
5. `tail zilliqa-00001-log.txt`
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNetwork/P2PComm.cpp
+++ b/src/libNetwork/P2PComm.cpp
@@ -80,6 +80,7 @@ static bool comparePairSecond(
 
 P2PComm::P2PComm()
     : m_sendQueue(SENDQUEUE_SIZE)
+    , m_eventBase(event_base_new(), event_base_free)
 {
     auto func = [this]() -> void {
         std::vector<unsigned char> emptyHash;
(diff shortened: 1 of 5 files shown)


src/libNetwork/P2PComm.cpp
```diff
@@ -642,6 +643,34 @@ void P2PComm::AcceptConnectionCallback([[gnu::unused]] evconnlistener* listener,
...
+void P2PComm::RegisterExitSignal(std::vector<int>&& sigs)
+{
+    for (const auto& sig : sigs)
+    {
+        auto signal_event = std::make_unique<struct event>();
+        event_assign(signal_event.get(), m_eventBase.get(), sig,
+                     EV_SIGNAL | EV_PERSIST, OnExitingSignal,
+                     signal_event.get());
+        event_add(signal_event.get(), nullptr);
+        // push it to m_sigalEvents to let P2PComm destructor to handle the deallocation
```
bb111189: small typo here. `m_signalEvents`

Gnnng:
This PR is obsolete. Reopen when we need this again.

PR #545, opened 2018-08-27
Dual PoW difficulty levels for DS committee and shards

bb111189:
## Note
Do not merge till kubernetes configuration is also updated 

## Description

With the merging of PoW1 and PoW2, there is currently only 1 PoW required in the protocol. This PR implements the mechanism for choosing new node into the DS committee. 

In this implementation, there will be two difficulty levels, known as DS difficulty and difficulty (for shard). Both will use the same PoW process. The former is expected to be harder than the later. 

- If a node solution meets the difficulty level, it may be chosen into shard. 
- If a node solution meets the DS difficulty level, it may be chosen to be part of the DS committee (if not, it may be chosen into shard). 
- A node may submit multiple solution*, but the solution with the hash value being the least, will be taken into account. 

\* Spammer will be blocked by reputation manager and a max submission limit.

If multiple nodes solution meet DS difficulty level, the solution with the least hash value will chosen into the DS committee. 

Todo:
- ~Node to send the difficulty level at which pow is performed~
- ~DS to verify with the correct difficulty level~
- Shard to verify ds leader + sharding structure

Other changes:
- Remove magic numbers from genesis ds block
- Rename `shouldMine` to `m_shouldMine` in libPoW
- Change `m_shouldMine` to `atomic`
- Refactor `StartPoW()` to separate the logic of computng PoW and Sending out PoW message
- Cleanup `libNode/PowProcessing.cpp`
- Remove duplicated mutex `m_mutexAllPoWs`  and unused mutex `m_mutexDSBlockConsensus` (Fix https://github.com/Zilliqa/Issues/issues/180) 

Other todos:
- Update internal testnet constants file
- Update explorer to display dual difficulty
- Dynamic DS difficulty level
- Validator for ds leader

Known issue (for core team reference):
- https://github.com/Zilliqa/Issues/issues/179
- https://github.com/Zilliqa/Issues/issues/193
- https://github.com/Zilliqa/Issues/issues/194

A follow up PR will implements multiple DS nodes joining and leaving the DS committee. 

## Review Suggestion

Do a local run and check CPU utilitization

### Implementation

- [x] **ready for review**

- [x] local machine test (commit: 94e02a3ae3057598287dfcb8286bb327d14b0215)
- [x] medium-scale cloud test (commit: 94e02a3ae3057598287dfcb8286bb327d14b0215)
- [x] medium-scale cloud test (commit: aaf6507558df6a6ccda27899f41413839a275502)
- [x] medium-scale cloud test (commit: 5f8df331ad9a88486da7e1f3ab964cf5eacca035)


src/libNode/PoWProcessing.cpp
```diff
@@ -84,59 +81,109 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t difficulty,
...
-        cur_offset += BLOCK_HASH_SIZE;
-        powmessage.insert(powmessage.end(), mixhash_vec.begin(),
-                          mixhash_vec.end());
-        cur_offset += BLOCK_HASH_SIZE;
-
-        Signature sign;
-        if (!Schnorr::GetInstance().Sign(powmessage, m_mediator.m_selfKey.first,
-                                         m_mediator.m_selfKey.second, sign))
+        if (POW::GetInstance().CheckSolnAgainstsTargetedDifficulty(
+                (std::string)winning_result.result, difficulty))
```
ShengguangXiao: Does here need to check against DS difficulty?

src/libNode/PoWProcessing.cpp
```diff
@@ -84,59 +81,109 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t difficulty,
...
         {
-            LOG_GENERAL(WARNING, "Failed to sign PoW");
+            LOG_GENERAL(INFO,
+                        "soln does not meet ds committee criteria. Will keep "
+                        "doing more pow");
+
+            ethash_mining_result ds_pow_winning_result
+                = POW::GetInstance().PoWMine(
+                    block_num, difficulty, rand1, rand2,
```
ShengguangXiao: Does the 2nd round mine need to use DS difficulty?

src/libNode/PoWProcessing.cpp
ShengguangXiao: Here should check against DS difficulty?

src/libNode/PoWProcessing.cpp
ShengguangXiao: Here should use DS difficulty?

bb111189: You are right! Didn't notice it. Thanks

src/libDirectoryService/DirectoryService.h
```diff
@@ -73,10 +73,11 @@ class DirectoryService : public Executable, public Broadcastable
     std::condition_variable cv_scheduleFinalBlockConsensus;
 
     // PoW common variables
-    std::mutex m_mutexAllPoWs;
-    std::map<PubKey, Peer> m_allPoWConns;
     std::mutex m_mutexAllPoWConns;
+    std::map<PubKey, Peer> m_allPoWConns;
     std::vector<unsigned char> m_PoWConsensusMessage;
+    std::mutex m_mutexAllPoWCounter;
+    std::map<PubKey, uint8_t> m_AllPoWCounter;
```
ShengguangXiao: What does this m_AllPoWCounter use for?

bb111189: number of PoW submissions. As there is 2 level of difficult, a node may submit twice or more. The counter is there to track the number of submission and prevent spamming of submission

src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -90,30 +90,61 @@ bool DirectoryService::VerifyPoWSubmission(
...
+                            << " Expected: " << to_string(expectedDSDiff)
+                            << " or " << to_string(expectedDiff));
+
+            // TODO: penalise sender in reputation manager
+            return false;
+        }
+    }
+    else if (block_num == 1)
+    {
+        if (difficultyLevel != POW_DIFFICULTY
```
KaustubhShamshery: What is this check for?

src/libNode/PoWProcessing.cpp
```diff
@@ -91,63 +89,131 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t difficulty,
...
-            = {MessageType::DIRECTORY, DSInstructionType::POWSUBMISSION};
-        unsigned int cur_offset = MessageOffset::BODY;
+        // Possibles scenario
+        // 1. Found solution that meet ds difficulty and difficulty
+        // - Submit solution
+        // 2. Found solution that met only diffiulty
+        // - Submit solution and continue to do PoW till DS difficulty met or
+        //   ds block received. (stopmining())
+        if (POW::GetInstance().CheckSolnAgainstsTargetedDifficulty(
+                (std::string)winning_result.result, ds_difficulty))
```
ShengguangXiao: The winning_result.result is already std::string type, here no need to do a conversion.

src/libNode/PoWProcessing.cpp
```diff
@@ -91,63 +89,131 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t difficulty,
...
-        P2PComm::GetInstance().SendMessage(peerList, powmessage);
+    Signature sign;
+    if (!Schnorr::GetInstance().Sign(powmessage, m_mediator.m_selfKey.first,
+                                     m_mediator.m_selfKey.second, sign))
+    {
+        LOG_GENERAL(WARNING, "Failed to sign PoW");
     }
+    sign.Serialize(powmessage, cur_offset);
 
+    deque<Peer> peerList;
```
ShengguangXiao: If only need to push_back, why don't use vector instead of deque, vector if more efficient if only need to push_back.

src/libDirectoryService/PoWProcessing.cpp
bb111189: there is two difficulty level. To ensure ds committee don't waste time verifying on the difficulty, the node will declare the difficulty used. If it matched (strictly equal) either ds diff or diff, the ds committee will need to proceed to verify at the respective diff level

src/libNode/PoWProcessing.cpp
bb111189: good poing

src/libNode/PoWProcessing.cpp
bb111189: alright. will do.

PR #550, opened 2018-08-30
Feature/remove look up macro dependency

### LINKED ISSUE #240: Use CMake multiple targets to build lookup
We can simply add a new target for lookup build instead of using separate tooling: https://stackoverflow.com/questions/6665887/different-compile-flags-for-same-file-in-different-targets

This not only removes the redundancy of `build_lookup.sh`, but also improves the maintainability. 

We could also take this chance to sort out the use of `#ifdef IS_LOOKUP_NODE`.

### LINKED ISSUE #514: Avoid using MACRO like IS_LOOKUP_NODE
We can create another constant (read it from the constant.xml file) and if-else condition to replace #ifndef IS_LOOKUP_NODE.

sandipbhoir:
## Description
Remove the dependency of preprocessor  - "IS_LOOKUP_NODE" from the code base.
Instead we use the new flag - "LOOKUP_NODE_MODE" read from constants.xml.

In order to run the normal node , set the <LOOKUP_NODE_MODE>false</LOOKUP_NODE_MODE>.
In order to run the lookup node , set the <LOOKUP_NODE_MODE>true</LOOKUP_NODE_MODE>.

We need not use separate builds for lookup and normal node anymore.

Moving this new flag ( few existing ones as well ) from constant.xml as an command-line option is something to considered with separate PR.

This PR will close #240 and fix #514.
### Implementation

- [x] To modify the test_node_simple.sh and test_node_lookup.h such that it SET the correct flag for LOOKUP_NODE_MODE in constant_local.xml. This will avoid manual modification of constant.xml required for local test.

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


Gnnng:
Hey @sandipbhoir, I just added one more commit to make some changes in CI/CD.

src/depends/libDatabase/LevelDB.cpp
```diff
@@ -99,6 +75,14 @@ leveldb::Slice toSlice(boost::multiprecision::uint256_t num)
     return (leveldb::Slice)h.ref();
 }
 
+string LevelDB::GetDBName() 
+{
+        if (LOOKUP_NODE_MODE)
```
ansnunez: For consistency with existing style, please enclose branched code in braces, even if it's a single line.
Reference: https://github.com/Zilliqa/Zilliqa/wiki/Coding-Guidelines#formatting-and-coding-style

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -974,27 +1086,27 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
...
         StoreFinalBlock(txBlock);
 
-#ifndef IS_LOOKUP_NODE
-        BlockStorage::GetBlockStorage().PutMetadata(MetaType::DSINCOMPLETED,
-                                                    {'0'});
-        BlockStorage::GetBlockStorage().PopFrontTxBodyDB();
-#endif // IS_LOOKUP_NODE
+        if (!LOOKUP_NODE_MODE)
+        {
+            BlockStorage::GetBlockStorage().PopFrontTxBodyDB();
```
ansnunez: I don't know how this code works, but just wondering why the order of the two lines was flipped.

src/libNode/Node.h
```diff
@@ -283,7 +281,7 @@ class Node : public Executable, public Broadcastable
     bool ProcessVCBlock(const vector<unsigned char>& message,
                         unsigned int cur_offset, const Peer& from);
 
-#ifndef IS_LOOKUP_NODE
+    //#ifndef IS_LOOKUP_NODE -- normal nodes specific
```
ansnunez: Can probably just remove these lines, to be consistent with the other header file changes?

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -487,6 +534,14 @@ bool DirectoryService::DSBlockValidator(
 
 bool DirectoryService::RunConsensusOnDSBlockWhenDSBackup()
 {
+    if (LOOKUP_NODE_MODE)
+    {
+        LOG_GENERAL(WARNING,
+                    "DirectoryService::RunConsensusOnDSBlockWhenDSBackup not "
+                    "expected to be called from LookUp node.");
+        return true;
```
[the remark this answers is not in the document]
sandipbhoir: Most of the existing functions which are not relevant to Lookup nodes but are callable for them return true. So as to be inline with it, better to stick with returning true. However, in general they should not be called by code path of lookup nodes.

ansnunez:
For consistency, we should probably keep `LOG_MARKER();` as the first line in each function, instead of the check for `LOOKUP_NODE_MODE`. But I would say not mandatory to change right now.

ansnunez:
src/libNode/Coinbase.cpp was removed from master, and I think unintentionally put back in your latest commit

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -105,6 +112,14 @@ void DirectoryService::StartSynchronization()
 
 bool DirectoryService::CheckState(Action action)
 {
+    if (LOOKUP_NODE_MODE)
+    {
+        LOG_GENERAL(WARNING,
+                    "DirectoryService::CheckState not "
+                    "expected to be called from LookUp node.");
+        return true;
```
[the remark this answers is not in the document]
sandipbhoir: Almost all of the existing functions return true by default when its not lookup node, so try to stick to same for above changes.

PR #549, opened 2018-08-29
DS Sort the Sharding Structure based on the pow winning hash.

KaustubhShamshery:
## Description

The ds leader propose the sharding structure sorted by H(hash_of_last_final_block, pow_winning_hash) and then each ds backup independently verify the uniqueness and ordering.

### Implementation

- [x] **ready for review**

### Integration Test (Core Team)

- [x] local machine test

- [x] small-scale cloud test

- [x] 200 node cloud test (commit: 3c04615)

PR #552, opened 2018-08-30
Randomized DS Leader selection

KaustubhShamshery:
## Description
Earlier, the DS Leader would be the PoW winner, but now the DS leader is elected randomly with modular operations on the lastTxBlockHash

## Review Suggestion

Dependent  : #549 

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -212,23 +212,8 @@ void DirectoryService::SendDSBlockToShardNodes(unsigned int my_shards_lo,
 
 void DirectoryService::UpdateMyDSModeAndConsensusId()
 {
-    // If I was DS primary, now I will only be DS backup
-    if (m_mode == PRIMARY_DS)
-    {
-        LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-                  "I am now just a backup DS");
-        LOG_EPOCHINFO(to_string(m_mediator.m_currentEpochNum).c_str(),
-                      DS_BACKUP_MSG);
-        m_mode = BACKUP_DS;
-        m_consensusMyID++;
-
-        LOG_STATE("[IDENT][" << setw(15) << left
-                             << m_mediator.m_selfPeer.GetPrintableIPAddress()
-                             << "][" << setw(6) << left << m_consensusMyID
-                             << "] DSBK");
-    }
     // Check if I am the oldest backup DS (I will no longer be part of the DS committee)
-    else if ((uint32_t)(m_consensusMyID + 1)
+    if ((uint32_t)(m_consensusMyID + 1)
              == m_mediator.m_DSCommittee->size())
     {
         LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
@@ -241,7 +226,39 @@ void DirectoryService::UpdateMyDSModeAndConsensusId()
                              << m_mediator.m_selfPeer.GetPrintableIPAddress()
                              << "][      ] IDLE");
     }
-    // Other DS nodes continue to remain DS backups
+
+    else
+    {
+        uint32_t dsIndex = lastBlockHash  %  (m_mediator.m_DSCommittee->size());
+        //if dsIndex == 0 , that means the pow Winner is the DS Leader
+        if(dsIndex > 0 && m_mediator.m_DSCommittee->at(dsIndex - 1).first == m_mediator.m_selfKey.second)
+        {
+            LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
+                  "I am now Leader DS");
+            LOG_EPOCHINFO(to_string(m_mediator.m_currentEpochNum).c_str(),
+                      DS_LEADER_MSG);
+            m_mode = PRIMARY_DS;
+        }
+        else
+        {
+            LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
+                  "I am now backup DS");
+            LOG_EPOCHINFO(to_string(m_mediator.m_currentEpochNum).c_str(),
+                      DS_BACKUP_MSG);
+            m_mode = BACKUP_DS;
+        }
+        
+        
+        m_consensusMyID++;
+
+        LOG_STATE("[IDENT][" << setw(15) << left
+                             << m_mediator.m_selfPeer.GetPrintableIPAddress()
+                             << "][" << setw(6) << left << m_consensusMyID
+                             << "] DSBK");
+    }
+    
+    
+    /*// Other DS nodes continue to remain DS backups
     else
     {
         m_consensusMyID++;
@@ -250,7 +267,7 @@ void DirectoryService::UpdateMyDSModeAndConsensusId()
                              << m_mediator.m_selfPeer.GetPrintableIPAddress()
                              << "][" << setw(6) << left << m_consensusMyID
                              << "] DSBK");
-    }
+    }*/
 }
 
 void DirectoryService::UpdateDSCommiteeComposition(const Peer& winnerpeer)
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -32,6 +32,7 @@
 #include "libNetwork/Whitelist.h"
 #include "libUtils/DataConversion.h"
 #include "libUtils/DetachedFunction.h"
+#include "libUtils/HashUtils.h"
 #include "libUtils/Logger.h"
 #include "libUtils/SanityChecks.h"
 
@@ -213,8 +214,9 @@ void DirectoryService::SendDSBlockToShardNodes(unsigned int my_shards_lo,
 void DirectoryService::UpdateMyDSModeAndConsensusId()
 {
     // Check if I am the oldest backup DS (I will no longer be part of the DS committee)
-    if ((uint32_t)(m_consensusMyID + 1)
-             == m_mediator.m_DSCommittee->size())
+    uint16_t lastBlockHash = HashUtils::SerializableToHash16Bits(
+        m_mediator.m_txBlockChain.GetLastBlock());
+    if ((uint32_t)(m_consensusMyID + 1) == m_mediator.m_DSCommittee->size())
     {
         LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
                   "I am the oldest backup DS -> I am now just a shard node"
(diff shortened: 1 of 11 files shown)


src/libUtils/HashUtils.h
```diff
@@ -0,0 +1,57 @@
...
+* Some programs in this code are governed by the GNU General Public License v3.0 (available at 
+* https://www.gnu.org/licenses/gpl-3.0.en.html) (‘GPLv3’). The programs that are governed by 
+* GPLv3.0 are those programs that are located in the folders src/depends and tests/depends 
+* and which include a reference to GPLv3 in their program files.
+**/
+
+#ifndef __HASH_UTILS__
+#define __HASH_UTILS__
+
+#include "DataConversion.h"
```
ansnunez: Not sure why this header is needed

src/libUtils/HashUtils.h
```diff
@@ -0,0 +1,57 @@
...
+        sha2.Update(vec);
+        const vector<unsigned char>& resVec = sha2.Finalize();
+
+        return resVec;
+    }
+    static uint16_t SerializableToHash16Bits(const Serializable& sz)
+    {
+        const vector<unsigned char>& vec = SerializableToHash(sz);
+
+        uint32_t size = vec.size() - 1;
```
ansnunez: `size` a bit misleading, better name might be `lsb`.  But not too important right now, so hold off on changing this unless other reviewers have requested changes to this file too.

src/libNode/DSBlockProcessing.cpp
```diff
@@ -528,23 +530,43 @@ bool Node::ProcessDSBlock(const vector<unsigned char>& message,
             m_mediator.m_ds->ProcessTxnBodySharingAssignment();
 
             // Update my DS mode and ID
-            m_mediator.m_ds->m_consensusMyID = 0;
             m_mediator.m_ds->m_consensusID
                 = m_mediator.m_currentEpochNum == 1 ? 1 : 0;
-            m_mediator.m_ds->m_mode = DirectoryService::Mode::PRIMARY_DS;
 
-            // (We're getting rid of this eventually) Clean up my txn list since I'm a DS node now
```
ansnunez: Should keep this comment?

src/libNode/DSBlockProcessing.cpp
```diff
@@ -528,23 +530,43 @@ bool Node::ProcessDSBlock(const vector<unsigned char>& message,
...
             m_mediator.m_node->CleanCreatedTransaction();
 
-            LOG_EPOCHINFO(to_string(m_mediator.m_currentEpochNum).c_str(),
-                          DS_LEADER_MSG);
-            LOG_STATE("[IDENT]["
```
ansnunez: Should we keep this log message?

KaustubhShamshery: The point is this node might not be the DS Leader, there is a further condition. check line 552.

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -268,6 +269,13 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
         m_consensusMyID++;
     }
     m_consensusLeaderID = 0;
+    if (m_mediator.m_currentEpochNum > 1)
```
ansnunez: Or, alternatively, since SetPrimary feeds us the Peer of the leader, couldn't we just look through the current `m_DSCommittee` to know the leader index?  Or perhaps we can do both to cross-check?

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -608,6 +723,11 @@ bool DirectoryService::DSBlockValidator(
     // [Sharding structure]
     curr_offset = PopulateShardingStructure(message, curr_offset);
 
+    if (!VerifyPoWOrdering())
+    {
+        LOG_GENERAL(INFO, "Failed to verify ordering");
```
KaustubhShamshery: Need to return False here when enabling the shard verification.

src/libDirectoryService/DirectoryService.cpp
KaustubhShamshery: This check is not needed to be honest. Since SetPrimary should only be invoked in epoch 1. I just add this for symmetry.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -262,24 +263,14 @@ void DirectoryService::UpdateMyDSModeAndConsensusId()
         return;
     }
 
-    // If I was DS primary, now I will only be DS backup
-    if (m_mode == PRIMARY_DS)
+    // Check if I am the oldest backup DS (I will no longer be part of the DS committee)
```
ansnunez: This comment applies to the code further down

KaustubhShamshery: noted.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -154,18 +161,126 @@ void DirectoryService::ComputeSharding(
...
     {
         const PubKey& key = kv.second;
-        map<PubKey, Peer>& shard = m_shards.at(i % numOfComms);
-        shard.emplace(key, m_allPoWConns.at(key));
-        m_publicKeyToShardIdMap.emplace(key, i % numOfComms);
+        const array<unsigned char, BLOCK_HASH_SIZE>& powHash = kv.first;
+
+        // sort all PoW submissions according to H(last_block_hash, pow_hash)
+        vector<unsigned char> hashVec;
+        hashVec.resize(BLOCK_HASH_SIZE + BLOCK_HASH_SIZE);
```
ansnunez: For the `powHash` portion I think should use `POW_SIZE` even though technically it's the same as `BLOCK_HASH_SIZE`

KaustubhShamshery: okay, noted.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -154,18 +161,126 @@ void DirectoryService::ComputeSharding(
...
+    m_allPoWs[m_mediator.m_DSCommittee->back().first]
+        = array<unsigned char, BLOCK_HASH_SIZE>();
+
+    vector<unsigned char> hashVec;
+    bool ret = true;
+    vector<unsigned char> vec(BLOCK_HASH_SIZE);
+    for (unsigned int i = 0; i < m_shards.size(); i++)
+    {
+
+        for (auto& j : m_shards.at(i))
```
ansnunez: Can be `const`

KaustubhShamshery: noted.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -154,18 +161,126 @@ void DirectoryService::ComputeSharding(
...
+            m_mediator.m_txBlockChain.GetLastBlock());
+    }
+    //Temporarily add the old ds to check ordering
+    m_allPoWs[m_mediator.m_DSCommittee->back().first]
+        = array<unsigned char, BLOCK_HASH_SIZE>();
+
+    vector<unsigned char> hashVec;
+    bool ret = true;
+    vector<unsigned char> vec(BLOCK_HASH_SIZE);
+    for (unsigned int i = 0; i < m_shards.size(); i++)
```
ansnunez: Can use the range-based for loop

KaustubhShamshery: noted.

src/libDirectoryService/DirectoryService.h
```diff
@@ -178,6 +180,7 @@ class DirectoryService : public Executable, public Broadcastable
     void ComputeSharding(
         const vector<pair<array<unsigned char, 32>, PubKey>>& sortedPoWSolns);
     void ComputeTxnSharingAssignments(const Peer& winnerpeer);
+    bool VerifyPoWOrdering();
```
ansnunez: Should this be called as well by the DS backup?

KaustubhShamshery: it is being called. DSBlockPreProcessing.cpp

ansnunez:
@bb111189 does this PR break the view change logic?  If so, we might need to write that up as an issue.

src/libDirectoryService/DirectoryService.cpp
ansnunez: I thought about that too, but wondered if you put that there because it might also be invoked when trying to reboot the whole network (in which case epoch would be X). But yeah I don't think we have startup code for setting `m_currentEpochNum` to whatever is the last block number in persistent storage plus one.

src/libNode/DSBlockProcessing.cpp
ansnunez: I meant should we also move the `LOG_STATE` message into the appropriate location.  Seems like only the `LOG_EPOCHINFO` one was retained.

src/libDirectoryService/DirectoryService.cpp
KaustubhShamshery: I think anyways SetPrimary is contradictory to random ds selection, maybe I can put a warning log instead there if SetPrimary is called not in epoch 1.

src/libNode/DSBlockProcessing.cpp
KaustubhShamshery: okay, noted.

src/libDirectoryService/DirectoryService.h
ansnunez: Oops missed that

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -154,18 +161,126 @@ void DirectoryService::ComputeSharding(
...
+                            << DataConversion::Uint8VecToHexStr(sortHashVec)
+                            << " " << j.first);
+            if (sortHashVec < vec)
+            {
+                LOG_GENERAL(
+                    WARNING,
+                    "Failed to Verify due to bad PoW ordering "
+                        << DataConversion::Uint8VecToHexStr(vec) << " "
+                        << DataConversion::Uint8VecToHexStr(sortHashVec));
+                ret = false;
```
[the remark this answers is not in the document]
bb111189: alright. Resolved.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -154,18 +161,126 @@ void DirectoryService::ComputeSharding(
...
+                break;
+            }
+            auto r = keyset.insert(j.first);
+            if (!r.second)
+            {
+
+                LOG_GENERAL(WARNING,
+                            "The key is not unique in the sharding structure "
+                                << j.first);
+                ret = false;
```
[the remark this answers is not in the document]
bb111189: Resolved. Should return false here

src/libDirectoryService/DSBlockPreProcessing.cpp
bb111189: Resolved.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -291,16 +282,50 @@ void DirectoryService::UpdateMyDSModeAndConsensusId()
...
+                      "I am now Leader DS");
+            LOG_EPOCHINFO(to_string(m_mediator.m_currentEpochNum).c_str(),
+                          DS_LEADER_MSG);
+            m_mode = PRIMARY_DS;
+        }
+        else
+        {
+            LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
+                      "I am now backup DS");
+            LOG_EPOCHINFO(to_string(m_mediator.m_currentEpochNum).c_str(),
```
bb111189: Syntax issue

bb111189: My bad. Not a syntax issue

bb111189:
> @bb111189 does this PR break the view change logic? If so, we might need to write that up as an issue.

I don't think it will break the existing view change logic. But at design level, view change should also be change.

PR #556, opened 2018-09-04
Final Block Consensus Buffer

KaustubhShamshery:
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -857,12 +853,22 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSBackup()
...
 
+    if (revertStateDelta)
+    {
+        LOG_GENERAL(WARNING,
+                    "Failed DS microblock consensus, revert state delta");
+        AccountStore::GetInstance().InitTemp();
+        AccountStore::GetInstance().DeserializeDeltaTemp(m_stateDeltaFromShards,
+                                                         0);
+    }
+    CommitFinalBlockConsensusBuffer();
```
kaikawaliu: I think this should be called after `if(ConsensusObjCreation)`

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -896,6 +896,7 @@ void DirectoryService::RunConsensusOnFinalBlock(bool revertStateDelta)
         SetState(FINALBLOCK_CONSENSUS);
         cv_finalBlockConsensusObject.notify_all();
     }
+    CommitFinalBlockConsensusBuffer();
```
kaikawaliu: In this case the cv below to trigger view change may always times out, as the triggering happens in processing the consensus message but in the same thread. I think currently we should detach this CommitFinalBlockConsensusBuffer first in order to let the cv get notified to avoid initiating view change

src/libDirectoryService/DirectoryService.h
```diff
@@ -94,6 +94,11 @@ class DirectoryService : public Executable, public Broadcastable
     std::unordered_map<uint64_t, std::vector<std::vector<unsigned char>>>
         m_MBSubmissionBuffer;
 
+    std::mutex m_mutexFinalBlockConsensusBuffer;
+    std::unordered_map<uint64_t,
```
kaikawaliu: Actually the first key for epoch num is not essential, can be removed

KaustubhShamshery: @kaikawaliu I will remove it.

PR #544, opened 2018-08-27
Release helper for upgrading protocol

ckyang:
## Description

Following upgrading protocol utilities (https://github.com/Zilliqa/Zilliqa/pull/534), the release helper should provide following features:

- [x]  Generate software version (major, minor, fix, expectedDS, commit) to a separated file "VERSION"
- [x]  Generate deb package
- [x]  Make SHA-256, put into "VERSION"
- [x]  Make multi-signature of SHA-256 hash, put into "VERSION"
- [x]  Put the deb package & [SHA-256 & multi-sig] onto GitHub

Usage:
source ./release.sh privateKeyFileName publicKeyFileName

## Review Suggestion

Release helper is an off-line independent script that will only be executed when a release needed to be generated. No impact to current code flow, so the review focus should be identifying the correctness of generated package (.deb) and relative version information (VERSION).

### Implementation

- [x] **ready for review**

### Integration Test (Core Team)

- [x] local machine test

- [x] small-scale cloud test (commit: 9f75730)

PR #558, opened 2018-09-04
Download procedure for upgrading protocol

ckyang:
## Description

Following release helper (https://github.com/Zilliqa/Zilliqa/pull/544), the download procedure should provide following features:

- [x] HasNewSW(): Check website, verify if multi-sig is valid && SHA-256 is new
- [x] DownloadSW(): Download SW from website, then update current SHA-256 value & m_latestSWInfo
- [x] Store current DS block, final block, and states into metadata (will be called in ReplaceNode())
- [x] Deploy downloaded software, and kill current node, then the recovery procedure will wake up node with stored data (will be called in ReplaceNode())
- [x] Write unit test for DownloadSW()

### Implementation

- [x] **ready for review**

- [x] local machine test
- [x] small-scale cloud test
- [x] unit test

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -13,7 +13,12 @@
 # GPLv3.0 are those programs that are located in the folders src/depends and tests/depends
 # and which include a reference to GPLv3 in their program files.
 
-# User configuration settings, mandatory to be filled in
+#######################################################################################
+# This script should be only used by Zilliqa Research,                                #
+# for releasing a draft version binary with relative version information onto GitHub. #
+#######################################################################################
+
+# [MUST BE FILLED IN] User configuration settings
 GitHubToken=""
 repoName="PreRelease"    # Change to Zilliqa after PreRelease is ok
 packageName=""
--- a/release.sh
+++ b/release.sh
@@ -107,7 +107,7 @@ echo -e "Making SHA-256 & multi-signature..."
 privKeyFile="$(realpath $1)"
 pubKeyFile="$(realpath $2)"
 cd ${releaseDir}
-sha="$(md5sum ${debFile}|cut -d ' ' -f1)"
+sha="$(sha256sum ${debFile}|cut -d ' ' -f1)"
 sed -i "${shaLine}s/.*/${sha}/" ${versionFile}
 signature="$(./bin/signmultisig ${sha} ${privKeyFile} ${pubKeyFile})"
 sed -i "${sigLine}s/.*/${signature}/" ${versionFile}
--- a/src/libUtils/UpgradeManager.cpp
+++ b/src/libUtils/UpgradeManager.cpp
@@ -15,9 +15,13 @@
 **/
 
 #include "UpgradeManager.h"
+#include "libCrypto/Schnorr.h"
 #include "libUtils/Logger.h"
 
 using namespace std;
+
+#define RELEASE_URL                                                            \
+    "https://api.github.com/repos/Zilliqa/Zilliqa/releases/latest"
 
 SWInfo::SWInfo()
     : m_major(0)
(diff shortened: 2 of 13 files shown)


src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,205 @@
...
+    return um;
 }
 
-SWInfo::SWInfo(const SWInfo& src)
-    : m_major(src.m_major)
-    , m_minor(src.m_minor)
-    , m_fix(src.m_fix)
-    , m_upgradeDS(src.m_upgradeDS)
-    , m_commit(src.m_commit)
+bool UpgradeManager::DownloadFile(const char* fileTail)
```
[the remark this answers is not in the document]
ckyang: No problem, I will look for alternative method to download files

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,307 @@
...
+        if (string::npos != s.rfind(fileTail))
+        {
+            downloadFilePath = s;
+            break;
+        }
+    }
 
-    return SIZE;
+    string cmd = downloadFilePath + " | wget -qi -";
+    return (system(cmd.c_str()) >= 0)
```
Gnnng: Calling `system()` directly is quite dangerous and a better way is to use libraries to handle the downloading.

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,307 @@
...
+
+    /// Store final block
+    vector<unsigned char> serializedTxBlock;
+    mediator.m_txBlockChain.GetLastBlock().Serialize(serializedTxBlock, 0);
+    BlockStorage::GetBlockStorage().PutTxBlock(
+        mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum(),
+        serializedTxBlock);
+
+    /// Deploy downloaded software
+    if (system("dpkg -i *.deb") < 0)
```
Gnnng: Same `system()` issue here

bb111189: in addition, using *.deb is a dangerous practice too.

bb111189: Concretely, using *.deb presents at least two issues
1. you may install other non related packages
2. installation of the first packages may fail. However, if sub-sequence packages passed, I am worried that the return result might be 0 instead of 1 (to be verify). This may bypass some checks.

We should just avoid using * completely..

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
 {
-    LOG_MARKER();
+    curl_global_init(CURL_GLOBAL_DEFAULT);
+    CURL* curl = curl_easy_init();
+
+    if (!curl)
+    {
+        curl_easy_cleanup(curl);
+        curl_global_cleanup();
+        return nullptr;
```
Gnnng: returning a nullptr would result in a segfault

ckyang: OK, I will fix it.

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
-                               unsigned int offset) const
+string UpgradeManager::DownloadFile(const char* fileTail)
 {
-    LOG_MARKER();
+    curl_global_init(CURL_GLOBAL_DEFAULT);
+    CURL* curl = curl_easy_init();
+
+    if (!curl)
+    {
+        curl_easy_cleanup(curl);
```
Gnnng: if the variable `curl` is nullptr, how would `curl_easy_cleanup` use it for cleanup? Perhaps its official documentation has better examples.

release.sh
```diff
@@ -107,7 +107,7 @@ echo -e "Making SHA-256 & multi-signature..."
 privKeyFile="$(realpath $1)"
 pubKeyFile="$(realpath $2)"
 cd ${releaseDir}
-sha="$(md5sum ${debFile}|cut -d ' ' -f1)"
+sha="$(sha256sum ${debFile}|cut -d ' ' -f1)"
```
bb111189: Can we do a sha at the protocol level? This assume the system have sha256sum install

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
 
-SWInfo::~SWInfo(){};
-
-/// Implements the Serialize function inherited from Serializable.
-unsigned int SWInfo::Serialize(std::vector<unsigned char>& dst,
-                               unsigned int offset) const
+string UpgradeManager::DownloadFile(const char* fileTail)
 {
-    LOG_MARKER();
+    curl_global_init(CURL_GLOBAL_DEFAULT);
```
Gnnng: https://curl.haxx.se/libcurl/c/libcurl-tutorial.html

> Repeated calls to `curl_global_init` and `curl_global_cleanup` should be avoided. They should only be called once each.

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
+        curl_easy_cleanup(curl);
+        curl_global_cleanup();
+        return nullptr;
+    }
 
-    unsigned int size_remaining = dst.size() - offset;
+    string curlRes;
+    curl_easy_setopt(curl, CURLOPT_URL, RELEASE_URL);
+    curl_easy_setopt(curl, CURLOPT_USERAGENT, USER_AGENT);
+    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, WriteString);
```
Gnnng: Seems using a custom callback is not necessary according to https://curl.haxx.se/libcurl/c/libcurl-tutorial.html
> libcurl offers its own default internal callback that will take care of the data if you don't set the callback with CURLOPT_WRITEFUNCTION.

src/libUtils/UpgradeManager.cpp
ckyang: Got it, I will moved them to class constructor/destructor.

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
+        return nullptr;
+    }
 
-UpgradeManager::UpgradeManager() {}
+    curl_easy_setopt(curl, CURLOPT_WRITEDATA, file);
+    res = curl_easy_perform(curl);
 
-UpgradeManager::~UpgradeManager() {}
+    if (res != CURLE_OK)
+    {
```
Gnnng: `fclose(file);` is missing here

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
-/// Less-than comparison operator.
-bool SWInfo::operator<(const SWInfo& r) const
-{
-    return tie(m_major, m_minor, m_fix, m_upgradeDS, m_commit)
-        < tie(r.m_major, r.m_minor, r.m_fix, r.m_upgradeDS, r.m_commit);
-}
+    if ((res != CURLE_OK) || ((response_code / 100) == 3))
+    {
+        char* location;
+        curl_easy_getinfo(curl, CURLINFO_REDIRECT_URL, &location);
```
Gnnng: Error checking is missing. This would cause memory overflow when the uninitialized variable `location` is assigned to `downloadFilePath` and `downloadFilePath` is used aftewards.

https://curl.haxx.se/libcurl/c/curl_easy_getinfo.html
> The data pointed-to will be filled in accordingly and can be relied upon only if the function returns CURLE_OK.

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
+        int line_no = 0;
+        string line;
+
+        /// Read major version
+        while (line_no != 2 && getline(versionFile, line))
+        {
+            ++line_no;
+        }
 
+        major = stoul(line);
```
Gnnng: `stoul` could throw exceptions when no conversion can be performed.

https://en.cppreference.com/w/cpp/string/basic_string/stoul

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,370 @@
...
+    /// Store final block
+    vector<unsigned char> serializedTxBlock;
+    mediator.m_txBlockChain.GetLastBlock().Serialize(serializedTxBlock, 0);
+    BlockStorage::GetBlockStorage().PutTxBlock(
+        mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum(),
+        serializedTxBlock);
+
+    string cmd = string("dpkg -i ") + m_packageFileName;
+    /// Deploy downloaded software
+    if (system(cmd.data()) < 0)
```
Gnnng: perhaps add a reference to the known issue as a reminder in the comment.

src/libUtils/UpgradeManager.cpp
ckyang: Thanks for the information, the WriteStream can be removed. However WriteString would be reserved because here the curl result should be put into string and apply some string tokenization.

release.sh
ckyang: It's a great advice, however need to provide one more utility. How about leave it as follow-up. (https://github.com/Zilliqa/Issues/issues/185)

src/libUtils/UpgradeManager.cpp
Gnnng: okay, looks okay to keep `WriteString`

release.sh
bb111189: There is no need to provide a utility. we can use openssl sha. 
https://stackoverflow.com/questions/2262386/generate-sha256-with-openssl-and-c

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,387 @@
...
-{
-    return tie(m_major, m_minor, m_fix, m_upgradeDS, m_commit)
-        == tie(r.m_major, r.m_minor, r.m_fix, r.m_upgradeDS, r.m_commit);
-}
+    /// Download the file
+    curl_easy_reset(m_curl);
+    curl_easy_setopt(m_curl, CURLOPT_URL, downloadFilePath.data());
+    curl_easy_setopt(m_curl, CURLOPT_VERBOSE, 1L);
+    curl_easy_setopt(m_curl, CURLOPT_NOPROGRESS, 1L);
+    FILE* file = fopen(fileName.data(), "wb");
```
Gnnng: RAII can make this more maintainable, see https://github.com/Zilliqa/Issues/issues/172

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,387 @@
...
+    curl_easy_setopt(m_curl, CURLOPT_WRITEFUNCTION, WriteString);
+    curl_easy_setopt(m_curl, CURLOPT_WRITEDATA, &curlRes);
+    CURLcode res = curl_easy_perform(m_curl);
 
-    if (size_remaining < SIZE)
+    if (res != CURLE_OK)
     {
-        dst.resize(SIZE + offset);
+        LOG_GENERAL(WARNING,
+                    "curl_easy_perform() failed: " << curl_easy_strerror(res));
```
Gnnng: perhaps add more information to the error log? e.g., the url being requested.

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,387 @@
...
-    , m_fix(fix)
-    , m_upgradeDS(upgradeDS)
-    , m_commit(commit)
+UpgradeManager::~UpgradeManager()
 {
+    if (m_curl)
+    {
+        curl_easy_cleanup(m_curl);
+        curl_global_cleanup();
+        m_curl = nullptr;
```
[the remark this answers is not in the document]
Gnnng: This is in the destructor of `UpgradeManager` and `m_curl` will be de-allocated no matter you reset it or not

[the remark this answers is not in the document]
ckyang: OK, it's true, I didn't notice that

src/libUtils/UpgradeManager.cpp
```diff
@@ -15,148 +15,391 @@
...
-    , m_minor(minor)
-    , m_fix(fix)
-    , m_upgradeDS(upgradeDS)
-    , m_commit(commit)
+UpgradeManager::~UpgradeManager()
 {
+    if (m_curl)
+    {
+        curl_easy_cleanup(m_curl);
+        curl_global_cleanup();
```
Gnnng: should call `curl_global_cleanup()` anyway in the destructor

src/libUtils/UpgradeManager.h
```diff
@@ -17,58 +17,20 @@
...
-    bool operator==(const SWInfo& r) const;
-};
-
 class UpgradeManager
 {
 private:
-    std::shared_ptr<SWInfo> m_curSWInfo;
+    std::shared_ptr<SWInfo> m_latestSWInfo;
+    std::vector<unsigned char> m_latestSHA;
+    CURL* m_curl;
```
Gnnng: we can also wrap `m_curl` in a `unique_ptr` to make life easier.

ckyang: Maybe need more time to investigate how to use unique_ptr with CURL because it's a "typedef."

PR #561, opened 2018-09-06
[Refactor] Integrate protobuf in DS and Node messages

ansnunez:
## Description

This is the first of several PRs to slowly replace the manual serialization/deserialization with protobuf.
In this PR, I covered the messages `POWSUBMISSION` for DirectoryService and `DSBLOCK` and `FINALBLOCK` for Node.

## Review Suggestion

Basically you will notice we now have a ZilliqaMessage.proto file for the protobuf message definitions. I will update this file every time new messages will be protobuf-ed in the next PRs.

The actual protobuf-ing can be found in Messenger.h and cpp.

I also removed the caching of the DSBlock and FinalBlock announcements, which was pretty clumsy and also could result in serialization issues (e.g., the cosig and bitmap had to be updated within the cached message.  Now `m_PoWConsensusMessage` and `m_finalBlockMessage` have been removed.

For DSBlock, there are still some functions that will be eventually removed when Consensus announcement has been protobuf-ed. These functions are `PopulateShardingStructure` and `SaveTxnBodySharingAssignment`. The next PRs will get rid of these.

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ansnunez:
Something wrong with DS Block serialization, causing 0 transactions to appear. Will submit separate PR to exclude this in the meantime.

PR #563, opened 2018-09-06
Handling Finalblock Consensus Error due to Missing Microblocks

kaikawaliu:
## Description

Basic Procedure:
- Let the node simply fetch the missing microblocks from the ds leader, use them save to the coinbase.
- At meanwhile, get the final version of state delta from the leader, initialize the AccountStoreTemp and use this one to update it, then calculate its hash and verify with the StateDeltaHash in finalblock header.
- If there is any Cosig2 message, just proceed normally, it leads to ProcessFinalBlockConsensusDone and will commit the state delta to account store.
- Of course if it never receive the Cosig2 message, the state delta will revert back and initiate the view change.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

- [x] medium-scale cloud test 122d50df5b53c4d305c6f236b134d2cfe08b3e4b


src/libData/AccountData/LogEntry.cpp
```diff
@@ -22,7 +22,7 @@ bool LogEntry::Install(const Json::Value& eventObj,
                        const Address& address) //, unsigned int& numIndexed)
 {
     m_eventObj = eventObj;
-    if (!m_eventObj.isMember("eventname") || !m_eventObj.isMember("params"))
+    if (!m_eventObj.isMember("_eventname") || !m_eventObj.isMember("params"))
```
ansnunez: I think for next time we can define these fields as constants somewhere to avoid these typos

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -573,32 +585,189 @@ bool DirectoryService::CheckMicroBlockHashes()
...
+        for (const auto& microBlock : microBlocks)
         {
             if (microBlock.GetHeader().GetTxRootHash()
-                    == microBlockHash.m_txRootHash
+                    == missingMicroBlockHashes[i].m_txRootHash
                 && microBlock.GetHeader().GetStateDeltaHash()
-                    == microBlockHash.m_stateDeltaHash)
+                    == missingMicroBlockHashes[i].m_stateDeltaHash)
             {
+                mb = microBlock;
```
ansnunez: I think we can make this more efficient by doing the serializing here directly so we don't create another temporary microblock copy.

bb111189:
For this PR, I will help to do a medium scale test using commit hash https://github.com/Zilliqa/Zilliqa/commit/122d50df5b53c4d305c6f236b134d2cfe08b3e4b

src/libDirectoryService/FinalBlockPreProcessing.cpp
kaikawaliu: Noted and will make all of these optimizing in the next PR

PR #413, opened 2018-07-10
Feature/gossip protocol

digiront:
## Description

This PR adds a Gossip protocol, specifically this [RandomizedRumorSpreading](https://github.com/digiront/GossipProtocols/tree/master/RandomizedRumorSpreading) algorithm, to the Zilliqa codebase. The goal is to provide an efficient replacement to the existing broadcasting mechanism.

The current broadcasting mechanism in `P2PComm::SendBroadcastMessage` is resource hungry; it sends `O(n^2)` messages, requires a lot of threads and opens a lot of TCP connections. 

The gossip algorithm, described in detail in this [paper](https://zoo.cs.yale.edu/classes/cs426/2013/bib/karp00randomized.pdf), provides a method to spread a message in `O(logn)` rounds and `O(ln(ln(n)))` messages, where `n` is the number of peers participating in the gossip.

Related open issues: https://github.com/Zilliqa/Zilliqa/issues/339

## Review Suggestion

The code was written in such a way to make reviewing easier.:

- The algorithm implementation is located in the `depends` directory. It was written in a way that allows testing in isolation.
- The core of the integration is the newly added `libNetwork/RumorManager` component.
- I had to do some refactoring to the existing `P2PComm` implementation inspired by [this PR](https://github.com/Zilliqa/Zilliqa/pull/369). There were no functional changes during this refactoring.
- I also had to write new code for the gossip transmission formats. Again, there were no functional changes to the original code!
- Also, I enabled edits from maintainers

### Implementation

- [X] TODOs ...
* Parameterize hardcoded variables
* Complete all bits and pieces for the the new gossip reception/transmission byte format
* Write some unit tests 
* Do a second iteration to spot easily optimizable code
- [ ] **ready for review**

Issue #339, opened 2018-06-14
Implement Gossip protocol to replace P2PComm::SendBroadcastMessage 


digiront:
The current implementation spawns a lot of threads to multicast a message to peers. 

A [gossip protocol](https://en.wikipedia.org/wiki/Gossip_protocol) implementation can be used to spread the message instead of multicast. This will spawn fewer threads at the cost of extra network hops.

### Resources
- https://zoo.cs.yale.edu/classes/cs426/2013/bib/karp00randomized.pdf
- https://github.com/maidsafe/safe_gossip
- https://github.com/apache/incubator-gossip
- https://www.youtube.com/watch?v=Gxf5glthqrk
- https://github.com/hashicorp/memberlist

### Concerns
- Make it easily pluggable in the Zillqa codebase
- Integration testing

digiront:
Work in progress.

satoshi101:
Further resources:

* https://github.com/ethereum/cpp-ethereum/tree/develop/libp2p
* https://github.com/ethereum/wiki/wiki/libp2p-Whitepaper
* https://en.wikipedia.org/wiki/Hole_punching_(networking)

Further concerns:

* Might require large refactoring on the rest of the code

bb111189:
On the further resources, 3rd link on hole punching. A PR (review and revision in progress) has been created to support upnp. https://github.com/Zilliqa/Zilliqa/pull/259

satoshi101:
Hello @bb111189 - I can see some hole punching [here](https://github.com/ethereum/cpp-ethereum/blob/40df049e8fffcd99f5d6b8cc7845ad15aaed8947/libp2p/Network.cpp#L176) as well. Do you prefer [miniupnp](http://miniupnp.free.fr/) over the libp2p solution, or it's a starting point and we're open to whatever is best?

digiront:
Follow https://github.com/Zilliqa/Zilliqa/pull/345 for updates on this issue.

digiront:
Forgot to mention: I will follow up with a commit on #345

digiront:
Got a working standalone implementation here https://github.com/digiront/GossipProtocols/tree/master/RandomizedRumorSpreading/libRumorSpreading

So, I will modify this https://github.com/satoshi101/Zilliqa/blob/c1671f9fdd6fb867b8024e82c616cf4860128058/src/libNetwork/P2PGossip.h to "gossip" instead of broadcast.

digiront:
[...]

PR #564, opened 2018-09-07
Provide P2PComm with efficient broadcasting mechanism using gossip/RRS protocol

sandipbhoir:
## Description
This PR adds a Gossip protocol / algorithm, to the Zilliqa codebase. The goal is to provide an efficient replacement to the existing broadcasting mechanism.  

The current broadcasting mechanism in P2PComm::SendBroadcastMessage is resource hungry; it sends O(n^2) messages, requires a lot of threads and opens a lot of TCP connections.

The gossip algorithm, described in detail in this(https://zoo.cs.yale.edu/classes/cs426/2013/bib/karp00randomized.pdf) paper, provides a method to spread a message in O(logn) rounds and O(ln(ln(n))) messages, where n is the number of peers participating in the gossip.

## Usage: 
_To be used by nodes who shares the peerlist/memberlist of similar node types._

**Obvious use case:**
- Announcement of consensus from shard leader to all other shard member. 

**Usecase - Not so upfront since they share different peerlist**

- Broadcasting Transaction packet from lookup to all shard nodes. Option could be to send packet from lookup node to shard leader and then broadcast it to all shard nodes using gossip protocol.
- Broadcasting Microblock from shard leader to all DS members. Option could be to send packet from shard leader to DS leader and then broadcast it to all DS members.
- Broadcasting FinalBlock from every DS member to specific shard. 

Related open issues: #339

## Review Suggestion

By default, the existing multicast protocol is being used.

**Local Test:**
To enable it, Set flag BROADCAST_GOSSIP_MODE in Constants_local.xml

**Small-Scale Test:**
To enable it, Set flag BROADCAST_GOSSIP_MODE in Constants.xml
Flag GOSSIP_CUSTOM_ROUNDS_SETTINGS is disabled by default. And will rely on number of peers in each shards to calculate the round setting at runtime. 

If shard-size is too small ( like one with local test ) will suggest to enable GOSSIP_CUSTOM_ROUNDS_SETTINGS flag.

### Implementation

- [x] Integrating existing libRumorSpreading (PR #413 ) to Zilliqa codebase by fixing minor compilation issues. 
- [x] Fix the libRumorSpreading with expected gossip protocol functionality.( Ex. we have to consider Rumor status /age/rounds before forwarding to next peer and/or decide on getting rid after fixed rounds/age is reached for that rumor. )
- [x] Implement RumorManager which will always be ready to add rumor/receive rumor and broadcast it to available peers. Think of it as having a function which runs a rounds in loop forever.
- [x] RumorManager to be started at start of DS Epoch ( or may be can restart every DS Epoch ). Why ?  Because we need peerlist which is pre-requiste for this protocol. 
- [x] Implement changes needed for new gossip reception/transmission byte format
- [x] Implement P2PComm function to insert rumor (sender) and changes in evt call back to receive messages(receiver) and direct it to RumorManager based on reception byte format.
- [x] Changes to use new broadcasting api for DS MicroBlock consensus and Shard MicroBlock Consensus. 
- [ ] Changes to use new broadcasting api for other use-cases to be implemented in separate PR.
- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


sandipbhoir:
Self-Reminder - Log the rumor received timings..

src/libRumorSpreading/RumorStateMachine.h
```diff
@@ -0,0 +1,90 @@
...
+        void advanceFromNew(const std::unordered_set<int>& membersInRound);
+
+        void advanceFromKnown();
+
+        void advanceToOld();
+
+    public:
+        // CONSTRUCTORS
+        // Default constructor. The returned state machine instance will be in an invalid state.
+        RumorStateMachine();
```
ansnunez: If we shouldn't use this constructor I think you can just declare it as `RumorStateMachine() = delete;` and remove the definition in the cpp.

src/libRumorSpreading/RumorStateMachine.cpp
```diff
@@ -0,0 +1,180 @@
+#include "RumorStateMachine.h"
+
+#define LITERAL(s) #s
+
+namespace RRS
+{
+
+    // STATIC MEMBERS
+    std::map<RumorStateMachine::State, std::string>
```
ansnunez: Did you mean to declare this as static?

sandipbhoir: yes, its more of tied to class rather than object itself. So is declared as static.

src/libNetwork/RumorManager.cpp
```diff
@@ -0,0 +1,298 @@
...
+                    if (l != m_peerIdPeerBimap.left.end())
+                    {
+                        LOG_GENERAL(DEBUG,
+                                    "Sending " << result.second.size()
+                                               << " push messages");
+                        SendMessages(l->second, result.second);
+                    }
+                }
+            } // end critical section
+            if (m_condStopRound.wait_for(
```
ansnunez: I'm curious if `stopRounds` is actually able to trigger `notify_all` because `m_continueRoundMutex` is held by this thread throughout its lifetime, and `stopRounds needs it`.  Were you able to check if this thread actually terminates when stopRounds is called?

src/libNetwork/RumorManager.cpp
```diff
@@ -0,0 +1,298 @@
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(ROUND_TIME_IN_MS));
+}
+
+// PUBLIC METHODS
+bool RumorManager::Initialize(const std::vector<Peer>& peers,
+                              const Peer& myself)
+{
+    LOG_MARKER();
+    {
+        std::lock_guard<std::mutex> guard(m_continueRoundMutex);
```
ansnunez: Same comment here. I wonder if `Initialize` can be called again after `startRounds` has been called once.

src/libNetwork/RumorManager.h
```diff
@@ -0,0 +1,68 @@
+#ifndef ZILLIQA_RUMORMANAGER_H
```
ansnunez: Following our header guard naming convention, this should be __RUMOR_MANAGER_H__.
Please update the other files similarly.

src/libNetwork/RumorManager.cpp
```diff
@@ -0,0 +1,298 @@
...
+        {
+            // Add raw message to outgoing message
+            cmd.insert(cmd.end(), m->second.begin(), m->second.end());
+            LOG_GENERAL(DEBUG,
+                        "Sending Non Empty - Gossip Message - "
+                            << k << " To Peer : " << toPeer);
+        }
+
+        // Send the message to peer .
+        P2PComm::GetInstance().SendMessageNoQueue(toPeer, cmd,
```
ansnunez: I would suggest not to use the `NoQueue` version anywhere in `RumorManager`. Historically, this function was added only for test code that didn't necessarily have all the classes instantiated, and where we basically just want to send a message. In our actual code, using the `NoQueue` version would skip using the sending queue, which is supposed to manage all the outgoing messages.

src/libNetwork/RumorManager.cpp
sandipbhoir: wait_for will actually release the m_continueRoundMutex for ROUND_TIME.

src/libRumorSpreading/NetworkConfig.cpp
```diff
@@ -0,0 +1,48 @@
...
+    // PUBLIC CONST METHODS
+    size_t NetworkConfig::networkSize() const { return m_networkSize; }
+
+    int NetworkConfig::maxRoundsInB() const { return m_maxRoundsInB; }
+
+    int NetworkConfig::maxRoundsInC() const { return m_maxRoundsInC; }
+
+    int NetworkConfig::maxRoundsTotal() const { return m_maxRoundsTotal; }
+
+    bool NetworkConfig::operator==(const NetworkConfig& other) const
```
ansnunez: Not sure this is actually used anywhere at the moment. If it isn't, maybe we can just remove it.

src/libNetwork/RumorManager.cpp
ansnunez: After discussion, will keep this way for now

src/libRumorSpreading/NetworkConfig.cpp
sandipbhoir: Fixed.

src/libNetwork/RumorManager.h
sandipbhoir: Fixed

sandipbhoir:
Created 3 other separate PR instead.

PR #572, opened 2018-09-10
Shard size calculator

bb111189:
## Description

This PR implements a static shard size calculator. This is to support future implementation(s) for shard size adjustment. This function is currently not in use. 

## Review Suggestion

Refer to the spreadsheet for the shard size calculation. 

### Implementation

- [x] **ready for review**

- [x] unit test 
- [ ] ~local machine test~

### DIFF
--- a/src/libDirectoryService/DirectoryService.h
+++ b/src/libDirectoryService/DirectoryService.h
@@ -327,6 +327,9 @@ class DirectoryService : public Executable, public Broadcastable
 
     uint8_t CalculateNewDifficulty(const uint8_t& prevDifficulty);
 
+    /// Calculate and return the min size of each for n number of nodes
+    uint32_t CalculateShardSize(uint32_t numberOfNodes);
+
 public:
     enum Mode : unsigned char
     {
--- a/src/libDirectoryService/DirectoryService.h
+++ b/src/libDirectoryService/DirectoryService.h
@@ -327,9 +327,6 @@ class DirectoryService : public Executable, public Broadcastable
 
     uint8_t CalculateNewDifficulty(const uint8_t& prevDifficulty);
 
-    /// Calculate and return the min size of required each shard for a specifc number of total nodes
-    uint32_t CalculateShardSize(const uint32_t numberOfNodes);
-
 public:
     enum Mode : unsigned char
     {
--- a/src/libDirectoryService/DirectoryService.h
+++ b/src/libDirectoryService/DirectoryService.h
@@ -327,7 +327,7 @@ class DirectoryService : public Executable, public Broadcastable
 
     uint8_t CalculateNewDifficulty(const uint8_t& prevDifficulty);
 
-    /// Calculate and return the min size of each for n number of nodes
+    /// Calculate and return the min size of required each shard for a specifc number of total nodes
     uint32_t CalculateShardSize(const uint32_t numberOfNodes);
 
 public:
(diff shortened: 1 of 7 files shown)


src/libDirectoryService/DirectoryService.h
```diff
@@ -291,6 +291,9 @@ class DirectoryService : public Executable, public Broadcastable
 
     uint8_t CalculateNewDifficulty(const uint8_t& prevDifficulty);
 
+    /// Calculate and return the min size of required each shard for a specifc number of total nodes
+    uint32_t CalculateShardSize(const uint32_t numberOfNodes);
```
ansnunez: We can make this static in case our test code needs to use this without actually going through a DS instance

src/libDirectoryService/ShardSizeCaluator.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+    else if (numberOfNodes >= 651)
+    {
+        return 651;
+    }
+    else if (numberOfNodes >= 1368)
+    {
+        return 684;
+    }
+    else if (numberOfNodes >= 2133)
```
sandipbhoir: For all cases, it will fall under only one of two if statements. Ex. numberOfNodes = 2133 will also go under second if statement. I think we should use < instead of >

ansnunez: I think the sign is correct but we need to flip the sequence of the checks (i.e., descending instead of ascending)

sandipbhoir: Yes, can be any way round. However, we can possibly consolidate the multiple if statements returning same value.

ansnunez:
Might want to consider this suggestion from @Gnnng : http://cpp.sh/7jeic
Binary search in a lookup table will be more efficient than sequential comparisons

bb111189:
The nested if else will most likely ended up in a jump table when it is compiled to binary. So it shouldn't be sequential comparisons. 

But the suggestion from @Gnnng  is neater and easier to maintain. Will modify it.

src/libUtils/ShardSizeCalculator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    auto constexpr range_size = std::extent<decltype(range)>::value;
+    auto constexpr result_size = std::extent<decltype(result)>::value;
+
+    static_assert(range_size == result_size,
+                  "mismatch in static shard size table");
+
+    auto it = std::upper_bound(range, range + range_size, numberOfNodes);
+    std::size_t index = it - range;
+
+    if (index < 1)
```
Gnnng: `index < 1` is actually not possible in this implementation. The `upper_bound` will return the first element that is greater than `numberOfNodes`, which guarantees `index` will be greater than 0 since `range[0]` is 0.

src/libUtils/ShardSizeCalculator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        LOG_GENERAL(WARNING, "Index index. Shard is too small.");
+    }
+
+    if (index >= result_size)
```
Gnnng: `index >= result_size` is also not necessary if we have unit test that covers all the possible inputs.

src/libUtils/ShardSizeCalculator.cpp
bb111189: sure. will remove unnecessary code.

src/libUtils/ShardSizeCalculator.cpp
bb111189: sure. will remove unnecessary code.

PR #573, opened 2018-09-11
TxnPacket/FBCons buffer refactoring & introduce MBCons Buffer

kaikawaliu:
### Implementation
- Refactoring on TxnPacket buffering and committing
- Refactoring on Finalblock buffering and commiting
- Introduce Buffer mechanism to Microblock consensus
- Renaming MICROBLOCK_CONSENSUS_PREP to WAITING_DSBLOCK

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud test (200 nodes)


src/libNode/Node.cpp
```diff
@@ -621,21 +631,34 @@ bool Node::ProcessTxnPacketFromLookup(
     {
         // check for recieval of new ds block
         // need to wait the ProcessDSBlock finish
-        lock_guard<mutex> g(m_mutexDSBlock);
+        lock_guard<mutex> g1(m_mutexDSBlock);
```
bb111189: should we lock both g1 and g2 together at the same time. If not careful, one may introduce codes in the future that cause a deadlock. i.e hold onto g2 first then g1.

kaikawaliu: g2 only lock when the if condition is true, as the two locks are for different purpose I still feel it's better to keep them separate

PR #578, opened 2018-09-12
Dual difficulty adjustment

ShengguangXiao:
## Description

Support two levels of difficulty adjustment

Add a ```CalculateNewDifficultyCore``` function, so the normal difficulty adjustment function ```CalculateNewDSDifficulty``` and DS difficulty adjustment function will both call ```CalculateNewDifficultyCore```.

## Review Suggestion

Check the ```DirectoryService::CalculateNewDSDifficulty```, ```DirectoryService::CalculateNewDifficultyCore``` and ```DirectoryService::CalculateNewDifficultyCore``` functions.

Build the project, run the test case ```./build/test/POW/Test_POW``` make sure auto test case can pass.
On local machine, run small network first, then run ```./tests/Node/test_node_late.sh``` to join new nodes, the difficulty will adjust.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


ShengguangXiao:
Tested with 200 nodes, able to run 29 DS blocks and 140 Tx blocks without node drops.

PR #579, opened 2018-09-12
[clang-tidy] Use range-based for-loop preferably

Gnnng:
## Description

This PR adds another clang-tidy check `modernize-loop-convert`. Now clang-tidy on Travis will suggest improvements on all the places that can use range-based for-loop.

## Review Suggestion

See the refactoring brought by clang-tidy.

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~

### DIFF
--- a/src/libDirectoryService/DirectoryService.cpp
+++ b/src/libDirectoryService/DirectoryService.cpp
@@ -232,14 +232,14 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
                                           sizeof(uint32_t));
         curr_offset += sizeof(uint32_t);
 
-        for (unsigned int i = 0; i < ds.size(); i++)
+        for (auto& d : ds)
         {
             // PubKey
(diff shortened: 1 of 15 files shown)


src/libDirectoryService/DirectoryService.cpp
```diff
@@ -232,14 +232,14 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
                                           sizeof(uint32_t));
         curr_offset += sizeof(uint32_t);
 
-        for (unsigned int i = 0; i < ds.size(); i++)
+        for (auto& d : ds)
```
ansnunez: I wonder why this wasn't changed to `const`

KaustubhShamshery:
Sometimes use "i" as iterator, sometimes used the singular of the loop element? Is there any convention for it ?

src/libDirectoryService/DirectoryService.cpp
Gnnng: I guess it's because our `Serialize` was not declared as const

PR #581, opened 2018-09-13
Integration for upgrading protocol

ckyang:
## Description

This is the final PR of upgrading protocol, which will integrate all utilities of it.
Following items should be done:

- [x] [DS leader] In DirectoryService::ComposeDSBlock(), check status of the trusted source website
- [x] [DS backup] In DirectoryService::DSBlockValidator(), DS committee check DSBlock to see if download is needed
- [x] [Shard node] In Node::ProcessDSBlock(), shard (normal) nodes check DSBlock to see if download is needed
- [x] In DirectoryService::ProcessFinalBlockConsensusWhenDone() / Node::ProcessFinalBlock, after DS / shard node set DSINCOMPLETED = '0', check if upgrade DS epoch is reached
- [x] Store m_DSCommittee inside ReplaceNode()
- [x] Retrieve m_DSCommittee inside Node::Install()

## Review Suggestion

To verify upgrading protocol, the simulated upgrading procedure needed to be done on AWS. We temporarily force downloading author's personal GitHub repo (instead of Zilliqa repo) to do verification testing, then observe if each upgrading step (files downloaded, deployed, process killed, recovered) happened smooth in each node.

Known issue: ValidateStates() would be failed to validate state root, thus we skip this function check (always return true), to verify this PR.

### Implementation

- [x] **ready for review**

- [x] small-scale cloud test (20 nodes, commit: ea69e20)

- [x] large-scale cloud test (200 nodes, commit: ea69e20)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -666,6 +675,18 @@ bool DirectoryService::DSBlockValidator(
         return false;
     }
 
+    if (m_mediator.m_curSWInfo != m_pendingDSBlock->GetHeader().GetSWInfo())
+    {
+        auto func = [this]() mutable -> void {
```
Gnnng: any race condition analysis of the memory access in this lambda function?

ckyang: OK, I will add some protection on m_mediator.m_curSWInfo

ckyang:
> I can't see what is the initial value of `m_curSWInfo`, could it be uninitalized and accessed accidentally?

The default constructor of SWInfo(SWInfo.cpp:22) would be called when mediator is created, and all the variables inside SWInfo will be set as zero.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -113,15 +114,23 @@ void DirectoryService::ComposeDSBlock(
                       << ", new difficulty " << std::to_string(difficulty));
     }
 
+    if (UpgradeManager::GetInstance().HasNewSW())
```
Gnnng: From what I have seen, `HasNewSW` and `DownloadSW` are blocking calls involving curl downloading. Do we have any timeout protection that deals with unideal network connections so that the node won't be stuck?

constants.xml
```diff
@@ -50,6 +50,8 @@
         <MAX_NEIGHBORS_PER_ROUND>10</MAX_NEIGHBORS_PER_ROUND>
         <NUM_NODE_INCR_DIFFICULTY>5000</NUM_NODE_INCR_DIFFICULTY>
         <MAX_SHARD_NODE_NUM>7500</MAX_SHARD_NODE_NUM>
+        <TERMINATION_COUNTDOWN>5</TERMINATION_COUNTDOWN>
+        <SHARD_DELAY_WAKEUP>5</SHARD_DELAY_WAKEUP>
```
ShengguangXiao: Maybe with unit of time is better?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -958,6 +959,21 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
     m_mediator.UpdateDSBlockRand();
     m_mediator.UpdateTxBlockRand();
 
+    {
+        lock_guard<mutex> g(m_mediator.m_mutexCurSWInfo);
+        if (0 == (m_mediator.m_currentEpochNum % NUM_FINAL_BLOCK_PER_POW)
+            && m_mediator.m_curSWInfo.GetUpgradeDS()
+                == ((m_mediator.m_currentEpochNum / NUM_FINAL_BLOCK_PER_POW)
+                    + 2))
```
ShengguangXiao: What does + 2 mean here? Should it change to a more meaningful name?

src/libPersistence/BlockStorage.cpp
```diff
@@ -323,6 +325,82 @@ bool BlockStorage::GetMetadata(MetaType type, std::vector<unsigned char>& data)
...
+    LOG_MARKER();
+
+    unsigned int index = 0;
+    consensusLeaderID = stoul(m_dsCommitteeDB->Lookup(index++));
+    LOG_GENERAL(INFO, "Retrieved DS leader ID: " << consensusLeaderID);
+    string dataStr;
+
+    while (m_dsCommitteeDB->Lookup(index) != "")
+    {
+        dataStr = m_dsCommitteeDB->Lookup(index);
```
ShengguangXiao: The same Lookup called for two times, can combine it to one?

src/libDirectoryService/DSBlockPreProcessing.cpp
ckyang: Cause we need to wait m_mediator.m_curSWInfo be updated, to compose a new DS block using latest SWInfo, we cannot detach HasNewSW() and DownloadSW() here. Will implement a timeout protection inside curl-related function in follow-up PR. (https://github.com/Zilliqa/Issues/issues/185)

PR #583, opened 2018-09-13
Fix inconsistent declaration parameter name

Gnnng:
## Description

Some of the parameter names in the declaration are inconsistent with the definition. clang-tidy fixes this by fixing the parameter names in declaration according to the definitions.

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~


src/libNode/Node.h
```diff
@@ -420,7 +419,7 @@ class Node : public Executable, public Broadcastable
     void StartSynchronization();
 
     /// Performs PoW mining and submission for DirectoryService committee membership.
-    bool StartPoW(const uint64_t& block_num, uint8_t dsDifficulty,
+    bool StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
```
bb111189: Although it is automated, I prefer it to be without  ```_``` for consistency with other code. same for block_num.

src/libNode/Node.h
```diff
@@ -239,10 +239,10 @@ class Node : public Executable, public Broadcastable
     bool ProcessFinalBlock(const std::vector<unsigned char>& message,
                            unsigned int offset, const Peer& from);
     bool ProcessForwardTransaction(const std::vector<unsigned char>& message,
-                                   unsigned int offset, const Peer& from);
+                                   unsigned int cur_offset, const Peer& from);
```
bb111189: Although it is automated, I prefer it to be without _ for consistency with other code. same for block_num.

src/libNode/Node.h
```diff
@@ -239,10 +239,10 @@ class Node : public Executable, public Broadcastable
...
     bool
     ProcessForwardTransactionCore(const std::vector<unsigned char>& message,
-                                  unsigned int offset);
+                                  unsigned int cur_offset);
```
bb111189: Although it is automated, I prefer it to be without _ for consistency with other code. same for block_num.

src/libNode/Node.h
```diff
@@ -155,8 +155,8 @@ class Node : public Executable, public Broadcastable
 
     // internal calls from ProcessStartPoW1
     bool ReadVariablesFromStartPoWMessage(
-        const vector<unsigned char>& message, unsigned int offset,
-        uint64_t& block_num, uint8_t& dsDifficulty, uint8_t& difficulty,
+        const vector<unsigned char>& message, unsigned int cur_offset,
+        uint64_t& block_num, uint8_t& ds_difficulty, uint8_t& difficulty,
```
bb111189: Although it is automated, I prefer it to be without _ for consistency with other code. same for block_num.

bb111189:
Very good improvement PR

src/libNode/Node.h
Gnnng: The name with `_` is taken from the definition of this function. It didn't invent this name here.

src/libNode/Node.h
Gnnng: The name is taken from the definition of `ProcessForwardTransaction`.

https://github.com/Zilliqa/Zilliqa/blob/70fbb16b9852844b21c9fcc3c081e2653c2b2fc5/src/libNode/FinalBlockProcessing.cpp#L1197-L1199

src/libNode/Node.h
bb111189: maybe we can do a minor clean up on these few cases. If not, we can also fix these kind of issues in future pr. I let you make the call for this.

Gnnng: perhaps let me find some checker tool for the naming of parameters and variables.

PR #586, opened 2018-09-14
Proof of Reputation - select nodes with better reputation to join network

ShengguangXiao:
## Description

Proof of Reputation - select nodes to join network with more successive contribution. Make hacker harder to control the network.

When the PoW submissions number exceed the defined maximum shard node number, then choose the node with better reputation to join the network.

## Review Suggestion

Review the DirectoryService::FindTopPriorityNodes and DirectoryService::CalculateNodePriority related functions.

Temporally disable state root check, set MAX_SHARD_NODE_NUM to 13,  build Zilliqa. Run local test, after a while run ./tests/Node/test_node_late.sh to join 5 new nodes, only 2 nodes can join, there will be only 13 shard nodes in the network. In the Zilliqa explorer can see 23 peers. The new node can join until some old nodes failed to do PoW.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -351,6 +352,12 @@ void DirectoryService::UpdateDSCommiteeComposition(const Peer& winnerpeer)
         m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetMinerPubKey());
 }
 
+/// Calculate node priority to determine which node has the priority to join the network.
+uint8_t DirectoryService::CalculateNodePriority(uint16_t reputation)
```
KaustubhShamshery: if (reputation == 0)
{
return 0 ;
}

ansnunez:
libMessage changes look ok

ShengguangXiao:
Run local test, run 60 DS blocks and 300 Tx blocks.
Run 200 nodes test, run to 26 DS blocks and 135 Tx blocks.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -176,6 +190,15 @@ void DirectoryService::ComputeSharding(
     for (const auto& kv : sortedPoWSolns)
```
ansnunez: I think this might have a performance issue once we hit, say, 10k nodes.
It might be better to have two loops: one for if `setTopPriorityNodes` is empty, and an else loop for otherwise.
In the if branch, you have a for loop over `sortedPoWSolns`.
In the else branch, you have a for loop over `setTopPriorityNodes`.

ShengguangXiao: At most of the time, the FindTopPriorityNodes function will not be called, so setTopPriorityNodes will be empty, and the checking setTopPriorityNodes is empty is very fast.

ansnunez:
@ShengguangXiao Don't forget to update the constants files in testnet repository

Issue #574, opened 2018-09-11
API Feature Request - Get Contract Address By Transaction ID


neeboo:
**Description**:

The Contract Address is created when `CreateTransaction` is done.  After deploying a SmartContract using `CreateTransaction` API, it returns TransactionID as the result.
However, we don't know if it the SmartContract is deployed currently, unless we do one more call to `GetContracts` using Account Address which previously deployed/invoked the contract. After that,  we get list of the contract, but we still have to 'guess' which one of the result is the correct one.

So I hope if the contract address is returned by CreateTransaction API. It would be much better.

Example:

```
curl -d '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "CreateTransaction",
    "params": [{
      "version": 0,
      "nonce": 5,
      "to": "c5a829596fb06a59e2b1ddb6589811c759025d52",
      "amount": 83,
      "pubKey": "026ad7a7cb2aa230af177ca0ea3c4112a902d296bf7a5f25f5ae4b3df72321f4d0",
      "gasPrice": 1,
      "gasLimit": 1,
      "signature": "508b838641a6d0f9a7114c5b359a3a6119072e213b06b0f3331f4a33a8c80cde09d3aa7f64c3d767024ab56abbcb2ebd553085d214133c0383d4007828bde0c4"
    }]
}' -H "Content-Type: application/json" -X POST "https://api-scilla.zilliqa.com/"
```

The above command returns JSON structured like this:
```

{
  "id": "1",
  "jsonrpc": "2.0",
  "result":{
     "txnId":"a5e238bca907c28d015e50acc76788f8bad79882bf40459215722c414a7b940b",
     "contractAddr:"sssssssssssssssssssssssssssssssssssssssss"
  }
}
// result should be normalized as Objects,
// with "txnId" and "contractAddress" as their keys.
```

**Furthermore**:
Because transaction and contract are related in a way. What if we consider using TransactionID as parameter to request, and get the correct Contract Address?

like this:

```
curl -d '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetContractAddressByTransaction",
    "params": ["43104643E5F35455B71F3CFCB471CE4A4C8B96E00F8E8B77733075662E5FC905"]
}' -H "Content-Type: application/json" -X POST "https://api-scilla.zilliqa.com/"

```
The above command returns JSON structured like this:

```
{
  "id": "1",
  "jsonrpc": "2.0",
[...]

PR #587, opened 2018-09-15
Add/Modify Server APIs for feature request.

KaustubhShamshery:
## Description
This PR modifies CreateTransaction and adds GetContractAddressFromTransactionID APIs for ease of knowing the Contract Address.

* ### CreateTransaction
     * #### Returns a JSON containing :
     1. Error, if there is an error in creating Transaction
     2. Info, tells if the Transaction was a contract transaction and if it went to a ds shard.
     3. ContractAddress, if the transaction was a contract transaction, contains the contract Address.
* ### GetContractAddressFromTransactionID
    * Given Valid Contract Transaction ID, returns the ContractAddress (string)

## Review Suggestion
Should be merged with corresponding changes in Wallet.
Related issue : #574 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/Server.cpp
+++ b/src/libServer/Server.cpp
@@ -75,23 +75,27 @@ string Server::GetNetworkId() { return "TestNet"; }
 
 string Server::GetProtocolVersion() { return "Hello"; }
 
-string Server::CreateTransaction(const Json::Value& _json)
+Json::Value Server::CreateTransaction(const Json::Value& _json)
 {
     LOG_MARKER();
+
+    Json::Value ret;
 
     try
     {
 
         if (!JSONConversion::checkJsonTx(_json))
         {
-            return "Invalid Tx Json";
+            ret["Error"] = "Invalid Tx Json";
+            return ret;
         }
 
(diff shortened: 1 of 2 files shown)


iantanwx:
@KaustubhShamshery sorry for the delay.

I'm not sure if these changes address #574. It seems that the feature requested is not information on the nature of the transaction (i.e., what category the TX falls into), but rather, where the transaction is actually one where a contract is to be _created_, what the resulting address of the deployed contract is.

However, having said that, I'm not sure if it's even possible to implement the feature requested at this level; we wouldn't know whether the contract was actually successfully deployed at the time of the call to `CreateTransaction`.

Am I wrong? Maybe @kaikawaliu can weigh in.

KaustubhShamshery:
@iantanwx it provides an API given a txn, it will return the contract address.

jiayaoqijia:
@KaustubhShamshery Can you run an AWS test?

iantanwx:
Non-contract transactions work as expected.

PR #588, opened 2018-09-16
Fix issue of new ds leader crashing

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/DSBlockProcessing.cpp
+++ b/src/libNode/DSBlockProcessing.cpp
@@ -527,9 +527,10 @@ bool Node::ProcessDSBlock(const vector<unsigned char>& message,
             // Process txn sharing assignments as a DS node
             m_mediator.m_ds->ProcessTxnBodySharingAssignment();
 
-            // Update my DS mode and ID
+            // Update my ID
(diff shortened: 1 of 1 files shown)


src/libNode/DSBlockProcessing.cpp
```diff
@@ -527,9 +527,10 @@ bool Node::ProcessDSBlock(const vector<unsigned char>& message,
             // Process txn sharing assignments as a DS node
             m_mediator.m_ds->ProcessTxnBodySharingAssignment();
 
-            // Update my DS mode and ID
+            // Update my ID
             m_mediator.m_ds->m_consensusID
                 = m_mediator.m_currentEpochNum == 1 ? 1 : 0;
+            m_mediator.m_ds->m_consensusMyID = 0;
```
ShengguangXiao: If with line 533, then the line 531 and 532 should be removed right?

KaustubhShamshery: no, if you see minutely m_consensusID and m_consensusMyId is different

Issue #571, opened 2018-09-10
Remove ConsensusUser from the Zilliqa dispatcher list


ansnunez:
This was initially to test the consensus library, but should be removed from the final list of message types recognized.

If the type values are changed (e.g., moved up by one) for the others, take care to update all the other components that also send messages to the node.

bb111189:
Already implemented

PR #590, opened 2018-09-16
[Refactor] Integrate protobuf in consensus messages

ansnunez:
## Description

This PR integrates Protobuf into all the consensus messages from announcement to final collective sig.

## Review Suggestion

The changes for commit, response, challenge, and collective sig are pretty much straightforward.
The tricky part to review is the announcement.

In the old way, `StartConsensus` required two parameters: the announcement message and the length within that message that will be co-signed by the peers (Usually this will be the block header).
In the new way, we let the `Messenger` class give us back a separate vector for the message to be co-signed.

Another tricky thing is that `ProcessFinalBlockConsensus` and `ProcessMicroblockConsensus` need to pre-read the consensus ID part of the incoming consensus message to know if it should be buffered or not. So, I have had to provide a `GetConsensusID` interface through `ConsensusCommon` and then through `Messenger` for all possible consensus message types.

Finally, I have chosen to remove the test `ConsensusUser` instead of updating it as well. Its position in the `msg_handlers` will have to be taken out eventually (#571).

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] mid-scale cloud test (200 nodes)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -509,7 +484,21 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary()
...
         << "] BGIN");
 
-    cl->StartConsensus(PoWConsensusMessage, DSBlockHeader::SIZE);
+    auto announcementGeneratorFunc =
+        [this](vector<unsigned char>& dst, unsigned int offset,
+               const uint32_t consensusID,
+               const vector<unsigned char>& blockHash, const uint16_t leaderID,
+               const pair<PrivKey, PubKey>& leaderKey,
+               vector<unsigned char>& messageToCosign) mutable -> bool {
+        const auto& winnerPeer = m_allPoWConns.find(
```
Gnnng: `m_allPoWConns` needs a mutex to be thread-safe

src/libNode/MicroBlockProcessing.cpp
```diff
@@ -1560,10 +1583,18 @@ bool Node::MicroBlockValidator(const vector<unsigned char>& microblock,
...
+            microBlock, messageToCosign))
+    {
+        LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
+                  "Messenger::GetNodeMicroBlockAnnouncement failed.");
+        return false;
+    }
 
-    // [TODO] To put in the logic
-    m_microblock = make_shared<MicroBlock>(MicroBlock(microblock, 0));
+    m_microblock.reset(new MicroBlock(microBlock));
```
Gnnng: If we wrap the `microBlock` in a unique_ptr, we could directly assign it to `m_microblock`  with the same effect but better efficiency.

src/libMessage/ZilliqaMessage.proto
```diff
@@ -208,4 +231,95 @@ message LookupGetStartPoWFromSeed
...
+
+message ConsensusAnnouncement
+{
+    message ConsensusInfo
+    {
+        required uint32 consensusid           = 1;
+        required bytes blockhash              = 2; // 32 bytes
+        required uint32 leaderid              = 3; // only lower 2 bytes used    
+    }
+    required ConsensusInfo data               = 1;
```
Gnnng: The field name `data` is a bit too general.

src/libMessage/Messenger.cpp
```diff
@@ -56,8 +56,204 @@ namespace
...
+        {
+            LOG_GENERAL(
+                WARNING,
+                "ConsensusAnnouncement.ConsensusInfo initialization failed.");
+            return false;
+        }
+
+        // Sign the announcement
+
+        vector<unsigned char> tmp;
```
Gnnng: Yes, it's a `tmp` variable, but certainly we can have better alternatives.

src/libMessage/Messenger.cpp
```diff
@@ -56,8 +56,204 @@ namespace
...
+
+        if (announcement.has_dsblock()
```
Gnnng: could use the protobuf-generated member function `announcement_case()`  to do a `switch` instead of multiple `if-else`

src/libDirectoryService/DSBlockPreProcessing.cpp
ansnunez: The mutex `m_mutexAllPoWConns` at line 411 should still be held by the time this lambda is invoked inside `StartConsensus`

src/libMessage/Messenger.cpp
ansnunez: Cool, didn't know about this. Will update it. Thanks

Gnnng:
Do we have any plan to follow the protobuf style guide [here](https://developers.google.com/protocol-buffers/docs/style)? If the field name in `.proto` file could be `this_is_field_name`, then the generated function name would also look better in C++ side.

There're also linter tools available that we can enforce these style rules easily
- https://github.com/ckaznocha/protoc-gen-lint
- https://github.com/uber/prototool

ansnunez:
@Gnnng not opposed to it. Mind if we do in a separate PR? I can write up as a to-do issue.

Gnnng:
> Mind if we do in a separate PR? I can write up as a to-do issue.

Sounds good. Thanks!

src/libMessage/Messenger.cpp
```diff
@@ -302,894 +560,1935 @@ bool Messenger::GetNodeDSBlock(const vector<unsigned char>& src,
...
 
-    result.set_listenport(listenPort);
+    Signature signature;
 
-    if (!result.IsInitialized())
+    ProtobufByteArrayToSerializable(result.signature(), signature);
+
+    if (!Schnorr::GetInstance().Verify(tmp, signature, leaderKey))
     {
-        LOG_GENER
```
ckyang: How about using "const auto i&"?

ckyang: How about using for(const auto i : result.consensusinfo().bitmap())?

src/libNode/MicroBlockProcessing.cpp
```diff
@@ -1601,8 +1630,4 @@ bool Node::MicroBlockValidator(const vector<unsigned char>& microblock,
     }
 
     return valid;
```
ckyang: Just feel uncomfortable on this kind of checking, with a strange while(false). (I know it's not the modification of this PR). I think only one if condition is needed for the checking. If possible, could you please help to modify?

src/libZilliqa/Zilliqa.cpp
```diff
@@ -97,7 +98,7 @@ Zilliqa::Zilliqa(const std::pair<PrivKey, PubKey>& key, const Peer& peer,
     , m_ds(m_mediator)
     , m_lookup(m_mediator)
     , m_n(m_mediator, syncType, toRetrieveHistory)
-    , m_cu(key, peer)
+    //    , m_cu(key, peer)
```
ckyang: Why preserve m_cu? For future possible usage?

src/libMessage/Messenger.cpp
ckyang: How about using for(const auto i : result.consensusinfo().bitmap())?

ckyang: How about using "const auto i&"?

src/libZilliqa/Zilliqa.cpp
ansnunez: Nope, more like just to remember we have to do some clean-up work. Anyway this will be taken care of in #571 hopefully by next week.

src/libNode/MicroBlockProcessing.cpp
ansnunez: Historically, this was because we would have each check in a separate if condition, and any time that condition fails, we would break out of the while loop. I can remove the while loop now.

src/libZilliqa/Zilliqa.cpp
ckyang: Got it, thanks.

src/libMessage/Messenger.cpp
ansnunez: ok

src/libNode/MicroBlockProcessing.cpp
Gnnng: There're similar patterns in other locations in this PR. A large object is defined first, and copied over to a dynamically allocated object.

```cpp
LargeObject obj;

if (...)
{
    // do something with obj
}

m_obj_unique_ptr.reset(new LargeObject(object)); // an uncessary copy will happen if LargeObject doesn't have a move constructor
```

PR #593, opened 2018-09-17
Fix build error after remove "using namespace std" in Miner.h

ShengguangXiao:
## Description

The project failed to build GPU mining project after remove "using namespace std" in Miner.h.

Put std:: in front of std library types.

## Review Suggestion

Check the CUDAMiner.h/cpp and CLMiner.h/cpp.

Build the project with "./build.sh cuda", and "./build.sh opencl", the build should can pass.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


bb111189:
Verified for ```./build.sh cuda```

PR #599, opened 2018-09-19
Deserialization error fix due to wrong offset.

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ansnunez:
https://github.com/Zilliqa/Zilliqa/blob/98e14e1f0ed4365bb23f6ca90232a1e0b2ba14fd/src/libDirectoryService/MicroBlockProcessing.cpp#L463

@KaustubhShamshery this line should use `sizeof(unsigned char)`, not `MessageOffset::INST`.  It was just pure coincidence that `MessageOffset::INST` is 1.

This means you should also use `sizeof(unsigned char)` in your code change.

PR #601, opened 2018-09-19
Fix node drop after micro block consensus failed

ShengguangXiao:
## Description

Fix node drop out the network after micro block fail because state not correct problem.

Add condition check when transit to WAITING_DSBLOCK state for node.

https://github.com/Zilliqa/Issues/issues/206

## Review Suggestion

Check libNode/PoWProcessing.cpp

Set DS priority to 7, run local test, , there should be no node drop out problem.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/PoWProcessing.cpp
+++ b/src/libNode/PoWProcessing.cpp
@@ -105,11 +105,6 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


ShengguangXiao:
Tested with 200 cloud nodes, run 27 DS blocks, 125 Tx Blocks, didn't find node drop out problem.

PR #603, opened 2018-09-20
Fix CheckMicroBlocks in FinalBlockConsensus

kaikawaliu:
## Description

Previously, we only check the txnRootHash and the stateDeltaHashRoot of the microblocks, but in case they are all null, the backup may miss some microblocks. Thus in the vacuous epoch running coinbase, the node may miss some rewardee thus cannot pass the consensus.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ansnunez:
Don't forget to update the constants files in the testnet repo

PR #607, opened 2018-09-21
Added some functions for supporting the functioning of Archival Nodes.

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -812,6 +812,10 @@ bool Lookup::ProcessGetDSBlockFromSeed(const vector<unsigned char>& message,
                           .GetHeader()
                           .GetBlockNum();
     }
+    else if (lowBlockNum == 0)
+    {
+        lowBlockNum = 1;
+    }
 
     if (highBlockNum == 0)
     {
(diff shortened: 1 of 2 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -990,6 +994,10 @@ bool Lookup::ProcessGetTxBlockFromSeed(const vector<unsigned char>& message,
                           .GetHeader()
                           .GetBlockNum();
     }
+    else if (lowBlockNum == 0)
+    {
+        lowBlockNum = 1;
```
Gnnng: I was trying to understand the reason why `lowBlockNum` is overwritten to 1. Can you clarify a bit?

KaustubhShamshery: so, as the current convention goes (1,0) ---> gives you the latest blocknum, and giving a request (0,0) --> gives you all the blocks in the blockchain. so keeping that in mind, when lookup recv 0 as low blocknum, it starts to give you the blocknums starting from 0.

Gnnng: Okay, got it. Perhaps we can add some comments there together with the code.

src/libLookup/Lookup.cpp
```diff
@@ -1133,6 +1143,76 @@ bool Lookup::ProcessGetTxBodyFromSeed(const vector<unsigned char>& message,
...
+
+    uint32_t i = 0;
+    for (const auto& shard : shards)
+    {
+        LOG_GENERAL(INFO, "Size of shard " << i << " " << shard.size());
+        i++;
+    }
+    lock_guard<mutex> g(m_mutexShards);
+
+    m_mediator.m_ds->m_shards = shards;
```
ansnunez: I think you can do a std::move here to be more efficient

PR #613, opened 2018-09-24
Handle Trie exception to avoid crash

ShengguangXiao:
## Description

Fix software crash because exception throwed from AccountStore::RetrieveFromDisk.

Catch the exception and log the details.

https://github.com/Zilliqa/Issues/issues/166

## Review Suggestion

Check the try..catch added.

Need to run large scale test, to see if number of the crash nodes decreased.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


ShengguangXiao:
Run 200 nodes test, run to 25 DS Blocks and 120 Tx Blocks without problem.

PR #616, opened 2018-09-25
 Fix interruption caused by late ProcessFinalBlockConsensus

kaikawaliu:
## Review Suggestion

Previously in `ProcessFinalBlockConsensusWhenDone()`, when the vacuous epoch comes, the `m_consensusID` is updated after ethash was set up, which may be quite late, but the state was already changed to `POW_SUBMISSION`.
However under current design, when a ds receive finalblock consensus message, it will check whether it's under the correct state, otherwise it will force it to be changed to the right state for processing finalblock, and triggers `RunConsensusOnFinalBlock`.
In this case, if a node happens to send the consensus message slightly late, it will interrupt the normal procedure of the protocol.

The fix here is to solve this issue.

And there is also a minor fix on the logging of processed_count, as it's tiny, was also included here.

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/FinalBlockPostProcessing.cpp
+++ b/src/libDirectoryService/FinalBlockPostProcessing.cpp
@@ -379,6 +379,10 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone()
             LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
                       "[PoW needed]");
 
+            m_consensusID = 0;
+            m_mediator.m_node->m_consensusID = 0;
+            m_mediator.m_node->m_consensusLeaderID = 0;
+
             CleanFinalblockConsensusBuffer();
 
             m_mediator.m_node->CleanCreatedTransaction();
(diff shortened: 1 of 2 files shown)


src/libNode/Node.cpp
```diff
@@ -735,22 +735,24 @@ bool Node::ProcessTxnPacketFromLookupCore(
                 else
                 {
                     compIdx.insert(tx);
-                    txn_sent_count++;
                 }
+                txn_sent_count++;
```
ShengguangXiao: Put ++ in front will be more efficient.

Gnnng: I'm afraid there's no difference in terms of the efficiency between `txn_sent_count++` and `++txn_sent_count`.
This is an example in `-O2` (also the same in `-O2`) https://godbolt.org/z/ci-DHi. Let's just trust the compiler can do better work on this.

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -535,6 +536,8 @@ bool DirectoryService::ProcessFinalBlockConsensus(
         LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
                   "Process final block arrived early, saved to buffer");
 
+        lock_guard<mutex> g(m_mutexConsensus);
```
bb111189: Not sure whether is it necessary here. Any insight on this?

kaikawaliu: For every invoking of ProcessFinalBlockConsensus, when it comes to ProcessMessage and finally enters the WhenDone function, it will be covered by this mutex, it protect the race condition of the changes on m_consensusID and so on. While the following code deals with the m_consensusID, and eventually tries to RunConsensus, it's necessary to have this mutex for protection

PR #617, opened 2018-09-25
Disallow bootstrap function after blockchain started

ShengguangXiao:
## Description

Disallow bootstrap function after blockchain started. Improve security of the blockchain. Malicious user may use this function to attack the network.

Check the block number when call these functions, if larger than 1, then don't process function.

https://github.com/Zilliqa/Issues/issues/176

## Review Suggestion

Check the file changes.

Run local 20 nodes test. After bootstrap, run the below script to send bootstrap message to the nodes again, from the running log, should can find "...is a bootstrap function, it shouldn't be called after blockchain started" in the log files.

> for ds in {1..10}
> do
>     python tests/Zilliqa/test_zilliqa_local.py sendcmd $ds 01000000000000000000000000000100007F00001389
> done
> sleep 10
> 
> for node in {11..20}
> do
>     python tests/Zilliqa/test_zilliqa_local.py startpow $node 10 0000000000000001 05 03 2b740d75891749f94b6a8ec09f086889066608e4418eda656c93443e8310750a e8cc9106f8a28671d91e2de07b57b828934481fadf6956563b963bb8e5c266bf
> done

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DirectoryService.cpp
+++ b/src/libDirectoryService/DirectoryService.cpp
@@ -175,6 +175,16 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
     // Message = [Primary node IP] [Primary node port]
     LOG_MARKER();
 
+    if (m_mediator.m_currentEpochNum > 1)
+    {
+        // TODO: Get the IP address of who send this message, and deduct its reputation.
+        LOG_GENERAL(WARNING,
+                    "DirectoryService::ProcessSetPrimary is a bootstrap "
+                    "function, it shouldn't be called after blockchain "
+                    "started.");
+        return false;
+    }
+
     // Peer primary(message, offset);
     Peer primary;
     if (primary.Deserialize(message, offset) != 0)
@@ -215,7 +225,6 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
     // TODO: Refactor this code
     if (primary == m_mediator.m_selfPeer)
     {
-
         PeerStore& dsstore = PeerStore::GetStore();
         dsstore.AddPeerPair(
             m_mediator.m_selfKey.second,
--- a/src/libNode/PoWProcessing.cpp
+++ b/src/libNode/PoWProcessing.cpp
@@ -314,6 +314,15 @@ bool Node::ProcessStartPoW(const vector<unsigned char>& message,
                                        .GetBlockNum()
                       + 1);
 
+    if (m_mediator.m_currentEpochNum > 1)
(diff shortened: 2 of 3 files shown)


src/libDirectoryService/DirectoryService.cpp
```diff
@@ -175,6 +175,15 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
     // Message = [Primary node IP] [Primary node port]
     LOG_MARKER();
 
+    if (m_mediator.m_currentEpochNum > 1)
+    {
+        LOG_GENERAL(WARNING,
+                    "DirectoryService::ProcessSetPrimary is a bootstrape "
```
bb111189: should be bootstrap

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -175,6 +175,15 @@ bool DirectoryService::ProcessSetPrimary(const vector<unsigned char>& message,
     // Message = [Primary node IP] [Primary node port]
     LOG_MARKER();
 
+    if (m_mediator.m_currentEpochNum > 1)
+    {
+        LOG_GENERAL(WARNING,
```
bb111189: add a comment to deduct score from the reputation manager

src/libNode/PoWProcessing.cpp
```diff
@@ -314,6 +314,14 @@ bool Node::ProcessStartPoW(const vector<unsigned char>& message,
                                        .GetBlockNum()
                       + 1);
 
+    if (m_mediator.m_currentEpochNum > 1)
+    {
+        LOG_GENERAL(WARNING,
+                    "Node::ProcessStartPoW is a bootstrape function, it "
```
bb111189: similar comments as above

src/libDirectoryService/DirectoryService.cpp
ShengguangXiao: If a attacker send a lot of bootstrape message to this node, then this node's reputation will decrease a lot.

ShengguangXiao:
> Can you list all the bootstrap functions to be disabled in this PR?

Actually only two. ProcessStartPoW for shard node and ProcessSetPrimary for DS node

src/libDirectoryService/DirectoryService.cpp
bb111189: Maybe for consistency with other todo in the codebase, maybe we can replace "Work need to do" with TODO. Thanks

Issue #555, opened 2018-09-03
Test GPU mining on a rig with 6 cards and result is only use one GPU cards


dtompichayut:
They are follow my instruction here for local test with Nvidia GPU
[https://t.me/zilliqaminer/1395](url)

And the result of gpu resource usage is
[https://t.me/zilliqaminer/1385](url)

Is it need advance config or it is a bug?

ShengguangXiao:
Now the Zilliqa only support using the first GPU. We can change the software to let user select which GPU to use, so user can run multiple Zilliqa software using different GPU. But it need some time.

dtompichayut:
Thanks for answer, I will keep this issue open and let me know for if it is available.
This feature is important to get support from miner that own some rig to be come zilliqa nodes

dtompichayut:
Hi @jiayaoqijia after testing it still look not working as expected, It look like a node use all gpu instead of using one, it should can assign a gpu for each node to use like @ShengguangXiao mentioned

PR #618, opened 2018-09-25
Support using multiple GPUs to do mining together

ShengguangXiao:
## Description

Support using multiple GPUs to do mining together

Update the POW library. Create a new thread for each GPU to do mining.

https://github.com/Zilliqa/Zilliqa/issues/555

## Review Suggestion

Check the libPOW.

Use a PC with more than 1 GPUs, set NUM_DEVICE_TO_USE in constants to 2. If you are using OpenCL GPU, set OPENCL_GPU_MINE to true, if you are using CUDA gpu, set CUDA_GPU_MINE to true, build the software with ./build.sh opencl or ./build.sh cuda command. Then run cd build && ./tests/POW/Test_POW see if the test case can pass.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -15,6 +15,7 @@
 **/
 
 #include <boost/algorithm/string/predicate.hpp>
+#include <chrono>
 #include <ctime>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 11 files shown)


ShengguangXiao:
Run 20 nodes test, 2 nodes using 2 AMD gpu together to do PoW, run to 23 DS Blocks and 105 Tx Blocks without problem.

src/libPOW/pow.cpp
```diff
@@ -250,40 +251,89 @@ ethash_mining_result_t POW::MineFull(ethash_full_t& full,
...
     dev::eth::WorkPackage wp;
     wp.blockNumber = blockNum;
     wp.boundary = (dev::h256)(dev::u256)((dev::bigint(1) << 256)
                                          / (dev::u256(1) << difficulty));
 
     wp.header = dev::h256{header_hash.b, dev::h256::ConstructFromPointer};
-    wp.startNonce = std::time(0);
+
+    constexpr uint32_t NONCE_SEGMENT_WIDTH = 40;
+    const uint64_t NONCE_SGEMENT = (uint64_t)pow(2, NONCE_SEGMENT_WIDTH);
```
Gnnng: typo `NONCE_SGEMENT` --> `NONCE_SEGMENT`

src/libPOW/pow.cpp
```diff
@@ -250,40 +251,89 @@ ethash_mining_result_t POW::MineFull(ethash_full_t& full,
...
+                                << index);
     dev::eth::WorkPackage wp;
     wp.blockNumber = blockNum;
     wp.boundary = (dev::h256)(dev::u256)((dev::bigint(1) << 256)
                                          / (dev::u256(1) << difficulty));
 
     wp.header = dev::h256{header_hash.b, dev::h256::ConstructFromPointer};
-    wp.startNonce = std::time(0);
+
+    constexpr uint32_t NONCE_SEGMENT_WIDTH = 40;
```
Gnnng: perhaps `const` will be good enough here. Just for code tidiness.

src/libPOW/pow.h
```diff
@@ -102,7 +102,11 @@ class POW
     ethash_light_t ethash_light_client;
     uint64_t currentBlockNum;
     std::atomic<bool> m_shouldMine;
-    std::unique_ptr<dev::eth::Miner> m_miner;
+    std::vector<dev::eth::MinerPtr> m_miners;
+    std::vector<ethash_mining_result_t> m_vecMiningResult;
+    std::atomic<int> m_minerIndex;
+    std::condition_variable m_cvMineResult;
```
Gnnng: perhaps we can be consistent on the namings? Using the word `MiningResult` is better.

src/libPOW/pow.h
```diff
@@ -102,7 +102,11 @@ class POW
...
+    std::mutex m_mutexMineResult;
```
Gnnng: The same naming suggestion here, `MiningResult`

src/libPOW/pow.h
ShengguangXiao: Ok, I will update it.

PR #621, opened 2018-09-26
Support for Archival Nodes 2

KaustubhShamshery:
* Dispatch Microblocks to lookup
* Storage for Microblocks
* fetching microblocks from lookups

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (commit: e61bb7c)
- [x] small-scale cloud test (commit: 08e2ba6)


src/libMessage/ZilliqaMessage.proto
```diff
@@ -439,3 +439,31 @@ message ConsensusCommitFailure
     required ConsensusInfo consensusinfo = 1;
     required ByteArray signature         = 2;
 }
+
+message MicroBlockInfo
```
ansnunez: Please make this a nested message inside `LookupGetMicroBlockFromLookup`

src/libMessage/ZilliqaMessage.proto
```diff
@@ -439,3 +439,31 @@ message ConsensusCommitFailure
...
+{
+    required uint64 blocknum      = 1;
+    repeated uint32 shards        = 2;
+}
+
+message LookupGetMicroBlockFromLookup
```
ansnunez: Please move these messages right after `LookupSetShardsFromSeed`

ansnunez: Same goes for Messenger.h and Messenger.cpp

src/libMessage/Messenger.h
```diff
@@ -409,6 +409,41 @@ class Messenger
         const uint32_t consensusID, const std::vector<unsigned char>& blockHash,
         uint16_t& backupID, std::vector<unsigned char>& errorMsg,
         const std::deque<std::pair<PubKey, Peer>>& committeeKeys);
+
+    static bool SetLookupGetMicroBlockFromLookup(
+        std::vector<unsigned char>& dest, unsigned int offset,
```
ansnunez: Sorry to be fussy, but to be consistent with the other functions, please rename `dest` to `dst`, and make offset `const`, and make the other arguments `const` too.

src/libMessage/Messenger.h
```diff
@@ -409,6 +409,41 @@ class Messenger
...
+        const std::map<uint64_t, std::vector<uint32_t>>& microBlockInfo,
+        uint32_t portNo);
+
+    static bool GetLookupGetMicroBlockFromLookup(
```
ansnunez: For the `GetLookupGetxxx` functions, same comment about `offset`

src/libMessage/Messenger.cpp
```diff
@@ -3214,3 +3214,213 @@ bool Messenger::GetConsensusCommitFailure(
...
+{
+    LOG_MARKER();
+
+    LookupGetTxnsFromLookup result;
+
+    result.set_portno(portNo);
+
+    for (const auto& txhash : txnhashes)
+    {
+        copy(txhash.asArray().begin(), txhash.asArray().end(),
```
ansnunez: For consistency, I propose you copy the way `repeated bytes tranhashes` in `ProtoMicroBlock` is serialized. (I'll update the Protobuf guide for this)

src/libMessage/ZilliqaMessage.proto
KaustubhShamshery: @ansnunez I needed to make use of this type (in declaration) so brought it outside.

src/libPersistence/BlockStorage.cpp
```diff
@@ -175,6 +176,58 @@ bool BlockStorage::PutTxBody(const dev::h256& key,
...
+                                      sizeof(uint32_t));
+
+    return DataConversion::Uint8VecToHexStr(vec);
+}
+
+bool BlockStorage::PutMicroBlock(const uint64_t& blockNum,
+                                 const uint32_t& shardId,
+                                 const vector<unsigned char>& body)
+{
+    if (!LOOKUP_NODE_MODE)
```
Gnnng: We can remove this check if we already know where this function would be called.

src/libPersistence/BlockStorage.cpp
```diff
@@ -175,6 +176,58 @@ bool BlockStorage::PutTxBody(const dev::h256& key,
...
+
+    return (ret == 0);
+}
+
+bool BlockStorage::GetMicroBlock(const uint64_t& blockNum,
+                                 const uint32_t& shardId,
+                                 MicroBlockSharedPtr& microblock)
+{
+
+    if (!LOOKUP_NODE_MODE)
```
Gnnng: We can remove this check if we already know where this function would be called.

src/libPersistence/BlockStorage.cpp
```diff
@@ -175,6 +176,58 @@ bool BlockStorage::PutTxBody(const dev::h256& key,
...
+    }
+
+    string key = MakeKey(blockNum, shardId);
+    string blockString = m_microBlockDB->Lookup(key);
+
+    if (blockString.empty())
+    {
+        return false;
+    }
+    microblock = MicroBlockSharedPtr(new MicroBlock(
```
Gnnng: Use 'microblock = make_shared<...>(...)' for performance reason

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -691,6 +693,34 @@ void Node::CallActOnFinalblock()
     BroadcastTransactionsToLookup(txns_to_send);
 }
 
+void Node::BroadcastMicroBlockToLookup()
```
Gnnng: The suggestion for entire function is: do the check `m_microblock == nullptr` first and have a return if it failed, then proceed with the heavy logic.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -691,6 +693,34 @@ void Node::CallActOnFinalblock()
...
+        curr_offset += sizeof(uint64_t);
+        m_microblock->Serialize(msg, curr_offset);
+        m_mediator.m_lookup->SendMessageToLookupNodes(msg);
+    }
+    else if (m_microblock == nullptr)
+    {
+        LOG_GENERAL(WARNING, "MicroBlock is null");
+    }
+    else
+    {
```
Gnnng: This clause is unreachable.

src/libLookup/Lookup.cpp
```diff
@@ -1218,6 +1218,237 @@ bool Lookup::ProcessSetSeedPeersFromLookup(const vector<unsigned char>& message,
     return true;
 }
 
+bool Lookup::AddMicroBlockToStorage(const uint64_t& blocknum,
+                                    const MicroBlock& microblock)
+{
+    if (!LOOKUP_NODE_MODE)
```
Gnnng: We can remove this check if we already know where this function would be called.

src/libLookup/Lookup.cpp
```diff
@@ -1218,6 +1218,237 @@ bool Lookup::ProcessSetSeedPeersFromLookup(const vector<unsigned char>& message,
...
+    {
+        LOG_GENERAL(INFO, "not expected to be called from non-lookup");
+        return false;
+    }
+    uint32_t id = microblock.GetHeader().GetShardID();
+
+    TxBlock txblk = m_mediator.m_txBlockChain.GetBlock(blocknum);
+    unsigned int i = 0;
+    if (!(txblk == TxBlock()))
```
Gnnng: The entire if-statement has a better form

```cpp
if (txblk == TxBlock())
{
    LOG_GENERAL(...);
    return false;
}

for (...)
{
    ...
}

if (i == txblk.GetShardIDs().size())
{
    LOG_GENERAL(...);
    return false;
}

if (TxRootHash mismatched && StateDeltaHash mismatched)
{
    LOG_GENERAL(...);
    return false;
}

if (!BlockStorage::GetBlockStorage().PutMicroBlock(blocknum, id, body))
{
    LOG_GENERAL(...);
    return false;  // This return seems to be missing in the code.
}

return true;
```

src/libLookup/Lookup.cpp
```diff
@@ -1248,6 +1248,243 @@ bool Lookup::ProcessSetSeedPeersFromLookup(const vector<unsigned char>& message,
...
+
+    SendMessageToRandomLookupNode(msg);
+}
+
+void Lookup::CommitMicroBlockStorage()
+{
+    lock_guard<mutex> g(m_mutexMicroBlocksBuffer);
+    const uint64_t& currentEpoch = m_mediator.m_currentEpochNum;
+
+    for (uint64_t i = 1; i <= currentEpoch; i++)
```
kaikawaliu: This loop will be very very long if have large epoch num in the future. you can just use an iterate to loop among the map for only existing value

src/libDirectoryService/Coinbase.cpp
```diff
@@ -73,6 +73,7 @@ bool DirectoryService::SaveCoinbaseCore(const vector<bool>& b1,
     for (const auto& kv : shard)
     {
         const auto& pubKey = std::get<SHARD_NODE_PUBKEY>(kv);
+        LOG_GENERAL(INFO," "<<pubKey);
```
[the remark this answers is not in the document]
KaustubhShamshery: this commit will be roll backed

src/libLookup/Lookup.cpp
KaustubhShamshery: noted

src/libLookup/Lookup.cpp
```diff
@@ -1248,6 +1248,247 @@ bool Lookup::ProcessSetSeedPeersFromLookup(const vector<unsigned char>& message,
...
+
+    return true;
+}
+
+bool Lookup::ProcessGetMicroBlockFromLookup(
+    const vector<unsigned char>& message, unsigned int offset, const Peer& from)
+{
+    LOG_MARKER();
+
+    if (!LOOKUP_NODE_MODE)
```
Gnnng: Same question here for the check. Can we remove it if we know this function won't be called wrongly? Or can we do the check in the caller so that the function can be cleaner?

src/libLookup/Lookup.cpp
```diff
@@ -1248,6 +1248,247 @@ bool Lookup::ProcessSetSeedPeersFromLookup(const vector<unsigned char>& message,
...
+                                          unsigned int offset, const Peer& from)
+{
+    //message = [epochNum][microblock]
+
+    unsigned int curr_offset = offset;
+
+    uint64_t epochNum = Serializable::GetNumber<uint64_t>(message, curr_offset,
+
+                                                          sizeof(uint64_t));
+    if (!LOOKUP_NODE_MODE)
```
Gnnng: Same question for the check.

PR #622, opened 2018-09-26
Add ThreadSanitizer and Address Sanitizer option to build system 

Gnnng:
## Description

This PR adds two new options to our build system. Some refactoring is also being done in `build.sh` and `CMakeLists.txt` to have better error checking.

The refactored build system will accept the following categories of the option combination

- zero options to have a default build
- one of the GPU mining options 
- one of the Sanitizer options
- one of the Sanitizer options and one of the Sanitizer options

## Review Suggestion

See how the options passed to `build.sh` are parsed and subsequently consumed by `CMakeLists.txt`

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~


ShengguangXiao:
Can check the options in the build script? Because if user input both tsan and asan, it need long time to find out why the build is failed.

Gnnng:
> Can check the options in the build script? Because if user input both tsan and asan, it need long time to find out why the build is failed.

I don't think it's necessary to make the `build.sh` complicated. It's just a place for shortcuts that save the time of 80% developers. For all the error checking of wrong options, it's better to leave them to CMake. The issue of long waiting time is because of the order of commands in `CMakeLists.txt`. If we want to have a fast failure on option checking errors, we'd better move the them before other time-consuming commands in `CMakeLists.txt`, which perhaps is out of the scope of this PR.

PR #626, opened 2018-09-26
Fallback Solution

kaikawaliu:
#### PR for implementing Fallback solution, includes:

- FallbackPreConsensus
- FallbackPostConsensus
- FallbackBlockProcessing

#### Details worth mentioning during implementation:

- Use deque for m_shards, for the ease to pop_front

- Adopted Protobuf for all message used in fallback solution.

- Refactoring MicroblockConsensus into two files, MicroblockPreProcessing & MicroblockPostProcessing, no functional change

- **Unifying the m_consensusID to libDS and libNode, can be accessed via mediator.**

- **Put the logic of determining vacuous epoch into mediator**

- New node joining will fetch the latest consensusID from lookup, based on which the timing to fetch latest states, fetch ds info, init mining and so on will be determined.

- Move the place of resetting consensusID and consensusLeader to zero from vacuous epoch finalblock processing to libDS/DSBlockPostProcessing and libNode/DSBlockProcessing, as to provide the latest consensus ID to new node.

## Review Suggestion

- For testing with the fallback feature, indicate the epoch when the ds will be killed by setting `FALLBACK_TEST_EPOCH` in constants, then run `./build.sh fallbacktest` to build the code, then start lookup and normal node as usual

### Implementation

- [x] **ready for review**

- [x] local machine test (including new node joining)

- [X] small-scale cloud test (200 nodes)


src/libData/BlockData/BlockHeader/FallbackBlockHeader.cpp
```diff
@@ -0,0 +1,208 @@
...
+    curOffset += m_leaderNetworkInfo.Serialize(dst, curOffset);
+    curOffset += m_leaderPubKey.Serialize(dst, curOffset);
+    SetNumber<uint32_t>(dst, curOffset, m_shardId, sizeof(uint32_t));
+    curOffset += sizeof(uint32_t);
+    SetNumber<uint256_t>(dst, curOffset, m_timestamp, UINT256_SIZE);
+    curOffset += UINT256_SIZE;
+    return size_needed;
+}
+
+int FallbackBlockHeader::Deserialize(const vector<unsigned char>& src,
```
KaustubhShamshery: Should not you use protobuf ?

kaikawaliu: It's used, but the old school serialization is also implemented

src/libNode/MicroBlockPostProcessing.cpp
```diff
@@ -0,0 +1,385 @@
+/**
```
kaikawaliu: No functional changes, just refactoring

src/libData/BlockData/BlockHeader/FallbackBlockHeader.cpp
kaikawaliu: Any example to refer to?

KaustubhShamshery: perhaps @ansnunez can guide what is correct

ansnunez: Ah never mind I see you actually already implemented and used protobuf

src/libNode/PoWProcessing.cpp
```diff
@@ -152,7 +153,7 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
         }
     }
 
-    if (m_state != MICROBLOCK_CONSENSUS)
+    if (m_state != MICROBLOCK_CONSENSUS_PREP || m_state != MICROBLOCK_CONSENSUS)
```
ansnunez: This looks familiar. Is it a fix, or an accidental rollback? Just want to make sure we didn't roll back something that was actually fixed.

src/libNode/Node.cpp
```diff
@@ -625,7 +627,8 @@ bool Node::ProcessTxnPacketFromLookup(
         return false;
     }
 
-    if (m_mediator.m_currentEpochNum % NUM_FINAL_BLOCK_PER_POW == 0
+    if (m_mediator.m_consensusID
```
ansnunez: Looks like the logic changed here but I can't understand why. Can you confirm it's what you intended?

constants.xml
```diff
@@ -45,6 +45,7 @@
         <FALLBACK_INTERVAL_STARTED>60</FALLBACK_INTERVAL_STARTED>
         <FALLBACK_INTERVAL_WAITING>3600</FALLBACK_INTERVAL_WAITING>
         <FALLBACK_EXTRA_TIME>10</FALLBACK_EXTRA_TIME>
+		<FALLBACK_TEST_EPOCH>2</FALLBACK_TEST_EPOCH>
```
ansnunez: wrong indentation (I think our travis might flag this?)

src/libNode/PoWProcessing.cpp
kaikawaliu: It's a fix. Previously we deleted it for simplicity, however as the MICROBLOCK_CONSENSUS is set when node finished RunConsensusOnMicroBlock, who may also fail, I feel that it's essential to have a PREP state which can act as a buffer to accept finalblock or fallbackblock even though the consensus obj failed to be created. Otherwise the state will always be stuck at  WAITING_DSBLOCK, then the recovery mechanisms such as sync with finalblock or fallbackblock cannot work properly.

src/libNode/Node.cpp
kaikawaliu: Yes, this is for realizing `"Once the PoW is done and a new DS block is broadcast, the network should continue with the last valid txn epoch number, thus the way of determining vacuous epoch will change."`. For example, we have 50 txBlocks per DS epoch, if the fallback happens at epoch 30, then the next vacuous epoch will be on epoch 80, this is determined by the consensus id, as we reset it after fallback, so it will increases from 0 at epoch 30 to 50 at epoch 80.

constants.xml
kaikawaliu: yes I also noticed

src/libNode/Node.cpp
ansnunez: What if we kept the current logic and just had fewer tx blocks in this particular epoch? Anyway it is an extraordinary case as fallback shouldn't happen occasionally. My concern is that the link between consensus ID and the epoch stuff is a very loose one. If we randomize the ID in the future then this logic will be broken.

src/libNode/FallbackPreProcessing.cpp
```diff
@@ -0,0 +1,536 @@
...
+                cv_lk, std::chrono::seconds(FALLBACK_INTERVAL_WAITING))
+            == std::cv_status::timeout)
+        {
+            if (m_mediator.m_ds->m_mode != DirectoryService::IDLE)
+            {
+                return;
+            }
+
+            if (m_state != FALLBACK_CONSENSUS_PREP
+                || m_state != FALLBACK_CONSENSUS)
```
ckyang: This condition checking seems always true

src/libNode/PoWProcessing.cpp
ckyang: Always return true?

src/libData/BlockData/BlockHeader/FallbackBlockHeader.cpp
```diff
@@ -0,0 +1,208 @@
...
+                    header.m_fallbackState, header.m_hash,
+                    header.m_leaderConsensusId, header.m_shardId,
+                    header.m_timestamp)
+        > std::tie(m_fallbackEpochNo, m_fallbackDSEpochNo, m_fallbackState,
+                   m_hash, m_leaderConsensusId, m_shardId, m_timestamp);
+}
+
+bool FallbackBlockHeader::operator>(const FallbackBlockHeader& header) const
+{
+    return !((*this == header) || (*this < header));
```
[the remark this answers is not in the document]
kaikawaliu: Ok noted

src/libNode/FallbackPreProcessing.cpp
kaikawaliu: Ok noted

src/libNode/Node.cpp
kaikawaliu: Ok I think that make sense, and also doable, will change

PR #636, opened 2018-10-01
P2pComm Network layer - manager/wrapper for gossip  library

sandipbhoir:
## Description
This PR cover the P2PComm network layer changes which acts as an manager /wrapper for gossip library (libRumorSPreading).

## Review Suggestion
Main network APIs used to broadcast using Gossip:

1) Initialize the setup using :

P2PComm::InitializeRumorManager ( PeerList )
    Stops Existing Round if any
    Initializes the RumorManager which manages all the rumors and their state.
    Starts Rounds. ( that runs after every ROUND_TIME_IN_MS ms )
     
    1. thats checks the state of every rumor in RumorHolder and sends
    to MAX_NEIGHBORS_PER_ROUND random peers if not old enough.

    2. Decision of state being old is made using Median Counter algorithm
    as explained in paper ( section 3 )
    

2) Start Sending message using :

P2PComm::SpreadRumor ( Message )

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


constants.xml
```diff
@@ -107,7 +110,7 @@
     -->
     </lookups>
     <gpu>
-        <NUM_DEVICE_TO_USE>1</NUM_DEVICE_TO_USE>
+        <NUM_DEVICE_TO_USE>1</NUM_DEVICE_TO_USE>
```
ShengguangXiao: Extra blank should be removed at the end of the line.

src/libNetwork/P2PComm.cpp
```diff
@@ -482,15 +492,18 @@ void P2PComm::EventCallback(struct bufferevent* bev, short events,
...
-        LOG_GENERAL(WARNING, "Incorrect message length.");
-        return;
-    }
-
     if (startByte == START_BYTE_BROADCAST)
     {
+        LOG_PAYLOAD(INFO, "Incoming broadcast message from " << from, message,
+                    Logger::MAX_BYTES_TO_DISPLAY);
+
+        // Check for length consistency
```
ansnunez: Not a major comment, but we can put this length check out of the if-else block. Consider it if there are other changes to be made, otherwise make a note of it for your next PR.

src/libNetwork/RumorManager.cpp
```diff
@@ -0,0 +1,405 @@
...
+        {
+        case 1:
+            return RRS::Message::Type::PUSH;
+        case 2:
+            return RRS::Message::Type::PULL;
+        case 3:
+            return RRS::Message::Type::EMPTY_PUSH;
+        case 4:
+            return RRS::Message::Type::EMPTY_PULL;
+        case 5:
```
ShengguangXiao: Can use
if (type > 0 && type < COUNT)
   return static_cast<RRS::Message::Type>(type);
else
    return RRS::Message::Type::UNDEFINED;

src/libNetwork/RumorManager.cpp
```diff
@@ -0,0 +1,405 @@
...
+            INFO,
+            "Received Rumor from peer which does not exist in my peerlist "
+                << from);
+        return false;
+    }
+
+    int64_t recvdRumorId = -1;
+    RRS::Message::Type t = convertType(type);
+    bool toBeDispatched = false;
+    if (t == RRS::Message::Type::EMPTY_PUSH
```
ShengguangXiao: Put the constant value in front or change the "RRS::Message::Type t" to const.

src/libNetwork/P2PComm.cpp
```diff
@@ -583,6 +606,81 @@ void P2PComm::EventCallback(struct bufferevent* bev, short events,
...
+        {
+            LOG_GENERAL(
+                WARNING,
+                "Gossip Msg Type and/or Gossip Round and/or SNDR LISTNR "
+                "Port is missing (messageLength = "
+                    << messageLength << ")");
+            return;
+        }
+
+        std::vector<unsigned char> gossipMsgTyp(message.begin() + HDR_LEN,
```
ansnunez: It seems like we could have just used `unsigned char gossipMsgType = message.at(HDR_LEN);`

src/libNetwork/P2PComm.cpp
```diff
@@ -583,6 +606,81 @@ void P2PComm::EventCallback(struct bufferevent* bev, short events,
...
+                                                message.begin() + HDR_LEN
+                                                    + GOSSIP_MSGTYPE_LEN);
+
+        std::vector<unsigned char> tmp(
```
ansnunez: Similarly, maybe going through a tmp buffer is something we can avoid. I think at this low layer we need to squeeze out as much efficiency as we can.

src/libNetwork/P2PComm.cpp
```diff
@@ -583,6 +606,81 @@ void P2PComm::EventCallback(struct bufferevent* bev, short events,
...
+        else if (p2p.m_rumorManager.RumorReceived((unsigned int)gossipMsgTyp[0],
+                                                  gossipMsgRound, rumor_message,
+                                                  from))
+        {
+            std::pair<vector<unsigned char>, Peer>* raw_message
+                = new pair<vector<unsigned char>, Peer>(rumor_message, from);
+
+            // Queue the message
+            m_dispatcher(raw_message);
+        }
```
ansnunez: Perhaps we can add an else case here, and log an error msg.

src/libNetwork/RumorManager.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+#include "RumorManager.h"
+
+#include <chrono>
+#include <thread>
+
+#include "P2PComm.h"
+#include "common/Messages.h"
+
+const unsigned char START_BYTE_GOSSIP = 0x33;
```
ansnunez: Hmm we had to define this twice in separate cpp files. Maybe we can move the constants to Messages.h

src/libNetwork/P2PComm.cpp
sandipbhoir: Its not error in general but just that we don't need to dispatch it. Reason for not dispatching is already logged in RumorReceived() for ex: when gossip is old one.

src/libNetwork/P2PComm.cpp
sandipbhoir: True.

src/libNetwork/P2PComm.cpp
sandipbhoir: I don't think we can do it over here. we need multiple chars.. similar to way we read the message length at start.

src/libNetwork/RumorManager.cpp
sandipbhoir: This will require quite a work to move all existing constants to message.h or may be new constant header file. Other clean approach i think of adding to p2pcomm.h 
extern const unsigned char START_BYTE_GOSSIP and get rid of definition from RumorManager.cpp

PR #637, opened 2018-10-01
Usecases of Gossip Protocol

sandipbhoir:
## Description
This PR covers the following use-cases of gossip:

ANNOUNCE = 0x00, (The leader multicasts to all the backup nodes)
CHALLENGE = 0x02, (The leader multicasts to all the backup nodes)
COLLECTIVESIG = 0x04, (The leader multicasts to all the backup nodes)
FINALCHALLENGE = 0x06, (The leader multicasts to all the backup nodes)
FINALCOLLECTIVESIG = 0x08, (The leader multicasts to all the backup nodes)
MICROBLOCKSUBMISSION = 0x03,  (A shard leader multicasts to all the DS nodes in DS shard)
FINALBLOCK = 0x04 ( DS nodes sends the final block to X shard nodes and X nodes multicast to all the nodes in that shard)
FORWARDTXNBLOCK = 0x09 (A lookup node multicasts a list of txns to X shard nodes and X nodes multicast to all the nodes in that shard) 

## Review Suggestion
By default, the existing multicast protocol is being used.

Local Test:
To enable it, Set flag BROADCAST_GOSSIP_MODE in Constants_local.xml

Small-Scale Test:
To enable it, Set flag BROADCAST_GOSSIP_MODE in Constants.xml
Flag GOSSIP_CUSTOM_ROUNDS_SETTINGS is disabled by default. And will rely on number of peers in each shards to calculate the round setting at runtime.

If shard-size is too small ( like one with local test ) will suggest to enable GOSSIP_CUSTOM_ROUNDS_SETTINGS flag.

### Implementation

- [x] **ready for review**

Enable 

- [x] local machine test
      - [x] local machine test (commit: 3fab16b) 
- [x] small-scale cloud test
      - [x] small-scale cloud test (commit: 3fab16b)


src/libConsensus/ConsensusLeader.cpp
```diff
@@ -627,14 +640,24 @@ bool ConsensusLeader::StartConsensus(
     // Multicast to all nodes in the committee
     // =======================================
 
-    deque<Peer> peer;
-
-    for (auto const& i : m_committee)
+    if (useGossipProto)
     {
-        peer.push_back(i.second);
+        // Other Nodes if too slow, dont get their RumorManager started yet.
```
ansnunez: Is this tmp fix still applicable? Or can we remove this commented-out code?

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -438,6 +439,20 @@ void DirectoryService::StartFirstTxEpoch()
...
         m_dsStartedMicroblockConsensus = false;
 
+        if (BROADCAST_GOSSIP_MODE)
+        {
+            std::vector<Peer> peers;
+            for (const auto& i : *m_mediator.m_node->m_myShardMembers)
+            {
+                if (i.second.m_listenPortHost != 0)
+                {
+                    peers.push_back(i.second);
```
ansnunez: Not critical, but let's use emplace_back more frequently instead of push_back from now on.

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -184,8 +184,30 @@ void DirectoryService::SendFinalBlockToShardNodes(
                 + 1
             << "] FBBLKGEN");
 
-        P2PComm::GetInstance().SendBroadcastMessage(shard_peers,
-                                                    finalblock_message);
+        if (BROADCAST_GOSSIP_MODE)
```
ansnunez: Hmm if we have gossip enabled, then the code block at line 143 is partially useless and the log messages in it misleading.  I think we can optimize this by avoiding going through a temp shard_peers if gossip is enabled. Also, we only need to log the actual peers stored inside the shardFinalBlockReceivers for the gossip case.

src/libNetwork/P2PComm.cpp
```diff
@@ -492,13 +492,6 @@ void P2PComm::EventCallback(struct bufferevent* bev, short events,
     const uint32_t messageLength = (message[2] << 24) + (message[3] << 16)
         + (message[4] << 8) + message[5];
 
-    // Check for length consistency
```
ansnunez: Was this code block inadvertently removed?

src/libNode/DSBlockProcessing.cpp
```diff
@@ -405,9 +405,38 @@ void Node::StartFirstTxEpoch()
         }
     }
 
+    // Choose N other DS nodes to be recipient of microblock
+    m_DSMBReceivers.clear();
+    unsigned int numOfMBReceivers = NUM_MICROBLOCK_GOSSIP_RECEIVERS;
```
ansnunez: Not critical, but for brevity and clarity I suggest in the future to use std::min, as I think this code pattern was used at least twice in this PR.

src/libNode/DSBlockProcessing.cpp
```diff
@@ -405,9 +405,38 @@ void Node::StartFirstTxEpoch()
...
+    {
+        numOfMBReceivers = m_mediator.m_DSCommittee->size();
+    }
+
+    for (unsigned int i = 0; i < numOfMBReceivers; i++)
+    {
+        m_DSMBReceivers.emplace_back(m_mediator.m_DSCommittee->at(i).second);
+    }
+
+    m_consensusLeaderID = 0;
```
ansnunez: Can you double-check the need for this line 421? I just remember @kaikawaliu had a PR which involved moving this line to somewhere else, if I'm not mistaken.

PR #638, opened 2018-10-02
Put block number in consensus message and validate it

ShengguangXiao:
## Description

Put block number in consensus message and validate it

Put block number in consensus message and validate it

The block number is not used in the consensus, which may give the chance for hacker to attack our network.

## Review Suggestion

Check the file changes.

Run local test.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libConsensus/ConsensusBackup.cpp
+++ b/src/libConsensus/ConsensusBackup.cpp
@@ -83,9 +83,10 @@ bool ConsensusBackup::ProcessMessageAnnounce(
     // =======================================
 
     std::vector<unsigned char> errorMsg;
-    if (!m_msgContentValidator(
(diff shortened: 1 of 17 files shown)


ansnunez:
Changes look ok. Just wondering why epoch number is used for the block number in all consensuses though. Should we just say epoch number instead?

ShengguangXiao:
> Changes look ok. Just wondering why epoch number is used for the block number in all consensuses though. Should we just say epoch number instead?

Yes, they are actually the same. But in our software, in mediator it is named as epoch number, in block headers, it is named as block number. I think we may need another PR to unify the naming.

ansnunez:
But technically for microblock consensus, epoch number is not the same as block number (I think). So yes I agree let's unify the naming in a follow-up PR.

PR #643, opened 2018-10-03
Upgrading protocol on look-up nodes

ckyang:
## Description

Upgrading protocol should also be applied on look-up nodes.

### Implementation

- [x] **ready for review**

- [x] small-scale cloud test (20 nodes, commit: 373a7d8)

- [x] small-scale cloud test (200 nodes, commit: 373a7d8)


src/libUtils/UpgradeManager.cpp
```diff
@@ -394,37 +394,46 @@ bool UpgradeManager::ReplaceNode(Mediator& mediator)
...
-    {
-        LOG_GENERAL(INFO,
-                    "DS backup node, upgrade after "
-                        << TERMINATION_COUNTDOWN_IN_SECONDS + 1
+                    "Lookup node, upgrade after "
+                        << TERMINATION_COUNTDOWN_IN_SECONDS + 3
                         << " seconds...");
         this_thread::sleep_for(
-            chrono::seconds(TERMINATION_COUNTDOWN_IN_SECONDS + 1));
+            chrono::seconds(TERMINATION_COUNTDOWN_IN_SECONDS + 3));
```
ShengguangXiao: Why use + 1, + 2, + 3 here? How does the delay is determined?

ckyang: Ok, I will define some meaningful variables for them.

PR #644, opened 2018-10-03
Verify block hash in consensus to enhance security

ShengguangXiao:
## Description

Verify block hash in consensus to enhance security

Use last block hash in the consensus message, and validate it.

## Review Suggestion

Check the files changed.

Run local test.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
+++ b/src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
@@ -218,6 +218,18 @@ const BlockHash& TxBlockHeader::GetDSBlockHeader() const
     return m_dsBlockHeader;
 }
 
+BlockHash TxBlockHeader::GetMyHash() const
+{
+    SHA2<HASH_TYPE::HASH_VARIANT_256> sha2;
+    std::vector<unsigned char> vec;
+    Serialize(vec, 0);
+    sha2.Update(vec);
+    const std::vector<unsigned char>& resVec = sha2.Finalize();
+    BlockHash blockHash;
+    std::copy(resVec.begin(), resVec.end(), blockHash.asArray().begin());
+    return blockHash;
+}
+
 bool TxBlockHeader::operator==(const TxBlockHeader& header) const
 {
     return std::tie(m_type, m_version, m_gasLimit, m_gasUsed, m_prevHash,
--- a/src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
+++ b/src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
@@ -218,18 +218,6 @@ const BlockHash& TxBlockHeader::GetDSBlockHeader() const
     return m_dsBlockHeader;
 }
 
-BlockHash TxBlockHeader::GetMyHash() const
-{
-    SHA2<HASH_TYPE::HASH_VARIANT_256> sha2;
-    std::vector<unsigned char> vec;
-    Serialize(vec, 0);
-    sha2.Update(vec);
-    const std::vector<unsigned char>& resVec = sha2.Finalize();
-    BlockHash blockHash;
-    std::copy(resVec.begin(), resVec.end(), blockHash.asArray().begin());
-    return blockHash;
-}
-
 bool TxBlockHeader::operator==(const TxBlockHeader& header) const
 {
     return std::tie(m_type, m_version, m_gasLimit, m_gasUsed, m_prevHash,
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -519,7 +519,6 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary()
     ComputeTxnSharingAssignments(winnerPeer->second);
 
     // Create new consensus object
-    // Dummy values for now
     uint32_t consensusID = 0;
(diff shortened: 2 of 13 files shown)


src/libData/BlockData/BlockHeader/TxBlockHeader.cpp
```diff
@@ -218,6 +218,18 @@ const BlockHash& TxBlockHeader::GetDSBlockHeader() const
...
+    SHA2<HASH_TYPE::HASH_VARIANT_256> sha2;
+    std::vector<unsigned char> vec;
+    Serialize(vec, 0);
+    sha2.Update(vec);
+    const std::vector<unsigned char>& resVec = sha2.Finalize();
+    BlockHash blockHash;
+    std::copy(resVec.begin(), resVec.end(), blockHash.asArray().begin());
+    return blockHash;
+}
+
```
ckyang: Since GetMyHash() logic are identical between Ds and Tx, it would be nice to put onto BlockHeaderBase class.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -79,12 +65,17 @@ void DirectoryService::ComposeDSBlock(
...
     uint8_t dsDifficulty = DS_POW_DIFFICULTY;
     uint8_t difficulty = POW_DIFFICULTY;
     if (m_mediator.m_dsBlockChain.GetBlockCount() > 0)
     {
         DSBlock lastBlock = m_mediator.m_dsBlockChain.GetLastBlock();
         blockNum = lastBlock.GetHeader().GetBlockNum() + 1;
+        prevHash = lastBlock.GetHeader().GetMyHash();
+
+        LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
+                  "Prev DS block hash as per leader " << prevHash.hex());
```
ckyang: Seems "prevHash" only be used inside the scope for logging, maybe we can directly call lastBlock.GetHeader().GetMyHash() inside logging function to avoid extra variable used.

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -275,8 +271,10 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSPrimary()
     // Create new consensus object
     // Dummy values for now
```
ckyang: comment need to be modified

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -1133,8 +1131,10 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSBackup()
     // Create new consensus object
 
     // Dummy values for now
```
ckyang: comment need to be modified

src/libDirectoryService/ViewChangePreProcessing.cpp
```diff
@@ -328,8 +328,10 @@ bool DirectoryService::RunConsensusOnViewChangeWhenCandidateLeader()
     uint32_t consensusID = m_viewChangeCounter;
     // Create new consensus object
     // Dummy values for now
```
ckyang: comment need to be modified

src/libDirectoryService/DSBlockPreProcessing.cpp
ShengguangXiao: The preHash is used to initialize the DSBlock below.

ckyang: ok, didn't notice that

src/libDirectoryService/FinalBlockPreProcessing.cpp
ansnunez: Looks like this wasn't addressed, but let's forget it for now

src/libMessage/Messenger.cpp
```diff
@@ -658,12 +658,22 @@ namespace
             return false;
         }
 
-        if ((announcement.consensusinfo().blockhash().size()
-             != blockHash.size())
+        std::vector<unsigned char> remoteBlockHash;
```
ansnunez: Hmm I think going through `remoteBlockHash` is unnecessary. We can just say `if ((announcement.consensusinfo().blockhash().size() != blockHash.size()) || !equal(blockHash.begin(), blockHash.end(), announcement.consensusinfo().blockhash().begin()))`

ansnunez: In the event that the check fails and you need to log the hash, then yeah I think we can do `remoteBlockHash` inside the LOG_GENERAL, but not until the check fails

ShengguangXiao: I find if we don't covert it to std::vector<unsigned char> remoteBlockhash first, this comparison always fail, i don't know why.

ansnunez: Looks like this will make it work: `if (equal(blockHash.begin(), blockHash.end(), announcement.consensusinfo().blockhash().begin(), [](const unsigned char left, const char right) -> bool { return left == (unsigned char)right; }))`

ShengguangXiao: Let me try this solution.

ansnunez: Latest commit looks ok. As discussed, we will do another improvement later on (#648)

PR #645, opened 2018-10-04
Archival Node Functionality 

KaustubhShamshery:
## Description
Add new type of node, namely archival which fetch data from a lookup node after a prescribed amount of time. They use mongoDB to store information. No change in the functioning of normal/lookup nodes.

Explorer DB : Stores tx block , ds block, txns as jsons.
Archival DB : Stores tx block, ds block, txns as bytes

Note that only one of these will be active in an instance of archival node

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (commit : 9ad9017)

- [x] Add Script to smoothen MongoDB installation


src/libDB/Archival.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+bool Archival::RemoveFromFetchMicroBlockInfo(const uint64_t& blockNum,
+                                             const uint32_t shardId)
+{
+    LOG_MARKER();
+
+    lock_guard<mutex> g(m_mutexMicroBlockInfo);
+
+    vector<uint32_t>& shard_ids = m_fetchMicroBlockInfo[blockNum];
```
kaikawaliu: Should use `contains()` to check whether the key exist otherwise it will create one with the using of `[]`, and here better to use `at()` to provide a boundary check and exception handling

src/libDB/Archival.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+using namespace std;
+
+unsigned int REFRESH_DELAY = 5;
+
+void Archival::InitSync()
+{
+    auto func = [this]() -> void {
+        m_synchronizer.FetchDSInfo(m_mediator.m_lookup);
+        m_synchronizer.FetchOfflineLookups(m_mediator.m_lookup);
```
kaikawaliu: For the usage of FetchOfflineLookups, you can refer to https://github.com/Zilliqa/Zilliqa/blob/f8cf5417033f68ce152293555a790cd4940d3f42/src/libNode/Node.cpp#L352 where all the other queries is being blocked by a cv below until the offline lookups being received.

CMakeLists.txt
```diff
@@ -123,6 +123,15 @@ if (ENABLE_COVERAGE AND CMAKE_COMPILER_IS_GNUCXX)
...
 
+find_package(libmongocxx REQUIRED)
+find_package(libbsoncxx REQUIRED)
+include_directories(${LIBMONGOCXX_INCLUDE_DIR})
+include_directories(${LIBBSONCXX_INCLUDE_DIR})
+include_directories("/usr/local/include/mongocxx/v_noabi")
+include_directories("/usr/local/include/bsoncxx/v_noabi")
+include_directories("/usr/local/include/libmongoc-1.0")
+include_directories("/usr/local/include/libbson-1.0")
+include_directories("/usr/local/lib")
```
Gnnng: Need some clean-ups on these `include_directories` commands. Try to keep it minimum.

src/libDB/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+find_package(libmongocxx REQUIRED)
+find_package(libbsoncxx REQUIRED)
+include_directories(${LIBMONGOCXX_INCLUDE_DIR})
+include_directories(${LIBBSONCXX_INCLUDE_DIR})
+include_directories("/usr/local/include/mongocxx/v_noabi")
+include_directories("/usr/local/include/bsoncxx/v_noabi")
+include_directories("/usr/local/include/libmongoc-1.0")
+include_directories("/usr/local/include/libbson-1.0")
+include_directories("/usr/local/lib")
```
Gnnng: Need some clean-ups on these `include_directories` commands. Try to keep it minimum.

Gnnng: Also better to make a decision on which file to include these commands. It doesn't look good to me that both top-level CMakeLists.txt and `src/libDB/CMakeLists.txt` have these lines.

CMakeLists.txt
KaustubhShamshery: I dont know why but does not work without this

jiayaoqijia: I think the find_package helps you to include some libs and you can remove those from "include".

PR #646, opened 2018-10-04
Refactoring on UnavailableMicroBlocks

kaikawaliu:
## Description

No major functional changes, mainly on stricting the checks for UnavailableMicroBlocks.
With some clean and refactoring as well.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -103,26 +103,18 @@ void Node::StoreFinalBlock(const TxBlock& txBlock)
         << "] RECV");
 }
 
-bool Node::IsMicroBlockTxRootHashInFinalBlock(
-    TxnHash microBlockTxRootHash, StateHash microBlockStateDeltaHash,
-    const uint64_t& blocknum, bool& isEveryMicroBlockAvailable)
+bool Node::IsMicroBlockTxRootHashInFinalBlock(const ForwardedTxnEntry& entry,
+                                              bool& isEveryMicroBlockAvailable)
 {
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-              "Deleting unavailable "
-                  << "microblock txRootHash " << microBlockTxRootHash
-                  << " stateDeltaHash " << microBlockStateDeltaHash
-                  << " for blocknum " << blocknum);
+              "Deleting unavailable microblock: " << entry);
     lock_guard<mutex> g(m_mutexUnavailableMicroBlocks);
-    auto it = m_unavailableMicroBlocks.find(blocknum);
-    bool found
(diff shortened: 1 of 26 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -134,105 +126,113 @@ bool Node::LoadUnavailableMicroBlockHashes(const TxBlock& finalBlock,
...
+    if (doRejoin || m_doRejoinAtFinalBlock)
+    {
+        // LOG_GENERAL(WARNING,
+        //             "Failed the last microblock consensus but "
+        //             "still found my shard microblock, "
+        //             " need to Rejoin");
+        // RejoinAsNormal();
+        // return false;
     }
+
```
ckyang: Since there is nothing to do inside the condition, how about comment out the condition checking, or remove them directly (if not be used in the future)?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -134,105 +126,113 @@ bool Node::LoadUnavailableMicroBlockHashes(const TxBlock& finalBlock,
...
 }
 
 bool Node::RemoveTxRootHashFromUnavailableMicroBlock(
-    const uint64_t& blocknum, const TxnHash& txnRootHash,
-    const StateHash& stateDeltaHash)
+    const ForwardedTxnEntry& entry)
 {
-    for (auto it = m_unavailableMicroBlocks[blocknum].begin();
-         it != m_unavailableMicroBlocks[blocknum].end(); it++)
+    for (auto it = m_unavailableMicroBlocks[entry.m_blockNum].begin();
```
ckyang: Since "at()" will do boundary-check and throw exception, it's nice to use "at()" instead of "[]" for accessing container.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1329,22 +1209,17 @@ void Node::CommitForwardedTransactionBuffer()
...
             }
             m_forwardedTxnBuffer.erase(it);
-            break;
-        }
-        else
-        {
-            it++;
+            continue;
         }
+        it++;
```
ckyang: This line seems never be executed, since the loop always be continued before here.

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: Good point, but still need to keep

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: Noted

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: Noted

PR #650, opened 2018-10-05
Treebased Cluster Broadcasting

sandipbhoir:
## Description
In the new DS epoch, before receiving the DS block, a shard node doesn’t know the information of the other nodes in the same shard. Thus, we should leverage multicast sing treebased clustering to broadcast the DS block to the nodes within a shard. Also use the same to VC Block submissions triggered during final block consensus.

## Review Suggestion
Please refer the algorithm proposed in design document.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test
 - [x] small-scale cloud test (commit: 74b7cf8)


jiayaoqijia:
You need to run 200-node testnet.

src/libNode/DSBlockProcessing.cpp
```diff
@@ -647,6 +647,11 @@ bool Node::ProcessDSBlock(const vector<unsigned char>& message,
                 return false;
             }
 
+            if (BROADCAST_TREEBASED_CLUSTER_MODE)
```
KaustubhShamshery: !LOOKUP_NODE_MODE needed, causing lookup to crash

sandipbhoir: Lookup node won't reach that code statement. we already check !LOOKUP_NODE_MODE at line 550

KaustubhShamshery: The comment is dislocated. Refers to VCblock and FInal block. Resolved

sandipbhoir:
@KaustubhShamshery , i have added additonal logging but its upto programmer to make sure we use same constant (NUM_FORWARDED* or any other new constant) across DS node ( sender) and Shard Node (receiver) so that we have correct cluster-size.

PR #651, opened 2018-10-05
Feature Gas Mechanism

kaikawaliu:
## Description

This PR is for interfacing with the scilla interpreter for the new implementation of calculating gas.
So the interpreter will have one more field in the input as `-gaslimit XXX`, and the output of the interpreter will indicate the gas remained in the json object as `gas_remaining`

Besides, there is also another change on the nonce. Previously when a user calls a failure contract transaction, we will deduct the gas fee from him but the nonce didn't update. However in this case the user can generate multiple transaction with the same nonce. So in this PR we also increase the nonce of the user whenever its balance get changed.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -208,14 +215,16 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
             {
                 this->IncreaseNonce(fromAddr);
                 receipt.SetResult(false);
-                receipt.SetCumGas(CONTRACT_CREATE_GAS);
+                receipt.SetCumGas(transaction.GetGasLimit() - gasRemained);
```
KaustubhShamshery: how are we sure this wont be < 0

PR #652, opened 2018-10-05
Dynamically calculate sharding size

ckyang:
## Description

The sharding size is defined by following rule:
1. If shard node number < 651, use COMM_SIZE defined in constants.xml
2. Otherwise, apply ShardSizeCalculator::CalculateShardSize() to look for a proper sharding size.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


constants.xml
```diff
@@ -6,6 +6,7 @@
         <DS_MULTICAST_CLUSTER_SIZE>10</DS_MULTICAST_CLUSTER_SIZE>
         <TX_SHARING_CLUSTER_SIZE>10</TX_SHARING_CLUSTER_SIZE>
         <COMM_SIZE>200</COMM_SIZE>
+        <!-- Only take effect when shard node number < 651-->
```
ansnunez: Can you put this comment in the same line as COMM_SIZE? Although I'm not sure if clang-tidy will reformat it.

src/libMediator/Mediator.cpp
```diff
@@ -242,3 +243,19 @@ void Mediator::IncreaseEpochNum() {
...
+
+  if (useShardStr) {
+    for (const auto& shard : m_ds->m_shards) {
+      shardNodeNum += shard.size();
+    }
+  } else {
+    shardNodeNum = m_ds->m_allPoWs.size();
+  }
+
+  return (shardNodeNum < 651)
```
ansnunez: I think maybe we can move this decision-making logic into `CalculateShardSize`, so we don't use the magic number 651 in two places.

constants.xml
ckyang: Actually clang-tidy/clang-format will check this, but anyway I will move the comment above <COMM_SIZE>.

src/libDirectoryService/DirectoryService.h
```diff
@@ -378,6 +375,10 @@ class DirectoryService : public Executable, public Broadcastable {
   /// committee.
   std::atomic<Mode> m_mode;
 
+  std::mutex m_mutexAllPOW;
```
bb111189: Why was this moved?

src/libMediator/Mediator.cpp
```diff
@@ -242,3 +243,17 @@ void Mediator::IncreaseEpochNum() {
 }
 
 bool Mediator::GetIsVacuousEpoch() { return m_isVacuousEpoch; }
+
+uint32_t Mediator::GetShardSize(const bool& useShardStr) const {
```
bb111189: useShardStr may not be intuitive and may be confusing

src/libDirectoryService/DirectoryService.h
ckyang: Move from private to public, to allow mediator access it to calculate shard node number

src/libMediator/Mediator.cpp
ckyang: OK, I will change to a better name

PR #653, opened 2018-10-05
Add COPYRIGHT banner, checker, and fixup

Gnnng:
## Description

This PR fixes Zilliqa/Issues#182. The column limit of COPYRIGHT banner is set to 80 to prepare for the next PR of coding style reformating.

## Review Suggestion

See if `scripts/copyright_checker.sh` covers all the files that need to be checked.

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~

### DIFF
--- a/tests/depends/Trie/Test_TxnTrie.cpp
+++ b/tests/depends/Trie/Test_TxnTrie.cpp
@@ -1,7 +1,21 @@
-/**
-* Copyright (c) 2018 Zilliqa 
-* This is an alpha (internal) release and is not suitable for production.
-**/
+/*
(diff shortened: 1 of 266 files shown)


src/libUtils/RsyncThread.h
```diff
@@ -0,0 +1,18 @@
+/*
```
bb111189: Good catch. thanks

tests/depends/Trie/Test_TxnTrie.cpp
```diff
@@ -1,18 +1,21 @@
-/**
```
bb111189: includes got overwritten

PR #654, opened 2018-10-06
Reformat using clang-format

Gnnng:
## Description

This PR is a follow-up for #653 and google flavor is used when reformatting the code. Also a dummy author (maintainers@zilliqa.com) is used when the reformatting is applied.

The `.clang-format` is generated in the following manner
```
clang-format-7 -dump-config -style=Google > .clang-format
```

Also, the clang-format version we used is 7.0, which is the latest release from LLVM at the time this PR is created.

Other important changes include
- remove clang-format check on Travis Mac build as llvm 7.0 installation has not been added.
- make the `clang-format-fix` optional in `./build.sh`. Use `./build.sh llvmextra` when `clang-format-fix` is desired

## Review Suggestion

As this is a relatively huge PR, kindly following the commit history to see the change. Only the last commit (afc9bae082a619643be59873fb1c089113eeb274) applies the reformatting.

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~


jiayaoqijia:
You may disable the persistent-storage test and run 200-node testnet for regression test.

PR #659, opened 2018-10-08
Develop verify lookup node and messages

ShengguangXiao:
## Description

Develop verify lookup node and messages to enhance the security of Zilliqa blockchain.

Add lookup public key to the constants.xml, add signature and pubkey in lookup messages, so we can verify the anthentic of the messages.

## Review Suggestion

Check source code modified, the constants.xml and test_zilliqa_lookup.py.

Run a local test.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -90,7 +90,10 @@ void Lookup::SetLookupNodes() {
       inet_aton(v.second.get<string>("ip").c_str(), &ip_addr);
       Peer lookup_node((uint128_t)ip_addr.s_addr,
                        v.second.get<uint32_t>("port"));
-      m_lookupNodes.emplace_back(lookup_node);
+      PubKey pubKey(
+          DataConversion::HexStrToUint8Vec(v.second.get<std::string>("pubkey")),
+          0);
+      m_lookupNodes.emplace_back(pubKey, lookup_node);
     }
   }
 }
@@ -197,7 +200,7 @@ bool Lookup::GenTxnToSend(size_t num_txn,
   return true;
 }
 
-vector<Peer> Lookup::GetLookupNodes() {
+const VectorOfLookupNode& Lookup::GetLookupNodes() const {
   LOG_MARKER();
 
   return m_lookupNodes;
@@ -213,10 +216,11 @@ void Lookup::SendMessageToLookupNodes(
 
   for (const auto& node : m_lookupNodes) {
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-              "Sending msg to lookup node " << node.GetPrintableIPAddress()
-                                            << ":" << node.m_listenPortHost);
+              "Sending msg to lookup node "
+                  << node.second.GetPrintableIPAddress() << ":"
+                  << node.second.m_listenPortHost);
 
-    allLookupNodes.emplace_back(node);
+    allLookupNodes.emplace_back(node.second);
   }
 
   P2PComm::GetInstance().SendBroadcastMessage(allLookupNodes, message);
@@ -230,12 +234,13 @@ void Lookup::SendMessageToLookupNodesSerial(
   // to_string(m_mediator.m_currentEpochNum).c_str())
   vector<Peer> allLookupNodes;
 
-  for (auto node : m_lookupNodes) {
+  for (const auto& node : m_lookupNodes) {
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-              "Sending msg to lookup node " << node.GetPrintableIPAddress()
-                                            << ":" << node.m_listenPortHost);
+              "Sending msg to lookup node "
+                  << node.second.GetPrintableIPAddress() << ":"
+                  << node.second.m_listenPortHost);
 
-    allLookupNodes.emplace_back(node);
+    allLookupNodes.emplace_back(node.second);
   }
 
   P2PComm::GetInstance().SendMessage(allLookupNodes, message);
@@ -249,7 +254,7 @@ void Lookup::SendMessageToRandomLookupNode(
   // - NUM_LOOKUP_USE_FOR_SYNC;
   int index = rand() % m_lookupNodes.size();
 
-  P2PComm::GetInstance().SendMessage(m_lookupNodes[index], message);
+  P2PComm::GetInstance().SendMessage(m_lookupNodes[index].second, message);
 }
 
 void Lookup::SendMessageToSeedNodes(
@@ -357,13 +362,6 @@ vector<unsigned char> Lookup::ComposeGetDSBlockMessage(uint64_t lowBlockNum,
 // low and high denote the range of blocknumbers being requested(inclusive).
 // use 0 to denote the latest blocknumber since obviously no one will request
 // for the genesis block
-bool Lookup::GetDSBlockFromSeedNodes(uint64_t lowBlockNum,
-                                     uint64_t highBlockNum) {
-  LOG_MARKER();
-  SendMessageToSeedNodes(ComposeGetDSBlockMessage(lowBlockNum, highBlockNum));
-  return true;
-}
-
 bool Lookup::GetDSBlockFromLookupNodes(uint64_t lowBlockNum,
                                        uint64_t highBlockNum) {
   LOG_MARKER();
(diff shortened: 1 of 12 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -393,13 +391,6 @@ vector<unsigned char> Lookup::ComposeGetTxBlockMessage(uint64_t lowBlockNum,
 // low and high denote the range of blocknumbers being requested(inclusive).
```
ansnunez: I think you can remove the comments too

src/libLookup/Lookup.cpp
```diff
@@ -357,13 +362,6 @@ vector<unsigned char> Lookup::ComposeGetDSBlockMessage(uint64_t lowBlockNum,
...
```
ansnunez: I think you can remove the comments too

src/libMessage/Messenger.cpp
```diff
@@ -850,6 +850,25 @@ bool SerializeToArray(const T& protoMessage, vector<unsigned char>& dst,
                                        protoMessage.ByteSize());
 }
 
+template <class T>
+bool RepeatableToArray(const T& repeatable, vector<unsigned char>& dst,
+                       const unsigned int offset) {
+  int tempOffset = offset;
+  for (int i = 0; i < repeatable.size(); ++i) {
```
ansnunez: Will it work if we change to range-based for loop? `for (const auto& element : repeatable)`

src/libMessage/Messenger.cpp
```diff
@@ -850,6 +850,25 @@ bool SerializeToArray(const T& protoMessage, vector<unsigned char>& dst,
...
+    const int length_available = dst.size() - tempOffset;
+    const auto& element = repeatable.Get(i);
+    if (length_available < element.ByteSize()) {
```
ansnunez: Since SerializeToArray already does the size check and resize, you don't need to do it here anymore

src/libMessage/Messenger.cpp
```diff
@@ -850,6 +850,25 @@ bool SerializeToArray(const T& protoMessage, vector<unsigned char>& dst,
...
+      dst.resize(dst.size() + element.ByteSize() - length_available);
+    }
+
+    if (!SerializeToArray(element, dst, tempOffset)) {
+      return false;
```
ansnunez: Probably good to have an error message here

src/libMessage/Messenger.cpp
```diff
@@ -2227,6 +2248,27 @@ bool Messenger::SetNodeForwardTxnBlock(
     txnsGeneratedCount++;
   }
 
+  Signature signature;
+  if (result.transactions().size() > 0) {
+    // Serialize the transactions to temp buffer, later use it to calculate the
+    // signature.
+    std::vector<unsigned char> tmp;
+    int tempOffset = 0;
+    for (int i = 0; i < result.transactions().size(); ++i) {
```
ansnunez: Can probably use range-based for loop here. Please check the other parts of the code too for similar changes.

src/libMessage/Messenger.cpp
```diff
@@ -2227,6 +2248,27 @@ bool Messenger::SetNodeForwardTxnBlock(
...
+      tmp.resize(tmp.size() + result.transactions(i).ByteSize());
```
ansnunez: SerializeToArray already does a resize if necessary. Please check the other parts of the code too for similar changes.

ShengguangXiao:
Run 200 nodes test, run to 42 DS Blocks, 205 Tx Blocks, have 192 peers left.

src/libMessage/Messenger.cpp
```diff
@@ -2552,6 +2615,23 @@ bool Messenger::SetLookupSetSeedPeers(vector<unsigned char>& dst,
...
+
+  Signature signature;
+  if (result.candidateseeds().size() > 0) {
+    // Serialize the candidate seeds to temp buffer, later use it to calculate
+    // the signature.
+    std::vector<unsigned char> tmp;
+    RepeatableToArray(result.candidateseeds(), tmp, 0);
+    if (!Schnorr::GetInstance().Sign(tmp, lookupKey.first, lookupKey.second,
+                                     signature)) {
+      LOG_GENERAL(WARNING, "Failed to sign transactions.");
```
ansnunez: Error message is wrong. Please check the other parts of the code too for similar changes.

src/libMessage/Messenger.cpp
```diff
@@ -2710,22 +2844,40 @@ bool Messenger::GetLookupGetDSBlockFromSeed(const vector<unsigned char>& src,
...
 
+  SerializableToProtobufByteArray(lookupKey.second, *result.mutable_pubkey());
+
   for (const auto& dsblock : dsBlocks) {
     DSBlockToProtobuf(dsblock, *result.add_dsblocks());
   }
 
+  Signature signature;
+  if (result.dsblocks().size() > 0) {
+    // Serialize the transactions to temp buffer, later use it to calculate the
```
ansnunez: Comment is wrong. Please check the other parts of the code too for similar changes.

src/libMessage/Messenger.cpp
```diff
@@ -3233,6 +3493,42 @@ bool Messenger::GetLookupGetStartPoWFromSeed(const vector<unsigned char>& src,
   return true;
 }
 
+bool Messenger::SetLookupSetStartPoWFromSeed(
```
ansnunez: I think for empty messages we can still sign something, because it's still important to verify the lookup in these cases. We can either just define some dummy bytes to sign, or have something more meaningful like maybe the epoch number.

src/libLookup/Lookup.cpp
```diff
@@ -1939,22 +2046,38 @@ bool Lookup::ProcessSetLookupOnline(const vector<unsigned char>& message,
...
-  if (!Messenger::GetLookupSetLookupOnline(message, offset, portNo)) {
+  PubKey lookupPubKey;
+  if (!Messenger::GetLookupSetLookupOnline(message, offset, portNo,
+                                           lookupPubKey)) {
     LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
               "Messenger::GetLookupSetLookupOnline failed.");
     return false;
   }
 
+  auto iter =
```
ansnunez: Hmm this code block is repeated quite a few times. Maybe we can define it as a function somewhere.

src/libLookup/Lookup.cpp
ShengguangXiao: I think this comments is quite helpful, from it i know the the meaning of LowBlockNumber and HighBlockNumber.

src/libMessage/Messenger.cpp
ShengguangXiao: Ok, i will update.

src/libMessage/Messenger.cpp
ShengguangXiao: Good point, i will update it.

src/libNode/Node.cpp
```diff
@@ -706,15 +706,29 @@ bool Node::ProcessTxnPacketFromLookup(
...
   if (!Messenger::GetNodeForwardTxnBlock(message, offset, epochNumber, shardId,
-                                         transactions)) {
-    LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
+                                         lookupPubKey, transactions)) {
+    LOG_EPOCH(WARNING, std::to_string(m_mediator.m_currentEpochNum).c_str(),
               "Messenger::GetNodeForwardTxnBlock failed.");
     return false;
   }
 
+  auto lookupNodes = m_mediator.m_lookup->GetLookupNodes();
```
Gnnng: `lookupNodes` is a reference so the operation is now on a shared resource if the function `ProcessTxnPacketFromLookup` is invoked from multiple threads. Can you check if any race condition could happen here?

ShengguangXiao: Ok, i will put it to a const reference, there will be only query from it.

src/libLookup/Lookup.cpp
ShengguangXiao: Good point, i will do it.

src/libMessage/Messenger.cpp
ShengguangXiao: I think can do.

src/libLookup/Lookup.cpp
ShengguangXiao: I put the check into a function VerifyLookupNode, it will be more readable and easy to maintain.

src/libMessage/Messenger.cpp
```diff
@@ -1010,6 +1010,12 @@ bool RepeatableToArray(const T& repeatable, vector<unsigned char>& dst,
   return true;
 }
 
+template <class T>
+void NumberToArray(const T& number, vector<unsigned char>& dst,
+                   const unsigned int offset) {
+  Serializable::SetNumber<T>(dst, offset, number, sizeof(T));
```
ansnunez: Oops, `sizeof(T)` won't work for `uint128_t` and `uint256_t`.  You will have to copy the way `NumberToProtobufByteArray` is written.

ShengguangXiao: Ok

src/libLookup/Lookup.cpp
```diff
@@ -2706,3 +2831,13 @@ bool Lookup::GetIsServer() {
 
   return m_isServer;
 }
+
+bool Lookup::VerifyLookupNode(const VectorOfLookupNode& vecLookupNodes,
```
Gnnng: still, putting a const reference seems not the solution to the thread-safety issue. i remember that `m_lookupNode` will have write operations somewhere. Can you take a look again?

PR #660, opened 2018-10-08
Stop Timing for Fallback if PoW Failed & Other fixes & Add checkers & CheckStateRoot in every epoch

kaikawaliu:
## Description

- Put FallbackStop in ClearVariables()
- If a node didn't find itself in the sharding structure, also rejoin.
- Simplified the logic when lookup receives forward txn from shard
- Add Transaction Receipt Hash checker in ProcessForwardTransactionCore
- Removed obsolete code marked as FIXME for creating dummy account with huge balance
- Add real state root hash in every finalblock and add CheckStateRoot in every epoch
- If a node time-out when doing the PoW at default difficulty no successful answer obtained, and receiving any ds block, then it will fetch is there any new block from lookup if so then rejoin.
- Removed POW_BACKUP_TIME_WINDOW since we already implemented logic for handling early receiving of DS Block announcement in DS backup
- Did update on consensusId in libNode when VC done

### Implementation

- [x] **ready for review**

- [x] **local machine test**

- [x] **middle-scale cloud test** (200 aws nodes cluster)


src/libData/AccountData/AccountStoreBase.tpp
```diff
@@ -127,14 +127,15 @@ bool AccountStoreBase<MAP>::UpdateAccounts(const Transaction& transaction,
...
+    LOG_GENERAL(WARNING, "sender " << fromAddr.hex() << " not exist");
     // FIXME: remove this, temporary way to test transactions, should return
     // false
-    LOG_GENERAL(WARNING,
-                "AddAccount... FIXME: remove this, temporary way to "
-                "test transactions, should return false in the future");
-    this->AddAccount(fromAddr, {10000000000, 0});
-    fromAccount = this->GetAccount(fromAddr);
-    // return false;
+    // LOG_GENERAL(WARNING,
```
ansnunez: Can we just remove the commented out code entirely?

kaikawaliu: sure

bb111189:
@KaustubhShamshery good finding. 

Indeed, rejoining shouldn't be trigger if the node code execute enter 
`if (winning_result.success)`. This means it has find a valid solution (but not yet accepted by the network). It is in a state of waiting for DS block.

As briefly discussed with @kaikawaliu, We can put a else condition (not perfect solution* on a second thoughts) to trigger the rejoining, as it implies the node failed to find a valid solution. However, there is a caveat. A typical condition (but there may be others) for nodes not able to find the solution is when the node triggered `StopMining()`. This is triggered when the node received DS block under `ProcessDSBlock`.

The chicken and egg problem comes when the node failed to find a solution -> did not receive DS/Sharding block (as not part of network) -> did not stop mining ->  did not enter the else condition -> unable to rejoin. 

\* Not perfect solution. Cornercase: what if the node found the solution, enter the `if (winning_result.success)` but did not get accepted by the network due to various reasons like too many shards, network issues, too late, etc. It will not receive the DS block and will not enter the `else condition`.

In my opinion, a better solution will be a node should be `self-aware` of the PoW window (heart beat like) and stop mining when necessary. Re-sync (Get latest blocks) and try PoW at the next window.  It shouldn't rely on DS committee (or DS Block) as an indicator. @jiayaoqijia  What do you think?

src/libDB/Archival.cpp
```diff
@@ -31,24 +31,14 @@ unsigned int REFRESH_DELAY = 5;
 
 void Archival::InitSync() {
   auto func = [this]() -> void {
-    m_synchronizer.FetchOfflineLookups(m_mediator.m_lookup);
+    if (!m_mediator.m_node->GetOfflineLookups()) {
```
KaustubhShamshery: I dont think so we should "return" here as it will return from the whole sync thread then

kaikawaliu: The success of this data fetching is very elementary for rejoining. We can adjust the times of retrying, but if it always fail, then I think it's better to terminate the rejoin first.

KaustubhShamshery: but archival does not "rejoin", it is in pull mode

kaikawaliu: I shall change the language, for "synchronizing"

PR #664, opened 2018-10-08
Benchmarking of gossip

sandipbhoir:
## Description
This PR covers following:

- Logging of each rumor with required details that could help benchmarking.
- Some additional changes in libRumorSpreading to capture exact rounds spent in each state.
- Small change to avoid additional round when transiting from B-State to C-State.
- Small change to set the rumor to C-State directly if rumor received was from C-State.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


sandipbhoir:
Constant parameters used for testing with 1300 nodes ( DS Nodes - 600, Normal Shard nodes - 700 in 1 shard) with gossip and treebased cluster broadcasting.
[constants.txt](https://github.com/Zilliqa/Zilliqa/files/2468466/constants.txt)

PR #666, opened 2018-10-09
Minor fix to CreateTransaction

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


iantanwx:
Thanks, @KaustubhShamshery.

On a separate (?) note, I think it would be good for us to standardise on error messages, like this:

https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h

PR #667, opened 2018-10-09
Node recovery

ckyang:
## Description

After upgrading protocol and heartbeat rejoining, we need to implement node recovery in advance, for balancing network loading across nodes.
That is, any node (DS / shard / lookup) should be recovered well after it is destroyed anytime by some reason.
(Detail documentation: https://docs.google.com/document/d/1fx-zruwTuuF20VjutbuV1qqUub4K6qekrCxUKKxppbU/edit#heading=h.i34k42ood3s4)

- [x] Use daemon instead of docker monitor (https://github.com/Zilliqa/Zilliqa/pull/264)
- [x] For every node, store final-block state-delta for lastest 2 ds epochs
- [x] For every node, store shard structure for latest DS epoch
- [x] Upgrade::Replace() should send some special signal to kill process, to separate from process killed abnormally
- [x] Retrieve data (DS blocks, Tx blocks, DS committee, shard structure, final-block state-delta) from persistence
- [x] If re-launched process is in vacuous epoch, or still within the first DS epoch, do rejoin instead of recovery
- [x] When retrieving Ds blocks, do not delete the incomplete ds block
- [x] When retrieving Tx blocks, ask the relative final-block state-delta from persistence immediately
- [x] For the lacked Tx blocks and relative final-block state-delta, ask from lookup node
- [x] Store the final-block state-delta back to account store (DirectoryService::ProcessStateDelta / Node::ProcessStateDeltaFromFinalBlock)
- [x] When doing retrieving jobs, temporarily disable incoming message, or do nothing for every incoming message
- [x] After retrieving necessary data, wait for final-block announcement (DirectoryService::RunConsensusOnFinalBlock) for ds node, or process final-block (bottom part of Node::ProcessFinalBlock) for shard node

### Implementation

- [x] **ready for review**

- [x] large-scale cloud test (200 nodes)


src/libPersistence/Retriever.cpp
```diff
@@ -75,23 +75,32 @@ void Retriever::RetrieveDSBlocks(bool& result) {
...
-    blocks.pop_back();
-    hasIncompletedDS = true;
   }
 
+  m_mediator.m_blocklinkchain.Reset();
+  uint64_t index = 0;
+
   for (const auto& block : blocks) {
     m_mediator.m_dsBlockChain.AddBlock(*block);
+    m_mediator.m_blocklinkchain.AddBlockLink(
```
KaustubhShamshery: this code would be wrong if there is an VC/FB block generated.

ckyang: That's keep the logic here, and once your new AddBlockLink function finished, I will help to test the node recovery functionality.

src/libLookup/Lookup.cpp
```diff
@@ -402,6 +419,14 @@ bool Lookup::GetTxBlockFromLookupNodes(uint64_t lowBlockNum,
   return true;
 }
 
+bool Lookup::GetStateDeltaFromLookupNodes(uint64_t blockNum) {
```
KaustubhShamshery: can use const&

src/libNode/Node.cpp
```diff
@@ -76,133 +76,92 @@ Node::Node(Mediator& mediator, [[gnu::unused]] unsigned int syncType,
...
-      }
+    if (!StartRetrieveHistory(wakeupForUpgrade)) {
+      return false;
+    }
 
-      BlockStorage::GetBlockStorage().GetDSCommittee(
-          m_mediator.m_DSCommittee, m_mediator.m_ds->m_consensusLeaderID);
-      m_mediator.UpdateDSBlockRand();
-      m_mediator.UpdateTxBlockRand();
+    m_mediator.m_currentEpochNum =
```
KaustubhShamshery: dont need typecast

src/libNode/Node.cpp
```diff
@@ -271,43 +231,232 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
...
 
-  tDS.join();
-  tTx.join();
+  if (!tx_result) {
+    return false;
+  }
+
+  /// Retrieve lacked Tx blocks from lookup nodes
+  uint64_t oldTxNum = m_mediator.m_txBlockChain.GetBlockCount();
+  m_mediator.m_lookup->GetTxBlockFromLookupNodes(
```
KaustubhShamshery: Can use synchronizer function instead?

src/libNode/Node.cpp
```diff
@@ -271,43 +228,254 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
...
+  LOG_GENERAL(INFO, "Retrieve final block from lookup node, please wait...");
 
-  tDS.join();
-  tTx.join();
+  unique_lock<mutex> lock(m_mediator.m_lookup->m_MutexCVSetTxBlockFromSeed);
+  if (m_mediator.m_lookup->cv_setTxBlockFromSeed.wait_for(
+          lock, chrono::seconds(RECOVERY_SYNC_TIMEOUT)) == cv_status::timeout) {
+    LOG_GENERAL(WARNING,
+                "GetTxBlockFromLookupNodes Timeout, cannot get final block "
+                "from lookup node!");
```
kaikawaliu: If failed should fetch again. also better to refer to the logic of `StartSynchronization` to get Offline Lookups first

src/libNode/Node.cpp
```diff
@@ -271,43 +228,254 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
...
+  }
+
+  if (m_mediator.m_txBlockChain.GetBlockCount() > oldTxNum + 1) {
```
kaikawaliu: Is that mean recovery can only work for delay of one epoch only?

src/libNode/Node.cpp
```diff
@@ -271,43 +228,254 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
...
+        "Retrieve final block state delta from lookup node, please wait...");
+    unique_lock<mutex> lock(
+        m_mediator.m_lookup->m_MutexCVSetStateDeltaFromSeed);
+    if (m_mediator.m_lookup->cv_setStateDeltaFromSeed.wait_for(
+            lock, chrono::seconds(RECOVERY_SYNC_TIMEOUT)) ==
+        cv_status::timeout) {
+      LOG_GENERAL(WARNING,
+                  "GetStateDeltaFromLookupNodes Timeout, cannot get final "
+                  "block state delta from lookup node!");
+    }
```
kaikawaliu: same timeout issue as above

src/libNode/Node.cpp
```diff
@@ -271,43 +228,254 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
...
+  }
+
+  /// Save coin base for final block, from last DS epoch to current TX epoch
+  for (uint64_t blockNum =
+           m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() -
+           (m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() %
+            NUM_FINAL_BLOCK_PER_POW);
```
kaikawaliu: If a fallback happened, which will result a ds block being generated at any epoch, this starting `blockNum` may not fit. Alternatively, You can use the epoch number obtained from DSBlockHeader after #788 was merged

ckyang: ok, will do that then, in a separated PR

src/libNode/Node.cpp
```diff
@@ -271,43 +228,257 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
...
+      (m_mediator.m_txBlockChain.GetBlockCount()) % NUM_FINAL_BLOCK_PER_POW;
+
+  /// Save coin base for micro block, from last DS epoch to current TX epoch
+  std::list<MicroBlockSharedPtr> microBlocks;
+  if (BlockStorage::GetBlockStorage().GetRangeMicroBlocks(
+          m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() -
+              (m_mediator.m_txBlockChain.GetLastBlock()
+                   .GetHeader()
+                   .GetBlockNum() %
+               NUM_FINAL_BLOCK_PER_POW),
```
kaikawaliu: Same change will be applied here. Btw in 788, MicroBlock will have GetEpochNum rather than GetBlockNum, which is to keep the same convention with ds/vc/fallback block.

ckyang: ok

PR #668, opened 2018-10-09
Changed data block types serialization to completely use Messenger

ansnunez:
## Description

Before this PR, we already had protobuf-equivalent types for each data block type (`DSBlock`, `MicroBlock`, `TxBlock`, `VCBlock`, and `FallbackBlock`).

However, we only used these protobuf-equivalent types when sending out messages (e.g., the DSBlock inside DSBlock Announcement). There are other cases (e.g., storing to persistent storage, doing header cosig verification, ...) that continued to use the `Serialize` and `Deserialize` methods inside the data blocks.

In this PR, I am keeping those methods but changing them to also use our `Messenger` to do protobuf-type serialization. A lot of other methods like `GetSerializedSize` and variables like `SIZE` are now not needed anymore as a result.

Temporarily, this PR has introduced `SerializableDataBlock` because I changed the return type of `Serialize` and `Deserialize` to `bool`. When the other primitives (e.g., `Peer`, `PubKey`, ...) are changed in a similar way, we can get rid of the old `Serializable`.

Local testing has been done to check the data block types are working correctly, but I have yet to test `VCBlock` and `FallbackBlock`. I propose a code review at this stage and we will be able to revisit later the testing for these two block types, since it is not easy to trigger the conditions for them.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/common/Serializable.h
+++ b/src/common/Serializable.h
@@ -77,4 +77,59 @@ class Serializable {
   }
 };
 
+// This is a temporary class for use with data blocks
+class Serializable2 {
+ public:
+  /// Serializes internal state to destination byte stream.
+  virtual bool Serialize(std::vector<unsigned char>& dst,
+                         unsigned int offset) const = 0;
+
+  /// Deserializes source byte stream into internal state.
+  virtual bool Deserialize(const std::vector<unsigned char>& src,
+                           unsigned int offset) = 0;
+
+  /// Virtual destructor.
+  virtual ~Serializable2() {}
+
+  /// Template function for extracting a number from the source byte stream at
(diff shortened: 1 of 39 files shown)


src/common/Serializable.h
```diff
@@ -77,4 +77,59 @@ class Serializable {
   }
 };
 
+// This is a temporary class for use with data blocks
+class Serializable2 {
```
ShengguangXiao: Can change the Serializable2 to a more meaningful name?

ansnunez: Changed to `SerializableDataBlock`

PR #669, opened 2018-10-10
Benchmarking of TreeBased Cluster Broadcasting

sandipbhoir:
## Description
This PR cover adding additional logging for TreeBased cluster broadcasting to help benchmarking.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -231,8 +231,11 @@ void DirectoryService::SendDSBlockToShardNodes(
       std::vector<Peer> shardDSBlockReceivers;
 
       LOG_GENERAL(
-          INFO, "Sending to {NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD} peers : "
-                    << NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD);
+          INFO,
+          "Sending message with hash: ["
+              << DataConversion::Uint8VecToHexStr(this_msg_hash).substr(0, 6)
+              << "] to NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD:"
+              << NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD << " shard peers");
 
       unsigned int numOfDSBlockReceivers = std::min(
           NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD, (uint32_t)p->size());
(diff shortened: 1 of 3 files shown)


src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -231,8 +231,11 @@ void DirectoryService::SendDSBlockToShardNodes(
       std::vector<Peer> shardDSBlockReceivers;
 
       LOG_GENERAL(
-          INFO, "Sending to {NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD} peers : "
-                    << NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD);
+          INFO,
+          "Sending message with hash: ["
+              << DataConversion::Uint8VecToHexStr(this_msg_hash).substr(0, 6)
+              << "] to NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD:"
+              << NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD << " shard peers");
```
ShengguangXiao: I suggest this message to move down to log numOfDSBlockReceivers, because the value of  NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD we can know from constants.xml.

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -590,10 +590,16 @@ void DirectoryService::SendBlockToShardNodes(
...
+        SHA2<HASH_TYPE::HASH_VARIANT_256> sha256;
+        sha256.Update(block_message);
+        vector<unsigned char> this_msg_hash = sha256.Finalize();
+
+        LOG_GENERAL(
+            INFO,
+            "Sending message with hash: ["
+                << DataConversion::Uint8VecToHexStr(this_msg_hash).substr(0, 6)
+                << "] to NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD:"
+                << NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD << " shard peers");
```
ShengguangXiao: Suggest to log value of numOfBlockReceivers

src/libDirectoryService/DSBlockPostProcessing.cpp
sandipbhoir: This was added to log after suggestion from previous review to clearly show which constants we are using on DS side (sender) and Shard side (receiver) so that we know we are not using different constants.

PR #673, opened 2018-10-10
Ignore failed contract related transaction

kaikawaliu:
## Description

Previously we still append the transaction which failed CheckCreatedTransaction. But now as we already have the mechanism to cost the sender gas fee once the interpreter was invoked, this part of logic can be removed. 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


iantanwx:
@kaikawaliu having tested this PR on a testnet with 40 nodes and looking at the implementation, it seems that the transaction is ignored (i.e. completely not included in the final `TxBlock`). This means the client will be left polling the lookup node for some time, without receiving an answer as to whether the transaction failed or succeeded. Can we still append the transaction even in the case of an error, but with `success: false`? Otherwise the client will keep polling the lookup for no reason.

iantanwx:
@kaikawaliu, @KaustubhShamshery will do the check at the libServer layer so it should not be necessary for you to do anything.

PR #674, opened 2018-10-10
DS to send vcds message instead of vc block (before ds block consensus) to shard nodes

bb111189:
## Description

This PR delays the sending of VC blocks (for ds block consensus)  to shard nodes. The sending of the vcblocks have been postponed and combined with the sending of DS Blocks. The benefit of such delays is that the DS Committee will have the sharding structure before transmission of the blocks. 

This will allows for the implementation for efficient broadcasting (tree based) of vc block(s). 

Known issue that hinder testing of this PR
- Lookup consuming too much cpu utilization on startup (https://github.com/Zilliqa/Issues/issues/224)
- https://github.com/Zilliqa/Issues/issues/225

With this issue, it is hard to observe lookup performing changing its view. However, it used the same function as the same as shard nodes. Changing of view is tested to be successful on DS Committee and Shard nodes. 

TODO (refactors):
- [x] Rename SetNodeDSBock to SetNodeVCDSBlocksMessage
- [x] Rename GetNodeDSBock to GetNodeVCDSBlocksMessage
- [x] Rename ProcessDSblock to ProcessVCDSBlocksMessage

**Message visualization for - DS -> node : 
Before:** 
- DS -> node DS Block
`[DSBlock][Sharding structure][misc info]
`
- All VC 
`[VC Block]
`

**After:**
- (Renamed) VCDSMBlocksMessage to node 
`[DSBlock][Vector of VCBlocks][Sharding structure][misc info]
`
- VC after DS Block: 
`[VC Block]
`
## Review Suggestion

Note: A revisit and follow PR for view change is required. Follow up is for various scenario of view change. 

### Implementation

- [x] **ready for review** 
- [x] Retest rebase

**Local test**
- [x] No view change standard test (commit: e750a27cdc1ba64c65507b1055b1a248ec9ddd78)
- [x] 1x view change before ds block (commit: f152a042bd199511663d9322b99a4ada24567899)
- [ ] 3x view change before ds block (commit: f152a042bd199511663d9322b99a4ada24567899)

**Medium scale test**
- [x] medium-scale cloud test (No view change) (commit: f152a042bd199511663d9322b99a4ada24567899) - 15 DS Epoch

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -628,6 +628,7 @@ void DirectoryService::ProcessDSBlockConsensusWhenDone(
       << m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() + 1
       << "] AFTER SENDING DSBLOCK");
 
+  ClearVCBlockVector();
   UpdateDSCommiteeComposition();
   UpdateMyDSModeAndConsensusId();
 
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -92,8 +92,9 @@ void DirectoryService::SendDSBlockToLookupNodes() {
   vector<unsigned char> dsblock_message = {MessageType::NODE,
                                            NodeInstructionType::DSBLOCK};
   if (!Messenger::SetNodeDSBlock(dsblock_message, MessageOffset::BODY, 0,
-                                 *m_pendingDSBlock, m_shards, m_DSReceivers,
-                                 m_shardReceivers, m_shardSenders)) {
+                                 *m_pendingDSBlock, m_VCBlockVector, m_shards,
+                                 m_DSReceivers, m_shardReceivers,
+                                 m_shardSenders)) {
     LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
               "Messenger::SetNodeDSBlock failed.");
     return;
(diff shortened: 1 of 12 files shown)


src/libDirectoryService/ViewChangePostProcessing.cpp
```diff
@@ -229,16 +226,22 @@ void DirectoryService::ProcessViewChangeConsensusWhenDone() {
...
-    m_mediator.m_lookup->SendMessageToLookupNodes(vcblock_message);
-    LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-              "I the part of the subset of DS committee that have sent the "
-              "VCBlock to the lookup nodes");
+  // Broadcasting vcblock to lookup nodes iff view change do not occur before ds
+  // block consensus. This is to be consistent with how normal node process the
+  // vc block (before ds block).
+  if (viewChangeState != DSBLOCK_CONSENSUS &&
+      viewChangeState != DSBLOCK_CONSENSUS_PREP) {
+    unsigned int nodeToSendToLookUpLo = m_mediator.GetShardSize(true) / 4;
```
sandipbhoir: This logic fails in general as we observed earlier if ds committe size is smaller than shard_size/4. (Ex. Ds committee with 100 nodes and Shard-Size of 500). I think this should be fine for this PR, but we need to fix this logic in general wherever applicable.

sandipbhoir: Created an issue https://github.com/Zilliqa/Issues/issues/226 to track this in future.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -91,11 +91,12 @@ void DirectoryService::SendDSBlockToLookupNodes() {
 
   vector<unsigned char> dsblock_message = {MessageType::NODE,
                                            NodeInstructionType::DSBLOCK};
-  if (!Messenger::SetNodeDSBlock(dsblock_message, MessageOffset::BODY, 0,
-                                 *m_pendingDSBlock, m_shards, m_DSReceivers,
-                                 m_shardReceivers, m_shardSenders)) {
+  if (!Messenger::SetNodeVCDSBlocksMessage(
+          dsblock_message, MessageOffset::BODY, 0, *m_pendingDSBlock,
+          m_VCBlockVector, m_shards, m_DSReceivers, m_shardReceivers,
```
ansnunez: Do we need to use the mutex here and elsewhere below for `m_VCBlockVector`? Just checking.

PR #675, opened 2018-10-10
Use current account nonce to compute contract address

iantanwx:
## Description

Make `Server.cpp` and `AccountStoreSC.tpp` return a consistent address when a contract is being deployed.

`Server.cpp` and `AccountStoreSC.tpp` return different contract addresses as the former uses the transaction nonce, while the latter uses the account nonce _prior_ to update. As a result, clients end up calling non-existent contracts if they rely on the response from the RPC server.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libServer/Server.cpp
```diff
@@ -102,11 +102,11 @@ Json::Value Server::CreateTransaction(const Json::Value& _json) {
 
     const PubKey& senderPubKey = tx.GetSenderPubKey();
     const Address fromAddr = Account::GetAddressFromPublicKey(senderPubKey);
+    const Account* sender = AccountStore::GetInstance().GetAccount(fromAddr);
```
kaikawaliu: Should check whether `sender == nullptr` here

KaustubhShamshery: good point.

iantanwx:
@KaustubhShamshery @jiayaoqijia this passes my integration tests and can be merged. Thanks all!

PR #683, opened 2018-10-12
Random shard leader selection

kaikawaliu:
## Description

- Applied random selection of the leader based on last txBlock Hash for the next consensus round.
- Fixed bug for using wrong function when received missing txn and about to reprocess the message.
- Unified the usage of BlockHash with the one obtained get by using GetBlockHash().
- Reduce the new ds members elected constaints in local test to 2

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/DataConversion.cpp
+++ b/src/libUtils/DataConversion.cpp
@@ -94,3 +94,10 @@ const std::string DataConversion::CharArrayToString(
 
   return ret;
 }
+
+uint16_t DataConversion::charArrTo16Bits(
+    const std::vector<unsigned char>& hex_arr) {
+  uint32_t lsb = hex_arr.size() - 1;
+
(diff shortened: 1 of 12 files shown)


src/libUtils/DataConversion.cpp
```diff
@@ -94,3 +94,10 @@ const std::string DataConversion::CharArrayToString(
 
   return ret;
 }
+
+uint16_t DataConversion::charArrTo16Bits(
+    const std::vector<unsigned char>& hex_arr) {
+  uint32_t lsb = hex_arr.size() - 1;
```
bb111189: can check whether is hex_arr empty or not. Else they may be an underflow

kaikawaliu: Fixed

bb111189:
Pending 200 nodes test

PR #685, opened 2018-10-12
Fix new node cannot join problem

ShengguangXiao:
## Description

The new node cannot join because verify start pow message verification bug and variable not initialize bug.

Fix the bug.

## Review Suggestion

Check file changes, only a few lines changed.

Run a local test, and run ./tests/Node/test_node_late.sh to rejoin 5 nodes.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ShengguangXiao:
Tested on local machine, 5 new node can successfully join, 25 node together can run to 55 Tx blocks.

PR #695, opened 2018-10-16
Enable verification of PoW ordering to check the sharding structure from DS leader

ShengguangXiao:
## Description

Enable verification of PoW ordering to check the sharding structure from DS leader

Optimize the code, fix verify PoW ordering fail problem.

## Review Suggestion

Check files modified

Run a local test

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -44,9 +44,7 @@ using namespace std;
 using namespace boost::multiprecision;
 
 unsigned int DirectoryService::ComputeDSBlockParameters(
-    const vector<pair<array<unsigned char, 32>, PubKey>>& sortedDSPoWSolns,
-    std::vector<std::pair<std::array<unsigned char, 32>, PubKey>>&
-        sortedPoWSolns,
+    const VectorOfPoWSoln& sortedDSPoWSolns, VectorOfPoWSoln& sortedPoWSolns,
     map<PubKey, Peer>& powDSWinners, uint8_t& dsDifficulty, uint8_t& difficulty,
     uint64_t& blockNum, BlockHash& prevHash) {
   LOG_MARKER();
@@ -120,8 +118,7 @@ unsigned int DirectoryService::ComputeDSBlockParameters(
   return numOfElectedDSMembers;
 }
 
-void DirectoryService::ComputeSharding(
-    const vector<pair<array<unsigned char, 32>, PubKey>>& sortedPoWSolns) {
+void DirectoryService::ComputeSharding(const VectorOfPoWSoln& sortedPoWSolns) {
   if (LOOKUP_NODE_MODE) {
     LOG_GENERAL(WARNING,
                 "DirectoryService::ComputeSharding not expected to be "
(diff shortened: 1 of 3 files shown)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -219,28 +299,53 @@ bool DirectoryService::VerifyPoWOrdering(const DequeOfShard& shards) {
...
         m_mediator.m_txBlockChain.GetLastBlock().GetBlockHash().asBytes();
   }
-  // Temporarily add the old ds to check ordering
-  m_allPoWs[m_mediator.m_DSCommittee->back().first] =
-      array<unsigned char, BLOCK_HASH_SIZE>();
 
-  vector<unsigned char> hashVec;
+  constexpr float MISORDER_TOLERENCE = 0.02f;  // 2 Percent tolerance
+  const uint32_t MAX_MISORDER_NODE =
+      std::ceil(m_allPoWs.size() * MISORDER_TOLERENCE);
```
KaustubhShamshery: should u take some constant like the ds_committee size here instead of all pows size

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -219,28 +299,53 @@ bool DirectoryService::VerifyPoWOrdering(const DequeOfShard& shards) {
...
+                     m_pendingDSBlock->GetHeader().GetDSPoWWinners().size());
+  if (DEBUG_LEVEL >= 5) {
+    for (const auto& pairPoWKey : sortedPoWSolns) {
+      LOG_GENERAL(INFO, "[POWS]"
+                            << DataConversion::charArrToHexStr(pairPoWKey.first)
+                            << " " << pairPoWKey.second);
+    }
+  }
+
+  vector<unsigned char> hashVec(BLOCK_HASH_SIZE + BLOCK_HASH_SIZE);
```
KaustubhShamshery: why twice the size ?

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -405,6 +515,20 @@ void DirectoryService::ComputeTxnSharingAssignments(
   }
 }
 
+VectorOfPoWSoln DirectoryService::SortPoWSoln(const MapOfPubKeyPoW& mapOfPoWs) {
+  std::map<array<unsigned char, 32>, PubKey> PoWOrderSorter;
+  for (const auto& powsoln : mapOfPoWs) {
```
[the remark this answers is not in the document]
ShengguangXiao: It is the same as Jun Hao's original method, it is not only for sort, also to switch the pair of <PubKey, PoW> to <PoW, PubKey>.

src/libDirectoryService/DSBlockPreProcessing.cpp
ShengguangXiao: No, this is to verify the PoW submissions, not the ds comittee.

KaustubhShamshery: yes but the tolerance level depends upon the PoW submissions recvd, should not it be a factor of the ds committee size?

ShengguangXiao: The ds committee size has no relationship with number of PoW submissions.

KaustubhShamshery: I edited my comment, please see again

ShengguangXiao: From my test result,  the number of PoW submissions for DS leader and DS backup is the same, they are the same as total nodes in sharding structure. And only some received DS PoW, some didn't. And even the number of PoW is a little different, after multiply the factor of 0.02, the tolerance will be the same, unless the number of PoW difference over 50.

Issue #691, opened 2018-10-15
Test GPU mining on a rig with 6 cards and result is every node use all cards instead of a card


dtompichayut:
From current update local test to have number of gpu in constant.xml file.
The result is all nodes use all gpu cards to run.

It should be a node only run on a gpu and could set what gpu assign to what node.

such as (in simple node test with 20 nodes)
node 0001-0003 use gpu 0
node 0004-0006 use gpu 1
node 0007-0010 use gpu 2
node 0011-0013 use gpu 3
node 0014-0016 use gpu 4
node 0017-0020 use gpu 5

It should change the number of gpus to run  in constant.xml on each node to the gpu ID to assign, or add parameter to zilliqa command to assign gpu to run on that node.

dtompichayut:
code changed, now I can select gpu to run

PR #704, opened 2018-10-17
Support user select use which GPU for mine

ShengguangXiao:
## Description

Let user select which GPU use for mine.

Add an GPU_TO_USE variable in constants.xml, let user choose which GPU used for mine.

https://github.com/Zilliqa/Zilliqa/issues/691

## Review Suggestion

Check file changed.

Find a machine with 2 GPUs, then change the constants.xml GPU_TO_USE option, and run ./build/tests/POW/Test_POW

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -136,7 +136,7 @@
     -->
     </lookups>
     <gpu>
-        <!-- Which GPU to use, can use multile GPU, for example: 0, 2, 4 -->
+        <!-- Which GPU to use, can use multiple GPU, for example: "0, 2, 4" -->
         <GPU_TO_USE>0</GPU_TO_USE>
         <opencl>
             <LOCAL_WORK_SIZE>128</LOCAL_WORK_SIZE>
--- a/constants.xml
+++ b/constants.xml
@@ -136,7 +136,8 @@
     -->
     </lookups>
     <gpu>
-        <NUM_DEVICE_TO_USE>1</NUM_DEVICE_TO_USE>
+        <!-- Which GPU to use, can use multile GPU, for example: 0, 2, 4 -->
+        <GPU_TO_USE>0</GPU_TO_USE>
         <opencl>
             <LOCAL_WORK_SIZE>128</LOCAL_WORK_SIZE>
             <GLOBAL_WORK_SIZE_MULTIPLIER>8192</GLOBAL_WORK_SIZE_MULTIPLIER>
--- a/src/common/Constants.cpp
+++ b/src/common/Constants.cpp
@@ -78,7 +78,7 @@ unsigned int ReadGpuConstants(std::string propertyName) {
   return pt.get<unsigned int>("node.gpu." + propertyName);
 }
 
-std::string ReadGpuVariable(std::string propertyName) {
+std::string ReadGPUVariableFromConstantsFile(std::string propertyName) {
   auto pt = PTree::GetInstance();
   return pt.get<std::string>("node.gpu." + propertyName);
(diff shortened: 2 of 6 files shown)


constants.xml
```diff
@@ -136,7 +136,8 @@
     -->
     </lookups>
     <gpu>
-        <NUM_DEVICE_TO_USE>1</NUM_DEVICE_TO_USE>
+        <!-- Which GPU to use, can use multile GPU, for example: 0, 2, 4 -->
```
ansnunez: Can you revise to `Which GPU to use, can use multiple GPU, for example: "0, 2, 4"`
(i.e., Fixed the misspelled "multiple" and added quotes in the example to make it clear it is the actual value you want to put in this field)

src/common/Constants.cpp
```diff
@@ -78,6 +78,11 @@ unsigned int ReadGpuConstants(std::string propertyName) {
   return pt.get<unsigned int>("node.gpu." + propertyName);
 }
 
+std::string ReadGpuVariable(std::string propertyName) {
```
ansnunez: Can you change the parameter to `const std::string&` and the function name to `ReadGPUVariableFromConstantsFile`?

constants.xml
ShengguangXiao: Ok

constants.xml
```diff
@@ -136,7 +136,8 @@
     -->
     </lookups>
     <gpu>
-        <NUM_DEVICE_TO_USE>1</NUM_DEVICE_TO_USE>
+        <!-- Which GPU to use, can use multiple GPU, for example: "0, 2, 4" -->
```
ckyang: Maybe update the new GPU usage in wiki

src/libPOW/pow.cpp
```diff
@@ -499,3 +510,14 @@ void POW::InitCUDA() {
...
 #endif
 }
+
+std::vector<unsigned int> POW::GetGpuToUse() {
+  std::vector<unsigned int> vecGpuIndex;
+  std::stringstream ss(GPU_TO_USE);
+  std::string item;
+  while (std::getline(ss, item, ',')) {
+    unsigned int index = strtol(item.c_str(), NULL, 10);
+    vecGpuIndex.push_back(index);
```
ckyang: Need to check duplicated value before inserting into container

constants.xml
ShengguangXiao: OK

src/libPOW/pow.cpp
```diff
@@ -474,21 +476,23 @@ void POW::InitCUDA() {
...
-
   if (!CUDAMiner::configureGPU(CUDA_BLOCK_SIZE, CUDA_GRID_SIZE, CUDA_STREAM_NUM,
                                CUDA_SCHEDULE_FLAG, 0, 0, false, false)) {
     LOG_GENERAL(FATAL, "Failed to configure CUDA GPU, please check hardware");
   }
 
   CUDAMiner::setNumInstances(UINT_MAX);
-  for (uint32_t i = 0; i < NUM_DEVICE_TO_USE; ++i) {
-    m_miners.push_back(std::make_unique<CUDAMiner>(i));
+  auto setGpuIndex = GetGpuToUse();
```
ckyang: The naming of variable should try to keep away of what the implementation it. Then, if someday we need to modify the implementation, we don't need to change the variable name again.

PR #705, opened 2018-10-17
Process the console prints when the execution of contract failed

kaikawaliu:
## Description

Process the console prints when the execution of contract failed.

e.g.
when we made a failed call, the interpreter will print out rather than generating a file like: 
```
{
  "gas_remaining": 0,
  "errors": [
    {
      "error_message": "Ran out of gas when parsing contract/init files.\n",
      "start_location": { "file": "", "line": 0, "column": 0 },
      "end_location": { "file": "", "line": 0, "column": 0 }
    }
  ]
}
```
Which includes the left gas_remaining and the reason why the contract is failed.
The gas_remaining will be used to determine the gas the user may left.
However we are not making the blockchain an error-reporting tool or a debugging tool. So the errors showing here will be ignored.

### Implementation

- [X] **ready for review**

- [x] local machine test (**with contract deployment and invocation**)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -486,15 +497,19 @@ bool AccountStoreSC<MAP>::ParseCreateContractJsonOutput(
   }
   gasRemained = atoi(_json["gas_remaining"].asString().c_str());
 
-  if (!_json.isMember("message") || !_json.isMember("states")) {
-    LOG_GENERAL(WARNING,
-                "The json output of this contract is corrupted or create "
-                "contract failed");
+  if (!_json.isMember("message") || !_json.isMember("states") ||
```
ansnunez: Should it be `&&`?

ansnunez: Another one below as well

kaikawaliu: && is for both of the condition doesn't meet. || here means if any of the condition doesn't meet.

PR #710, opened 2018-10-18
Fix DS backup verify difficulty fail problem

ShengguangXiao:
## Description

Sometimes the DS backup failed to verify DS difficulty because received different number of DS PoW, it will trigger view change which can be avoid

Change the compare algorithm. Previous the DS leader POW window time is longer than DS backup, so the difficulty calculated should be equal or bigger than DS backup. Now the window time is the same, so DS backup may calculate the difficulty bigger than DS leader.

https://github.com/Zilliqa/Issues/issues/245

## Review Suggestion

Check files changed.

Run a 200 nodes clout test, see if find difficulty mismatch problem.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -263,8 +263,9 @@ bool DirectoryService::VerifyDifficulty() {
   auto localDSDifficulty = CalculateNewDSDifficulty(
       m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetDSDifficulty());
   constexpr uint8_t DIFFICULTY_TOL = 1;
-  if (remoteDSDifficulty < localDSDifficulty ||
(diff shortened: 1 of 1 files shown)


ShengguangXiao:
Run 200 nodes test, run to 200 Tx blocks didn't find this problem again.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -263,8 +263,9 @@ bool DirectoryService::VerifyDifficulty() {
   auto localDSDifficulty = CalculateNewDSDifficulty(
       m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetDSDifficulty());
   constexpr uint8_t DIFFICULTY_TOL = 1;
```
kaikawaliu: Can this be put in constants?

PR #719, opened 2018-10-22
Fix Json Conversion to include gasPrice/Limit

KaustubhShamshery:
## Description
* Add gasPrice / gasLimit to JSON return by rpc server
* Change Amount to be a string from front-end

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


iantanwx:
@KaustubhShamshery can you change https://github.com/Zilliqa/Zilliqa/blob/8d196b7ad9fe5571a7f66019b4a66f8cd7cb8412/src/libServer/JSONConversion.cpp#L200? It should not be `isIntegral` anymore; I will send a string.

PR #742, opened 2018-10-24
Robust consensus using subsets

sandipbhoir:
## Description

Newly modified  consensus process as follows:

**Round 1 of consensus:**
    Leader sends announcement to all backups.
    Leader receives as many commits as it can before we reach timeout or all commits have been received 
    After COMMIT_WINDOW_IN_SECONDS or all commit received (whichever is earlier), leader generates NUM_CONSENSUS_SUBSETS subsets each with exactly numForConsensus backups assigned at random.
    Leader continues consensus (starting at challenge) within each subset. Challenge Message will have additional field - subsetID
    Backup will process the challenge as before and will send the response back t leader but with subsetID as received in challenge. 
    As soon as one subset reaches the end of consensus i.e. it receives enough responses, leader sends collectivesig to everyone and stops processing other subsets. 
**Round 2 of consensus:**
   After collectivesig is done, leader will receive as many commits from backup and process in "Round 1 of consensus" is repeated. So new subsets are created and same process follows until one of subset reach final consensus. After this other subsets consensus is stopped and leader sends out finalcollectivesig to backup.

COMMIT_WINDOW_IN_SECONDS and NUM_CONSENSUS_SUBSETS are configured through constants.xml

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (commit: 3000540 )


src/libConsensus/ConsensusBackup.cpp
```diff
@@ -47,6 +47,7 @@ bool ConsensusBackup::CheckState(Action action) {
                                       // get ignored by the node)
        {FINALCOMMIT_DONE, PROCESS_FINALCHALLENGE},
        {FINALCOMMIT_DONE, PROCESS_FINALCOLLECTIVESIG},
+       {FINALRESPONSE_DONE, PROCESS_FINALCHALLENGE},
```
ansnunez: Can you confirm that this action is really allowed in this state?

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -63,9 +63,207 @@ bool ConsensusLeader::CheckState(Action action) {
...
+      found = true;
+      break;
+    }
+  }
+
+  if (!found) {
+    LOG_GENERAL(WARNING, "SubsetID: " << subsetID << ", Action "
+                                      << GetActionString(action)
+                                      << " not allowed in state "
+                                      << GetStateString());
```
ansnunez: I think what you want to log is the `m_state` of the subset, but `GetStateString()` here will return the value of `ConsensusCommon::m_state`

ansnunez: Actually it might be useful to log both the leader's state as well as the state of the subset

src/libConsensus/ConsensusBackup.cpp
sandipbhoir: Yes, that’s because we need to process multiple challenges inspite of responding to first challenge received.. otherwise given backup will send response only to first subset challenge and ignore other subset’s challenge.. i had tested this scenario.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: Yes. I agree. Will log both overall and subset’s states

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -63,9 +63,207 @@ bool ConsensusLeader::CheckState(Action action) {
...
+  for (unsigned int index = 0; index < m_commitMap.size(); index++) {
+    if (m_commitMap.at(index) && index != m_myID) {
+      peersWhoCommitted.push_back(index);
+    }
+  }
+  // Generate NUM_CONSENSUS_SUBSETS lists (= subsets of peersWhoCommitted)
+  // If we have exactly the minimum num required for consensus, no point making
+  // more than 1 subset
+
+  const unsigned int numSubsets = (peersWhoCommitted.size() < m_numForConsensus)
```
ansnunez: If `peersWhoCommitted.size`  is less than the required minimum for consensus, should we still proceed with the protocol?

sandipbhoir: No. Good catch. Will add a check.

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -63,9 +63,207 @@ bool ConsensusLeader::CheckState(Action action) {
...
+      SetStateSubset(index, m_state);
+
+      // Add the leader to the responses
+      Response r(*m_commitSecret, subset.m_challenge, m_myPrivKey);
+      subset.m_responseData.emplace_back(r);
+      subset.m_responseDataMap.at(m_myID) = r;
+      subset.m_responseMap.at(m_myID) = true;
+      subset.m_responseCounter = 1;
+
+      // Multicast to all nodes who send validated commits
```
ansnunez: This comment now only applies to the `else` case below, so I suggest to move it there, and have a more accurate comment for the `if (BROADCAST_GOSSIP_MODE)` branch as well.

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -236,32 +398,37 @@ bool ConsensusLeader::ProcessMessageCommitFailure(
...
 
+  ConsensusSubset& subset = m_consensusSubsets.at(subsetID);
+
   // Aggregate commits
-  CommitPoint aggregated_commit = AggregateCommits(m_commitPoints);
+  CommitPoint aggregated_commit = AggregateCommits(subset.m_commitPoints);
   if (!aggregated_commit.Initialized()) {
-    LOG_GENERAL(WARNING, "AggregateCommits failed");
+    LOG_GENERAL(WARNING, "[Subset " << subsetID << "] AggregateCommits failed");
     m_state = ERROR;
```
ansnunez: I'm thinking maybe the errors here (and perhaps elsewhere) have to be re-evaluated.  Do we really want to set the overall `m_state` to `ERROR`, or set just `subset.m_state` to `ERROR`?

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -296,28 +463,58 @@ bool ConsensusLeader::ProcessMessageResponseCore(
   // Extract and check response message body
   // =======================================
 
+  // Format: [4-byte consensus id] [32-byte blockhash] [2-byte backup id]
```
ansnunez: We can remove these comments since they were for the pre-protobuf code

src/libConsensus/ConsensusLeader.cpp
ansnunez: Ah actually your check for `(m_commitCounter < m_numForConsensus)` in the timed functions below already check for it, so `(peersWhoCommitted.size() < m_numForConsensus)` should never happen.  I think you can change the `<` to `==`.

sandipbhoir: Now I remember, why I had used < instead of ==. If I had used ==,  I have to put this ‘peerwhocommitted.size() == m_numForConsus - 1’ becoz peerswhocommitted excludes commit by myself. And to avoid additional arithmetic op for - 1, I had stick to ‘peerswhocommitted.size() < m_numForConsensus’. So I think current way should be ok.

src/libConsensus/ConsensusLeader.h
```diff
@@ -80,16 +84,36 @@ class ConsensusLeader : public ConsensusCommon {
...
 
-  // Received responses
-  unsigned int m_responseCounter;
-  std::vector<Response> m_responseDataMap;
-  std::vector<Response> m_responseData;
+  // Tracking data for each consensus subset
+  // TODO: the vectors should be replaced by more space efficient DS
+  struct ConsensusSubset {
+    std::vector<bool> m_commitMap;
+    std::vector<CommitPoint> m_commitPointMap;  // Ordered list of commits of
```
ShengguangXiao: It looks the original m_commitPointMap and m_commitPoints still been used. Should we rename the variable names in ConsensusSubset so it can be distinguished from the original one?

sandipbhoir: Well, it’s still the commit maps but local to subset which itself distinguishes it from original.

PR #748, opened 2018-10-25
Adjust the order of check for DSBlock CommHash in ProcessVCDS

kaikawaliu:
## Description

The check for the dsblock commhash should be later than vc being processed, As the m_dscommittee will be updated after vc.

### Implementation

- [X] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


bb111189:
Verified this fix works

PR #753, opened 2018-10-26
Validation of Tx Blocks recvd from Lookup

KaustubhShamshery:
## Description
1. Check latest block co-sig
**Three condtions:**

- Invalid signature / The ds epoch is a prev one : Fetch Tx block again
- The ds comm is stale (of a previous epoch) : Save tx blocks and wait for dir blocks
- Valid : Commit to storage

2. Verify the other blocks using prev hash

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (commit : 1da9f45) 20 nodes + 5 new node join
- [x] small-scale cloud test (commit: 3c9b264)  20 nodes + 5 new node join


src/libValidator/Validator.cpp
```diff
@@ -350,3 +350,59 @@ bool Validator::CheckDirBlocks(
...
+  }
+
+  if (txBlocks.size() < 2) {
+    return TxBlockValidationMsg::VALID;
+  }
+
+  BlockHash prevBlockHash = latestTxBlock.GetHeader().GetPrevHash();
+  unsigned int sIndex = txBlocks.size() - 2;
+
+  for (unsigned int i = 0; i < txBlocks.size() - 1; i++) {
```
bb111189: use range based loop?

src/libLookup/Lookup.cpp
```diff
@@ -1597,73 +1597,97 @@ bool Lookup::ProcessSetTxBlockFromSeed(const vector<unsigned char>& message,
...
-            LOG_GENERAL(INFO, "MicroBlock of hash "
-                                  << txBlock.GetMicroBlockHashes()[i].hex()
-                                  << " empty");
-          }
-        }
+void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
+  LOG_GENERAL(INFO, "[TxBlockVerif]"
+                        << "Success");
+  for (const auto& txBlock : txBlocks) {
+    LOG_EPOCH(
```
bb111189: Can you extract this entire logging message to a new function for code readability

src/libValidator/Validator.cpp
KaustubhShamshery: need to run to size() - 1

KaustubhShamshery: OK

src/libLookup/Lookup.cpp
KaustubhShamshery: OK.

src/libValidator/Validator.cpp
KaustubhShamshery: on further investigation, as I use sIndex to fetch an address, this cannot be changed in to range-based.

PR #761, opened 2018-10-27
Feature/lib serverproto

robin-thomas:
## Description
* sh script to download and setup gRPC
* cmake module to find and build proto files using gRPC plugin
* wrote the server proto (grpc service)
* wrote the c++ server implementation


ansnunez:
@robin-thomas can you also update the Dependencies section in README.md?

Gnnng:
Thanks a lot for your work @robin-thomas. It seems that the way gRPC is being managed can be done similar to these examples. Is it possible to make some changes to gRPC downloading, installing and finding according to the existing way used in the following third-party dependencies?

- https://github.com/Zilliqa/Zilliqa/blob/master/cmake/InstallMongo.cmake 
- https://github.com/Zilliqa/Zilliqa/blob/master/cmake/InstallG3log.cmake

A high-level overview of this process would be
1. pin the version of gRPC git repo using `git submodule` under `src/depends/grpc`
2. add `cmake/InstallgRPC` to handle the downloading, building and installation. Also, ensure the installed version of `gRPCConfig.cmake` can be discovered by CMake.
3. change the top-level `CMakeLists.txt` to simply use `FindPackage()` for package discovery (see the example from gRPC: https://github.com/grpc/grpc/blob/ffac9d90b18cb076b1c952faa55ce4e049cbc9a6/examples/cpp/helloworld/CMakeLists.txt#L73)

robin-thomas:
> Thanks a lot for your work @robin-thomas. It seems that the way gRPC is being managed can be done similar to these examples. Is it possible to make some changes to gRPC downloading, installing and finding according to the existing way used in the following third-party dependencies?
> 
> * https://github.com/Zilliqa/Zilliqa/blob/master/cmake/InstallMongo.cmake
> * https://github.com/Zilliqa/Zilliqa/blob/master/cmake/InstallG3log.cmake
> 
> A high-level overview of this process would be
> 
> 1. pin the version of gRPC git repo using `git submodule` under `src/depends/grpc`
> 2. add `cmake/InstallgRPC` to handle the downloading, building and installation. Also, ensure the installed version of `gRPCConfig.cmake` can be discovered by CMake.
> 3. change the top-level `CMakeLists.txt` to simply use `FindPackage()` for package discovery (see the example from gRPC: https://github.com/grpc/grpc/blob/ffac9d90b18cb076b1c952faa55ce4e049cbc9a6/examples/cpp/helloworld/CMakeLists.txt#L73)

When I started, I tried to build with cmake. But it gave me a lot of issues with versions (since gRPC also comes with a protoc). So I just resorted to building from source for more granular control.

PR #762, opened 2018-10-27
[testnetv3-dev] Added tests for primitives in libMessage

ansnunez:
## Description

Minor modification of the test functions comparing to the previous pull request because of the separation of the test data generator functions into the standalone library testLib.
...............................
This PR adds tests for the functions under the "Primitives" section in Messenger.h.

It also removes the unused files Message.proto and Test_Message.cpp, which were only used during initial stage of protobuf work.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/Message/Test_Messenger_Primitives.cpp
```diff
@@ -0,0 +1,395 @@
...
+    std::numeric_limits<uint8_t>::min(), std::numeric_limits<uint8_t>::max());
+std::uniform_int_distribution<std::mt19937::result_type> distUint16(
+    std::numeric_limits<uint16_t>::min(), std::numeric_limits<uint16_t>::max());
+std::uniform_int_distribution<std::mt19937::result_type> distUint32(
+    std::numeric_limits<uint32_t>::min(), std::numeric_limits<uint32_t>::max());
+
+PubKey GenerateRandomPubKey() { return PubKey(PrivKey()); }
+
+Peer GenerateRandomPeer() {
+  uint128_t ip_address = distUint32(rng);
```
bb111189: Use min and max inmt19937 to bound it to a valid ip address and port

jendis:
I definitely agree on moving the helper functions to separate *.h/cpp file so everyone can use it - I consider them really beneficial :).
I have already created the similar one so I can switch to yours or continue on developing the additional ones so we don't have to explore the wheel again :).

From my side the small suggestion would be to the part:
`
std::mt19937 rng;
std::uniform_int_distribution<std::mt19937::result_type> dist1to99(1, 99);
std::uniform_int_distribution<std::mt19937::result_type> distUint8(
    std::numeric_limits<uint8_t>::min(), std::numeric_limits<uint8_t>::max());
std::uniform_int_distribution<std::mt19937::result_type> distUint16(
    std::numeric_limits<uint16_t>::min(), std::numeric_limits<uint16_t>::max());
std::uniform_int_distribution<std::mt19937::result_type> distUint32(
    std::numeric_limits<uint32_t>::min(), std::numeric_limits<uint32_t>::max());
std::uniform_int_distribution<std::mt19937::result_type> distUint64(
    std::numeric_limits<uint64_t>::min(), std::numeric_limits<uint64_t>::max());
`

This can be conveniently compressed in one function :
`
std::mt19937 rng;
template <typename T>
T generateFromNtoM(T n, T m){
  return std::uniform_int_distribution<T> {n, m}(rng);
}
`
So after that this can be finally used as for example:
`
uint8_t epoch_size = generateFromNtoM<uint8_t>(1,100);
`

jendis:
I would especially appreciate the review of the changes in CMakeList.txt.
I modified it based on the existing ones because of my poor available time. I haven't payed attention on the build setup too much yet so I suppose there will be useless includes etc.

PR #768, opened 2018-10-29
[testnetv3-dev] Fix the Bad Pow Ordering

sandipbhoir:
## Description
When we have missing pow , DS backup fetches them from DS block announcement.
However, it later results in bad pow ordering.

Issue - https://github.com/Zilliqa/Issues/issues/263

## Review Suggestion

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -335,6 +335,7 @@ bool DirectoryService::VerifyPoWOrdering(
             return item.second == toFind;
           });
 
+      std::array<unsigned char, 32> result;
       if (it == sortedPoWSolns.cend()) {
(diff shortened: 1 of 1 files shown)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -335,6 +335,7 @@ bool DirectoryService::VerifyPoWOrdering(
             return item.second == toFind;
           });
 
+      std::array<unsigned char, 32> result;
       if (it == sortedPoWSolns.cend()) {
```
ShengguangXiao: The logic is correct. But i suggest to put the switch the if else condition check. Because most of the time will run to (it != sortedPoWSolns.cend()), then the code will be more efficient.

PR #775, opened 2018-10-30
Protobuf Account and AccountStore

ansnunez:
## Description

This PR does the conversion to Protobuf for both `Account` and `AccountStore`.

`AccountStore` is quite tricky because of the inheritance structure starting from `AccountStoreBase`.

Here is some information to help with the review:

1. Because `AccountStoreBase` is a template class, we end up in a recursive header inclusion logic between `AccountStoreBase` and `Messenger`.  To remedy this, I had to move out the functions that `AccountStoreBase` needs from `Messenger` into another file MessengerAccountStoreBase.h.

2. In my build afterwards, the inclusion of MemoryDB.h in some files somehow triggered the unused parameter clang error. This header actually seems unused anyway, so I removed it from some files.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libPersistence/BlockStorage.h
```diff
@@ -27,6 +27,7 @@
 
 #include "common/Singleton.h"
 #include "depends/libDatabase/LevelDB.h"
+#include "libData/AccountData/TransactionReceipt.h"
```
ShengguangXiao: Redundant include, removed still can build

src/libServer/JSONConversion.h
```diff
@@ -24,6 +24,7 @@
 #include <array>
 #include <vector>
 
+#include "libData/AccountData/TransactionReceipt.h"
```
ShengguangXiao: Redundant include, removed still can build

src/libMessage/MessengerAccountStoreBase.h
```diff
@@ -0,0 +1,40 @@
...
+ * are governed by GPLv3.0 are those programs that are located in the folders
+ * src/depends and tests/depends and which include a reference to GPLv3 in their
+ * program files.
+ */
+#ifndef __MESSENGERACCOUNTSTOREBASE_H__
+#define __MESSENGERACCOUNTSTOREBASE_H__
+
+#include <vector>
+
+class MessengerAccountStoreBase {
```
ShengguangXiao: No class inherit from MessengerAccountStoreBase, why it named with Base postfix?

src/libMessage/ZilliqaMessage.proto
```diff
@@ -11,6 +11,34 @@ message ByteArray
...
+    optional bytes codehash        = 5;
+    optional uint64 createblocknum = 6;
+    optional bytes initdata        = 7;
+    optional bytes code            = 8;
+    repeated StorageData storage   = 9;
+}
+
+message ProtoAccountStore
+{
+    message Account
```
ShengguangXiao: This message included address and Account, can change it to AddressAccount?

src/libMessage/MessengerAccountStoreBase.h
ansnunez: I named it `MessengerAccountStoreBase` because only the class `AccountStoreBase` uses this class.  If I name it `Aux`, it won't be so obvious that it's for that `Base` class.

src/libMessage/ZilliqaMessage.proto
ansnunez: Ok will do

src/libPersistence/BlockStorage.h
ansnunez: Ok will remove

src/libServer/JSONConversion.h
ansnunez: Ok will remove

src/libMessage/Messenger.cpp
```diff
@@ -111,6 +115,227 @@ void NumberToArray(const T& number, vector<unsigned char>& dst,
...
+
+  bool fullCopy = false;
+
+  if (oldAccount == nullptr) {
+    oldAccount = &acc;
+    fullCopy = true;
+  }
+
+  int256_t balanceDelta =
+      int256_t(newAccount.GetBalance()) - int256_t(oldAccount->GetBalance());
```
bb111189: Can we use safemath here?

src/libMessage/Messenger.cpp
```diff
@@ -111,6 +115,227 @@ void NumberToArray(const T& number, vector<unsigned char>& dst,
...
+  protoAccount.set_numbersign(balanceDelta > 0);
+
+  uint256_t balanceDeltaNum(abs(balanceDelta));
+  NumberToProtobufByteArray<uint256_t, UINT256_SIZE>(
+      balanceDeltaNum, *protoAccount.mutable_balance());
+
+  uint256_t nonceDelta = newAccount.GetNonce() - oldAccount->GetNonce();
```
bb111189: use safe math

src/libMessage/Messenger.cpp
```diff
@@ -111,6 +115,227 @@ void NumberToArray(const T& number, vector<unsigned char>& dst,
...
+  ProtobufByteArrayToNumber<uint256_t, UINT256_SIZE>(protoAccount.nonce(),
+                                                     tmpNumber);
+  account.IncreaseNonceBy(tmpNumber);
+
+  if (protoAccount.code().size() > 0) {
+    bool doInitContract = false;
+
+    if (fullCopy) {
+      vector<unsigned char> tmpVec;
+      tmpVec.resize(protoAccount.code().size());
```
bb111189: should check that size() is not excessively large before doing resizing

src/libMessage/Messenger.cpp
ansnunez: On second thought, will not change this for now since SafeMath doesn't support signed type. Issue created: https://github.com/Zilliqa/Zilliqa/issues/792

src/libMessage/Messenger.cpp
ansnunez: ok

PR #784, opened 2018-10-31
Upgrade/release should refer to account/repo defined in constant file for binary host

ckyang:
## Description

For the binary URL, both UpgradeManager & relase.sh should refer to account/repo defined inside constant file, in order to be consistent, and be more flexible.
Besides, this PR also fix a minor bug caused from code refactoring before.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


README.md
```diff
@@ -72,7 +72,7 @@ To run Zilliqa, we recommend the following minimum system requirements:
     sudo apt-get install git libboost-system-dev libboost-filesystem-dev libboost-test-dev \
         libssl-dev libleveldb-dev libjsoncpp-dev libsnappy-dev cmake libmicrohttpd-dev \
         libjsonrpccpp-dev build-essential pkg-config libevent-dev libminiupnpc-dev \
-        libprotobuf-dev protobuf-compiler libcurl4-openssl-dev
+        libprotobuf-dev protobuf-compiler libcurl4-openssl-dev xml-twig-tools
```
Gnnng: Can we be conservative when adding new dependencies? This is actually not a code build dependency and could be avoided by using some workaround. Also, you have to check whether `xml_grep` is available on Mac or not

ckyang: OK, I'll try a workaround without using this external tool

PR #796, opened 2018-11-01
Get Directory Blocks From Persistence

KaustubhShamshery:
## Description
Instead of retrieving just DS Blocks all VC/FB/DS should be retrieved

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -268,7 +268,7 @@ bool Node::StartRetrieveHistory(bool& wakeupForUpgrade) {
 
   /// Retrieve DS blocks
   bool ds_result;
-  m_retriever->RetrieveDirectoryBlocks(ds_result, wakeupForUpgrade);
+  m_retriever->RetrieveBlockLink(ds_result, wakeupForUpgrade);
 
   /// Retrieve Tx blocks, relative final-block state-delta from persistence
   bool st_result = m_retriever->RetrieveStates();
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -233,6 +233,20 @@ bool Node::StartRetrieveHistory(bool& wakeupForUpgrade) {
 
   m_mediator.m_txBlockChain.Reset();
   m_mediator.m_dsBlockChain.Reset();
+  m_mediator.m_blocklinkchain.Reset();
+  m_mediator.m_blocklinkchain.GetBuiltDSComm().clear();
+  {
+    lock_guard<mutex> lock(m_mediator.m_mutexInitialDSCommittee);
+    if (m_mediator.m_initialDSCommittee->size() != 0) {
+      for (const auto& initDSCommKey : *m_mediator.m_initialDSCommittee) {
+        m_mediator.m_blocklinkchain.GetBuiltDSComm().emplace_back(initDSCommKey,
+                                                                  Peer());
+        // Set initial ds committee with null peer
+      }
+    } else {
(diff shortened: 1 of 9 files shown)


src/libPersistence/Retriever.cpp
```diff
@@ -167,6 +167,171 @@ void Retriever::RetrieveTxBlocks(bool& result, const bool& wakeupForUpgrade) {
...
+        LOG_GENERAL(WARNING, "Could not delete VC block");
+      }
+    } else if (std::get<BlockLinkIndex::BLOCKTYPE>(blocklink) ==
+               BlockType::FB) {
+      if (!BlockStorage::GetBlockStorage().DeleteFallbackBlock(
+              std::get<BlockLinkIndex::BLOCKHASH>(blocklink))) {
+        LOG_GENERAL(WARNING, "Could not deleteLoop  FB block");
+      }
+    }
+  }
```
ckyang: When toDelete is true, it means not only blockstorage should be removed, but also new block should not be added to m_blocklinkchain. We should consider toDelete before calling m_blocklinkchain.AddBlockLink().

src/libNode/Node.cpp
```diff
@@ -233,6 +233,20 @@ bool Node::StartRetrieveHistory(bool& wakeupForUpgrade) {
...
+  {
+    lock_guard<mutex> lock(m_mediator.m_mutexInitialDSCommittee);
+    if (m_mediator.m_initialDSCommittee->size() != 0) {
+      for (const auto& initDSCommKey : *m_mediator.m_initialDSCommittee) {
+        m_mediator.m_blocklinkchain.GetBuiltDSComm().emplace_back(initDSCommKey,
+                                                                  Peer());
+        // Set initial ds committee with null peer
+      }
+    } else {
+      LOG_GENERAL(WARNING, "Initial DS comm size 0 ");
```
bb111189: I think this should be upgraded to FATAL as initial DS comm cannot be 0

src/libPersistence/BlockStorage.cpp
```diff
@@ -408,6 +418,36 @@ bool BlockStorage::GetAllTxBodiesTmp(std::list<TxnHash>& txnHashes) {
...
+      LOG_GENERAL(WARNING, "Lost one blocklink in the chain");
+      delete it;
+      return false;
+    }
+    BlockLink blcklink;
+    if (!Messenger::GetBlockLink(
+            vector<unsigned char>(blockString.begin(), blockString.end()), 0,
+            blcklink)) {
+      LOG_GENERAL(WARNING, "Deserialization of blockLink failed " << bns);
+      return false;
```
bb111189: need to delete it here?

src/libPersistence/Retriever.cpp
```diff
@@ -167,6 +167,172 @@ void Retriever::RetrieveTxBlocks(bool& result, const bool& wakeupForUpgrade) {
...
+      LOG_GENERAL(INFO, "Has incompleted DS Block, remove it");
+      toDelete = true;
+    }
+  }
+
+  uint64_t lastDsIndex = std::get<BlockLinkIndex::DSINDEX>(blocklinks.back());
+  const BlockType lastType =
+      std::get<BlockLinkIndex::BLOCKTYPE>(blocklinks.back());
+  if (lastType != BlockType::DS) {
+    lastDsIndex--;
```
bb111189: Should check for underflow

PR #806, opened 2018-11-02
Fix retrieve history failed but genesis account not added to blockchain problem

ShengguangXiao:
## Description

Fix retrieve history failed but genesis account not added to blockchain problem

If StartRetrieveHistory return false, don't directly return the Install function, continue add genesis account. 

https://github.com/Zilliqa/Issues/issues/282

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/BlockChainData/BlockLinkChain.h
+++ b/src/libData/BlockChainData/BlockLinkChain.h
@@ -122,6 +122,7 @@ class BlockLinkChain {
   void SetBuiltDSComm(const std::deque<std::pair<PubKey, Peer>>& dsComm) {
     m_builtDsCommittee = dsComm;
   }
+
   const BlockLink& GetLatestBlockLink() {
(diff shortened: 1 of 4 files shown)


src/libData/BlockChainData/BlockLinkChain.h
```diff
@@ -122,6 +122,7 @@ class BlockLinkChain {
   void SetBuiltDSComm(const std::deque<std::pair<PubKey, Peer>>& dsComm) {
     m_builtDsCommittee = dsComm;
   }
+
```
KaustubhShamshery: the only change in this file seems to be space

ShengguangXiao: Yeah, i add a empty line because there should be a empty line between two functions. It is easier to tell these are two functions.

PR #814, opened 2018-11-06
Lookup Rewards (+ Placeholder for Gas Price)

KaustubhShamshery:
## Description
Reward Lookup nodes using coinbase 

Each lookup is rewarded a constant percentage of the reward equally.

## Review Suggestion
The commented code in `DirectoryService::LookupCoinbase` is for when we need to provide the lookup rewards on the basis of the new node it made to join

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (commit : 5a9411c )
- [x] small-scale cloud test (commit: c6d40bd)


constants.xml
```diff
@@ -69,6 +69,7 @@
         <NUM_CONSENSUS_SUBSETS>5</NUM_CONSENSUS_SUBSETS>
         <MISORDER_TOLERANCE_IN_PERCENT>5</MISORDER_TOLERANCE_IN_PERCENT>
         <MAX_CODE_SIZE_IN_BYTES>20480</MAX_CODE_SIZE_IN_BYTES>
+        <LOOKUP_REWARD>5</LOOKUP_REWARD>
```
ansnunez: I suggest to rename `LOOKUP_REWARD_IN_PERCENT`

src/libDirectoryService/Coinbase.cpp
```diff
@@ -137,8 +202,17 @@ void DirectoryService::InitCoinbase() {
   }
 
   LOG_MARKER();
+
+  VectorOfLookupNode vecLookup = m_mediator.m_lookup->GetLookupNodes();
```
ansnunez: Hmm perhaps we can change `GetLookupNodes()` to return `const VectorOfLookupNode&`, then here (and everywhere else it is called) we just get the reference instead of making a new vector?

src/libDirectoryService/Coinbase.cpp
```diff
@@ -175,34 +254,56 @@ void DirectoryService::InitCoinbase() {
...
     for (auto const& shardId : epochNum.second) {
       LOG_GENERAL(INFO, "[CNBSE] Rewarding " << shardId.first << " shard");
-
-      for (auto const& addr : shardId.second) {
-        if (!AccountStore::GetInstance().UpdateCoinbaseTemp(
-                addr, genesisAccount, reward_each)) {
-          LOG_GENERAL(WARNING, "Could Not reward " << addr);
-        } else {
-          suc_counter++;
+      if (shardId.first == -2) {
```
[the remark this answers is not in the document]
ansnunez: I agree.  As it is used in multiple places.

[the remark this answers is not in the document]
ansnunez: Did we designate -1 for DS committee? Also, if that one didn't have a const variable either, let's make one for -1 as well.

src/libDirectoryService/Coinbase.cpp
```diff
@@ -137,8 +202,17 @@ void DirectoryService::InitCoinbase() {
...
   lock_guard<mutex> g(m_mutexCoinbaseRewardees);
 
+  for (const auto& lookupNode : vecLookup) {
+    m_coinbaseRewardees[epochNum][-2].push_back(
+        Account::GetAddressFromPublicKey(lookupNode.first));
+  }
+
   if (m_coinbaseRewardees.size() < NUM_FINAL_BLOCK_PER_POW) {
     LOG_GENERAL(INFO, "[CNBSE]"
                           << "Less then expected rewardees "
```
ansnunez: I think we have to change these logs from "expected rewardees" to "expected tx epochs for rewarding" or something like that, and change INFO to WARNING. Also, probably good to actually log the size and NUM_FINAL_BLOCK_PER_POW.

src/libDirectoryService/Coinbase.cpp
```diff
@@ -175,34 +254,56 @@ void DirectoryService::InitCoinbase() {
...
+          } else {
+            suc_counter++;
+          }
         }
       }
     }
   }
 
-  uint256_t balance_left = total_reward - (suc_counter * reward_each);
+  uint256_t balance_left = total_reward - (suc_counter * reward_each) -
```
ansnunez: Just checking: Is it possible - especially with the division operations earlier - that somehow `suc_counter * reward_each` and `suc_lookup_counter * reward_each_lookup` end up being a little more than `total_reward`?  In that case, we have an overflow here.  Also, should we use safe math?

src/libDirectoryService/Coinbase.cpp
```diff
@@ -157,13 +231,18 @@ void DirectoryService::InitCoinbase() {
   Address genesisAccount(GENESIS_WALLETS[0]);
 
   uint256_t sig_count = 0;
-
+  uint64_t lookup_count = 0;
```
ansnunez: `lookupId` in `struct PoWSolution` is `uint32_t`, whereas for the count we use `uint64_t` here

src/libMessage/Messenger.cpp
```diff
@@ -530,6 +532,7 @@ void ProtobufToShardingStructureAnnouncement(
     for (const auto& proto_member : proto_shard.members()) {
       PubKey key;
       Peer peer;
+      Signature signature;
```
ansnunez: Can remove?

src/libMessage/Messenger.cpp
```diff
@@ -2890,6 +2903,7 @@ bool Messenger::GetDSDSBlockAnnouncement(
     PubKey key;
     std::array<unsigned char, 32> result;
     std::array<unsigned char, 32> mixhash;
+    Signature signature;
```
ansnunez: Can remove?

src/libDirectoryService/Coinbase.cpp
KaustubhShamshery: There is a mutex involved with the var returned. so I am not sure

src/libDirectoryService/Coinbase.cpp
KaustubhShamshery: Already used safemath to divide to get the each reward values

src/libDirectoryService/Coinbase.cpp
ansnunez: I see, then better not change for now

PR #827, opened 2018-11-07
Call scilla checker during contract deployment

kaikawaliu:
## Description

Add scilla checker during the processing of transaction that trying to deploy a contract

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -155,6 +155,18 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
 
     ExportCreateContractFiles(*toAccount);
 
+    // Undergo scilla checker
+    bool ret_checker = true;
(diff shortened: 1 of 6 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -443,6 +463,26 @@ std::string AccountStoreSC<MAP>::GetCallContractCmdStr(
   return ret;
 }
 
+template <class MAP>
+bool AccountStoreSC<MAP>::ParseContractCheckerOutput(
```
ansnunez: Do we plan to do something with the checker output eventually?  It currently seems unused apart from seeing if it's parse-able here.

kaikawaliu: The parse will only check whether the output from Scilla-Checker can form a JSON object or not. If can then it means passed, otherwise it will be a error message which can not be converted to JSON

PR #828, opened 2018-11-07
Avoid commit window when NUM_CONSENSUS_SUBSETS=1

ansnunez:
## Description

Previous behavior before the consensus subsets were implemented was to proceed with challenge phase when we receive the minimum 2/3 commits.

After subsets was implemented, we now have a time window to receive as many commits as we want, even above the minimum 2/3.

This PR brings back the old behavior (just receive minimum commits, no timer) when we set `NUM_CONSENSUS_SUBSETS` to 1 in the constants file.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libConsensus/ConsensusLeader.cpp
+++ b/src/libConsensus/ConsensusLeader.cpp
@@ -103,7 +103,6 @@ void ConsensusLeader::SetStateSubset(uint16_t subsetID, State newState) {
 
 void ConsensusLeader::GenerateConsensusSubsets() {
   LOG_MARKER();
-  lock_guard<mutex> g(m_mutex);
 
   // Get the list of all the peers who committed, by peer index
   vector<unsigned int> peersWhoCommitted;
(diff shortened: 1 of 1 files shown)


src/libConsensus/ConsensusLeader.cpp
```diff
@@ -328,12 +326,21 @@ bool ConsensusLeader::ProcessMessageCommitCore(
...
+      // notify the waiting thread to start with subset creations and subset
+      // consensus.
+      if (m_commitCounter == m_committee.size()) {
+        lock_guard<mutex> g(m_mutexAnnounceSubsetConsensus);
+        m_allCommitsReceived = true;
+        cv_scheduleSubsetConsensus.notify_all();
+      }
+    } else {
+      if (m_commitCounter == m_numForConsensus) {
+        LOG_GENERAL(INFO, "Sufficient commits obtained. Required/Actual = "
```
ShengguangXiao: This log looks doesn't log the required and actual commits.

ansnunez: As they are equal, we only need to display one value

ShengguangXiao: Ok, i see

PR #829, opened 2018-11-07
View change prechecking

bb111189:
## Description

This PR implements View change prechecking. In the prechecking phase, DS node(s) will query lookup to check whether is there any new DS block(s) or Final block(s). If there is, it implies that the DS node is lagging behind. Hence, instead of doing view change, it will rejoin as a DS node. 

Known issue:
- RejoinAsDS() is currently not working

## Review Suggestion

Check documentation on view change

### Implementation

- [x] **ready for review**

Normal view change
- [x] [vc1] normal view change at ds consensus
- [x] [vc3] normal view change at final block consensus

1 Node lagging behind
- [x] [vc7] 1 node lag behind and enter view change mode. It will rejoin as DS node
- [x] [vc8] 1 node lag behind and enter view change mode. It will rejoin as DS node


src/libLookup/Lookup.cpp
```diff
@@ -383,11 +371,11 @@ vector<unsigned char> Lookup::ComposeGetTxBlockMessage(uint64_t lowBlockNum,
   vector<unsigned char> getTxBlockMessage = {
       MessageType::LOOKUP, LookupInstructionType::GETTXBLOCKFROMSEED};
 
-  if (!Messenger::SetLookupGetTxBlockFromSeed(
+  if (!Messenger::GetLookupGetTxBlockFromSeed(
```
KaustubhShamshery: Why use Get here ?

bb111189: Thanks. Accidental change.

PR #847, opened 2018-11-11
Rumor Spreading test preparation

Jazzz42:
## Description
This is a preparation for further testing of Rumor Spreading open for review. Later on the code might be moved from Mediator.h to libTestUtils although that would mean change in /src/libMediator/CMakeLists.txt

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libMediator/Mediator.h
```diff
@@ -33,12 +33,52 @@
...
+};
+
 /// A mediator class for providing access to global members.
 class Mediator {
   std::mutex m_mutexHeartBeat;
   unsigned int m_heartBeatTime;
 
  public:
+  /// SysTest data
+  SysTestData* m_systest_data;
```
ansnunez: Perhaps you can just have a `SysTestData` member, not `SysTestData*`

src/libMediator/Mediator.h
```diff
@@ -33,12 +33,52 @@
...
+  bool IsANode = false;
+  uint64_t TestStartEpoch = 4;
+  uint64_t CurrentEpoch = 0;
+  uint16_t MyNodeId = 0;
+  uint16_t FirstAId = 0;
+
+ public:
+  /// Constructor
+  SysTestData() {
+    TestStartEpoch = TEST_START_EPOCH;
```
ansnunez: Do you intend to change these values some time during operation? If not, we can just directly use the constants.

src/libMediator/Mediator.h
mollar2020: !!!

PR #849, opened 2018-11-12
Migrate recent unit tests from dev branch to master

ansnunez:
## Description

These are recent unit tests developed together with @Jazzz42 and @jendis along another branch.  Migrating all the changes to master in this PR.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/Mediator/constants.xml
```diff
@@ -0,0 +1,164 @@
+<?xml version="1.0"?>
```
ShengguangXiao: Why add constants.xml in Mediator test folder?

tests/Utils/Test_ShardSizeCalculator.cpp
```diff
@@ -17,131 +17,87 @@
...
-
-BOOST_AUTO_TEST_CASE(test_max_shards_lower_bound) {
-  INIT_STDOUT_LOGGER();
-  uint32_t result = ShardSizeCalculator::CalculateShardSize(19584);
-  BOOST_CHECK_MESSAGE(result == 816,
-                      "Expected: 816. Result: " + to_string(result));
+typedef std::vector<uint32_t> TestDataSet;
+typedef std::map<uint32_t, TestDataSet> ShardSizeMap;
+
+void prepareTestdata(ShardSizeMap& testData) {
```
bb111189: There is a need to test boundary case such as the shard size low and upper bound for each shard size range. This can be done as a future todos

tests/Mediator/constants.xml
ansnunez: To override constants for some unit tests.
But anyway Mediator is disabled for now.
We will change this test some other time.

PR #851, opened 2018-11-12
Implementation of Gas Pricer

kaikawaliu:
## Description

The gas price decides the minimum gas price can be accepted for the transactions received during the current ds epoch.

The detailed gas price calculation algorithm is in libDirectoryService/GasPricer.cpp
Document (only visible to core members): https://docs.google.com/document/d/1eQ_j1f2lYAtPY5hmKKR6Bszr8EWOBWlxh2GZM5Unu0U/edit

Some implementation details:
- To enable the referring to the latest several ds epoch, we force the fetching ds block function in liblookup to fetch as much as the gas pricer needed.
- Added a new message type for user proposing customized gas price
- The DEFAULT_MIN_GAS_PRICE is replaced by PRECISION_MIN_VALUE which is to provide valid math operations based on desired precision for increasing or decreasing of gas price. PRECISION_MIN_VALUE = 2 means 10 powers 2 which means the precision is 1/(10^2)
- SafeMath::pow

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test


src/libDirectoryService/GasPricer.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+uint256_t DirectoryService::GetHistoricalMeanGasPrice() {
+  uint64_t curDSBlockNum =
+      m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum();
+  uint64_t lowDSBlockNum = (curDSBlockNum > MEAN_GAS_PRICE_DS_NUM)
+                               ? (curDSBlockNum - MEAN_GAS_PRICE_DS_NUM)
+                               : 0;
+  uint64_t totalBlockNum = 0;
+  uint256_t totalGasPrice = 0;
+  for (uint64_t i = curDSBlockNum; i >= lowDSBlockNum;) {
```
bb111189: put i-- in the for loop?

src/libDirectoryService/GasPricer.cpp
```diff
@@ -0,0 +1,205 @@
...
+    if (!SafeMath<uint256_t>::add(
+            totalGasPrice,
+            m_mediator.m_dsBlockChain.GetBlock(i).GetHeader().GetGasPrice(),
+            totalGasPrice)) {
+      break;
```
bb111189: should this be continue instead?

src/libNode/Node.cpp
```diff
@@ -997,6 +997,43 @@ bool Node::ProcessTxnPacketFromLookupCore(const vector<unsigned char>& message,
...
+    return true;
+  }
+
+  if (IsMessageSizeInappropriate(message.size(), offset, UINT256_SIZE)) {
+    LOG_GENERAL(WARNING,
+                "Message size for ProcessProposeGasPrice is too short");
+    return false;
+  }
+
+  if (string(from.GetPrintableIPAddress()) != "127.0.0.1") {
```
bb111189: what is the rationale for this check?

src/libUtils/SafeMath.h
```diff
@@ -41,6 +41,55 @@ class SafeMath {
...
+        return false;
+      }
+      --count;
+    }
+
+    result = temp;
+    return true;
+  }
+
+  // be careful when using
```
bb111189: can describe more?

src/libNode/Node.cpp
kaikawaliu: The current check is only allow the user send the command from local machine. Yup I agree we can configure that in the constants so let the user to decide which ip is legal to propose gas price for it.

src/libUtils/SafeMath.h
ansnunez: I think we can just keep one `pow` and have an optional `bool isCritical = true` parameter?

bb111189: Maybe describe the use case in the inline comments

src/libNode/Node.cpp
ansnunez: After discussing with @kaikawaliu, another option is we can have the node's user affix a signature to the end of this message, so in effect the node verifies its own signature.

constants_local.xml
```diff
@@ -94,7 +93,15 @@
...
-        <DEFAULT_MIN_GAS_PRICE>1</DEFAULT_MIN_GAS_PRICE>
+        <GAS_CONGESTION_PERCENT>80</GAS_CONGESTION_PERCENT>
+        <UNFILLED_PERCENT_LOW>10</UNFILLED_PERCENT_LOW>
+        <UNFILLED_PERCENT_HIGH>70</UNFILLED_PERCENT_HIGH>
+        <GAS_PRICE_PRECISION>2</GAS_PRICE_PRECISION>
+        <GAS_PRICE_DROP_RATIO>10</GAS_PRICE_DROP_RATIO>
+        <GAS_PRICE_RAISE_RATIO_LOWER>5</GAS_PRICE_RAISE_RATIO_LOWER>
+        <GAS_PRICE_RAISE_RATIO_UPPER>15</GAS_PRICE_RAISE_RATIO_UPPER>
+        <GAS_PRICE_TOLERANCE>10</GAS_PRICE_TOLERANCE>
+        <MEAN_GAS_PRICE_DS_NUM>5</MEAN_GAS_PRICE_DS_NUM>
```
KaustubhShamshery: NO legal gas price IP constant here

kaikawaliu: fixed

src/libUtils/SafeMath.h
```diff
@@ -41,6 +41,57 @@ class SafeMath {
     return true;
   }
 
+  static bool power(const T& a, const T& b, T& result) {
```
KaustubhShamshery: can you rename exponent and base ?

src/libValidator/Validator.cpp
```diff
@@ -124,6 +124,17 @@ bool Validator::CheckCreatedTransactionFromLookup(const Transaction& tx) {
     }
   }
 
+  if (tx.GetGasPrice() <
+      m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetGasPrice()) {
```
KaustubhShamshery: Why fetch it everytime from blockchain, can be ineffecient for 1000s of txns

src/libUtils/SafeMath.h
kaikawaliu: okay fixed

bb111189:
@kaikawaliu Do fix the merge conflicts

PR #854, opened 2018-11-13
Fix normal node failing PoW and getting stuck.

KaustubhShamshery:
## Description
Node rejoins the network after it looses PoW. 

## Review Suggestion
Node is able to rejoin if COMM_SIZE = now_of_nodes_doing_pow - 1. and check if it does PoW in the next epoch.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (commit: ebd60ec)

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -2063,7 +2063,8 @@ bool Lookup::InitMining(uint32_t lookupIndex) {
 
   // Check whether is the new node connected to the network. Else, initiate
   // re-sync process again.
-  this_thread::sleep_for(chrono::seconds(POW_WINDOW_IN_SECONDS));
+  this_thread::sleep_for(
+      chrono::seconds(POW_WINDOW_IN_SECONDS + 2 * NEW_NODE_SYNC_INTERVAL));
   m_startedPoW = false;
   if (m_syncType != SyncType::NO_SYNC) {
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -1519,8 +1519,7 @@ bool Lookup::ProcessSetDSBlockFromSeed(const vector<unsigned char>& message,
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
               "I already have the block");
   } else {
-    if (m_syncType == SyncType::NO_SYNC &&
-        m_mediator.m_node->m_stillMiningPrimary) {
+    if (m_syncType == SyncType::NO_SYNC) {
       m_fetchedLatestDSBlock = true;
       cv_latestDSBlock.notify_all();
       return true;
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -1487,10 +1487,6 @@ bool Lookup::ProcessSetDSBlockFromSeed(const vector<unsigned char>& message,
 
   LOG_MARKER();
 
-  if (AlreadyJoinedNetwork()) {
-    return true;
-  }
-
   unique_lock<mutex> lock(m_mutexSetDSBlockFromSeed);
 
   uint64_t lowBlockNum;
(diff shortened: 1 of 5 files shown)


src/libNode/PoWProcessing.cpp
```diff
@@ -220,6 +227,28 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
     }
   }
 
+  if (sentToDs) {
+    auto func = [this]() mutable -> void {
+      unique_lock<mutex> lk(m_mutexCVWaitDSBlock);
+      if (cv_waitDSBlock.wait_for(
+              lk,
+              chrono::seconds(NEW_NODE_SYNC_INTERVAL + POW_WINDOW_IN_SECONDS +
+                              FALLBACK_EXTRA_TIME)) == cv_status::timeout) {
```
kaikawaliu: Do you mean the sleep should be longer than enough for triggering view change? if so should use FALLBACK_INTERVAL_WAITING instead

src/libLookup/Lookup.cpp
```diff
@@ -1519,8 +1515,7 @@ bool Lookup::ProcessSetDSBlockFromSeed(const vector<unsigned char>& message,
     LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
               "I already have the block");
   } else {
-    if (m_syncType == SyncType::NO_SYNC &&
```
kaikawaliu: Why changed this condition

src/libLookup/Lookup.cpp
```diff
@@ -1487,10 +1487,6 @@ bool Lookup::ProcessSetDSBlockFromSeed(const vector<unsigned char>& message,
 
   LOG_MARKER();
 
-  if (AlreadyJoinedNetwork()) {
```
kaikawaliu: What's the intention of putting this down?

satoshi101:
Ναι, αυτό είναι το code coverage. Είναι καλό.

Ok-επειδή είδα PoW νόμιζα ότι θα έχει τίποτα erformance related στο hashing
που κάνει. Νομίζω θέλουν GPU mining - δεν ξέρω αν το έχουν κάνει ήδη.

Ναι, αυτό είναι ένα συμπαθητικό project με το οποίο έχω δουλέψει στο
παρελθόν.

src/libLookup/Lookup.cpp
KaustubhShamshery: because it will never check the condition if SyncTyepe == NoSync and return from here.

src/libLookup/Lookup.cpp
KaustubhShamshery: When the node sent the solution to ds, it has stopped mining.

src/libNode/PoWProcessing.cpp
KaustubhShamshery: The sleep must be longer than POW_WINDOW + DS block consensus

src/libLookup/Lookup.cpp
kaikawaliu: But this is ProcessSetDSBlockFromSeed, which is the response when the node calls GetDSBlockFromLookup, what's the relationship of sending the solution to ds here?

src/libLookup/Lookup.cpp
kaikawaliu: The original intention for this check is for discarding the extra ProcessSetDSBlock message

src/libLookup/Lookup.cpp
KaustubhShamshery: New node sends ds pow solution -> waits ->checks if ds block has been mined

src/libLookup/Lookup.cpp
KaustubhShamshery: yes, it would still discard it but after notifying the cv

src/libNode/PoWProcessing.cpp
kaikawaliu: For that purpose, should use some other constants, as this FALLBACK_EXTRA_TIME is used for different purpose

src/libLookup/Lookup.cpp
kaikawaliu: I see, seems we no longer use this for synchronization. specifically used for checking pow

src/libLookup/Lookup.cpp
kaikawaliu: ok I see

src/libNode/PoWProcessing.cpp
```diff
@@ -220,6 +227,28 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
...
+        LOG_GENERAL(WARNING, "Time out while waiting for DS Block");
+        if (GetLatestDSBlock()) {
+          LOG_GENERAL(INFO, "fDS block created, means I lost PoW");
```
kaikawaliu: typo here

PR #856, opened 2018-11-13
Improvised gossip using hashes only instead of raw message

sandipbhoir:
## Description
This PR covers : 
1. Now only hash of raw message will be gossiped instead of raw message using LAZY_PULL or LAZY_PUSH message ( Previously were named as PULL or PUSH instead )
2. Receiver of hash message (node N1) ask the sender (node N2) of hash message for the corresponding raw message using PULL message
3. In response to PULL message from node N1, node N2 sends the corresponding raw message to node N1 using PUSH message.
3. If the sender of hash message (N2) himself don't have the corresponding raw message, he adds the (N1 + hash) to subscriber list. As soon as he himself receives the raw message, he sends the same to N1 using PUSH message
4. If node N1 receives same hash message from other nodes say N3 and he have not yet received Raw message from N2 , then he will also ask N3 for raw message using PULL message. This way, if N2 is faulty, N1 will still receive real raw message from other peers ( N3 in this case )

## Review Suggestion
Please check files changed.
Run local test.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNetwork/RumorManager.cpp
```diff
@@ -42,7 +42,7 @@ RRS::Message::Type convertType(uint8_t type) {
 RumorManager::RumorManager()
     : m_peerIdPeerBimap(),
       m_peerIdSet(),
-      m_rumorIdRumorBimap(),
+      m_rumorIdHashBimap(),
```
KaustubhShamshery: should it be Bitmap ?

src/libNetwork/RumorManager.cpp
```diff
@@ -157,49 +157,53 @@ bool RumorManager::Initialize(const std::vector<Peer>& peers,
...
 bool RumorManager::AddRumor(const RumorManager::RawBytes& message) {
   LOG_MARKER();
-  {
-    std::lock_guard<std::mutex> guard(m_continueRoundMutex);
-    if (!m_continueRound) {  // Seems logical error. Round should have started.
-      if (message.size() > 0) {
-        SHA2<HASH_TYPE::HASH_VARIANT_256> sha256;
-        sha256.Update(message);  // raw_message hash
+  if (message.size() > 0) {
+    SHA2<HASH_TYPE::HASH_VARIANT_256> sha256;
```
KaustubhShamshery: can use hash utils function here

src/libNetwork/RumorManager.cpp
```diff
@@ -329,84 +333,185 @@ bool RumorManager::RumorReceived(uint8_t type, int32_t round,
...
+      // check if we have received the real message for this old rumor.
+      auto it = m_rumorHashRawMsgBimap.left.find(message);
+      if (it == m_rumorHashRawMsgBimap.left.end()) {
+        // didn't receive real message (PUSH) yet :( Lets ask this peer.
+        RRS::Message pullMsg(RRS::Message::Type::PULL, recvdRumorId, -1);
+        SendMessage(from, pullMsg);
+      }
+    }
+  } else if (RRS::Message::Type::PULL == t) {
+    // Now that sender wan't the real message, lets send it to him.
```
KaustubhShamshery: wan't -> want

src/libNetwork/RumorManager.cpp
sandipbhoir: These were been used in previous version. Which hash utils function are you referring?

KaustubhShamshery: static const std::vector<unsigned char> BytesToHash(
      const std::vector<unsigned char>& vec) {
    SHA2<HASH_TYPE::HASH_VARIANT_256> sha2;

    sha2.Update(vec);
    const std::vector<unsigned char>& resVec = sha2.Finalize();

    return resVec;
  }

PR #858, opened 2018-11-13
Use log fatal instead of assert in rumourholder.cpp

bb111189:
## Description

As the build is in release mode, `assert` will not be trigger. This PR replace `assert` in `rumourholder.cpp` with LOG FATAL. 

Note: I did a scan of usage of assert in the Zilliqa codebase (exclude depends) and this is the only place it was found.

### Implementation

- [x] **ready for review**

- [x] local machine test (commit: 085907608b9e3073de8a751eceaf405f10d8d775)

- [ ] small-scale cloud test

### DIFF
--- a/src/libRumorSpreading/RumorHolder.cpp
+++ b/src/libRumorSpreading/RumorHolder.cpp
@@ -101,7 +101,7 @@ RumorHolder::RumorHolder(const std::unordered_set<int>& peers,
   if (networkConfig.networkSize() != peers.size()) {
     LOG_GENERAL(
         FATAL,
-        "size of netoworkConfig does not match size of pees. networkConfig: "
+        "size of netoworkConfig does not match size of peers. networkConfig: "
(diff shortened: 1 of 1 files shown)


src/libRumorSpreading/RumorHolder.cpp
```diff
@@ -99,7 +98,12 @@ RumorHolder::RumorHolder(const std::unordered_set<int>& peers,
       m_nextMemberCb(),
       m_statistics(),
       m_maxNeighborsPerRound(1) {
-  assert(networkConfig.networkSize() == peers.size());
+  if (networkConfig.networkSize() != peers.size()) {
+    LOG_GENERAL(
+        FATAL,
+        "size of netoworkConfig does not match size of pees. networkConfig: "
```
ansnunez: Maybe fix the typos?

bb111189: Fixed

PR #861, opened 2018-11-13
Remove the dependency on GENESIS_KEYS

Gnnng:
## Description

Lookup node does not require genesis keys to generate the testing transaction anymore. This PR removes the use of GENESIS_KEYS in lookup code, as a necessary step to get https://github.com/Zilliqa/testnet/pull/190 work.

## Review Suggestion

N/A

### Implementation

- [x] **ready for review**

- [x] ~~local machine test~~

- [x] small-scale cloud test (commit: 74f637a)

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -130,18 +130,15 @@ bool Lookup::GenTxnToSend(size_t num_txn,
                           uint32_t numShards) {
   vector<Transaction> txns;
 
-  if (GENESIS_KEYS.size() == 0) {
-    LOG_GENERAL(WARNING, "No genesis keys found");
+  if (GENESIS_WALLETS.size() == 0) {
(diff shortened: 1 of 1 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -130,18 +130,15 @@ bool Lookup::GenTxnToSend(size_t num_txn,
                           uint32_t numShards) {
   vector<Transaction> txns;
 
-  if (GENESIS_KEYS.size() == 0) {
-    LOG_GENERAL(WARNING, "No genesis keys found");
+  if (GENESIS_WALLETS.size() == 0) {
```
KaustubhShamshery: The question of keys is such that the txn should be signed in gentxn. do we use a spearate constants.xml for that purpose then??

Gnnng: Don't we already load the signed transactions from the files?

PR #867, opened 2018-11-14
Fix schnorr memory allocation handling

bb111189:
## Description

In `Schnorr.cpp`, when memory allocation failures happen, in certain cases, it will proceeds to execute normally. Such behavior can be unpredictable. This PR properly address these cases.

### Implementation

- [x] **ready for review**

- [x] local machine test (commit: 5d1ccdf496bfe0a2742b61c10a6cf2e44fbcd4f5)

- [ ] small-scale cloud test


src/libCrypto/Schnorr.cpp
```diff
@@ -279,6 +281,7 @@ PubKey::PubKey(const PrivKey& privkey)
   if (m_P == nullptr) {
     LOG_GENERAL(WARNING, "Memory allocation failure");
     // throw exception();
+    return;
   } else if (!privkey.Initialized()) {
     LOG_GENERAL(WARNING, "Private key is not initialized");
```
jiayaoqijia: return; ?

bb111189: Resolved

PR #877, opened 2018-11-15
Update DS Block & remove placeholders

KaustubhShamshery:
## Description
**New fields (DS Block)**:
* difficultyDS -> number
* gasPrice -> string
* PoWWinners -> Array of String

**New API**
* GetMinimumGasPrice -> returns string of the numeric gas price
* GetPrevDiff -> integer containing difficulty (shard) of previous epoch
* GetPrevDSDiff -> integer containing difficulty (DS) of previous epoch

### Implementation

- [x] **Separate API for difficulty**

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libServer/Server.cpp
```diff
@@ -287,7 +260,10 @@ Json::Value Server::GetTxBlock(const string& blockNum) {
   }
 }
 
-string Server::GetGasPrice() { return "Hello"; }
+string Server::GetMinimumGasPrice() {
+  DSBlock latestDsBlock = m_mediator.m_dsBlockChain.GetLastBlock();
+  return latestDsBlock.GetHeader().GetGasPrice().str();
```
ckyang: How about "return m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetGasPrice().str();" to avoid temporarily memory used?

KaustubhShamshery: ok

iantanwx:
Tested, works.

PR #878, opened 2018-11-15
Fix DS rejoin.

KaustubhShamshery:
## Description
* Add multiple tries for fetching DS info
* Calculate leader based on Tx/VC block

## Status
- [x] **Remove MACRO**

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.h
```diff
@@ -169,6 +169,7 @@ class Lookup : public Executable, public Broadcastable {
   // TODO: move the Get and ProcessSet functions to Synchronizer
   bool GetSeedPeersFromLookup();
   bool GetDSInfoFromSeedNodes();
+  bool FetchDSInfoLoop();
```
kaikawaliu: This function is better to be located inside Synchronizer.h

KaustubhShamshery: Synchronizer does not have access to mediator

kaikawaliu: You can pass the mediator in as a variable.

src/libData/BlockChainData/BlockChain.h
```diff
@@ -71,13 +71,14 @@ class BlockChain {
...
-                               << " absent, a dummy block will be used and "
-                                  "abnormal behavior may happen!");
+    if (m_blocks.size() > 0 &&
+        (m_blocks.back().GetHeader().GetBlockNum() < blockNum)) {
+      LOG_GENERAL(WARNING,
+                  "BlockNum too high " << blockNum << " Dummy block used");
       return T();
-    } else if (blockNum + m_blocks.capacity() < m_blocks.size()) {
+    }
+
```
ckyang: Why there is a blank line here?

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -86,7 +86,14 @@ void DirectoryService::StartSynchronization() {
...
   };
 
+  auto func2 = [this]() -> void {
+    bool result = m_mediator.m_lookup->FetchDSInfoLoop();
+    if (!result) {
+      LOG_GENERAL(WARNING, "Unable to fetch DS info");
+    }
+  };
   DetachedFunction(1, func);
+  DetachedFunction(1, func2);
```
ckyang: FetchDSInfoLoop() will call GetDSInfoFromLookupNodes(), and need offline lookup node as well. Thus, We need to make sure FetchDSInfoLoop() be called after GetOfflineLookups() finished, just like FetchLatestTxBlocks().

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -86,7 +86,14 @@ void DirectoryService::StartSynchronization() {
     }
   };
 
+  auto func2 = [this]() -> void {
+    bool result = m_mediator.m_lookup->FetchDSInfoLoop();
+    if (!result) {
```
ckyang: We don't need a temporary variable "result", suggest to just use if(!m_mediator.m_lookup->FetchDSInfoLoop())

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -1206,8 +1206,8 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSBackup() {
     return true;
   }
 
-#ifdef VC_TEST_VC_PRECHECK_2
-  if (m_consensusMyID == 3) {
+#if VC_TEST_VC_PRECHECK_2
+  if (m_consensusMyID == 3 && m_mediator.m_currentEpochNum == 2) {
```
ckyang: What's the meaning of "3" and "2"?

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -1027,7 +1027,7 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSBackup() {
   }
 
 #ifdef VC_TEST_VC_PRECHECK_1
-  if (m_consensusMyID == 3) {
+  if (m_consensusMyID == 3 && m_mediator.m_currentEpochNum == 2) {
```
ckyang: What's the meaning of "3" and "2"?

src/libDirectoryService/ViewChangePreProcessing.cpp
```diff
@@ -454,13 +454,13 @@ bool DirectoryService::NodeVCPrecheck() {
       return true;
     }
 
-    for (const auto& dsblock : m_vcPreCheckDSBlocks) {
+    /*for (const auto& dsblock : m_vcPreCheckDSBlocks) {
```
ckyang: Why still preserve the unnecessary code?

src/libLookup/Lookup.cpp
```diff
@@ -2455,6 +2455,34 @@ void Lookup::StartSynchronization() {
...
+    GetDSInfoFromLookupNodes();
+    this_thread::sleep_for(chrono::seconds(NEW_NODE_SYNC_INTERVAL));
+    {
+      lock_guard<mutex> g(m_mediator.m_mutexDSCommittee);
+      if (m_mediator.m_DSCommittee->size() > 0) {
+        return true;
+      }
+    }
+    counter++;
+  }
```
ckyang: How about using cv instead of polling asking? It should be more efficient. You can refer to this: https://github.com/Zilliqa/Zilliqa/blob/master/src/libNode/Node.cpp#L316

src/libDirectoryService/FinalBlockPreProcessing.cpp
KaustubhShamshery: this is from previous versions of VCTEST

src/libDirectoryService/DSBlockPreProcessing.cpp
KaustubhShamshery: same as above

src/libDirectoryService/ViewChangePreProcessing.cpp
KaustubhShamshery: actually I want JH to review this

ckyang: @bb111189 Could you please review on this to see if this piece of code necessary to keep?

src/libDirectoryService/FinalBlockPreProcessing.cpp
ckyang: Is it possible to give a meaningful name?

src/libDirectoryService/DirectoryService.cpp
KaustubhShamshery: if you see the function start sync it is there.

src/libDirectoryService/DirectoryService.cpp
KaustubhShamshery: ok

src/libData/BlockChainData/BlockChain.h
KaustubhShamshery: ok

src/libDirectoryService/ViewChangePreProcessing.cpp
```diff
@@ -278,9 +278,9 @@ void DirectoryService::RunConsensusOnViewChange() {
   if (dsCurBlockNum != 0 && txCurBlockNum != 0) {
     VCFetchLatestDSTxBlockFromLookupNodes();
     if (!NodeVCPrecheck()) {
-      LOG_GENERAL(
-          WARNING,
-          "Failed the vc precheck. Node is lagging behind the whole network.");
+      LOG_GENERAL(WARNING,
+                  "[RDS]Failed the vc precheck. Node is lagging behind the "
```
bb111189: What is RDS?

KaustubhShamshery: Recovery DS

src/libDirectoryService/DSBlockPreProcessing.cpp
bb111189: Why the vc test have been scope to so specific test?

KaustubhShamshery: I just wanted one node to die. I can revert it however

src/libDirectoryService/ViewChangePreProcessing.cpp
bb111189: Why remove this piece of code?

src/libData/BlockChainData/BlockChain.h
```diff
@@ -71,13 +71,14 @@ class BlockChain {
   T GetBlock(const uint64_t& blockNum) {
     std::lock_guard<std::mutex> g(m_mutexBlocks);
 
-    if (blockNum >= m_blocks.size()) {
-      LOG_GENERAL(WARNING, "Block number "
-                               << blockNum
-                               << " absent, a dummy block will be used and "
-                                  "abnormal behavior may happen!");
+    if (m_blocks.size() > 0 &&
+        (m_blocks.back().GetHeader().GetBlockNum() < blockNum)) {
```
kaikawaliu: As the vc block and fallback block doesn't have `GetBlockNum()`, maybe we can overidely declare the function `GetBlock` for that two specific block types as deleted or return a dummy value.

src/libLookup/Lookup.cpp
```diff
@@ -2455,6 +2455,43 @@ void Lookup::StartSynchronization() {
...
+  while (counter <= FETCH_LOOKUP_MSG_MAX_RETRY) {
+    GetDSInfoFromLookupNodes();
+    unique_lock<mutex> lk(m_mutexDSInfoUpdation);
+    if (cv_dsInfoUpdate.wait_for(lk, chrono::seconds(NEW_NODE_SYNC_INTERVAL)) ==
+        cv_status::timeout) {
+      counter++;
+
+    } else {
+      break;
+    }
```
ckyang: Maybe we don't need to include GetDSInfoFromLookupNodes() inside the waiting loop.
Besides, the counter and cv.wait_for could be merged into the same while loop.

src/libData/BlockChainData/BlockChain.h
kaikawaliu: But in another case, if we deleted that, will it affect the block fetching by using BlockIndexChain?

kaikawaliu: Or we can have a blockNum for vc or fb block incrementally, which I feels it the best solution for now

KaustubhShamshery: I dont think so this is in the scope of this PR

kaikawaliu: Yeah it's not for add block num. But firstly should avoid vc and fb call GetBlockNum in this PR. otherwise it's undefined behavior

KaustubhShamshery: VC/FB are not attached for now in the blockchain.h

kaikawaliu: VC/FBBlockChain

src/libLookup/Lookup.cpp
ckyang: After thinking more, we don't need to modify this piece of code.

PR #879, opened 2018-11-16
Add stimulation delay option to stimulate network lag for gossip

sandipbhoir:
## Description
This PR adds stimulation delay option before sending message to stimulate the network lag if required.
Default is no lag.
Additionally, added shard-id info for log-state of TXNPROC.

## Review Suggestion
Please check files changed.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libNetwork/RumorManager.cpp
```diff
@@ -491,6 +491,8 @@ void RumorManager::SendMessage(const Peer& toPeer,
   }
 
   // Send the message to peer .
+  std::this_thread::sleep_for(
```
ShengguangXiao: Maybe check if (SIMULATED_NETWORK_DELAY_IN_MS == 0) then do a sleep_for, otherwise sleep 0 will still do the thread context switch.

PR #884, opened 2018-11-16
libPOW coverage improvement (mining with full dataset)

Jazzz42:
## Description
Improvement of libPOW coverage with test of mining with fullDataset

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/POW/test_POW.cpp
+++ b/tests/POW/test_POW.cpp
@@ -711,17 +711,17 @@ BOOST_AUTO_TEST_CASE(mining_and_verification) {
   uint8_t difficultyToUse = 10;
   uint8_t blockToUse = 0;
   ethash_mining_result_t winning_result = POWClient.PoWMine(
-      blockToUse, difficultyToUse, rand1, rand2, ipAddr, pubKey, false, 0, 0);
+      blockToUse, difficultyToUse, rand1, rand2, ipAddr, pubKey, 0, 0, false);
   bool verifyLight =
       POWClient.PoWVerify(blockToUse, difficultyToUse, rand1, rand2, ipAddr,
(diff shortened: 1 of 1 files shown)


Jazzz42:
The test run has frozen the same way as it freezes on local machine when <OPENCL_GPU_MINE>true</OPENCL_GPU_MINE> is set

tests/POW/test_POW.cpp
```diff
@@ -740,6 +740,38 @@ BOOST_AUTO_TEST_CASE(mining_and_verification) {
       blockToUse, difficultyToUse, rand1, rand2, ipAddr, pubKey, 0, 0, false,
       winning_nonce, winning_result.result, winning_result.mix_hash);
   BOOST_REQUIRE(!verifyWinningNonce);
+
+  // Full Dataset mine and verify
+  winning_result = POWClient.PoWMine(blockToUse, difficultyToUse, rand1, rand2,
```
ShengguangXiao: Please change it to  winning_result = POWClient.PoWMine(blockToUse, difficultyToUse, rand1, rand2,
                                     ipAddr, pubKey, 0, 0, true);
The last parameter is fullDataset mining

Jazzz42: Thanks for pointing this out. After doublechecking it seems, that this needs to be corrected also in multiple PoWVerify calls in this test suite. I fixed that in commit dec3bdfe0772ad780d86a8a2ccf4c514a072aa56 but when enabling the OPENCL I still get stalled test. Does this work on your machine? I'm not sure if this problem is caused by my graphics card driver or not.

ShengguangXiao: Yes, the others i will handle in my pull request. You only need to fix this one.

PR #888, opened 2018-11-19
Add partialRetrival flag for RetrieveDSBlocks

kaikawaliu:
## Description

Add flag for retrieving microblocks partially.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -782,7 +782,8 @@ void Lookup::RetrieveDSBlocks(vector<DSBlock>& dsBlocks, uint64_t& lowBlockNum,
     lowBlockNum = 1;
```
KaustubhShamshery: lowBlockNum = 1 and partialRetrieve = true, should return only the latest block ?

PR #890, opened 2018-11-19
Verify the finalblock node before forcing the state change

kaikawaliu:
## Description

Normally, once a shard node received a finalblock, it should skip the microblock consensus state and start processing the finalblock. But this mechanism didn't cover if the node received duplicate or malicious finalblock. So we will let the verification of the finalblock being done first before triggering the state change.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -643,32 +643,6 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
                              [[gnu::unused]] const Peer& from) {
   LOG_MARKER();
 
-  if (!LOOKUP_NODE_MODE) {
-    if (m_lastMicroBlockCoSig.first != m_mediator.m_currentEpochNum) {
-      std::unique_lock<mutex> cv_lk(m_MutexCVFBWaitMB);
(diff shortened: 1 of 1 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -643,38 +643,14 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
...
-      << m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() + 1
-      << "] RECEIVED FINAL BLOCK");
-
   uint32_t shardId = std::numeric_limits<uint32_t>::max();
   uint64_t dsBlockNumber = 0;
   uint32_t consensusID = 0;
   TxBlock txBlock;
   vector<unsigned char> stateDelta;
 
+  lock_guard<mutex> g(m_mutexFinalBlock);
```
ansnunez: Can we move obtaining the mutex to after doing `GetNodeFinalBlock`?

kaikawaliu: yeah, make sense

PR #896, opened 2018-11-19
 Wrap operation about m_createdTransaction with mutex in some corner case

kaikawaliu:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -1786,6 +1786,7 @@ bool Lookup::ProcessSetStateFromSeed(const vector<unsigned char>& message,
 
   if (!AccountStore::GetInstance().Deserialize(accountStoreBytes, 0)) {
     LOG_GENERAL(WARNING, "Deserialize AccountStore Failed");
+    return false;
   }
 
   if (ARCHIVAL_NODE) {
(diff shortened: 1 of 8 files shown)


src/libData/AccountData/AccountStore.cpp
```diff
@@ -53,7 +53,7 @@ void AccountStore::Init() {
 void AccountStore::InitSoft() {
   LOG_MARKER();
 
-  lock_guard<mutex> g(m_mutexPrimary);
+  unique_lock<shared_timed_mutex> g(m_mutexPrimary);
```
ShengguangXiao: Why change the mutex to shared_timed_mutex?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -85,11 +85,20 @@ AccountStore& AccountStore::GetInstance() {
   return accountstore;
 }
 
+bool AccountStore::Serialize(vector<unsigned char>& src,
+                             unsigned int offset) const {
+  LOG_MARKER();
+
+  shared_lock<shared_timed_mutex> lock(m_mutexPrimary);
```
ShengguangXiao: Why other places use unique_lock but here use shared_lock for m_mutexPrimary?

kaikawaliu: Because this is the only function we allow multithread accessing concurrently.

PR #902, opened 2018-11-20
Guard mode

bb111189:
## Description

Under development. TBD

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


KaustubhShamshery:
Have to change the leader calculation FinishRejoinAsDS too.

src/libZilliqa/Zilliqa.cpp
```diff
@@ -109,6 +109,10 @@ Zilliqa::Zilliqa(const std::pair<PrivKey, PubKey>& key, const Peer& peer,
 
 {
   LOG_MARKER();
+
+  // Setting the guard upon process launch
+  Guard::GetInstance().Init();
```
KaustubhShamshery: if (GUARD MODE) ??

bb111189: Have implemented the fix within init() function

bb111189:
> Have to change the leader calculation FinishRejoinAsDS too.

Implemented

PR #912, opened 2018-11-21
Add TxnRootHash for per MicroBlockInfo in TxBlock

kaikawaliu:
## Description

Add TxnRootHash in MicroBlockInfo attached to TxBlock.
So that the forwarded transactions received by lookup from each shard can get verified

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libUtils/RootComputation.cpp
```diff
@@ -87,13 +91,21 @@ TxnHash ComputeRoot(const list<Transaction>& receivedTransactions,
                     const list<Transaction>& submittedTransactions) {
   LOG_MARKER();
 
+  if (receivedTransactions.empty() && submittedTransactions.empty()) {
```
ansnunez: I think we can just make `ConcatTranAndHash` do the check for all the inputs if they are empty?  That way the risk is lower of us failing to check if one of the containers is empty.

PR #914, opened 2018-11-21
Fix lookup dead loop when node retrieve blocks before bootstrap

ShengguangXiao:
## Description

Fix lookup dead loop when node retrieve blocks before bootstrap

Check if the block chain is initilized before proceesing the request.

https://github.com/Zilliqa/Issues/issues/327

## Review Suggestion

Run a 200 nodes cloud test to check if have lookup dead issue.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -772,6 +772,13 @@ void Lookup::RetrieveDSBlocks(vector<DSBlock>& dsBlocks, uint64_t& lowBlockNum,
 
   uint64_t curBlockNum =
       m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum();
+
+  if (INIT_BLOCK_NUMBER == curBlockNum) {
+    LOG_GENERAL(WARNING,
+                "Blockchain is still bootstraping, no ds blocks available.");
+    return;
+  }
+
   uint64_t minBlockNum = (curBlockNum > MEAN_GAS_PRICE_DS_NUM)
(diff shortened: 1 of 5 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -772,6 +772,13 @@ void Lookup::RetrieveDSBlocks(vector<DSBlock>& dsBlocks, uint64_t& lowBlockNum,
 
   uint64_t curBlockNum =
       m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum();
+
+  if (INIT_BLOCK_NUMBER == curBlockNum) {
+    LOG_GENERAL(INFO,
```
ckyang: Suggest to use WARNING for this kind of error that would lead to abnormal behavior.

src/libLookup/Lookup.cpp
```diff
@@ -918,6 +925,12 @@ void Lookup::RetrieveTxBlocks(vector<TxBlock>& txBlocks, uint64_t& lowBlockNum,
         m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum();
   }
 
+  if (INIT_BLOCK_NUMBER == highBlockNum) {
+    LOG_GENERAL(INFO,
```
ckyang: Suggest to use WARNING for this kind of error that would lead to abnormal behavior.

PR #917, opened 2018-11-21
Test/lib pow

Jazzz42:
## Description
I added a very simple base for parameter range test of difficulty adjustment algorithm.
I plan to improve the test before merging. The goal of this PR is to get feedback on which variables should be tested and how. The percentage used in test was derived from existing test description, but may be changed upon request.

You can copy the code and run it on your machine to inspect the output csv. It looks pretty predictable, so I assume we should try to cover various POW submission difference percentages besides -20 , -5 , 0 , +5 , +20
The question is, if that is enough, or also other input values should be changed and what is your idea about test scope. For example I would expect less linear behavior of the difficulty adjustment algorithm.

### Implementation
This PR is still under development.

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ShengguangXiao:
@Jazzz42 Can you rebase all the commits into one commit and fix the conflicts first?

ShengguangXiao:
Staled pull request

PR #925, opened 2018-11-22
Cleanup hashes during re-initialization

sandipbhoir:
## Description
Fix: Cleanup the gossip hash-rawmessage map during re-initialization.

## Review Suggestion

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNetwork/RumorManager.cpp
+++ b/src/libNetwork/RumorManager.cpp
@@ -44,6 +44,7 @@ RumorManager::RumorManager()
     : m_peerIdPeerBimap(),
       m_peerIdSet(),
       m_rumorIdHashBimap(),
+      m_rumorHashRawMsgBimap(),
       m_tmpRumorHashSet(),
       m_selfPeer(),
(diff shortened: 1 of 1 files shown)


src/libNetwork/RumorManager.cpp
```diff
@@ -134,6 +135,7 @@ bool RumorManager::Initialize(const std::vector<Peer>& peers,
   m_peerIdSet.clear();
   m_selfPeer = myself;
   m_tmpRumorHashSet.clear();
+  m_rumorHashRawMsgBimap.clear();
```
ShengguangXiao: The default constructor should initialize empty bitmap already, why here need clean again?

sandipbhoir: the initialize function is called every ds epoch.

ShengguangXiao: Ok, i see, i thought the clear is in constructor function.

PR #926, opened 2018-11-22
Don't init full mine context when use GPU to mine

ShengguangXiao:
## Description

Don't init full mine context when use GPU to mine, because it will take 1GB RAM but didn't been used.

Check if using GPU to do mining, if yes, then don't initialize full mine context.

## Review Suggestion

Check libPOW.cpp

Run ./build/tests/POW/Test_POW

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -41,7 +41,7 @@ POW::POW() {
   m_epochContextLight =
       ethash::create_epoch_context(ethash::get_epoch_number(m_currentBlockNum));
 
-  if (FULL_DATASET_MINE) {
+  if (FULL_DATASET_MINE && !CUDA_GPU_MINE && !OPENCL_GPU_MINE) {
     m_epochContextFull = ethash::create_epoch_context_full(
         ethash::get_epoch_number(m_currentBlockNum));
   }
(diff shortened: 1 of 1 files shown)


deepgully:
When I try GPU mining on my 1080ti, it takes 3-4s to finish the 1st POW, this caused I lost the POW and can not join the testnet. 

The ethash DAG memory for GPU was generated at the 1st mine call, 
it spent to much time I think, am I right? 
https://github.com/Zilliqa/Zilliqa/blob/4719859b36ccf1c2e8400dfafd7b6768d613d698/src/depends/libethash-cuda/CUDAMiner.cpp#L434-L443

If the answer is yes, could you init GPU DAG earlier? otherwise GPU miner will waste at least 1 epoch time.

If the answer is no, could you help to investigate which part is slow.

@ShengguangXiao

ShengguangXiao:
@deepgully Hi deepgully, yes, this init will takes 3~4 seconds to finish for the first time. But our PoW window time is 150 seconds, so this part should not affect your GPU to finish the PoW.

ShengguangXiao:
@deepgully It is not depends on which PoW comes first. It depends on submitted result. The smaller one can join first.
And the difficulty not base on average PoW time, it depends on how many nodes want to join the network.

deepgully:
@ShengguangXiao 
got it, thx for explain.

deepgully:
@ShengguangXiao  
one thought fyi: 
If I change code to do PoW  1) as long as possible with-in PoW window 2) use higher difficulty to do PoW, does that means I have more possibility to beat other miners who are using official software?

ShengguangXiao:
@deepgully Yes, theoratically this way can make you join netwok faster. But it will not get reward for do PoW, reward is based on how many transactions the node has signed.

deepgully:
yes, in theory, I can manipulate DS committee with my malicious nodes.

deepgully:
I just join DS committee at epoch 8100 with this change. @ShengguangXiao 

> [INFO   ][19848][02:35:41:534][ProcessVCDSBlocksMessage      ][Epoch 8100] I won DS PoW. Currently, one of the new ds committee member with id 204

ShengguangXiao:
Yes, currently the mining difficulty is too low, so can easily finish the PoW. We will increase difficulty very soon.

PR #933, opened 2018-11-25
Test/data

jendis:
## Description
Increase coverage of libData/AccountData/Account.

## Status
Lines: 93.3%
Functions: 100%
Reaching 100% of line coverage in reasonable time would require mockig of several methods of Messenger and RLP class.
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/Data/Test_Account.cpp
+++ b/tests/Data/Test_Account.cpp
@@ -43,8 +43,7 @@ BOOST_AUTO_TEST_CASE(testInitEmpty) {
   Account acc1(TestUtils::DistUint64(), 0);
   std::vector<unsigned char> data;
   acc1.InitContract(data);
-  acc1.SetInitData(data);  // Satisfy coverage (coverage not taken into account
-                           // when called implicitly from InitContract)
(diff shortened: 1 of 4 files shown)


tests/Data/Test_Account.cpp
```diff
@@ -43,33 +43,104 @@ BOOST_AUTO_TEST_CASE(testInitEmpty) {
...
+  acc1.InitStorage();  // Improve coverage
 }
 
 BOOST_AUTO_TEST_CASE(testBalance) {
   INIT_STDOUT_LOGGER();
   LOG_MARKER();
 
-  uint128_t balance = TestUtils::DistUint32();
-  Account acc1(balance, 0);
+  uint128_t BALANCE = TestUtils::DistUint32();
```
Jazzz42: Why the capital letters when its not a constant? It is just a formal thing but it seems to me, that all capital names are usually used by constants in Zillia code. This comment is valid for multiple places in this file.

PR #937, opened 2018-11-26
Fix new node failed to rejoin problem

ShengguangXiao:
## Description

Fix new node failed to rejoin problem.

1. Fix when boostrap failed to join testnet, and start to rejoin with retrieve history, but no history to retrieve and didn't add genesis block problem.
2. Fix duplicated rejoin procedure when in StartPoW function.

Run a local 20 nodes test, set COMM_SIZE to 8.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/PoWProcessing.cpp
+++ b/src/libNode/PoWProcessing.cpp
@@ -107,36 +107,7 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
   LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
             "Current dsblock is " << block_num);
 
-  bool isNoSync = false;
-
-  if (m_mediator.m_lookup->m_syncType == SyncType::NO_SYNC) {
-    m_stillMiningPrimary = true;
-    isNoSync = true;
-
-    auto func = [this]() mutable -> void {
-      this_thread::sleep_for(chrono::seconds(NEW_NODE_SYNC_INTERVAL +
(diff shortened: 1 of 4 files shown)


ShengguangXiao:
Local test with some node always try to rejoin can run to Tx blocks 175.

src/libNode/PoWProcessing.cpp
```diff
@@ -107,36 +107,7 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
...
-              " rejoin");
-          RejoinAsNormal();
-        } else {
-          LOG_GENERAL(INFO, "Didn't get the latest DSBlock, what to do???");
-        }
-      }
-    };
-
-    DetachedFunction(1, func);
-  }
```
kaikawaliu: Why to remove this part? Is it causing problem?

ShengguangXiao: Yes, it will also cause rejoin, but now if wait for DS block timeout then trigger rejoin already.

KaustubhShamshery: This check here is for different purpose, if node is still mining for a long duration. The other thread only activates when node has finished mining.

ShengguangXiao: Then we should check the mining time, if the mining takes too long, then should stop mining. In my local test case, this piece of code make node always fetching latest Tx blocks, then it need another fix.

PR #938, opened 2018-11-27
LookupArchival to send txns to Lookups

KaustubhShamshery:
## Description

The PR adds two other type of lookup nodes (upper_seed and lower_seed). The difference between such nodes is that nodes will now not send directly send data to lookups, rather they will operate a lower level seed (controlled by client) which will forward txns to upper_level seed(controlled by us) and then to the lookup.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


constants_local.xml
```diff
@@ -188,6 +192,8 @@
             <pubkey/>
         </peer>
     </lookups>
+    <upper_seed/>
```
ShengguangXiao: What does upper_seed and lower_seed used for? Can put together with <seed>?

KaustubhShamshery: The idea behind separating them is there because the lower_seed will send transactions to upper_seed.

src/libLookup/Lookup.cpp
```diff
@@ -3232,6 +3309,50 @@ bool Lookup::VerifyLookupNode(const VectorOfLookupNode& vecLookupNodes,
...
+      lock_guard<mutex> g(m_mediator.m_ds->m_mutexShards);
+      shard_size = m_mediator.m_ds->m_shards.size();
+    }
+
+    if (shard_size == 0) {
+      LOG_GENERAL(WARNING, "Shard size 0");
+      return false;
+    }
+
+    for (auto txn : txns) {
```
ShengguangXiao: Use for (const auto& txn : txns)?

constants.xml
```diff
@@ -11,6 +11,10 @@
         <ARCHIVAL_NODE>false</ARCHIVAL_NODE>
         <DB_HOST/>
     </archival>
+    <seed>
+        <ARCHIVAL_LOOKUP>false</ARCHIVAL_LOOKUP>
+        <SEED_TXN_COLLECTION_TIME>5</SEED_TXN_COLLECTION_TIME>
```
ShengguangXiao: Add time unit in name to easy understand

constants.xml
```diff
@@ -210,6 +214,8 @@
         </peer>
     -->
     </lookups>
+    <upper_seed/>
+    <lower_seed/>
```
ckyang: Supposedly, the naming rule inside constant file should be \<sectionName\>\<TAG1\>...\</TAG1\>\<TAG2/\>...\</sectionName\>, is it possible to make it consistent?

constants_local.xml
```diff
@@ -188,6 +192,8 @@
...
+    <lower_seed/>
```
ckyang: Supposedly, the naming rule inside constant file should be \<sectionName\>\<TAG1\>...\</TAG1\>\<TAG2/\>...\</sectionName\>, is it possible to make it consistent?

src/libLookup/Lookup.cpp
```diff
@@ -144,12 +232,13 @@ bool Lookup::GenTxnToSend(size_t num_txn,
 
   unsigned int NUM_TXN_TO_DS = num_txn / GENESIS_WALLETS.size();
 
+  if (numShards == 0) {
+    return false;
```
ckyang: Is it possible to give a warning message before returning false, to let user know what happened?

constants.xml
KaustubhShamshery: these would contain Ips. as you can see lookup ip is also outside every scope

PR #939, opened 2018-11-27
Provide option to disable responding with message for request:LAZY_PUSH

sandipbhoir:
## Description
Currently, given receives two types of incoming gossip messages: EMPTY_PUSH and LAZY_PUSH

1) EMPTY_PUSH is received when sender don't have any new message and want receiver to respond with new messages if any.

2) LAZY_PUSH is received when sender had some active message and have sent the same. 
Receiver on receiving this new message, sends the messages he have back to receiver as kindness.
This option exploites the number of messages that flows when there are many different gossip messages in network.
This PR is about making this FEATURE  (2) as optional through flag : SEND_RESPONSE_FOR_LAZY_PUSH which defaulted to true.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


constants.xml
```diff
@@ -91,6 +91,7 @@
         <GET_INITIAL_DS_FROM_REPO>false</GET_INITIAL_DS_FROM_REPO>
         <UPGRADE_HOST_ACCOUNT>Zilliqa</UPGRADE_HOST_ACCOUNT>
         <UPGRADE_HOST_REPO>Zilliqa</UPGRADE_HOST_REPO>
+        <SEND_RESPONSE_FOR_LAZY_PUSH>true</SEND_RESPONSE_FOR_LAZY_PUSH>
```
ckyang: Seems more and more gossip-related options in constant file. Is it possible to group them?

sandipbhoir: created issue - https://github.com/Zilliqa/Issues/issues/339

PR #943, opened 2018-11-27
Shard guard enhancements

bb111189:
## Description

TBD

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libNetwork/Guard.cpp
```diff
@@ -213,11 +213,23 @@ void Guard::AddToExclusionList(const uint128_t& ft, const uint128_t& sd) {
   }
 }
 
+void Guard::ValidateRunTimeEnvironment() {
+  unsigned int nodeReplacementLimit = COMM_SIZE - (COMM_SIZE - (COMM_SIZE / 3));
```
ShengguangXiao: Why not just use nodeReplacementLimit = COMM_SIZE / 3, the result is the same.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -658,6 +680,17 @@ VectorOfPoWSoln DirectoryService::SortPoWSoln(const MapOfPubKeyPoW& mapOfPoWs,
       // 5. Finally, sort "FilteredPoWOrderSorter" and stored result in
       // "PoWOrderSorter"
 
+      unsigned int trimmedGuardCount = numNodesTrimmed - (numNodesTrimmed / 3);
```
ansnunez: Should we perhaps use the same formula used in `ConsensusCommon::NumForConsensus`?

src/libNetwork/Guard.cpp
bb111189: Miror the logic from the re balancing side. Not sure whether replacing with `COMM_SIZE / 3` will have a corner case or not. What do you think?

PR #946, opened 2018-11-27
Refactor sync type and stop mining if timeout

ShengguangXiao:
## Description

Refactor sync type and stop mining if wait for DS block timeout

Refactor sync type and stop mining if wait for DS block timeout

Fix node started rejoin but still mining problem.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


ShengguangXiao:
Tested with 200 nodes test, 60 DS nodes, shard size 60. 20 nodes try to rejoin, run to Tx blocks 170 without problem. Even in the middle the shard nodes trimmed to 60 nodes, later other nodes still able to rejoin to form 2 shards which is 120 nodes.

PR #949, opened 2018-11-27
Move out the responsibilty of gossip spreading txn packet from ProcessTxnPacketFromLookupCore

sandipbhoir:
## Description

This PR cover following changes:

- Issue : https://github.com/Zilliqa/Issues/issues/340

- Additionally, includes buffering of rumor when rumor-manager is not running and node tried to initiate a rumor. And once rumor-manager is initialized, these buffered messages will be initiated for gossip. This way node won't miss out to initiate gossiping of txn-packet within shard in-case round is not running in this cooling period.

- Fix the timing issue with starting of first txn epoch (every ds epoch) with delay X to allow other nodes to finish receving DS block consensus - final cosig. This also implies that sendign ds block to lookup and other shard nodes is also delayed by X .
- Fix the timing issue with lookup sending txn packet every first txn epoch ( every ds epoch) with having delay of Y to allow all nodes in network (non-ds shards) to receive ds block , update their sharding structure and re-start their rumor manager.
- Y = X + (time expected for all nodes to receive the ds block and re-initialize the rumor managerwith latest peerlist)
  X : DELAY_FIRSTXNEPOCH_IN_MS
  Y : LOOKUP_DELAY_SEND_TXNPACKET_IN_MS

## Review Suggestion

- Run small scale test (200 nodes) to see the stability.
- Check the files changed,

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -3116,6 +3116,16 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
     // return;
   }
 
+  // allow receving nodes to be ready with latest DS block ( Only** for first
+  // txn epoch of every ds epoch )
+  if ((m_mediator.m_currentEpochNum % NUM_FINAL_BLOCK_PER_POW == 1)) {
```
ShengguangXiao: This check may have problem, for epoch 1 is equal to 1 and if it is larger than 1, then should check (m_mediator.m_currentEpochNum % NUM_FINAL_BLOCK_PER_POW == 0)

src/libNetwork/RumorManager.cpp
```diff
@@ -158,20 +155,33 @@ bool RumorManager::Initialize(const std::vector<Peer>& peers,
   return true;
 }
 
+void RumorManager::SpreadBufferedRumors() {
+  LOG_MARKER();
+  if (m_continueRound) {
+    for (auto& i : m_bufferRawMsg) {
```
ShengguangXiao: can add a const in front of auto &i

PR #952, opened 2018-11-28
Enable sender thread on every lookup

KaustubhShamshery:
### Implementation

- [x] **ready for review**
-

- [x] local machine test

- [x] small-scale cloud test with two lookup nodes and 100 txns sent from genesis account via RPC

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3131,7 +3131,7 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
 
       LOG_GENERAL(INFO, "Transaction number generated: " << transactionNumber);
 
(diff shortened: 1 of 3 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3127,6 +3131,11 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
 
       LOG_GENERAL(INFO, "Transaction number generated: " << transactionNumber);
 
+      if (m_txnShardMap[i].size() == 0 && mp[i].size() == 0) {
```
ShengguangXiao: Can use m_txnShardMap[i].empty() instead of m_txnShardMap[i].size() == 0, it is more efficient

PR #953, opened 2018-11-28
Implementation of DataSender to send data based on the index in cosignatures 

kaikawaliu:
## Description

This PR is primarily to solve:
https://github.com/Zilliqa/Issues/issues/334

## Review Suggestion

It's quite a mess to look through this PR via git differ. Better to:
1. Review libNetwork/DataSender.* first.
2. Follow the places where the DataSender::GetInstance().SendDataToOthers was called
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libNetwork/DataSender.cpp
```diff
@@ -0,0 +1,257 @@
...
+    } else {
+      vector<Peer> shard_peers;
+
+      for (const auto& kv : *p) {
+        shard_peers.emplace_back(std::get<SHARD_NODE_PEER>(kv));
+        LOG_GENERAL(
+            INFO,
+            " PubKey: " << DataConversion::SerializableToHexStr(
+                               std::get<SHARD_NODE_PUBKEY>(kv))
+                        << " IP: "
```
bb111189: This log can be refine to just print std::get<SHARD_NODE_PEER>(kv) and it will be printed out as <IP:Port>

src/libNetwork/DataSender.cpp
```diff
@@ -0,0 +1,257 @@
...
+    const DequeOfShard& shards, const VectorOfLookupNode& lookups,
+    const BlockHash& hashForRandom,
+    const ComposeMessageForSenderFunc& composeMessageForSenderFunc,
+    const SendDataToLookupFunc& sendDataToLookupFunc,
+    const SendDataToShardFunc& sendDataToShardFunc) {
+  if (LOOKUP_NODE_MODE) {
+    LOG_GENERAL(WARNING,
+                "DataSender::SendDataToOthers not expected "
+                "to be called from LookUp node.");
+    return true;
```
bb111189: return false here?

src/libNetwork/DataSender.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if (indexB2 >= nodeToSendToLookUpLo && indexB2 < nodeToSendToLookUpHi) {
+      LOG_GENERAL(INFO,
+                  "Part of the committeement (assigned) that will send the "
+                  "Data to the lookup nodes");
+      if (sendDataToLookupFunc) {
+        sendDataToLookupFunc(lookups, message);
+      }
+    }
+
+    unsigned int my_cluster_num;
```
ShengguangXiao: Initialize variable when define it. Can change to unsigned int my_cluster_num = 0

src/libNetwork/DataSender.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+SendDataToLookupFunc SendDataToLookupFuncDefault =
+    [](const VectorOfLookupNode& lookups,
+       const std::vector<unsigned char>& message) mutable -> void {
+  SendDataToLookupNodesDefault(lookups, message);
+};
+
+SendDataToShardFunc SendDataToShardFuncDefault =
+    [](const std::vector<unsigned char>& message, const DequeOfShard& shards,
```
ansnunez: It seems like `message` is unused for any of the actual instances.  Should we just remove `message` from the definition?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -155,6 +155,8 @@ bool Node::LoadUnavailableMicroBlockHashes(const TxBlock& finalBlock,
           if (m_microblock->GetBlockHash() == info.m_microBlockHash) {
             if (m_microblock->GetHeader().GetTxRootHash() != TxnHash()) {
               if (info.m_txnRootHash != TxnHash()) {
+                // Update transaction processed
+                UpdateProcessedTransactions();
```
ansnunez: This seems out of place in this PR.  Is it a fix for some issue?

kaikawaliu: Yes it is. The purpose is to let the shard node update the processed txns only when they found their microblock in the finalblock, while the adoption now is update when the microblock consensus done, but it could happen even the consensus was done, but the microblock may still not be inside of the finalblock.

src/libNetwork/DataSender.cpp
kaikawaliu: It's used when P2P Sending Message

PR #954, opened 2018-11-28
Disallow DS to submit PoW

bb111189:
## Description

Drop PoW submission if 
- Soln come from a DS member (IP and port check)
- Soln come from a DS pubkey

### Implementation

- [X] **ready for review**

- [X] local machine test

### DIFF
--- a/src/libDirectoryService/PoWProcessing.cpp
+++ b/src/libDirectoryService/PoWProcessing.cpp
@@ -209,7 +209,10 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
     return false;
   }
 
-  // Todo: Reject PoW submissions from existing members of DS committee
+  // Reject PoW submissions from existing members of DS committee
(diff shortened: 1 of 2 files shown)


src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -354,6 +357,29 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
   return result;
 }
 
+bool DirectoryService::CheckSolnFromNonDSCommittee(
+    const PubKey& submitterPubKey, const Peer& submitterPeer) {
+  lock_guard<mutex> g(m_mediator.m_mutexDSCommittee);
+
+  for (const auto& dsMember : *m_mediator.m_DSCommittee) {
+    // Reject soln if any of the following condition is true
+    if (dsMember.first == submitterPubKey) {
```
ShengguangXiao: Why need to check both PubKey and Peer? Check PubKey should be enough?

PR #955, opened 2018-11-28
Make proof of reputation can work again

ShengguangXiao:
## Description

Make proof of reputation can work again

Choose the top reputation/priority nodes and verify top reputation/priority nodes before do the PoW trimming.

https://github.com/Zilliqa/Issues/issues/341

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


ShengguangXiao:
Local test with 5 nodes late join run to tx blocks 110 without problem. And some nodes cannot join because priority not enough, until some nodes failed to do PoW, then it can join.

PR #964, opened 2018-11-29
Refactor the GetBuiltDSComm function

ShengguangXiao:
## Description

Refactor the GetBuiltDSComm function

The get function shouldn't modify the property of class. Change to use set function to update it. It will be easier for people to read and debug the code.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


ShengguangXiao:
Tested with 200 nodes test, 60 DS nodes, 120 shard nodes, 20 to squeeze in. With guard mode enabled, run to 110 Tx blocks without problem.

PR #967, opened 2018-11-29
VerifyTxOrder Fix

KaustubhShamshery:
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -400,6 +400,14 @@ void Node::ProcessTransactionWhenShardLeader() {
       break;
     }
   }
+  //Put txns in map back into pool
+  for(const auto& kv: t_addrNonceTxnMap)
```
ansnunez: Since we do `t_createdTxns = m_createdTxns;` at the start of the function, what is the purpose of updating `t_createdTxns` again here?

PR #968, opened 2018-11-30
Recover all nodes at the same time

ckyang:
## Description

In real world, there is some probability that whole testnet stuck. Then we should make following process to make it alive again:

1. Shut down the testnet
2. Copy database from lookup node to other nodes (DS and shard node)
3. Re-start the testnet, the testnet keep going from previous status.

This PR fix some internal bug of this scenario.

Follow-up:
Provide a testnet script for step (2) and (3).
Noted that the re-start sequence should be: lookup -> ds -> shard.

### Implementation

- [x] **ready for review**

- [x] local machine test

<del>- [ ] small-scale cloud test</del>

### DIFF
--- a/src/libDirectoryService/DirectoryService.cpp
+++ b/src/libDirectoryService/DirectoryService.cpp
@@ -472,7 +472,7 @@ void DirectoryService::RejoinAsDS() {
     auto func = [this]() mutable -> void {
       m_mediator.m_lookup->SetSyncType(SyncType::DS_SYNC);
       m_mediator.m_node->CleanVariables();
-      m_mediator.m_node->Install(SyncType::DS_SYNC, true);
+      m_mediator.m_node->Install(SyncType::DS_SYNC);
       this->StartSynchronization();
     };
     DetachedFunction(1, func);
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -127,14 +127,6 @@ bool Node::Install(SyncType syncType, bool toRetrieveHistory) {
         if (m_mediator.m_DSCommittee->at(m_mediator.m_ds->m_consensusLeaderID)
                 .first == m_mediator.m_selfKey.second) {
           m_mediator.m_ds->m_mode = DirectoryService::PRIMARY_DS;
-
-          if (!wakeupForUpgrade) {
-            LOG_GENERAL(INFO,
-                        "Node recovery cannot be applied on DS leader, apply "
-                        "re-join process instead");
-            return false;
-          }
-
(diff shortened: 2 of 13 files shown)


src/libDirectoryService/DirectoryService.cpp
```diff
@@ -472,7 +472,7 @@ void DirectoryService::RejoinAsDS() {
     auto func = [this]() mutable -> void {
       m_mediator.m_lookup->SetSyncType(SyncType::DS_SYNC);
       m_mediator.m_node->CleanVariables();
-      m_mediator.m_node->Install(SyncType::DS_SYNC, true);
+      m_mediator.m_node->Install(SyncType::DS_SYNC);
```
ansnunez: I think all our `Install` calls are with `toRetrieveHistory = true`.
Do we still have a case where we might need `toRetrieveHistory = false`?
If not, maybe we can remove this parameter.

ckyang: Probably not, currently the bootstrap command in AWS/local will execute "zilliqa xxx 1 0 0", that is, Node::Install(x, false) is used.

src/libPersistence/BlockStorage.cpp
```diff
@@ -580,20 +580,10 @@ bool BlockStorage::PutShardStructure(const DequeOfShard& shards,
...
-  string strMyshardId = m_shardStructureDB->Lookup(index++);
-
-  if (strMyshardId.empty()) {
-    LOG_GENERAL(WARNING, "Cannot retrieve sharding structure!");
-    return false;
-  }
-
-  myshardId = stoul(strMyshardId);
-  LOG_GENERAL(INFO, "Retrieved shard ID: " << myshardId);
+  unsigned int index = 1;
```
KaustubhShamshery: Why use index, just protobuf the entire structure

ckyang: It's a good idea however considering the backward compatible, we should not change current database format but only can make some extension.

src/libNode/Node.cpp
```diff
@@ -568,6 +586,7 @@ void Node::WakeupForRecovery() {
     m_mediator.m_ds->SetState(
         DirectoryService::DirState::MICROBLOCK_SUBMISSION);
     auto func = [this]() mutable -> void {
+      this_thread::sleep_for(chrono::seconds(1));
```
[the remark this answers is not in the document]
ckyang: Actually the delay means waiting for other nodes' recovery, anyway I can make a meaningful word instead of magic number.

PR #969, opened 2018-11-30
Lookup send to leader

KaustubhShamshery:
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3172,6 +3172,16 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
 
           LOG_GENERAL(INFO, "Sent to node " << get<SHARD_NODE_PEER>(*it));
         }
+        if (m_mediator.m_ds->m_shards.at(i).empty()) {
+          continue;
+        }
+        uint16_t lastBlockHash = DataConversion::charArrTo16Bits(
+            m_mediator.m_txBlockChain.GetLastBlock().GetBlockHash().asBytes());
+        uint32_t leader_id =
+            lastBlockHash % m_mediator.m_ds->m_shards.at(i).size();
+        toSend.push_back(get<SHARD_NODE_PEER>(
+            m_mediator.m_ds->m_shards.at(i).at(leader_id)));
+        LOG_GENERAL(INFO, "leader id " << leader_id);
       }
 
       if (BROADCAST_GOSSIP_MODE) {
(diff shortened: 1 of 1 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3172,6 +3172,16 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
...
+          continue;
+        }
+        uint16_t lastBlockHash = DataConversion::charArrTo16Bits(
+            m_mediator.m_txBlockChain.GetLastBlock().GetBlockHash().asBytes());
+        uint32_t leader_id =
+            lastBlockHash % m_mediator.m_ds->m_shards.at(i).size();
+        toSend.push_back(get<SHARD_NODE_PEER>(
+            m_mediator.m_ds->m_shards.at(i).at(leader_id)));
+        LOG_GENERAL(INFO, "leader id " << leader_id);
       }
```
sandipbhoir: May be u need to change j < NUM_NODES_TO_SEND_LOOKUP --> j < NUM_NODES_TO_SEND_LOOKUP -1,
otherwise we end up sending to NUM_NODES_TO_SEND_LOOKUP + 1 nodes

src/libLookup/Lookup.cpp
```diff
@@ -3192,6 +3202,11 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
              j++, it++) {
           toSend.push_back(it->second);
         }
+        uint16_t lastBlockHash = DataConversion::charArrTo16Bits(
+            m_mediator.m_dsBlockChain.GetLastBlock().GetBlockHash().asBytes());
+        uint32_t leader_id = lastBlockHash % m_mediator.m_ds->m_shards.size();
+        toSend.push_back(m_mediator.m_DSCommittee->at(leader_id).second);
```
ckyang: It seems some problem here, put shard leader ID into ds committee?

src/libLookup/Lookup.cpp
```diff
@@ -3192,6 +3202,11 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
...
+        LOG_GENERAL(INFO, "ds leader id " << leader_id);
```
sandipbhoir: similarly, over here..

src/libLookup/Lookup.cpp
KaustubhShamshery: that is usually a low number like 3 right, anyways

PR #971, opened 2018-11-30
Add Tolerance to txn ordering

sandipbhoir:
## Description
This PR :
Adds Tolerance to mismatch in transaction ordering.

## Review Suggestion

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -592,28 +592,74 @@ bool Node::VerifyTxnsOrdering(const vector<TxnHash>& tranHashes) {
...
+  uint32_t misordertxns = 0;
+  auto leftIt = t_tranHashes.begin();
+  auto rightIt = tranHashes.begin();
+  while (leftIt != t_tranHashes.end() && rightIt != tranHashes.end()) {
+    if (*leftIt == *rightIt) {
+      orderedtxns++;
+      leftIt++;
+      rightIt++;
+    } else {
+      break;
```
bb111189: I think @jiayaoqijia mentioned that this algorithm is not correct and that this is a Knapsack problem. Hence, we need to implement the algorithm to handle it

jiayaoqijia: @bb111189 We can apply this simple and strict algo first and later we further optimize it.

PR #975, opened 2018-11-30
use hash for random

KaustubhShamshery:
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


ShengguangXiao:
@KaustubhShamshery Can we refactor all these into a function? The same code copied a lot of times

PR #976, opened 2018-11-30
Fix PoW send

KaustubhShamshery:
## Description
Send Pow submission to DS leader.
DS leader accepts the POW submission.

~~Blocker observed is that leader calculation of DS and normal node results in different leader id.~~

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/PoWProcessing.cpp
+++ b/src/libNode/PoWProcessing.cpp
@@ -293,7 +293,10 @@ bool Node::SendPoWResultToDSComm(const uint64_t& block_num,
     const auto& blocktype = get<BlockLinkIndex::BLOCKTYPE>(bl);
     if (blocktype == BlockType::DS) {
       uint16_t lastBlockHash = DataConversion::charArrTo16Bits(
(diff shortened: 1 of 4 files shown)


src/libNode/PoWProcessing.cpp
```diff
@@ -293,10 +293,7 @@ bool Node::SendPoWResultToDSComm(const uint64_t& block_num,
     const auto& blocktype = get<BlockLinkIndex::BLOCKTYPE>(bl);
     if (blocktype == BlockType::DS) {
       uint16_t lastBlockHash = DataConversion::charArrTo16Bits(
-          m_mediator.m_dsBlockChain.GetLastBlock()
-              .GetHeader()
-              .GetHashForRandom()
-              .asBytes());
+          m_mediator.m_dsBlockChain.GetLastBlock().GetBlockHash().asBytes());
```
KaustubhShamshery: This is the issue, kindly revert this part.

sandipbhoir: yes, will revert. But even that had not helped.

PR #986, opened 2018-12-03
Adjust DS Block wait time for shard node after PoW submission

ansnunez:
## Description

This PR puts in a more accurate DS Block waiting time for a shard node before attempting rejoin.
`TX_DISTRIBUTE_TIME_IN_MS` should not be used as it has nothing to do at this point in the protocol.
Also, `FALLBACK_EXTRA_TIME` has been set to 0 inside the constants, and later on we will completely remove all fallback-related code.

**Update:**
Initially, we set this time to account for fallback and multiple view changes (i.e., abnormal execution).
However, this proved difficult to manage in tests, as accounting for too many abnormal execution timeouts can easily cause our overall waiting time to overflow beyond one DS epoch, especially in local testnet runs.
Timeout value should be suited for normal operation more than abnormal cases.

### Implementation

- [x] **ready for review**

- [x] local machine test (commit: 2ab083e)

- [ ] small-scale cloud test


src/libNode/PoWProcessing.cpp
```diff
@@ -148,11 +148,12 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
...
-              lk, chrono::seconds(
-                      NEW_NODE_SYNC_INTERVAL + POW_WINDOW_IN_SECONDS +
-                      POWPACKETSUBMISSION_WINDOW_IN_SECONDS +
-                      FALLBACK_EXTRA_TIME + TX_DISTRIBUTE_TIME_IN_MS / 1000)) ==
-          cv_status::timeout) {
+              lk,
+              chrono::seconds(NEW_NODE_SYNC_INTERVAL + POW_WINDOW_IN_SECONDS +
+                              POWPACKETSUBMISSION_WINDOW_IN_SECONDS +
+                              FALLBACK_EXTRA_TIME + CONSENSUS_OBJECT_TIMEOUT +
+                              DELAY_FIRSTXNEPOCH_IN_MS / 1000 +
```
ShengguangXiao: DELAY_FIRSTXNEPOCH_IN_MS this one should added here only for the 1st epoch?

sandipbhoir:
@antonio, so DSBLOCK_EXTRA_WAIT_TIME is more about covering the time including consensus and broadcasting(gossip)?

ansnunez:
@sandipbhoir yes - consensus, broadcasting, and some additional buffer time which we will define ourselves.  The issue with using the other existing variables (e.g., viewchange and fallback timeout) is that it creates unnecessary circular dependencies - if viewchange timeout goes beyond one entire DS epoch (which was the case for my local testnet) then we have to adjust it, or else adjust the tx dist time to be longer, etc. etc.

PR #989, opened 2018-12-04
Log message name to track the size and process time

ShengguangXiao:
## Description

Log message name to track the size and process time. And develop grepperf tool to grep message size and processing time from Zilliqa log.

Add FormatMessageName function in Zilliqa class. Add unit test case for FormatMessageName. Add grepperf tool.

## Review Suggestion

Run follow command after run a local test.
./build/bin/grepperf ./build/local_run/node_0001/zilliqa-00001-log.txt ./build/local_run/node_0001/grepperf_result.txt

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libZilliqa/Zilliqa.cpp
+++ b/src/libZilliqa/Zilliqa.cpp
@@ -19,6 +19,7 @@
 
 #include <jsonrpccpp/common/exception.h>
 #include <jsonrpccpp/server/connectors/httpserver.h>
+#include <chrono>
 
 #include "Zilliqa.h"
 #include "common/Constants.h"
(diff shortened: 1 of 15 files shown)


src/libZilliqa/Zilliqa.cpp
```diff
@@ -77,9 +92,22 @@ void Zilliqa::ProcessMessage(pair<vector<unsigned char>, Peer>* message) {
         return;
       }
 
+      const auto ins_byte = message->first.at(MessageOffset::INST);
```
ansnunez: Should we add some kind of `ifdef` or check for a constant value around the newly added code here, so that we don't always include benchmarking in our binary?

ShengguangXiao: Maybe we can only print out bench mark information when debug level is higher, normally we don't log?

src/libZilliqa/Zilliqa.cpp
```diff
@@ -61,6 +62,20 @@ void Zilliqa::LogSelfNodeInfo(const std::pair<PrivKey, PubKey>& key,
...
 }
 
+/*static*/ std::string Zilliqa::FormatMessageName(unsigned char msgType,
+                                                  unsigned char instruction) {
+  const std::string InvalidMessageType = "INVALID_MESSAGE";
+  if (msgType >= ARRAY_SIZE(MessageTypeStrings)) return InvalidMessageType;
+
+  if (NULL == MessageTypeInstructionStrings[msgType]) return InvalidMessageType;
+
+  if (instruction >= MessageTypeInstructionSize[msgType])
```
sandipbhoir: You may want to keep the every statement for if and else in  {} as per our standard format.

PR #995, opened 2018-12-04
Increase coverage of AccountData/Transaction.cpp

jendis:
## Description
Increase coverage of AccountData/Transaction.cpp
Add more functions to libTestUtils

## Status
Lines 92% - further progress possible only when mocking is used
Functions 100%

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/libTestUtils/TestUtils.cpp
+++ b/tests/libTestUtils/TestUtils.cpp
@@ -59,6 +59,26 @@ uint8_t Dist1to99() { return RandomIntInRng<uint8_t>((uint8_t)1, (uint8_t)99); }
 
 PubKey GenerateRandomPubKey() { return PubKey(PrivKey()); }
 
+Address GetAddressFromPubKey(PubKey pubKey) {
+  Address address;
+
+  vector<unsigned char> vec;
+  pubKey.Serialize(vec, 0);
+  SHA2<HASH_TYPE::HASH_VARIANT_256> sha2;
+  sha2.Update(vec);
+
+  const vector<unsigned char>& output = sha2.Finalize();
+
+  if (output.size() != 32) {
+    LOG_GENERAL(WARNING, "assertion failed (" << __FILE__ << ":" << __LINE__
+                                              << ": " << __FUNCTION__ << ")");
+  }
+
+  copy(output.end() - ACC_ADDR_SIZE, output.end(), address.asArray().begin());
+
+  return address;
+}
+
 Peer GenerateRandomPeer() {
   uint128_t ip_address = DistUint32();
   uint32_t listen_port_host = DistUint32();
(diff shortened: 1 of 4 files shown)


tests/libTestUtils/TestUtils.cpp
```diff
@@ -59,6 +59,26 @@ uint8_t Dist1to99() { return RandomIntInRng<uint8_t>((uint8_t)1, (uint8_t)99); }
 
 PubKey GenerateRandomPubKey() { return PubKey(PrivKey()); }
 
+Address GetAddressFromPubKey(PubKey pubKey) {
```
kaikawaliu: there already is Account::GetAddressFromPublicKey for such purpose, can directly call that

tests/libTestUtils/TestUtils.cpp
```diff
@@ -209,18 +229,48 @@ DequeOfShard GenerateDequeueOfShard(size_t size) {
 
 CoSignatures GenerateRandomCoSignatures() { return CoSignatures(Dist1to99()); }
 
+// unsigned char randchar() {
```
kaikawaliu: are these commented out codes still useful, otherwise plz remove

tests/libTestUtils/TestUtils.cpp
ansnunez: As the classes under test are Account and Transaction, I think he made duplicate functions for `GetShardIndex` and `GetAddressFromPublicKey` to make sure they are computationally correct? @jendis can you confirm?

kaikawaliu: It doesn't make much sense to me to make new function to examine the correctness of existing function with same utility. The idea for unit test is by checking the output whether is desired by the input however

jendis: Well, Antonio told exactly what I meant. I think it make sense to keep it separate - it costs nothing but ensures the correctness of the computation in production code. I can imagine situation that can happen when inappropriate refactor in this production part is done.
On the other side, when any refactor is done intentionally in the end causes test failure, then it is absolutely right that you have to update even the unit test.
In this case I really copied the production code because I expected the code part computing the sha256 to be stable. Anyway I agree that it would make more sense to use directly the openssl lib and keep the constant ACC_ADDR_SIZE as a local constant. If this changes in constants.xml from any reason, then it is necessary to change it in test part, which I consider OK as I sad. If this changes "by accident", then test'll catch it.
So please let me know what do you think is the best way to go.

tests/libTestUtils/TestUtils.cpp
jendis: Fixed in incomming commit.

tests/libTestUtils/TestUtils.cpp
ansnunez: I think in keeping with the current way of doing unit tests elsewhere, and to avoid the necessary changes in test part, we can just do as @kaikawaliu says and stick to calling the production code for now.

PR #996, opened 2018-12-04
Bytes size check for tx hash

robin-thomas:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libMessage/Messenger.cpp
```diff
@@ -113,6 +113,20 @@ void NumberToArray(const T& number, vector<unsigned char>& dst,
   Serializable::SetNumber<T>(dst, offset, number, S);
 }
 
+template <class K, class V>
+bool CopyWithSizeCheck(const K& arr, V& result) {
```
ShengguangXiao: Can add some unit test case for this function?

src/libMessage/Messenger.cpp
```diff
@@ -113,6 +113,20 @@ void NumberToArray(const T& number, vector<unsigned char>& dst,
...
+  LOG_MARKER();
+
+  // Fixed length copying.
+  if (arr.size() != result.max_size()) {
```
ShengguangXiao: Why it is not if (arr.size() <= result.max_size()) ?

ansnunez: This function will be used for all our fixed-length types (the hashes).  As such, we shouldn't really expect to receive an array that is not exactly the size expected.  So it's better to have the strict `!=` check here.

src/libMessage/Messenger.cpp
ansnunez: It's a bit tricky I think to add unit test case for this as it is a local function that does not need to be exposed outside of Messenger.cpp.  For now I think the easiest way to test is to run testnet and trigger the call to this function in all the other functions (I'll do this for @robin-thomas)

src/libMessage/Messenger.cpp
ShengguangXiao: I see

src/libMessage/Messenger.cpp
ShengguangXiao: We can put this function in Messenger.h and then can add unit test for it in unit test cpp files.

ansnunez: Updated as suggested

PR #1000, opened 2018-12-05
Add option to disable Fallback

kaikawaliu:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


ansnunez:
I propose to put similar check for this flag before calling `FallbackTimerLaunch`, `FallbackTimerPulse`, etc.  Those functions consume CPU time (e.g., acquiring mutex).

PR #1001, opened 2018-12-05
Fix potential buffer overflow when using HexStrToStdArray

ansnunez:
## Description

This PR contains the following:
1. After calling `Messenger::GetDSPoWSubmission`, check for correct sizes of the PoW string components
2. Fix `HexStrToStdArray` and `HexStrToStdArray64` to prevent buffer overflow
3. Other minor improvements inside DataConversion class

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/PoWProcessing.cpp
+++ b/src/libDirectoryService/PoWProcessing.cpp
@@ -152,6 +152,21 @@ bool DirectoryService::ProcessPoWSubmission(
     return false;
   }
 
+  if (resultingHash.size() != 64) {
+    LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
+              "Wrong resultingHash size "
+                  << resultingHash.size() << " submitted by "
+                  << submitterPeer.GetPrintableIPAddress());
+    return false;
+  }
+
+  if (mixHash.size() != 64) {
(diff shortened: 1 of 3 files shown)


src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -152,6 +152,21 @@ bool DirectoryService::ProcessPoWSubmission(
     return false;
   }
 
+  if (resultingHash.size() != 64) {
```
ShengguangXiao: This size should be 32, and compare with sizeof(ethash_hash256) will be better.

src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -152,6 +152,21 @@ bool DirectoryService::ProcessPoWSubmission(
...
+    LOG_EPOCH(WARNING, to_string(m_mediator.m_currentEpochNum).c_str(),
+              "Wrong resultingHash size "
+                  << resultingHash.size() << " submitted by "
+                  << submitterPeer.GetPrintableIPAddress());
+    return false;
+  }
+
+  if (mixHash.size() != 64) {
```
ShengguangXiao: This size should be 32, and compare with sizeof(ethash_hash256) will be better.

PR #1008, opened 2018-12-05
Always buffer txns from lookup and only do gossip during txn distribution

kaikawaliu:
## Description

Related issue: https://github.com/Zilliqa/Issues/issues/366
### Implementation

- [x] **ready for review**

- [x] small-scale cloud test (200 aws nodes)

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -192,7 +192,7 @@ bool Lookup::GenTxnToSend(size_t num_txn,
   return true;
 }
 
-const VectorOfLookupNode& Lookup::GetLookupNodes() const {
+VectorOfLookupNode Lookup::GetLookupNodes() const {
   LOG_MARKER();
   lock_guard<mutex> lock(m_mutexLookupNodes);
   return m_lookupNodes;
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -192,10 +192,24 @@ bool Lookup::GenTxnToSend(size_t num_txn,
   return true;
 }
 
-VectorOfLookupNode Lookup::GetLookupNodes() const {
+const VectorOfLookupNode& Lookup::GetLookupNodes() const {
   LOG_MARKER();
   lock_guard<mutex> lock(m_mutexLookupNodes);
   return m_lookupNodes;
+}
+
+bool Lookup::IsLookupNode(const PubKey& pubKey) const {
+  return std::find_if(GetLookupNodes().begin(), GetLookupNodes().end(),
+                      [&pubKey](const std::pair<PubKey, Peer>& node) {
+                        return node.first == pubKey;
+                      }) != GetLookupNodes().end();
+}
+
+bool Lookup::IsLookupNode(const Peer& peerInfo) const {
+  return std::find_if(GetLookupNodes().begin(), GetLookupNodes().end(),
+                      [&peerInfo](const std::pair<PubKey, Peer>& node) {
+                        return node.second == peerInfo;
+                      }) != GetLookupNodes().end();
 }
 
 void Lookup::SendMessageToLookupNodes(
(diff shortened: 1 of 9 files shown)


src/libNode/Node.cpp
```diff
@@ -1115,7 +1129,27 @@ bool Node::ProcessTxnPacketFromLookupCore(const vector<unsigned char>& message,
   // gossip (txnpkt) from this function but will be done at gossip layer.
```
sandipbhoir: May be remove this comment since it is no more applicable.

src/libNode/Node.cpp
```diff
@@ -1069,26 +1070,28 @@ bool Node::ProcessTxnPacketFromLookup(
...
     }
   }
 
-  if (epochNumber < m_mediator.m_currentEpochNum) {
-    LOG_GENERAL(WARNING, "Txn packet from older epoch, discard");
-    return false;
-  } else if (epochNumber == m_mediator.m_currentEpochNum) {
-    return ProcessTxnPacketFromLookupCore(message, shardId, transactions);
-  } else {
+  //
```
sandipbhoir: We may still keep this check:

```
if (epochNumber < m_mediator.m_currentEpochNum) {
    LOG_GENERAL(WARNING, "Txn packet from older epoch, discard");
    return false;
}
```

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -682,7 +682,13 @@ bool Node::RunConsensusOnMicroBlockWhenShardLeader() {
...
               "Non-vacuous epoch, going to sleep for "
                   << TX_DISTRIBUTE_TIME_IN_MS << " milliseconds");
-    std::this_thread::sleep_for(chrono::milliseconds(TX_DISTRIBUTE_TIME_IN_MS));
+    std::this_thread::sleep_for(chrono::milliseconds(
+        TX_DISTRIBUTE_TIME_IN_MS +
+                (m_mediator.m_dsBlockChain.GetLastBlock()
+                     .GetHeader()
+                     .GetEpochNum() == m_mediator.m_currentEpochNum)
+            ? LOOKUP_DELAY_SEND_TXNPACKET_IN_MS
+            : 0));
```
sandipbhoir: Logging of sleep time and actual sleep time will differ for first txn epoch w.r.t current ds epoch.

src/libNode/Node.cpp
kaikawaliu: added

src/libLookup/Lookup.cpp
```diff
@@ -192,12 +192,26 @@ bool Lookup::GenTxnToSend(size_t num_txn,
   return true;
 }
 
-VectorOfLookupNode Lookup::GetLookupNodes() const {
+const VectorOfLookupNode& Lookup::GetLookupNodes() const {
```
ansnunez: I think it's uncommon to hold a mutex and then return a reference

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -679,10 +679,16 @@ bool Node::RunConsensusOnMicroBlockWhenShardLeader() {
 
   if (m_mediator.m_ds->m_mode == DirectoryService::Mode::IDLE &&
       !m_mediator.GetIsVacuousEpoch()) {
+    unsigned int sleep_time =
+        TX_DISTRIBUTE_TIME_IN_MS +
+        ((m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetEpochNum() ==
```
ansnunez: Is this line to check if we are in the first Tx epoch after a new DS epoch?  Doesn't seem correct.

src/libNode/Node.cpp
```diff
@@ -1069,26 +1070,32 @@ bool Node::ProcessTxnPacketFromLookup(
...
+
+  if (m_mediator.m_lookup->IsLookupNode(from)) {
```
ansnunez: Can you add some description here about why we only buffer if it's from lookup, and not call `ProcessTxnPacketFromLookupCore`?

PR #1013, opened 2018-12-06
Refactor the get DS leader peer into a function

ShengguangXiao:
## Description

Refactor the get DS leader peer into a function

Add Node::GetDSLeaderPeer function, so it can be reused in multiple places. Easy to maintain.

https://github.com/Zilliqa/Issues/issues/358

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3208,16 +3208,18 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
         if (Node::GetDSLeaderPeer(
                 m_mediator.m_blocklinkchain.GetLatestBlockLink(),
                 m_mediator.m_dsBlockChain.GetLastBlock(),
-                *m_mediator.m_DSCommittee, dsLeaderPeer)) {
+                *m_mediator.m_DSCommittee, m_mediator.m_currentEpochNum,
+                dsLeaderPeer)) {
           toSend.push_back(dsLeaderPeer);
         }
 
-        unsigned int count = 0;
         for (auto const& i : *m_mediator.m_DSCommittee) {
-          if (count < NUM_NODES_TO_SEND_LOOKUP && i.second != dsLeaderPeer) {
+          if (toSend.size() < NUM_NODES_TO_SEND_LOOKUP &&
+              i.second != dsLeaderPeer) {
             toSend.push_back(i.second);
-            count++;
-          } else {
+          }
+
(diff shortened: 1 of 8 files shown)


KaustubhShamshery:
Can you also change in FinishRejoinAsDS

src/libLookup/Lookup.cpp
```diff
@@ -3168,7 +3167,7 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
       {
         lock_guard<mutex> g(m_mediator.m_ds->m_mutexShards);
         auto it = m_mediator.m_ds->m_shards.at(i).begin();
-        //[FIX-ME]: Lookup sends to NUM_NODES_TO_SEND_LOOKUP + Leader
+        // Lookup sends to NUM_NODES_TO_SEND_LOOKUP + Leader
```
sandipbhoir: FIX ME - was added here to fix the number of total nodes to send including leader. right now its NUM_NODES_TO_SEND_LOOKUP + Leader which needs t be fixed eventually.

src/libNode/PoWProcessing.cpp
```diff
@@ -287,33 +287,13 @@ bool Node::SendPoWResultToDSComm(const uint64_t& block_num,
       break;
     }
   }
-  //[FIX-ME] Send to PoW PACKET_SENDERS + 1
+  // Send to PoW PACKET_SENDERS + DS Leader
```
sandipbhoir: FIX ME - was added here to fix the number of total nodes to send including leader. right now its NUM_NODES_TO_SEND_LOOKUP + Leader which needs t be fixed eventually.

src/libNode/Node.cpp
```diff
@@ -1668,3 +1669,42 @@ std::string Node::GetActionString(Action action) const {
...
+    if (!GUARD_MODE) {
+      leader_id = lastBlockHash % dsCommittee.size();
+    } else {
+      leader_id = lastBlockHash % Guard::GetInstance().GetNumOfDSGuard();
+    }
+    dsLeaderPeer = dsCommittee.at(leader_id).second;
+    LOG_EPOCH(INFO, to_string(epochNumber).c_str(),
+              "lastBlockHash " << lastBlockHash << ", current ds leader id "
+                               << leader_id << ", Peer "
+                               << std::string(dsLeaderPeer));
```
sandipbhoir: Do we need tostring() for printing peer?

ShengguangXiao: No need, i fixed it already.

PR #1021, opened 2018-12-08
fix 100% CPU problem

deepgully:
the sleep should inside the while loop.

## Description
PR #732 not fix the problem correctly.

### Implementation
- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


ansnunez:
Thanks @deepgully

PR #1026, opened 2018-12-11
Fix node rejoin and win DS PoW and then process sharding structure fail

ShengguangXiao:
## Description

An external node may use the same pubkey and try to join from different IP address, so when a node start to rejoin, it should clear all the old PoWs and pow connections, then can ProcessShardingStructure of the latest one succesffully.

Clean DS variables when start to rejoin as normal.

https://github.com/Zilliqa/Issues/issues/373

## Review Suggestion

Check files changed.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


KaustubhShamshery:
OK. this is a part of the fix. but when the described issue happens, should not the node accept leader's view

ShengguangXiao:
> OK. this is a part of the fix. but when the described issue happens, should not the node accept leader's view

Yes, now m_allPoWConns is cleared, so the new joined node will accept the view in the DS block.

PR #1033, opened 2018-12-11
Fix checking IsLookupNode & Add checking IsShardNode & Add mutex

kaikawaliu:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test (200 aws nodes)


src/libNode/Node.cpp
```diff
@@ -1092,12 +1091,20 @@ bool Node::ProcessTxnPacketFromLookup(
...
               << string(lookupPubKey).substr(0, 6) << "] RECVFROMLOOKUP");
     m_txnPacketBuffer.emplace_back(message);
-  } else {
+  } else if (IsShardNode(from)) {
     LOG_GENERAL(INFO,
                 "Packet received from a non-lookup node, "
                 "should be from gossip neightor and process it");
     return ProcessTxnPacketFromLookupCore(message, dsBlockNum, shardId,
                                           lookupPubKey, transactions);
+  } else if (ALLOW_TXN_PACKET_FROM_LOCAL &&
```
KaustubhShamshery: As we can see from the testnetv3, we cannot really rely on IPs of shard nodes.

PR #1034, opened 2018-12-11
Schnorr fix

AmritKumar:
## Description
 This PR fixes issues found by the auditors in our implementation of Schnorr signature scheme.
 

## Review Suggestion
 Reviewers are suggested to check out the security reports that proposed fixes.
 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


docker/Dockerfile
```diff
@@ -61,6 +61,13 @@ RUN curl -sL https://github.com/mongodb/mongo-c-driver/releases/download/1.13.0/
     && cmake --build mongo-cxx-build --target install \
     && rm -rf mongo-cxx-*
 
+# libCrypto requires openssl-1.1.x.
+RUN curl -sL https://www.openssl.org/source/openssl-1.1.1.tar.gz | tar -zx -C . \
```
bb111189: should we also add the instruction to our readme doc

scripts/ci_install_deps.sh
```diff
@@ -46,6 +49,20 @@ make -j -l4
...
+# only 1.1.x has the necessary BN_generate_dsa_nonce function in bn.h
+cd ${DEPENDS_DIR}
+curl -OL https://www.openssl.org/source/openssl-1.1.1.tar.gz
+tar -xzf openssl-1.1.1.tar.gz
+cd openssl-1.1.1
+./config -Wl,--enable-new-dtags,-rpath,'$(LIBRPATH)' > /dev/null
+make > /dev/null
+make install > /dev/null
+cd ..
+rm -rf openssl-1.1.1.tar.gz openssl-1.1.1.tar.gz
```
ansnunez: File listed twice

docker/Dockerfile
iantanwx: Yes, we should, for those who are running Ubuntu 16.04, they'll have to do this.

scripts/ci_install_deps.sh
iantanwx: Oops, I'll take care of this.

ansnunez:
@AmritKumar are these lines still considered dead code, and should we remove them?

```
       // Clear buffer
       fill(buf.begin(), buf.end(), 0x00);
 
       if (!err) {
         res = (BN_is_zero(result.m_r.get())) || (BN_is_zero(result.m_s.get()));
       }
```

AmritKumar:
> {

@ansnunez I don't think this is dead code. We need to check at some point whether `r` and `s` are non-zero. We are not doing that anywhere until at the end, so this code snippet is for sure reachable. I do admit that if for some reasons `r` is indeed zero, then instead of going ahead and computing `s` will be wasteful as we will have to regenerate `r` in the next iteration of the loop. Ideally, we should first check if `r` is zero in the middle of the while loop and if `true`, we should continue with the next iteration. The same holds for `s`.

AmritKumar:
@ansnunez @bb111189 Please do check for other dead code and uncaught exceptions. I haven't checked all of them.

README.md
```diff
@@ -73,12 +73,22 @@ To run Zilliqa, we recommend the following minimum system requirements:
         libssl-dev libleveldb-dev libjsoncpp-dev libsnappy-dev cmake libmicrohttpd-dev \
         libjsonrpccpp-dev build-essential pkg-config libevent-dev libminiupnpc-dev \
         libprotobuf-dev protobuf-compiler libcurl4-openssl-dev
+    
+    # Install openssl-1.1.1 on 16.04
```
ansnunez: Historically we haven't copied over some of our installation code (e.g., mongodb) into the readme, apart from the apt-get install and the brew install.  It might become too unwieldy I think.

iantanwx: @Gnnng will refactor this process into a cmake-based one.

Issue #1015, opened 2018-12-06
pow error when run on cuda gpu other than gpu 0


dtompichayut:
when run with gpu 0 it work correctly but when I tried on GPU 1 and 5 it is the same result, please check

$ tail -n 30 zilliqa-00001-log.txt 
[INFO   ][23946][10:51:30:153][InitMining                    ] BEGIN
[INFO   ][23946][10:51:30:153][UpdateDSBlockRand             ] BEGIN
[INFO   ][23946][10:51:30:153][Serialize                     ] BEGIN
[INFO   ][23946][10:51:30:153][Serialize                     ] END
[INFO   ][23946][10:51:30:153][UpdateDSBlockRand             ] END
[INFO   ][23946][10:51:30:153][GetStateRootHash              ] BEGIN
[INFO   ][23946][10:51:30:153][GetStateRootHash              ] END
[INFO   ][23946][10:51:30:153][CheckStateRoot                ] CheckStateRoot match
[INFO   ][23946][10:51:30:153][SetState                      ][Epoch 3100] Node State is now POW_SUBMISSION at epoch 3100
[INFO   ][23946][10:51:35:153][InitMining                    ][Epoch 3100] Starting PoW for new ds block number 32
[INFO   ][23946][10:51:35:153][StartPoW                      ] BEGIN
[INFO   ][23946][10:51:35:153][StartPoW                      ][Epoch 3100] Current dsblock is 32
[INFO   ][23946][10:51:35:154][PoWMine                       ] BEGIN
[INFO   ][26458][10:51:35:154][MineFullGPUThread             ] BEGIN
[INFO   ][26458][10:51:35:154][MineFullGPUThread             ] Difficulty : 3, miner index 0
[INFO   ][26458][10:51:38:960][MineFullGPUThread             ] END
[INFO   ][23946][10:51:38:960][PoWMine                       ] END
[INFO   ][23946][10:51:38:960][StartPoW                      ][Epoch 3100] Winning nonce   = 0x5c090d78
[INFO   ][23946][10:51:38:960][StartPoW                      ][Epoch 3100] Winning result  = 0x1273dc94d7994e512d23faf442fcf28d58827c2c772b3f082ffdbdd559b1a212
[INFO   ][23946][10:51:38:960][StartPoW                      ][Epoch 3100] Winning mixhash = 0xdc25536c67277831f4ba98e3cd28ee9ee9e390d809c70e815ef4c734b1517d54
[...]

PR #1042, opened 2018-12-12
Fix select GPU index other than 0 software crash problem

ShengguangXiao:
## Description

Fix select GPU index other than 0 software crash problem

Set the correct gpu going to use before do mining.

https://github.com/Zilliqa/Zilliqa/issues/1015

## Review Suggestion

The community member already help to test the branch on there 6 Nividia cards machine.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -406,9 +406,11 @@ void POW::InitOpenCL() {
     LOG_GENERAL(FATAL, "Failed to configure OpenCL GPU, please check hardware");
   }
 
-  CLMiner::setNumInstances(UINT_MAX);
   auto gpuToUse = GetGpuToUse();
   auto totalGpuDevice = CLMiner::getNumDevices();
+
+  CLMiner::setNumInstances(gpuToUse.size());
+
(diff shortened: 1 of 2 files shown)


src/libPOW/pow.cpp
```diff
@@ -433,19 +435,24 @@ void POW::InitCUDA() {
...
-  auto gpuToUse = GetGpuToUse();
+  CUDAMiner::setNumInstances(gpuToUse.size());
+
   auto totalGpuDevice = CUDAMiner::getNumDevices();
   for (const auto gpuIndex : gpuToUse) {
     if (gpuIndex >= totalGpuDevice) {
       LOG_GENERAL(FATAL, "Selected GPU "
                              << gpuIndex
-                             << " exceed the physical OpenCL GPU number "
+                             << " exceed the physical Nvidia GPU number "
```
ckyang: Not sure if it's proper to have other company's name inside our source code.

ShengguangXiao: I think should be ok. If you search Nvidia in all github, you can find 15 Million matchs.

PR #261, opened 2018-04-26
New Rejoining Mechanism | Lookup/DS Rejoining | Recovery by Daemon

kaikawaliu:
- Implemented DS backup node recovery mechanism
- Implemented Lookup node recovery mechanism
- Redesign and optimized the flow of new node joining
- Enabled recovery triggered by Daemon
- Fix Bugs
- Fix bootstrap for being compatible to latest version of cmakefile

PR #1043, opened 2018-12-12
 Remove sleep before staring POW.

deepgully:
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


deepgully:
@ShengguangXiao @ckyang 
why we sleep here for all nodes? some miner will take advantage from remove this delay.
if the delay is required for some nodes, can we check the state accurately?

ShengguangXiao:
@deepgully This delay is necessary, sometimes the DS nodes are not ready to process POW submissions, if submit too early, the POW maybe ignored. You can set to 0 to have a try.

deepgully:
@ShengguangXiao some thoughts FYI.

the startPOW message should be sent by the DS leader, at that time, there is at least 1 node(the leader ifself) can accept the POW result.

beside, a real miner should doing PoW as long as possible within POW window.

bb111189:
> @ShengguangXiao some thoughts FYI.
> 
> the startPOW message should be sent by the DS leader, at that time, there is at least 1 node(the leader ifself) can accept the POW result.
> 
> beside, a real miner should doing PoW as long as possible within POW window.

DS backup need to have the PoW result in order to validate what the DS leader proposed. As such, we need at least 2/3 of the DS committee to receive the PoW result and participate in the consensus.

deepgully:
if I read the code right, Node send Pow result to DS leader + 5 DS backup, these 6 DS will collecting Pow results from all nodes during POW window, them submit POW packages to DS committee for consensus.

https://github.com/Zilliqa/Zilliqa/blob/b4bb018b285b1e55dd3496b841293936dfee2eae/src/libNode/PoWProcessing.cpp#L295-L308

ansnunez:
Looks like this was added in #261 when we still had the two-phase PoW.  Should be ok to remove now, since we allow collecting submissions even if the DS nodes are a state behind POW_SUBMISSION.

PR #1045, opened 2018-12-13
Adjust the timing when lookup send out txn packet

kaikawaliu:
## Description

Now the lookup will send out the txn packet when TXN_DISTRIBUTE_TIME_IN_MS ended, saying during the microblock and finalblock consensus. With the previous implementation the lookup will avoid sending packet to the leader nodes which may congeste the consensus processing.

In this case, every first epoch in a DS epoch won't have any transaction to process, thus the waiting time before microblock consensus is reduced

And also a flag is introduced to fix the issue when the packet preparing is taking too long to flood the thread pool of the lookup.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -384,7 +384,9 @@ void Node::BeginNextConsensusRound() {
 
   ScheduleMicroBlockConsensus();
 
-  CommitTxnPacketBuffer();
+  if (!m_mediator.GetIsVacuousEpoch()) {
+    CommitTxnPacketBuffer();
+  }
 }
 
 bool Node::FindTxnInProcessedTxnsList(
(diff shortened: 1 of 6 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -824,8 +826,16 @@ bool Node::ProcessFinalBlock(const vector<unsigned char>& message,
 
     // Now only forwarded txn are left, so only call in lookup
     CommitMBnForwardedTransactionBuffer();
-    if (m_mediator.m_lookup->GetIsServer() && !isVacuousEpoch) {
-      m_mediator.m_lookup->SenderTxnBatchThread();
+    if (m_mediator.m_lookup->GetIsServer() && !isVacuousEpoch &&
+        !m_mediator.GetIsVacuousEpoch() &&
+        ((m_mediator.m_currentEpochNum + NUM_VACUOUS_EPOCHS + 1) %
```
kaikawaliu: the lookup won't initiate the packet thread if it's an epoch earlier than the vacuous epoch

PR #1046, opened 2018-12-13
Fix the deadlock on mutex m_mutexLookupNodes + Delay starting new lookup syncup

sandipbhoir:
## Description
This PR covers following:
 - Fix the deadlock on mutex m_mutexLookupNodes
- Added new lookyup sync delay - new lookup waits for DELAY sec before starting syncup.

## Review Suggestion

Check the file changed
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


constants.xml
```diff
@@ -43,6 +43,7 @@
         <POW_WINDOW_IN_SECONDS>150</POW_WINDOW_IN_SECONDS>
         <RECOVERY_SYNC_TIMEOUT>5</RECOVERY_SYNC_TIMEOUT>
         <TX_DISTRIBUTE_TIME_IN_MS>30000</TX_DISTRIBUTE_TIME_IN_MS>
+        <NEW_LOOKUP_SYNC_DELAY>180</NEW_LOOKUP_SYNC_DELAY>
```
ShengguangXiao: Can add a time unit at the end of the name?

PR #1047, opened 2018-12-13
Simplify log level output

ansnunez:
## Description

Just a minor enhancement for performance and log size - just reduce the level shown in the log to the first 4 characters.

Before this PR:
```
[INFO   ][ 3503][18-12-13T09:21:31.399][ProcessMessageCommitCore      ] BEGIN
[WARNING][ 3503][18-12-13T09:21:31.399][CheckState                    ] Action PROCESS_COMMIT not allowed in state CHALLENGE_DONE
```

After this PR:
```
[INFO][ 3503][18-12-13T09:21:31.399][ProcessMessageCommitCore      ] BEGIN
[WARN][ 3503][18-12-13T09:21:31.399][CheckState                    ] Action PROCESS_COMMIT not allowed in state CHALLENGE_DONE
```

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/Logger.cpp
+++ b/src/libUtils/Logger.cpp
@@ -27,15 +27,8 @@
 using namespace std;
 using namespace g3;
 
-string MyCustomFormatting(const LogMessage& msg) {
-  string res = string("[") + msg.level();
-
-  for (unsigned int i = msg.level().size(); i < WARNING.text.size(); ++i) {
-    res += " ";
-  }
-
-  res += "]";
-  return res;
+inline string MyCustomFormatting(const LogMessage& msg) {
+  return string("[") + msg.level().substr(0, 4) + "]";
 }
 
 namespace {


src/libUtils/Logger.cpp
```diff
@@ -27,15 +27,8 @@
...
-  string res = string("[") + msg.level();
-
-  for (unsigned int i = msg.level().size(); i < WARNING.text.size(); ++i) {
-    res += " ";
-  }
-
-  res += "]";
-  return res;
+inline string MyCustomFormatting(const LogMessage& msg) {
+  return string("[") + msg.level().substr(0, 4) + "]";
```
ShengguangXiao: If msg.level() length less than 4 software will crash

ansnunez: Nope, `string::substr` will only return as many characters as available in that case.
Right now INFO is our shortest level string.

ShengguangXiao: I see, then no problem.

PR #1049, opened 2018-12-13
Apply delay for lookup sending txn for every epoch

kaikawaliu:
…k whether the previous batch thread hasn't finish

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -3174,7 +3174,14 @@ void Lookup::SenderTxnBatchThread() {
...
   LOG_MARKER();
 
+  if (m_startedTxnBatchThread) {
+    LOG_GENERAL(WARNING,
+                "The last TxnBatchThread hasn't finished, discard this time");
+    return;
+  }
+
   auto main_func = [this]() mutable -> void {
+    m_startedTxnBatchThread = true;
```
KaustubhShamshery: Do you think  better to use mutex?

kaikawaliu: i have experienced piled up thread for sending txn batch which actually consumed all the cpu power, so it's better to discard it if too busy otherwise it may lead the untimely sending of txn packet

KaustubhShamshery: yes, my suggestion is rather than boolean, use mutex

KaustubhShamshery: use atomic bool.

kaikawaliu: okay

PR #1053, opened 2018-12-14
Removed unnecessary check from div_signint

ansnunez:
## Description

Once you have ruled out the problematic cases (the case b == 0; and the case b == -1 and a == MIN_VALUE), no overflow/underflow is actually possible, so the test a == b * c + a % b is unnecessary

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/SafeMath.tpp
+++ b/src/libUtils/SafeMath.tpp
@@ -271,10 +271,6 @@ bool SafeMath<T>::div_signint(const T& a, const T& b, T& result) {
 
   T c = a / b;
 
-  if (a != b * c + a % b) {
-    return false;
-  }
-
   result = c;
   return true;
 }


src/libUtils/SafeMath.tpp
```diff
@@ -271,10 +271,6 @@ bool SafeMath<T>::div_signint(const T& a, const T& b, T& result) {
 
   T c = a / b;
```
ShengguangXiao: Why didn't check if b is zero?

ansnunez: The check for that is in `SafeMath<T>::div`, which calls this function

ShengguangXiao: ok

PR #1057, opened 2018-12-14
DS Committee network information update and other various fixes

bb111189:
## Description

This PR enables ds guard to rejoin the ds committee upon termination of the pod. This PR provides a protocol for the updating and propagation  of new network information of ds guard. 

Refer to design documentation for the design of the updating process.

**Tasks:**
- [x] DS guard to update ds members and lookup
- [x] DS member to receive the updated network information
- [x] Lookup to receive the updated network information
- [x] Lookup to store the updated network information
- [x] Node to query the lookup at vacuous epoch
- [x] Node to receive updated information from lookup
- [x] DS guard to proceed with protocol
- [x] refactor local test script 
- [ ] large scale integration

**Other changes/fixes:**
- Removal of Heartbeat code
- Consistency fix in libmessage. `dest` to `dst`
- Log unknown instruction type and payload
- Update Zilliqa version message
- Update Zilliqa help message
- Grammar fixes

## Review Suggestion

Implementation is ready for review. 

### Implementation

- [x] **ready for review**

### Integration Test (Core Team)

- [x] local machine test
- [x] local machine re-test

- [ ] medium-scale cloud test

PR #1061, opened 2018-12-17
Feature/input param validation

jendis:
## Description
Validation of input parameter <IP address> into main.cpp - validates dotted decimal format in case of both IPv4 and IPv6.
IPv6 can be used now - added conversion from dotted decimal string into boost::multiprecision::uint128_t in network byte order.

### Implementation
For conversion between "dotted decimal" IP string into integer in network byte order is used library <boost/asio.hpp> because of slightly inconsistent results of <arpa/inet.h> depending on version - can accept not only the full digit IPs.

- [x] local machine test

- [x] small-scale cloud test (commit: b02f2b8)


src/libUtils/IPConverter.cpp
```diff
@@ -29,9 +32,58 @@ const std::string IPConverter::ToStrFromNumericalIP(
   return std::string(str);
 }
 
-const boost::multiprecision::uint128_t IPConverter::ToNumericalIPFromStr(
-    const std::string& ipStr) {
-  struct in_addr ip_addr;
-  inet_pton(AF_INET, ipStr.c_str(), &ip_addr);
-  return (boost::multiprecision::uint128_t)ip_addr.s_addr;
+void LogBrand() {
+  cout << "Copyright (C) Zilliqa. Version 1.0 (Durian). "
```
bb111189: It will be changed to "Copyright (C) Zilliqa. Version 3.0 (Durian - Mao Shan Wang)." in PR https://github.com/Zilliqa/Zilliqa/pull/1057

src/libUtils/IPConverter.cpp
```diff
@@ -29,9 +32,58 @@ const std::string IPConverter::ToStrFromNumericalIP(
...
+void LogUnsupported() {
+  LogBrand();
+  LogBugReport();
+  cout << "Error: Unknown address type - unsupported protocol\n" << endl;
+}
+
+void LogInvalidIP() {
+  LogBrand();
+  LogBugReport();
+  cout << "Error: listen_ip_address does not contain a character string "
```
bb111189: Able to log the invalid string?

src/libUtils/IPConverter.cpp
```diff
@@ -29,9 +32,58 @@ const std::string IPConverter::ToStrFromNumericalIP(
...
+          "representing a valid network address\n"
+       << endl;
+}
+
+void LogInternalErr() {
+  LogBrand();
+  LogBugReport();
+  std::cout << "Internal Error: cannot process the input IP address.\n"
```
bb111189: able to log the input ip?

src/libUtils/IPConverter.cpp
jendis: Done

src/libUtils/IPConverter.cpp
jendis: Done

src/libUtils/IPConverter.cpp
jendis: Done

tests/Utils/Test_IPConverter.cpp
```diff
@@ -45,8 +45,8 @@ BOOST_AUTO_TEST_CASE(test_IPNumericaltoString) {
 BOOST_AUTO_TEST_CASE(test_IPStringToNumerical) {
   INIT_STDOUT_LOGGER();
 
-  boost::multiprecision::uint128_t result =
-      IPConverter::ToNumericalIPFromStr("127.0.0.1");
+  boost::multiprecision::uint128_t result;
+  IPConverter::ToNumericalIPFromStr("127.0.0.1", result);
```
ansnunez: Should check for the result of `ToNumericalIPFromStr` wherever it's called

src/libUtils/IPConverter.cpp
```diff
@@ -29,9 +51,45 @@ const std::string IPConverter::ToStrFromNumericalIP(
...
+       << endl;
+}
+
+void LogInternalErr(const string& ip) {
+  LogBrandBugReport();
+  cerr << "Internal Error: cannot process the input IP address " << ip << ".\n"
+       << std::endl;
+}
+
+int ToNumericalIPFromStr(const std::string& ipStr,
```
ansnunez: Might be better to simplify return type to `bool` and we just consider `true` as successful conversion case.

tests/Utils/Test_IPConverter.cpp
jendis: Done

src/libUtils/IPConverter.cpp
jendis: Done

src/cmd/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ add_custom_command(TARGET zilliqa
     COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:zilliqa> ${CMAKE_BINARY_DIR}/tests/Zilliqa)
 target_include_directories(zilliqa PUBLIC ${CMAKE_SOURCE_DIR}/src)
 
-target_link_libraries(zilliqa PUBLIC Consensus Network Utils Zilliqa DirectoryService Node Validator NAT)
+target_link_libraries(zilliqa PUBLIC Consensus Network Utils Zilliqa DirectoryService Node Validator NAT ${Boost_LIBRARIES})
```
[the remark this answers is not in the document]
jendis: Done

jendis:
Added possibility to insert port number as part of address delimited by colon like "1.2.3.4:155", with necessary validation.

jendis:
boost::program_options applied together with more validation steps for cmd/*.
Changes in input arguments reflected to the execution scripts.

src/cmd/GetAddressFromPubKey.cpp
```diff
@@ -34,22 +35,78 @@
...
+    try {
+      key = PubKey::GetPubKeyFromString(pubk);
+    } catch (std::invalid_argument& e) {
+      std::cerr << e.what() << endl;
+      return ERROR_IN_COMMAND_LINE;
+    }
+
+    key.Serialize(message, 0);
+    sha2.Update(message, 0, PUB_KEY_SIZE);
+    const vector<unsigned char>& tmp2 = sha2.Finalize();
```
Gnnng: We prefer using `bytes` instead of `vector<unsigned char>` here. See the recent PR  https://github.com/Zilliqa/Zilliqa/pull/1094.

src/cmd/GetAddressFromPubKey.cpp
```diff
@@ -34,22 +35,78 @@
...
+      return ERROR_IN_COMMAND_LINE;
+    } catch (boost::program_options::error& e) {
+      SWInfo::LogBrandBugReport();
+      cerr << "ERROR: " << e.what() << endl << endl;
+      return ERROR_IN_COMMAND_LINE;
+    }
+
+    SHA2<HASH_TYPE::HASH_VARIANT_256> sha2;
+    sha2.Reset();
+    vector<unsigned char> message;
```
Gnnng: We prefer using bytes instead of vector<unsigned char> here. See the recent PR #1094.

Gnnng:
Thank @jendis for the update. We will let more team members take a look on some parts of this PR.

Can you make some further changes?
- Since https://github.com/Zilliqa/Zilliqa/pull/1094 we use `bytes` to replace `vector<unsigned char>` so let's just follow it. You may need to include the header if the type is not defined.
- enclose the boost header files within `<>` other than `""` to remain consistent

@ckyang, @KaustubhShamshery, @ShengguangXiao: Can you guys help to check the modification on these files and also see if all the affected code/scripts have been updated?
- `signmultisig.cpp` (@ckyang)
- `gen_sign_initial_ds.cpp` (@KaustubhShamshery)
- `grepperf.cpp` (@ShengguangXiao)

tests/Utils/Test_IPConverter.cpp
```diff
@@ -45,8 +45,12 @@ BOOST_AUTO_TEST_CASE(test_IPNumericaltoString) {
...
```
ansnunez: Can remove line 49

src/cmd/sendcmd.cpp
```diff
@@ -133,73 +143,148 @@ void process_cmd(int numargs, const char* progname, const char* cmdname,
...
-        processed = true;
-        break;
+      return SUCCESS;
+    } else {
+      string ip_;
+      if (IPConverter::GetIPPortFromSocket(ip, ip_, port)) {
+        ip = ip_;
       }
+      uint128_t remote_ip;
+      IPConverter::ToNumericalIPFromStr(ip, remote_ip);
```
ansnunez: Check result?

src/cmd/gen_sign_initial_ds.cpp
```diff
@@ -23,10 +23,18 @@
 
 #include <boost/property_tree/ptree.hpp>
 #include <boost/property_tree/xml_parser.hpp>
+#include "boost/program_options.hpp"
```
KaustubhShamshery: This program was inteded to be used with a single keypair. Any reason for changin?

jendis: The motivation for all binaries in _cmd/_ subfolder was the same - more input params validation together with incorporation of _boost/prograp_options_.
If you consider parameter flags being decrease in "user friendliness", then I can revert these changes. From perspective of scripts, only the _release.sh_ uses it and I have already done the adaptation here, so it shouldn't be a problem.

src/cmd/gen_sign_initial_ds.cpp
```diff
@@ -46,78 +55,121 @@ struct PTree {
...
+      Signature sig;
+      Schnorr::GetInstance().Sign(message, privKeys.at(i), pubKeys.at(i), sig);
+      bytes result;
+      sig.Serialize(result, 0);
+      sig_str = DataConversion::Uint8VecToHexStr(result);
     }
-  }
 
-  if (privKeys.size() != pubKeys.size()) {
-    if (pubKeys.size() != 1) {
```
KaustubhShamshery: I was talking about this check.

ckyang:
@jendis Yes indeed daemon_restart.py is a kind of testing, however it's more like application (system integration testing), rather than unit test. What I mention is, maybe we can add some test code inside /tests/cmd/, that calling signmultisig only. Thus, in the future when signmultisig has some bug, we can quickly locate where the bug is.

PR #1062, opened 2018-12-17
Cast uint128_t to int256_t when trying to get negative value

kaikawaliu:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libMessage/Messenger.cpp
+++ b/src/libMessage/Messenger.cpp
@@ -265,7 +265,9 @@ bool ProtobufToAccountDelta(const ProtoAccount& protoAccount, Account& account,
   ProtobufByteArrayToNumber<uint128_t, UINT128_SIZE>(protoAccount.balance(),
                                                      tmpNumber);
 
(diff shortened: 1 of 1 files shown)


src/libMessage/Messenger.cpp
```diff
@@ -265,7 +265,9 @@ bool ProtobufToAccountDelta(const ProtoAccount& protoAccount, Account& account,
   ProtobufByteArrayToNumber<uint128_t, UINT128_SIZE>(protoAccount.balance(),
                                                      tmpNumber);
 
-  int256_t balanceDelta = protoAccount.numbersign() ? tmpNumber : 0 - tmpNumber;
+  int256_t balanceDelta = protoAccount.numbersign()
```
ansnunez: Should we change to `int128_t`?  Also perhaps `Account::ChangeBalance` (currently takes in `int256_t`)?

ansnunez: Ah sorry, the recommendation is to keep 256

PR #1066, opened 2018-12-18
Add constants field for easier setting of min gas price

kaikawaliu:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


iantanwx:
Should it be 10^9 instead? I think the target gas price is 1000 `li`, which is 10^6 `qa`; so, the constant should be `1_000_000_000`.

jiayaoqijia:
> Should it be 10^9 instead? I think the target gas price is 1000 `li`, which is 10^6 `qa`; so, the constant should be `1_000_000_000`.

Yeah. You're correct. @kaikawaliu Can you revise the value accordingly? Thanks.

src/libServer/JSONConversion.cpp
```diff
@@ -43,11 +43,11 @@ const Json::Value JSONConversion::convertMicroBlockInfoArraytoJson(
     const vector<MicroBlockInfo>& v) {
   Json::Value mbInfosJson = Json::arrayValue;
   for (auto const& i : v) {
-    Json::Value mbInfoJson;
-    mbInfoJson["MicroBlockHash"] = i.m_microBlockHash.hex();
-    mbInfoJson["MicroBlockTxnRootHash"] = i.m_txnRootHash.hex();
-    mbInfoJson["MicroBlockShardId"] = i.m_shardId;
-    mbInfosJson.append(mbInfosJson);
+    Json::Value t_mbInfoJson;
```
KaustubhShamshery: remove this. already there.

PR #1076, opened 2018-12-19
Changed Protobuf definitions to allow backward compatibility in persistent storage

ansnunez:
## Description

There are 3 kinds of message classes defined in ZilliqaMessage.proto:
1. Classes for message exchanges between nodes
2. Classes for use in persistent storage
3. Classes for both #1 and #2

For #2, most of the fields in these message classes right now are of type `required`.
In the future, if we don't need these fields anymore, deserializing from persistent storage could become an issue.

This PR does the following:
1. Change fields for #2 from `required` to `optional` for classes used in persistent storage
2. Add `CheckFieldsInXXX` functions to enforce checks on the fields that were previously marked `required`
3. Change return type from `void` to `bool` for the affected functions (mostly the `ProtobufToXXX` functions)

Note:
For #2, extra care has to be taken to avoid enforcing the existence of fields that were previously defined as `repeated` or already `optional`.
`ProtoDSBlock` in particular had some already `optional` fields in the header.
`ProtobufToDSBlockHeader` has been updated to check for these already `optional` fields before accessing them (e.g., `dsdifficulty`).

These are the databases covered by this PR:
1. blockLinks
2. dsBlocks
3. dsCommittee
4. microBlocks
5. shardStructure
6. txBlocks
7. VCBlocks

These are the databases not covered by this PR:
1. contractCode
2. contractState
3. fallbackBlocks
4. metadata
5. state
6. stateDelta

A unit test has also been created under tests/Message/Test_Messenger_Compatibility, to prove that once we add new `optional` fields, we can still read the older fields from the old message format.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libMessage/Messenger.cpp
```diff
@@ -113,6 +117,162 @@ void NumberToArray(const T& number, vector<unsigned char>& dst,
   Serializable::SetNumber<T>(dst, offset, number, S);
 }
 
+// ============================================================================
+// Functions to check for fields in primitives that are used for persistent
+// storage. Remove fields from the checks once they are deprecated.
+// ============================================================================
+
+inline bool CheckFieldsInProtoBlockLink(const ProtoBlockLink& protoBlockLink) {
```
ckyang: Maybe we can change the API name to CheckRequiredFieldsxxx(), because this function is to replace the "required" field.

PR #1079, opened 2018-12-19
Fix verify node priority problem

ShengguangXiao:
## Description

Verify Priority fail because DS leader and DS backup choose different nodes with same priority

When verify node priority, need to check if the proposed priority smaller than the lowest priority, only it is smaller than the local smallest node priority, then consider it is a invalid shade node propose

https://github.com/Zilliqa/Issues/issues/381

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -511,14 +515,18 @@ bool DirectoryService::VerifyNodePriority(const DequeOfShard& shards) {
...
+          ++numOutOfMyPriorityList;
+          LOG_GENERAL(WARNING,
+                      "Node " << toFind << " is not in my top priority list");
+        }
       }
     }
   }
 
-  constexpr float tolerance = 0.02f;
+  constexpr float tolerance = 0.05f;
```
ckyang: Maybe we can put this experienced value into constant file.

src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -495,6 +512,7 @@ std::set<PubKey> DirectoryService::FindTopPriorityNodes() {
   for (size_t i = 0; i < MAX_SHARD_NODE_NUM && i < vecNodePriority.size();
        ++i) {
     setTopPriorityNodes.insert(vecNodePriority[i].first);
+    lowestPriority = vecNodePriority[i].second;
```
ansnunez: Should `lowestPriority` be equal to the smallest priority value inside `vecNodePriority`?
If yes, then we should maybe do this:
1. Outside the loop, set `lowestPriority` to `std::numeric_limits<uint8_t>::min`
2. Inside the loop, set `lowestPriority` to `max(lowestPriority, vecNodePriority[i].second)`

ShengguangXiao: vecNodePriority is sorted in descending order, so the last node should have the lowest priority.

ansnunez: Ok

PR #1080, opened 2018-12-20
Temporary blacklist node if it is unreachable or down

bb111189:
## Description

When a node is unreachable or down, it can cause connection to timeout after a time period. This will cause the network sendpool to clog up with messages. 

This PR blacklist temporary blacklist nodes that are unreachable or down. The blacklist is cleared at the start of each DS epoch or at the start of each VC (after precheck).

## Status
In progress 

### Implementation

- [ ] **ready for review**

- [ ] ~local machine test~

- [ ] medium-scale cloud test

### DIFF
--- a/src/libNetwork/P2PComm.cpp
+++ b/src/libNetwork/P2PComm.cpp
@@ -17,6 +17,9 @@
  * program files.
  */
 
+/* TCP error code:
+ * https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html */
+
 #include <errno.h>
 #include <event2/buffer.h>
 #include <event2/bufferevent.h>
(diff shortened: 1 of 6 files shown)


src/libNetwork/P2PComm.cpp
```diff
@@ -654,6 +683,13 @@ void P2PComm::AcceptConnectionCallback([[gnu::unused]] evconnlistener* listener,
   bufferevent_enable(bev, EV_READ | EV_WRITE);
 }
 
+bool P2PComm::IsHostDownOrUnreachable() {
```
ansnunez: Can reduce this to `return (errno == EHOSTUNREACH || errno == EHOSTDOWN);` and add `inline` to the function signature

src/libNetwork/P2PComm.cpp
```diff
@@ -134,6 +137,15 @@ uint32_t SendJob::writeMsg(const void* buf, int cli_sock, const Peer& from,
     ssize_t n = write(cli_sock, (unsigned char*)buf + written_length,
                       message_length - written_length);
 
+    if (P2PComm::IsHostDownOrUnreachable()) {
```
sandipbhoir: Is errno guaranteed to be for specific write sock attempt? SendJob::writeMsg can be called concurrently by different job handlers.

ansnunez: Should be okay I think: https://stackoverflow.com/questions/1694164/is-errno-thread-safe

PR #11, opened 2018-01-04
Remove CMake autogen files and copyright year changes

bb111189:

Issue #10, opened 2018-01-04
Hardcoded openssl path is failing to compile on Mac


ruchern:
https://github.com/Zilliqa/Zilliqa/blob/ba2d15d75861496fb9c5814c8734af53f313a0da/build.sh#L22

Installed `openssl` from `homebrew` at time of this issue has the version at `1.0.2n` and thus the build is failing to compile.

sidutta:
Issues related to mac are tracked under #7. Please use that thread. I think I had added the hardcoded path issue as a todo there. We won't be able to devote time to resolve those issues in the near future though. Feel free to make PRs and I could review them(I will comment on the one you have already made) or use a VM with ubuntu installed.

PR #1089, opened 2018-12-21
Fix potential race condition in CommitMicroBlockConsensusBuffer

ckyang:
## Description

In Node::CommitMicroBlockConsensusBuffer(), call-by-reference of a "vector" is used, and the memory address is passed to another thread. At the same time, this vector might grow (memory re-allocated) in a separated thread (Node::ProcessMicroblockConsensus()). Racing-condition might happened under this kind of scenario, thus we need to apply call-by-value to avoid this.
(reference: https://stackoverflow.com/questions/13793531/mysterious-segmentation-fault-on-destruction)

Below is detailed corestack:

[Current thread is 1 (Thread 0x7fe1e37e6700 (LWP 8763))]
(gdb) bt
#0  0x00007fe557c38428 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007fe557c3a02a in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007fe557c7a7ea in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007fe557c8337a in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007fe557c8753c in free () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x0000000000648204 in __gnu_cxx::new_allocator<unsigned char>::deallocate (this=0x7fe200389650, 
    __p=<optimized out>) at /usr/include/c++/5/ext/new_allocator.h:110
#6  std::allocator_traits<std::allocator<unsigned char> >::deallocate (__a=..., 
    __n=<optimized out>, __p=<optimized out>) at /usr/include/c++/5/bits/alloc_traits.h:517
#7  std::_Vector_base<unsigned char, std::allocator<unsigned char> >::_M_deallocate (
    __n=<optimized out>, __p=<optimized out>, this=0x7fe200389650)
    at /usr/include/c++/5/bits/stl_vector.h:178
#8  std::_Vector_base<unsigned char, std::allocator<unsigned char> >::~_Vector_base (
    this=0x7fe200389650, __in_chrg=<optimized out>) at /usr/include/c++/5/bits/stl_vector.h:160
#9  std::vector<unsigned char, std::allocator<unsigned char> >::~vector (this=0x7fe200389650, 
    __in_chrg=<optimized out>) at /usr/include/c++/5/bits/stl_vector.h:425
#10 std::pair<Peer, std::vector<unsigned char, std::allocator<unsigned char> > >::~pair (
    this=0x7fe200389620, __in_chrg=<optimized out>) at /usr/include/c++/5/bits/stl_pair.h:96
#11 Node::<lambda()>::~<lambda> (this=0x7fe200389610, __in_chrg=<optimized out>)
    at /zilliqa/src/libNode/MicroBlockPostProcessing.cpp:142
#12 std::_Bind<Node::CommitMicroBlockConsensusBuffer()::<lambda()>()>::~_Bind (this=0x7fe200389610, 
    __in_chrg=<optimized out>) at /usr/include/c++/5/functional:1058
#13 std::_Function_base::_Base_manager<std::_Bind<Node::CommitMicroBlockConsensusBuffer()::<lambda()>()> >::_M_destroy (__victim=...) at /usr/include/c++/5/functional:1726
#14 std::_Function_base::_Base_manager<std::_Bind<Node::CommitMicroBlockConsensusBuffer()::<lambda()>()> >::_M_manager(std::_Any_data &, const std::_Any_data &, std::_Manager_operation) (__dest=..., 
    __source=..., __op=<optimized out>) at /usr/include/c++/5/functional:1750

Update: There should be another reason for the crash, cause the default behavior of lambda capturing should be "copy by value". Need to dig out more for investigating this issue.

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/MicroBlockPostProcessing.cpp
+++ b/src/libNode/MicroBlockPostProcessing.cpp
@@ -138,7 +138,7 @@ bool Node::ProcessMicroblockConsensus(const vector<unsigned char>& message,
 void Node::CommitMicroBlockConsensusBuffer() {
   lock_guard<mutex> g(m_mutexMicroBlockConsensusBuffer);
 
-  for (const auto& i : m_microBlockConsensusBuffer[m_mediator.m_consensusID]) {
+  for (const auto i : m_microBlockConsensusBuffer[m_mediator.m_consensusID]) {
     auto runconsensus = [this, i]() {
       ProcessMicroblockConsensusCore(i.second, MessageOffset::BODY, i.first);
     };


src/libNode/MicroBlockPostProcessing.cpp
```diff
@@ -138,7 +138,11 @@ bool Node::ProcessMicroblockConsensus(const vector<unsigned char>& message,
 void Node::CommitMicroBlockConsensusBuffer() {
   lock_guard<mutex> g(m_mutexMicroBlockConsensusBuffer);
 
-  for (const auto& i : m_microBlockConsensusBuffer[m_mediator.m_consensusID]) {
+  /// Call-by-reference MUST NOT be used here, because later "i" (vector
+  /// element) will be passed to another thread, and this vector might grow
+  /// (memory re-allocated) in a separated thread (ProcessMicroblockConsensus),
+  /// then racing-condition could happened under this kind of scenario
+  for (const auto i : m_microBlockConsensusBuffer[m_mediator.m_consensusID]) {
     auto runconsensus = [this, i]() {
```
ansnunez: I think instead of capturing `i`, it has to be passed as a parameter

ansnunez: Can do like this:
```
for (const auto& i ...) {
   auto runconsensus = [this, i]() {
      vector<unsigned char> msg = i.second;
      ProcessMicroblockConsensusCore(msg, ...)
```

ckyang: There should be another reason for the crash, cause the default behavior of lambda capturing should be "copy by value". Need to dig out more for investigating this issue.

PR #1094, opened 2018-12-21
Using bytes instead of vector of unsigned char

ShengguangXiao:
## Description

Using bytes instead of vector of unsigned char

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/common/BaseType.h
+++ b/src/common/BaseType.h
@@ -17,16 +17,11 @@
  * program files.
  */
 
-#include "ContractStorage.h"
+#ifndef __BASETYPE_H__
+#define __BASETYPE_H__
 
-#include "libUtils/DataConversion.h"
+#include <vector>
 
-using namespace dev;
+using bytes = std::vector<unsigned char>;
 
-bool ContractStorage::PutContractCode(const h160& address, const bytes& code) {
-  return m_codeDB.Insert(address.hex(), code) == 0;
-}
-
-const bytes ContractStorage::GetContractCode(const h160& address) {
-  return DataConversion::StringToCharArray(m_codeDB.Lookup(address.hex()));
-}
+#endif  // __BASETYPE_H__
--- a/src/libConsensus/ConsensusBackup.cpp
+++ b/src/libConsensus/ConsensusBackup.cpp
@@ -70,8 +70,8 @@ bool ConsensusBackup::CheckState(Action action) {
   return true;
 }
 
-bool ConsensusBackup::ProcessMessageAnnounce(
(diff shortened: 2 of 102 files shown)


src/common/BaseType.h
```diff
@@ -0,0 +1,27 @@
...
+ * src/depends and tests/depends and which include a reference to GPLv3 in their
+ * program files.
+ */
+
+#ifndef __BASETYPE_H__
+#define __BASETYPE_H__
+
+#include <vector>
+
+using bytes = std::vector<unsigned char>;
```
ansnunez: Thanks, I've been meaning to make this change for a long while now.
Now that we have this, perhaps we should also change `unsigned char` to `uint8_t`, which is more clearer and concise.

src/libConsensus/ConsensusBackup.cpp
```diff
@@ -84,15 +84,15 @@ bool ConsensusBackup::ProcessMessageAnnounce(
...
                              m_blockNumber, m_blockHash, m_leaderID,
                              GetCommitteeMember(m_leaderID).first,
                              m_messageToCosign)) {
     LOG_GENERAL(WARNING, "Message validation failed");
 
     if (!errorMsg.empty()) {
-      vector<unsigned char> commitFailureMsg = {
+      bytes commitFailureMsg = {
           m_classByte, m_insByte,
           static_cast<unsigned char>(ConsensusMessageType::COMMITFAILURE)};
```
ansnunez: Can also change cases like this (`static_cast<unsigned char>`, `sizeof(unsigned char)`, and anywhere else where `unsigned char` is used).

src/libConsensus/ConsensusBackup.cpp
```diff
@@ -392,32 +390,31 @@ bool ConsensusBackup::ProcessMessageCollectiveSigCore(
...
 }
 
 ConsensusBackup::ConsensusBackup(uint32_t consensus_id, uint64_t block_number,
-                                 const vector<unsigned char>& block_hash,
-                                 uint16_t node_id, uint16_t leader_id,
-                                 const PrivKey& privkey,
+                                 const bytes& block_hash, uint16_t node_id,
+                                 uint16_t leader_id, const PrivKey& privkey,
                                  const deque<pair<PubKey, Peer>>& committee,
                                  unsigned char class_byte,
```
ansnunez: Maybe these also?

ShengguangXiao:
@KaustubhShamshery @ansnunez 200 nodes test is ok

PR #1099, opened 2018-12-24
Refactor DataConversion

bb111189:
## Description

This PR add exception handling to `DataConversion` utility. In addition, it changes the return value to `bool` to reflect the status of the conversion. This will prevent node crashing due to invalid hex value conversion using dataconversion lib. 

There are some area of code where the output (bool) is not validated. These are in cases when the input is always a valid hex value (eg. output of sha function)

Do not merge till medium-scale test pass.

## Review Suggestion

Review the changes in `Dataconversion` before reviewing other code changes.
**Reviewers are highly recommended to inspect all changes carefully. as there are numerous changes to numerous file.**
 
### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud test (commit: 027b130264ee2e94aa448117f08774813be325d9)

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -448,7 +448,8 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
     for (auto& ds : *m_mediator.m_DSCommittee) {
       string pubKey;
       if (!DataConversion::SerializableToHexStr(ds.first, pubKey)) {
-        return false;
+        LOG_GENERAL(WARNING, "Error converting pubkey to string");
+        continue;
       }
 
       if (ipMapping.find(pubKey) != ipMapping.end()) {
@@ -618,7 +619,8 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
           string pubKey;
           if (!DataConversion::SerializableToHexStr(
                   get<SHARD_NODE_PUBKEY>(node), pubKey)) {
-            return false;
+            LOG_GENERAL(WARNING, "Error converting pubkey to string");
+            continue;
           }
 
           if (ipMapping.find(pubKey) != ipMapping.end()) {
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -74,7 +74,7 @@ void addBalanceToGenesisAccount() {
 
   for (auto& walletHexStr : GENESIS_WALLETS) {
     bytes addrBytes;
-    if (!DataConversion::HexStrToUint8Vec(walletHexStr, addrBytes)){
+    if (!DataConversion::HexStrToUint8Vec(walletHexStr, addrBytes)) {
       continue;
     }
     Address addr{addrBytes};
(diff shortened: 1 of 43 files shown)


src/libServer/Server.cpp
```diff
@@ -504,7 +520,12 @@ Json::Value Server::GetSmartContracts(const string& address) {
       throw JsonRpcException(RPC_INVALID_PARAMETER,
                              "Address size not appropriate");
     }
-    bytes tmpaddr = DataConversion::HexStrToUint8Vec(address);
+    bytes tmpaddr;
+    if (!DataConversion::HexStrToUint8Vec(address, tmpaddr)) {
+      throw JsonRpcException(RPC_INVALID_ADDRESS_OR_KEY,
+                             "address is not a hex string");
```
sandipbhoir: you may want to use the same message `invalid address` that's used at other places

src/libNode/Node.cpp
```diff
@@ -441,10 +445,11 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
   GetIpMapping(ipMapping);
 
   if (!ipMapping.empty()) {
-    string pubKey;
-
     for (auto& ds : *m_mediator.m_DSCommittee) {
-      pubKey = DataConversion::SerializableToHexStr(ds.first);
+      string pubKey;
+      if (!DataConversion::SerializableToHexStr(ds.first, pubKey)) {
+        return false;
```
ansnunez: Do we want to `continue` here instead?

ansnunez: Or maybe add a warning message, as this code seems highly unlikely to be triggered, and if it ever were, then something is seriously wrong.

src/libNode/Node.cpp
```diff
@@ -608,12 +613,13 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
...
-      string pubKey;
-
       for (auto& shard : m_mediator.m_ds->m_shards) {
         for (auto& node : shard) {
-          pubKey = DataConversion::SerializableToHexStr(
-              get<SHARD_NODE_PUBKEY>(node));
+          string pubKey;
+          if (!DataConversion::SerializableToHexStr(
+                  get<SHARD_NODE_PUBKEY>(node), pubKey)) {
+            return false;
```
ansnunez: Same question here

src/libUtils/UpgradeManager.cpp
```diff
@@ -499,12 +514,14 @@ bool UpgradeManager::LoadInitialDS(vector<PubKey>& initialDSCommittee) {
       vector<std::string> tempDsComm_string{ReadDSCommFromFile()};
       initialDSCommittee.clear();
       for (auto ds_string : tempDsComm_string) {
-        initialDSCommittee.push_back(
-            PubKey(DataConversion::HexStrToUint8Vec(ds_string), 0));
+        bytes pubkeyBytes;
```
ansnunez: As most of the inputs here are read from the file (and so it can be possible for the file to be the source of error), maybe we need to add warning log messages wherever the dataconversion fails in this function

src/libNode/Node.cpp
ansnunez: Yeah should be sufficient I guess

PR #1100, opened 2018-12-24
Fix extra StateDelta being retrieved if TxBlock from DB is inconsistent

kaikawaliu:
## Description

1. Fix extra StateDelta being retrieved if TxBlock from DB is inconsistent
2. Fix GetBlock error e.g. `BlockNum : 3401 != GetBlockNum() : 3451, a dummy block will be used and abnormal behavior may happen!`
3. Fix unmatched type for shard_id when calling `SaveCoinBaseCore` in `SaveCoinBase` which caused integer underflow

### Implementation

- [ ] **ready for review**

- [x] local machine test (tested with new node join and rejoin after several ds epoch to replicate the scenario)

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/BlockChainData/BlockChain.h
+++ b/src/libData/BlockChainData/BlockChain.h
@@ -104,9 +104,9 @@ class BlockChain {
 
     if (blockNumOfExistingBlock < blockNumOfNewBlock ||
         INIT_BLOCK_NUMBER == blockNumOfExistingBlock) {
-      if (m_blocks.size() != 0) {
+      if (m_blocks.size() > 0) {
         uint64_t blockNumOfLastBlock =
-            m_blocks[blockNumOfNewBlock - 1].GetHeader().GetBlockNum();
+            m_blocks.back().GetHeader().GetBlockNum();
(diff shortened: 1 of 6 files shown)


ckyang:
Should fix travis failed first

src/libData/BlockChainData/BlockChain.h
```diff
@@ -104,6 +104,18 @@ class BlockChain {
 
     if (blockNumOfExistingBlock < blockNumOfNewBlock ||
         INIT_BLOCK_NUMBER == blockNumOfExistingBlock) {
+      if (m_blocks.size() != 0) {
```
KaustubhShamshery: can use empty

src/libData/BlockChainData/BlockChain.h
```diff
@@ -104,6 +104,18 @@ class BlockChain {
...
+        uint64_t blockNumOfLastBlock =
+            m_blocks[blockNumOfNewBlock - 1].GetHeader().GetBlockNum();
```
KaustubhShamshery: what if blockNumOfNewBlock -1 does not exist ? it will use dummy block (uint64_t - 1) ?

PR #1102, opened 2018-12-26
Verify Txns From S3

KaustubhShamshery:
## Description

The new seed node does not have historical txns, so it fetches the levelDB from s3 and a separate verification process is created. This process verifies the data fetched from s3 (see historicalDB.sh) using ValidateDB(). The process then relinquishes control of levelDB and hands it over to the lookup. 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNode/Node.cpp
```diff
@@ -239,6 +250,144 @@ void Node::AddGenesisInfo(SyncType syncType) {
   }
 }
 
+bool Node::ValidateDB() {
+  deque<pair<PubKey, Peer>> dsComm;
+  const string lookupIp = "127.0.0.1";
```
ShengguangXiao: Why use a local IP?

src/libNode/Node.cpp
```diff
@@ -239,6 +250,144 @@ void Node::AddGenesisInfo(SyncType syncType) {
...
+  }
+
+  if (!m_mediator.m_validator->CheckDirBlocks(dirBlocks, dsComm, 0, dsComm)) {
+    LOG_GENERAL(WARNING, "Fail to verify Dir Blocks");
+    return false;
+  }
+
+  vector<TxBlock> txBlocks;
+
+  for (auto txblock : txblocks) {
```
ShengguangXiao: use const auto& txBlock

src/libNode/Node.cpp
```diff
@@ -239,6 +250,144 @@ void Node::AddGenesisInfo(SyncType syncType) {
...
+  if (m_mediator.m_validator->CheckTxBlocks(txBlocks, dsComm,
+                                            blocklinks.back()) !=
+      ValidatorBase::TxBlockValidationMsg::VALID) {
+    LOG_GENERAL(WARNING, "Failed to verify TxBlocks");
+    return false;
+  }
+
+  for (uint i = 1; i < txBlocks.size(); i++) {
+    auto microblockInfos = txBlocks.at(i).GetMicroBlockInfos();
+    for (auto mbInfo : microblockInfos) {
```
ShengguangXiao: use const auto& mbInfo

src/libNode/Node.cpp
KaustubhShamshery: the verifier sends a  message  to inform the lookup machine (running on the same machine) that it has done the verification.

src/libNode/Node.cpp
KaustubhShamshery: the verifier sends a  message  to inform the lookup machine (running on the same machine) that it has done the verification.

ansnunez:
So the verifier process is another zilliqa node launched using the script historicalDB.sh?  Is the process terminated at some point?

KaustubhShamshery:
@ansnunez There is no easy way to terminate the process from within the code without using ugly sigkills

src/libNode/Node.cpp
```diff
@@ -243,6 +254,144 @@ void Node::AddGenesisInfo(SyncType syncType) {
...
+      dirBlocks;
+  for (const auto& blocklink : blocklinks) {
+    if (get<BlockLinkIndex::BLOCKTYPE>(blocklink) == BlockType::DS) {
+      auto blockNum = get<BlockLinkIndex::DSINDEX>(blocklink);
+      if (blockNum == 0) {
+        continue;
+      }
+      DSBlockSharedPtr dsblock;
+      if (!BlockStorage::GetBlockStorage().GetDSBlock(blockNum, dsblock)) {
+        LOG_GENERAL(WARNING, "Could not rertv DS Block " << blockNum);
```
ansnunez: "rertv" here, and "retrv" in other parts.  Maybe better to spell it out as "retrieve" so we don't have a hard time searching the logs.  Also, "Fail" and "Failed" both used.

src/libMessage/Messenger.cpp
```diff
@@ -6965,3 +6965,65 @@ bool Messenger::SetNodeGetNewDSGuardNetworkInfo(
 
   return true;
 }
+
+bool Messenger::SetSeedNodeHistoricalDB(
+    vector<unsigned char>& dst, const unsigned int offset,
```
ShengguangXiao: Replace vector<unsigned char> with bytes

src/libPersistence/BlockStorage.cpp
```diff
@@ -113,6 +113,27 @@ bool BlockStorage::PutMicroBlock(const BlockHash& blockHash,
...
+bool BlockStorage::GetTxnFromHistoricalDB(const dev::h256& key,
+                                          TxBodySharedPtr& body) {
+  std::string bodyString;
+
+  bodyString = m_txBodyDB->Lookup(key);
+
+  if (bodyString.empty()) {
+    return false;
+  }
+  body = TxBodySharedPtr(new TransactionWithReceipt(
```
ShengguangXiao: Use std::make_shared

PR #1107, opened 2018-12-26
 Change shardStructure db to store shards for every DS block number

ansnunez:
## Description

To have an easier way to retrieve sharding structure and track the history of each node across DS epochs, this PR modifies the shardStructure db:

`PutShardStructure`:
Previously: key = 0, value = my shard ID | key = 1, value = shards
Now: key = DS block number, value = my shard ID + shards

`GetShardStructure`:
Previously: returns value (shards) from key = 1
Now: returns value (shards) from latest entry in the db

A unit test has also been added: tests/Persistence/Test_ShardStructure.cpp

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPersistence/BlockStorage.cpp
+++ b/src/libPersistence/BlockStorage.cpp
@@ -533,53 +533,74 @@ bool BlockStorage::GetDSCommittee(
   return true;
 }
 
-bool BlockStorage::PutShardStructure(const DequeOfShard& shards,
-                                     const uint32_t myshardId) {
+bool BlockStorage::PutShardStructure(const uint64_t& blockNum,
+                                     const DequeOfShard& shards,
+                                     const uint32_t& myshardId) {
   LOG_MARKER();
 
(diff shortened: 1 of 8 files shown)


src/libPersistence/BlockStorage.cpp
```diff
@@ -533,53 +533,74 @@ bool BlockStorage::GetDSCommittee(
   return true;
 }
 
-bool BlockStorage::PutShardStructure(const DequeOfShard& shards,
-                                     const uint32_t myshardId) {
+bool BlockStorage::PutShardStructure(const uint64_t& blockNum,
```
ckyang: For backward-compatible, it's better to create some new API/DB to store the historical sharding structure, for both write/read sides.

ansnunez:
Closing this PR as a better solution has been implemented in #1112 after discussing with the team.

PR #1111, opened 2018-12-26
Feat/stress fungible token

iantanwx:
## Description

This PR adds on to the work done in https://github.com/Zilliqa/Zilliqa/pull/1050. It has two goals:
- ascertain a reasonably accurate picture of the time it takes to execute Scilla contracts with a large dataset (Map with >= 500k entries, nested `Map`s, multi-contract calls)
- provide an estimate of a reasonable microblock gas limit by simulating real-world usage of `FungibleToken`, `NonFungibleToken`, and `SimpleDEX`.

## Review Suggestion

- Checkout the branch and run the tests.

### Implementation

- [x] **Performance of `Transfer` on `FungibleToken` with 100-500k entries in `balances`**
- [x] **Performance of `TransferFrom` on `NonFungibleToken` with 100-500k entries in `balances`**
- [x] **Performance of `CreateOrder` on `SimpleDEX` with 100-500k hodlers in `token1` and `token2`, and 1000 orders in the order book**
- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -327,8 +327,8 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
                    runnerPrint)) {
       ret = false;
     }
-    LOG_GENERAL(DEBUG,
-                "Exec scilla-runner (microseconds) = " << r_timer_end(tnow));
+    LOG_GENERAL(DEBUG, "Executed root transition in " << r_timer_end(tnow)
+                                                      << " microseconds");
 
     if (ret && !ParseCallContract(gasRemained, runnerPrint)) {
       ret = false;
@@ -713,6 +713,7 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
     }
     return false;
   }
+  uint64_t startGas(gasRemained);
   gasRemained = atoi(_json["gas_remaining"].asString().c_str());
 
   if (!_json.isMember("_accepted")) {
@@ -778,6 +779,7 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
   }
 
   LOG_GENERAL(DEBUG, "LDB Write (microseconds) = " << r_timer_end(tnow));
+  LOG_GENERAL(DEBUG, "Gas used = " << (startGas - gasRemained));
 
   for (const auto& e : _json["events"]) {
     LogEntry entry;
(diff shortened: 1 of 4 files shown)


tests/Data/Test_Contract.cpp
```diff
@@ -427,4 +429,128 @@ BOOST_AUTO_TEST_CASE(testPingPong) {
...
+    Account* account = AccountStore::GetInstance().GetAccount(contrAddr);
+    // We should now have a new account.
+    BOOST_CHECK_MESSAGE(account != nullptr,
+                        "Error with creation of contract account");
+    nonce++;
+
+    // 2. Pre-generate and save a large map and save it to LDB
+    std::string initOwnerBalance;
+    for (unsigned int i = 0; i < hodlers; i++) {
+      unsigned char* hodler = new unsigned char[20];
```
vaivaswatha: I don't see `hodler` being deleted anywhere. Maybe you can just have a stack vector (`hodlerAddress` that you already have below) and pass it's `data()` to `RAND_Bytes`.

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -580,7 +602,14 @@ bool AccountStoreSC<MAP>::ParseCreateContractOutput(
     outStr = {std::istreambuf_iterator<char>(in),
               std::istreambuf_iterator<char>()};
   }
-  LOG_GENERAL(INFO, "Output: " << std::endl << outStr);
+
+  LOG_GENERAL(
```
ansnunez: You can avoid the check against max and just call `substr` in both cases, because `substr` will still work if you have less than max (http://www.cplusplus.com/reference/string/string/substr/)

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -695,6 +736,7 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
     }
     return false;
   }
+  uint64_t startGas(gasRemained);
```
ansnunez: Minor comment, maybe change to `uint64_t startGas = gasRemained;` just for coding consistency

tests/Data/Test_Contract.cpp
```diff
@@ -436,4 +444,878 @@ BOOST_AUTO_TEST_CASE(testPingPong) {
...
+  const uint128_t bal{std::numeric_limits<uint128_t>::max()};
+  uint64_t nonce = 0;
+  const unsigned int numDeployments = 10000;
+  const unsigned int numMapEntries = 1000;
+
+  ofstream report;
+  report.open("perf_report.csv");
+  report << "deployment_microsec,deployment_gas,invoke_microsec,invoke_gas\n";
+
+  INIT_STDOUT_LOGGER();
```
ansnunez: Might be better to move `INIT` and `MARKER` to the start of each function

src/libData/AccountData/AccountStoreSC.tpp
iantanwx: I used this check so that ellipses can be appended if the output string indeed has to be truncated.

ansnunez: Ok

PR #1113, opened 2018-12-27
Unify the type for consensusMyID and consensusLeaderID

kaikawaliu:
### Implementation

- [x] **ready for review**

- [x] small-scale cloud test


ShengguangXiao:
Can we change the consensusMyID and consensusLeaderID to private member and use get/set function to access it?

kaikawaliu:
> Can we change the consensusMyID and consensusLeaderID to private member and use get/set function to access it?

We can do this in another pr, so far this pr is not scoped on that

PR #1114, opened 2018-12-27
DS difficulty mining start with shard difficulty winning nonce

ShengguangXiao:
## Description

DS difficulty mining start with shard difficulty winning nonce

It is because the DS PoW start nonce use the current time, it has overlap with shard PoW nonce. The DS PoW should start from the nonce of the shard pow winning nonce, so more node can finish DS PoW, then there will be more nodes finish DS PoW, then the DS difficulty will be higher than shard difficulty.

https://github.com/Zilliqa/Issues/issues/389

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -321,11 +321,11 @@ ethash_hash256 POW::GenHeaderHash(
     const std::array<unsigned char, UINT256_SIZE>& rand2,
     const boost::multiprecision::uint128_t& ipAddr, const PubKey& pubKey,
     uint32_t lookupId, const boost::multiprecision::uint128_t& gasPrice) {
-  bytes sha3_result =
+  bytes sha2_result =
       ConcatAndhash(rand1, rand2, ipAddr, pubKey, lookupId, gasPrice);
 
   // Let's hash the inputs before feeding to ethash
-  return StringToBlockhash(DataConversion::Uint8VecToHexStr(sha3_result));
+  return StringToBlockhash(DataConversion::Uint8VecToHexStr(sha2_result));
 }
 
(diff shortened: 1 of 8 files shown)


ShengguangXiao:
Tested with 200 nodes, now the DS difficulty will already higher than shard difficulty

src/libPOW/pow.cpp
```diff
@@ -313,55 +316,50 @@ bytes POW::ConcatAndhash(const std::array<unsigned char, UINT256_SIZE>& rand1,
...
-ethash_mining_result_t POW::PoWMine(
-    uint64_t blockNum, uint8_t difficulty,
+ethash_hash256 POW::GenHeaderHash(
     const std::array<unsigned char, UINT256_SIZE>& rand1,
     const std::array<unsigned char, UINT256_SIZE>& rand2,
     const boost::multiprecision::uint128_t& ipAddr, const PubKey& pubKey,
-    uint32_t lookupId, const boost::multiprecision::uint128_t& gasPrice,
-    bool fullDataset) {
+    uint32_t lookupId, const boost::multiprecision::uint128_t& gasPrice) {
+  bytes sha3_result =
```
bb111189: sha2 not sha3

src/libPOW/pow.cpp
```diff
@@ -313,55 +316,50 @@ bytes POW::ConcatAndhash(const std::array<unsigned char, UINT256_SIZE>& rand1,
...
+      ConcatAndhash(rand1, rand2, ipAddr, pubKey, lookupId, gasPrice);
+
+  // Let's hash the inputs before feeding to ethash
+  return StringToBlockhash(DataConversion::Uint8VecToHexStr(sha3_result));
```
bb111189: sha2 not sha3

PR #1118, opened 2018-12-27
Periodic cleanup of gossip raw message store

sandipbhoir:
## Description
This PR addressess unbounded growth of below two storages:

- **m_rumorHashRawMsgBimap:** maps rumor hash to the full rumor message content
- **m_rumorIdHashBimap:** maps sequential rumorId integers to rumor message hashes

Below two need not to considered:
- **m_tmpRumorHashSet:** temporary rumor hash store to ensure duplicate rumors are not
distributed - **No more used in our codebase.**
- **m_hashesSubscriberMap:** if a node is asked for a message he does not have, the asking
node gets subscribed to this message (entries inside this map get deleted once subscribed
messages are delivered) - **which is good enough to control the growth of this container** 

## Review Suggestion

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libNetwork/RumorManager.cpp
```diff
@@ -516,4 +539,31 @@ void RumorManager::PrintStatistics() {
                               << " ], " << state);
     }
   }
+}
+
+void RumorManager::CleanUp() {
+  int count = 0;
+  auto now = std::chrono::high_resolution_clock::now();
+  while (not m_rumorRawMsgTimestamp.empty()) {
```
ansnunez: Please use `!` instead of `not`

src/libNetwork/RumorManager.cpp
```diff
@@ -152,6 +159,11 @@ bool RumorManager::Initialize(const std::vector<Peer>& peers,
     m_rumorHolder.reset(new RRS::RumorHolder(m_peerIdSet, 0));
   }
 
+  RAW_MESSAGE_EXPIRY_IN_MS = (KEEP_RAWMSG_FROM_LAST_N_ROUNDS < MAX_TOTAL_ROUNDS)
```
ansnunez: Maybe add a comment to explain the values used here?

src/libNetwork/RumorManager.cpp
```diff
@@ -516,4 +539,31 @@ void RumorManager::PrintStatistics() {
...
+
+void RumorManager::CleanUp() {
+  int count = 0;
+  auto now = std::chrono::high_resolution_clock::now();
+  while (not m_rumorRawMsgTimestamp.empty()) {
+    auto elapsed_milliseconds =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            now - m_rumorRawMsgTimestamp.front().second)
+            .count();
+    LOG_GENERAL(DEBUG, "elapsed_milliseconds:" << elapsed_milliseconds
```
ansnunez: Will this log too many messages?  If yes, should we replace with
```
if (DEBUG_LEVEL >= 5) {
        LOG_GENERAL(INFO, ...);
}
```

src/libNetwork/RumorManager.cpp
```diff
@@ -196,18 +208,25 @@ bool RumorManager::AddRumor(const RumorManager::RawBytes& message) {
       m_rumorIdHashBimap.insert(
           RumorIdRumorBimap::value_type(++m_rumorIdGenerator, hash));
 
-      m_rumorHashRawMsgBimap.insert(
-          RumorHashRumorBiMap::value_type(hash, message));
+      std::pair<RumorHashRumorBiMap::iterator, bool> result =
```
ShengguangXiao: Can use auto result = .....

src/libNetwork/RumorManager.h
```diff
@@ -58,13 +58,18 @@ class RumorManager {
   RumorHashesPeersMap m_hashesSubscriberMap;
   Peer m_selfPeer;
   std::vector<RawBytes> m_bufferRawMsg;
+  std::deque<std::pair<RumorHashRumorBiMap::iterator,
```
ShengguangXiao: Can define a alias type for this?  It is really quite long

src/libNetwork/RumorManager.h
```diff
@@ -58,13 +58,18 @@ class RumorManager {
...
+                       std::chrono::high_resolution_clock::time_point>>
+      m_rumorRawMsgTimestamp;
 
   int64_t m_rumorIdGenerator;
   std::mutex m_mutex;
   std::mutex m_continueRoundMutex;
   std::atomic<bool> m_continueRound;
   std::condition_variable m_condStopRound;
 
+  int32_t RAW_MESSAGE_EXPIRY_IN_MS;
```
ShengguangXiao: If it is not a constant value, it is better use m_rawMessageExpiryInMs

src/libNetwork/RumorManager.cpp
sandipbhoir: Fixed.

PR #1119, opened 2018-12-27
Change error signaling in libCrypto

jendis:
## Description
Add throw exception or return status for libCrypto when initialization or update fails when input parameters are incorrect.
Especially useful for more precise input parameter validation in executables.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


bb111189:
For this, we will need to calling function to gracefully handle the exception

PR #1122, opened 2018-12-28
 Develop remote mining using mining proxy

ShengguangXiao:
## Description

Develop remote mining using mining proxy

Add REMOTE_MINE option in constants.xml, if it is enabled, the node will send the pow work package to mining proxy, and GPU miner will get the work package from the proxy. After the miner finish PoW, it will send result to the proxy, and the zilliqa node get the result from proxy.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


ShengguangXiao:
Run 200 nodes using mining proxy to do pow run to 100 tx blocks without problem

src/libPOW/pow.cpp
```diff
@@ -238,8 +242,250 @@ ethash_mining_result_t POW::MineFullGPU(uint64_t blockNum,
...
+
+  ethash_mining_result_t miningResult{"", "", 0, false};
+  if (!SendWorkToProxy(pairOfKey, blockNum, headerHash, boundary)) {
+    LOG_GENERAL(WARNING, "Failed to send work package to mining proxy.");
+    return miningResult;
+  }
+
+  bool verifyResult = CheckMiningResult(pairOfKey, blockNum, headerHash,
+                                        boundary, miningResult);
+  if (!verifyResult) {
```
KaustubhShamshery: return from here??

ShengguangXiao: No, if verify fail, also need to send the result to Proxy, let proxy know it is a invalid solution.

src/libPOW/pow.cpp
```diff
@@ -238,8 +242,250 @@ ethash_mining_result_t POW::MineFullGPU(uint64_t blockNum,
...
+    try {
+      jsonrpc::Client client(*m_httpClient);
+      Json::Value ret = client.CallMethod("zil_checkWorkStatus", jsonValue);
+      LOG_GENERAL(INFO, "zil_checkWorkStatus return: " << ret);
+      bool workDone = ret[0].asBool();
+      if (!workDone) {
+        continue;
+      }
+
+      if (ret.size() < 4) {
```
KaustubhShamshery: can this four be some constant ? like NUM_JSON_ELEMENTS

ShengguangXiao: No need, this is defined in the communication protocal, https://docs.google.com/document/d/13KwqovblQ4gsZnXMsP6Lw7Yl3F5uzD6VqBheHgXB1wQ/edit

KaustubhShamshery: yeah, but i feel it is better as a constant, maybe one defined in this file. no need to define in constants.xml

ShengguangXiao: Ok, i use constant value already

src/libPOW/pow.cpp
```diff
@@ -238,8 +242,252 @@ ethash_mining_result_t POW::MineFullGPU(uint64_t blockNum,
...
+  bytes pubKeyData;
+  pairOfKey.second.Serialize(pubKeyData, 0);
+  jsonValue[0] = "0x" + BytesToHexString(pubKeyData.data(), pubKeyData.size());
+  tmp.insert(tmp.end(), pubKeyData.begin(), pubKeyData.end());
+
+  jsonValue[1] = "0x" + POW::BlockhashToHexString(headerHash);
+  tmp.insert(tmp.end(), headerHash.bytes,
+             headerHash.bytes + sizeof(ethash_hash256));
+
+  auto strBlcokNumber =
```
bb111189: strBlcokNumber -> strBlockNumber

ShengguangXiao: @bb111189 Fixed already

PR #1125, opened 2018-12-28
Signature verification on Gossip FORWARD Message type sent via SendRumorToForeignPeer

sandipbhoir:
## Description
This PR addresses issue raised in https://github.com/Zilliqa/Issues/issues/384

```
SendRumorToForeignPeer()
The original intention of this feature is to gossip across shard, to and fro ds committee and shard. i.e

    ~~Shard x to y~~ <-- We dont have this use case
    DS committee to shard
    Shard to DS committee

However, this function also allows external IP to gossip into the Zilliqa node. As such, guard mode leverage this unintentional feature.

    DS member (with changed IP) to DS committee

This method is problematic as it allows anyone in the internal to flood our network easily and choked up the threadpool.
```

So, we follow the below approach :
- Rumor Manager is initialized with PubKeys of entire network ( except that shard nodes won't PubKeys of nodes from other shard )
i.e.  - DS nodes will have PubKeys from DS committee  + all other shard nodes + lookup nodes
       - Shard nodes will have PubKeys from DS committee + all nodes from his shard + lookup nodes

- Node sending the FORWARD type gossip message via SendRumorToForeignPeer api, will to sign the raw message and add it to final message along with its public key.

- Node receiving the FORWARD type gossip message will verify if the PubKey in message exist in his store list. If Yes, we verify the signature of message using the pubkey received. Otherwise or in case of signature verification failure, the message received is discarded even before gossiping.

This will make gossip protocol robust against external attacker who target the SendRumorToForeignPeer api but at same time allow  DS member (with changed IP) to send message to DS committee.

Internal attacker is not in scope of this PR. 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNode/Node.cpp
```diff
@@ -1918,3 +1912,27 @@ std::string Node::GetActionString(Action action) const {
...
+
+void Node::GetEntireNetworkPeerInfo(std::vector<std::pair<PubKey, Peer>>& peers,
+                                    std::vector<PubKey>& pubKeys) {
+  peers.clear();
+  pubKeys.clear();
+
+  for (const auto& i : *m_myShardMembers) {
+    if (i.second.m_listenPortHost != 0) {
+      peers.emplace_back(i);
+    }
```
ansnunez: Can add `continue` and skip adding my own pubkey?

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -1087,4 +1081,28 @@ uint64_t DirectoryService::CalculateNumberOfBlocksPerYear() const {
...
+
+void DirectoryService::GetEntireNetworkPeerInfo(
+    std::vector<std::pair<PubKey, Peer>>& peers, std::vector<PubKey>& pubKeys) {
+  peers.clear();
+  pubKeys.clear();
+
+  for (const auto& i : *m_mediator.m_DSCommittee) {
+    if (i.second.m_listenPortHost != 0) {
+      peers.emplace_back(i);
+    }
```
ansnunez: Can add `continue` and skip adding my own pubkey?

src/libNode/Node.cpp
sandipbhoir: Fixed. Adding `continue` is not relevant though.

src/libDirectoryService/DirectoryService.cpp
sandipbhoir: Fixed. Adding `continue` is not relevant though.

PR #1127, opened 2018-12-31
Updated diagnostic data retrieval

ansnunez:
## Description

This PR adds a more convenient way to read out the contents of the diagnostic database.

The current way is not so easy to use as you have to iterate thru the db by inputting the block number.
Also, the database count isn't stored in the db, so it is zero when you launch your code (not a major issue right now, and will be addressed later).

With the new function in this PR, we can just dump out all the contents of the diagnostic data database for our monitoring tools to easily use.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPersistence/BlockStorage.cpp
+++ b/src/libPersistence/BlockStorage.cpp
@@ -683,7 +683,15 @@ void BlockStorage::GetDiagnosticData(
       continue;
     }
 
-    uint64_t dsBlockNum = stoul(dsBlockNumStr);
+    uint64_t dsBlockNum = 0;
+    try {
+      dsBlockNum = stoul(dsBlockNumStr);
+    } catch (...) {
+      LOG_GENERAL(WARNING,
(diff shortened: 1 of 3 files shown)


src/libPersistence/BlockStorage.cpp
```diff
@@ -663,6 +663,45 @@ bool BlockStorage::GetDiagnosticData(const uint64_t& dsBlockNum,
...
+    string dsBlockNumStr = it->key().ToString();
+    string dataStr = it->value().ToString();
+
+    if (dsBlockNumStr.empty() || dataStr.empty()) {
+      LOG_GENERAL(WARNING,
+                  "Failed to retrieve diagnostic data at index " << index);
+      continue;
+    }
+
+    uint64_t dsBlockNum = stoul(dsBlockNumStr);
```
ShengguangXiao: stoul may throw exception, should we put it into try... catch ... ?

PR #1131, opened 2018-12-31
Diagnostic Tool to show node movements between ds epochs

robin-thomas:
## Description

This tool reads the diagnostic DB and outputs a CSV formatted file, which shows the movement of an ip node (whether it's in ds committee or if it's in a shard, the shard index and the index within the shard) between various ds epochs

Make sure that the tool is run when the lookup node is not running.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/diagnostic/getnetworkhistory.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    output += row + "\n";
+  }
+}
+
+int main(int argc, char** argv) {
+  if (argc < 2) {
+    std::cout << "[USAGE] " << argv[0]
+              << " <output csv filename> <epoch width required>" << std::endl;
```
ansnunez: I think `<epoch width required>` is unused and can be removed?

ansnunez: I think it would be good to include the location of the database as a parameter

robin-thomas:
@Gnnng 
> It seems the Travis build had some issues. I will check further on this PR once it's been fixed.

Yeah. It was Travis issue.

> One comment for this PR is it's better to move this utility to `src/cmd`.

I did have that discussion with Antonio. Since there are plans to make many diagnostic tools, thought it would be better to place it not in `cmd`.

src/depends/libDatabase/LevelDB.cpp
```diff
@@ -87,21 +82,13 @@ LevelDB::LevelDB(const string & dbName, const string & subdirectory)
     leveldb::DB* db;
     leveldb::Status status;
 
-    if(m_subdirectory.empty())
+    string db_path = diagnostic ? ((m_subdirectory + (m_subdirectory.back() == '/' ? "" : "/")) + PERSISTENCE_PATH) : ("./"  + PERSISTENCE_PATH + (m_subdirectory.empty() ? "" : "/" + m_subdirectory));
```
ansnunez: To simplify this long statement, maybe remove the need for checking `m_subdirectory.back` by making sure getnetworkhistory removes any trailing slash before calling this constructor.

Also, might be good to put a comment here about why we handle diagnostic differently.  You can say that the bool is only ever true when used by the tool, and that the non-diagnostic path is preserved from the original code.

src/diagnostic/getnetworkhistory.cpp
robin-thomas: Done!

src/depends/libDatabase/LevelDB.cpp
robin-thomas: Done!

PR #1133, opened 2018-12-31
Add CV after submitting PoW during joining to check whether a new tx was mined or rejoined

kaikawaliu:
## Description

This will solve the issue if there is a vc during ds consensus, then the sleep (previously exists) times out causing the state check failed during processing vcds block message if received too late

Besides, DS rejoin is found not working under current master during testing, which is also fixed here. 

### Implementation

- [x] **ready for review**

- [x] local machine test (tested with new node join & ds rejoin)


src/libLookup/Lookup.cpp
```diff
@@ -2214,16 +2215,20 @@ bool Lookup::InitMining(uint32_t lookupIndex) {
...
-  // re-sync process again.
-  this_thread::sleep_for(chrono::seconds(
-      POW_WINDOW_IN_SECONDS + POWPACKETSUBMISSION_WINDOW_IN_SECONDS +
-      2 * NEW_NODE_SYNC_INTERVAL + (TX_DISTRIBUTE_TIME_IN_MS / 1000)));
-  m_startedPoW = false;
-  if (GetSyncType() != SyncType::NO_SYNC) {
-    LOG_EPOCH(INFO, to_string(m_mediator.m_currentEpochNum).c_str(),
-              "Not yet connected to network");
-    m_mediator.m_node->SetState(Node::SYNC);
+  uint64_t lastTxBlockNum =
```
KaustubhShamshery: m_startedPoW = false seems missing

kaikawaliu: Thanks

src/libNode/Node.h
```diff
@@ -65,7 +65,9 @@ class Node : public Executable, public Broadcastable {
   enum REJOINTYPE : unsigned char {
     ATFINALBLOCK = 0x00,
     ATNEXTROUND = 0x01,
-    ATSTATEROOT = 0x02
+    ATSTATEROOT = 0x02,
+    ATDSCONSENSUS = 0x03,   // For DS Rejoin
+    ATFINALCONSENSUS = 04,  // For DS Rejoin
```
KaustubhShamshery: 0x04

src/libDirectoryService/DirectoryService.h
```diff
@@ -535,6 +535,9 @@ class DirectoryService : public Executable, public Broadcastable {
       m_lookupStoreForGuardNodeUpdate;
   std::atomic_bool m_awaitingToSubmitNetworkInfoUpdate = {false};
 
+  bool m_doRejoinAtDSConsensus = false;
+  bool m_doRejoinAtFinalConsensus = false;
```
ShengguangXiao: Add data members as private, use function to access them

kaikawaliu: These two variables is only for debugging purpose. Will be removed eventually, it would be a little bit redundant to do so.

PR #1136, opened 2019-01-02
Sign and Verfiy Rest of gossip message type

sandipbhoir:
## Description
This PR covers following:

- All Gossip message are signed and verified before sending and receiving.
- Provide an option to enable/disable signing/verifying for two categories of gossip message. 
  i.e. Empty Push / Empty Pull message type and  Rest of NonEmpty message types.
- Additional check to verify if the pubkey in message actually maps to sender host ip/port. To handle case where sender got hold of some old random message and try to flood it later.

### Implementation

- [x] All Gossip message are signed and verified before sending and receiving.
- [x]  Provide an option to enable/disable signing/verifying for two categories of gossip message
- [x] Additional check to verify if the pubkey in message actually maps to sender host ip/port. 
- [x] Code refactor/ cleanup

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test - enable sig verification on both categories of msg type


src/libNetwork/P2PComm.cpp
```diff
@@ -469,16 +469,18 @@ void P2PComm::ClearBroadcastHashAsync(const bytes& message_hash) {
...
-  } else if (p2p.m_rumorManager.RumorReceived((unsigned int)gossipMsgTyp,
-                                              gossipMsgRound, rumor_message,
-                                              from)) {
-    std::pair<bytes, Peer>* raw_message =
-        new pair<bytes, Peer>(rumor_message, from);
+  } else {
+    auto resp = p2p.m_rumorManager.RumorReceived(
+        (unsigned int)gossipMsgTyp, gossipMsgRound, rumor_message, from);
+    if (resp.first) {
+      std::pair<bytes, Peer>* raw_message =
```
bb111189: can we try to use smart pointer here?

sandipbhoir: Tried to reuse the existing approach that we infact convert the smart_pointer to raw pointer before passing on to dispatcher 
https://github.com/Zilliqa/Zilliqa/blob/d6e7a2a24c5c56283b671d73a5db08782569febf/src/libNetwork/P2PComm.cpp#L606
Let me know your view.

PR #1137, opened 2019-01-02
Implement timeout for contract transition transaction

kaikawaliu:
## Description

Interrupt process for invoking scilla_checker/runner (interpreters) if a set timeout get triggered, this can avoid the protocol being stuck by processing huge contract contract while the gas limit may not cover.

A new popen function was implemented to get the pid of calling the interpreter, and a kill of negative pid was called once the timeout was triggered, which is to kill the process group with such pid, usually the process group id is named by the pid of the first process in the group.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -186,21 +191,78 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
...
 
     // Undergo scilla runner
     std::string runnerPrint;
-    if (ret && !SysCommand::ExecuteCmdWithOutput(
-                   GetCreateContractCmdStr(m_root_w_version, gasRemained),
-                   runnerPrint)) {
-      ret = false;
+    if (ret) {
+      int pid = -1;
+      auto func2 = [this, &runnerPrint, &ret, &pid,
```
ansnunez: This lambda seems to be used 3x in different places.  Might be better to refactor it as a function, for easier maintenance.

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -263,6 +263,27 @@ void Node::ProcessTransactionWhenShardLeader() {
   t_processedTransactions.clear();
   m_TxnOrder.clear();
 
+  bool txnProcTimeout = false;
+
+  auto txnProcTimer = [this, &txnProcTimeout]() -> void {
```
ansnunez: Same comment for this lambda

src/libData/AccountData/AccountStoreSC.tpp
kaikawaliu: no...they are all slightly different

src/libUtils/SysCommand.h
```diff
@@ -19,17 +19,86 @@
...
+#include <boost/algorithm/string/split.hpp>
+
 #include "Logger.h"
 
+#define READ 0
+#define WRITE 1
+
 class SysCommand {
  public:
+  static FILE* popen2(std::string command, std::string type, int& pid) {
```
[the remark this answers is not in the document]
kaikawaliu: Changed the name to popen_with_pid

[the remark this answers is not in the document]
bb111189: can we avoid using popen? Use execl instead to avoid running commands using a shell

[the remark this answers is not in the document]
kaikawaliu: it's not popen even has similar functionality actually. it's customized implemented via pipe, where you can also find execl command inside

src/libUtils/SysCommand.h
```diff
@@ -19,17 +19,86 @@
...
+      if (errno != EINTR) {
+        stat = -1;
+        break;
+      }
+    }
+
+    return stat;
+  }
+
   static bool ExecuteCmdWithoutOutput(std::string cmd) {
```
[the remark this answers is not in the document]
kaikawaliu: implemented

src/libUtils/SysCommand.h
```diff
@@ -19,17 +19,89 @@
...
 #include "Logger.h"
 
+#define READ 0
+#define WRITE 1
+
+static std::string cmd_output_place_holder;
+static int pid_place_holder;
+
 class SysCommand {
  public:
```
ckyang: The modification is fine. Since now only ExecuteCmd() be used, how about make other functions private?

kaikawaliu: static can't call member function, unless we make the syscommand a singleton class

ckyang: ok

PR #1146, opened 2019-01-07
Avoid asking for data from multiplier but only from lookup

sandipbhoir:
## Description
Zilliqa/Issues#392

The multiplier is treated as lookup in constants.xml and the chances are some node will send messages to the multiplier and expect responses.
When selecting lookup nodes to query, it should rule out the multiplier and never treat them as lookup node. It should only send messages to multiplier when no response is expected.

~~It is observed that PubKey is incorrectly considered as initialized which is currently blocking in identified the multiplier as uninitialized which was intended to be filtering character while sending message to lookup.~~

~~SetLookupNodes - Now we set empty PubKey() for Multiplier.
And expected to be unitialized when checked in SendMessageToLookupNodesSerial and SendMessageToRandomLookupNode. However, it happen to be initialized due to issue explained above (Details can be found here - Zilliqa/Issues#401 )~~

New approach:
Separate multiplier ip from lookup IPs - see constants.xml
Skip the m_multipliers from m_lookups while sending to random lookup nodes and lookupNodeSerials

Corresponding testnet changes - PR# https://github.com/Zilliqa/testnet/pull/249

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -130,7 +130,7 @@ void Lookup::SetLookupNodes() {
   ptree pt;
   read_xml("constants.xml", pt);
 
-  const vector<string> lookupTypes = {"node.lookups", "node.upper_seed",
+  const vector<string> lookupTypes = {"node.lookups", "node.multipliers",
```
KaustubhShamshery: remove lower_seed

PR #1159, opened 2019-01-09
Add getwork APIs for remote mining

deepgully:
## Description
Add ETH like getwork APIs to support remote mining on local GPU rigs.

Add a new mine option 'GETWORK_SERVER_MINE', when this option is true, the Node will run a JSONRPC server and support getwrok APIs: eth_getWork, eth_submitWork, and eth_submitHashrate(do nothing).

A GPU miner can connect to the Node using getWork protocol.

### Implementation
- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ShengguangXiao:
I locally run 20 nodes test with one node enable GETWORK_SERVER_MINE option and an ethminer connect to it, the node can delegate the PoW to ethminer, and the testnet can run.

constants_local.xml
```diff
@@ -141,6 +141,9 @@
         <MINING_PROXY_URL>http://127.0.0.1:4202/api</MINING_PROXY_URL>
         <!-- Every interval seconds to check if mining proxy has the PoW result -->
         <CHECK_MINING_RESULT_INTERVAL>5</CHECK_MINING_RESULT_INTERVAL>
+        <!-- Do PoW via getWork Server -->
```
ShengguangXiao: Change comment so  Make zilliqa node as an getWork server

src/libPOW/pow.cpp
```diff
@@ -157,7 +166,8 @@ bool POW::EthashConfigureClient(uint64_t block_number, bool fullDataset) {
     m_epochContextLight = ethash::create_epoch_context(epochNumber);
   }
 
-  bool isMineFullCpu = fullDataset && !CUDA_GPU_MINE && !OPENCL_GPU_MINE;
+  bool isMineFullCpu =
+      fullDataset && !CUDA_GPU_MINE && !OPENCL_GPU_MINE && !GETWORK_SERVER_MINE;
```
ShengguangXiao: Change to bool isMineFullCpu =
      fullDataset && !CUDA_GPU_MINE && !OPENCL_GPU_MINE && !GETWORK_SERVER_MINE && !REMOTE_MINE;

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,329 @@
...
+      return FAIL_RESULT;
+    }
+    if (boundary != m_curWork.boundary) {
+      LOG_GENERAL(WARNING, "Submit boundary diff with current work");
+      LOG_GENERAL(WARNING, "Current boundary: " << m_curWork.boundary);
+      LOG_GENERAL(WARNING, "Submit boundary: " << boundary);
+      return FAIL_RESULT;
+    }
+
+    auto header_hash = POW::StringToBlockhash(header);
```
ShengguangXiao: This function can replace with VerifyRemoteSoln function

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,329 @@
...
+using namespace std;
+using namespace jsonrpc;
+
+//////////////////////////////////////////////////
+// Helper functions
+//////////////////////////////////////////////////
+
+static ethash_mining_result_t FAIL_RESULT = {"", "", 0, false};
+
+bool HexStringToUint64(const std::string& s, uint64_t* res) {
```
ShengguangXiao: HexStringToUint64 and NormalizeHexString can add to DataConversion class. And add a few unit test in Test_DataConversion.cpp

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,329 @@
...
+  // wait_ms < 0
+  // wait until the first accept result
+  std::unique_lock<std::mutex> lk(m_mutexResult);
+  while (m_isMining && !m_curResult.success) {
+    m_cvGotResult.wait(lk);
+
+    // wait for a while miner is working
+    lk.unlock();
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    lk.lock();
```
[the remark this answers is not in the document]
deepgully: I am not sure the sleep will release the lock as expected.

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,329 @@
...
+ethash_mining_result_t GetWorkServer::GetResult(const int& wait_ms) {
+  if (wait_ms >= 0) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(wait_ms));
+    return m_curResult;
+  }
+
+  // wait_ms < 0
+  // wait until the first accept result
+  std::unique_lock<std::mutex> lk(m_mutexResult);
+  while (m_isMining && !m_curResult.success) {
```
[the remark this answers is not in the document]
deepgully: add m_isMining here to let other thread can stop the mining anytime.

src/libPOW/pow.cpp
deepgully: got it.

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+// VerifySubmit verify and create ethash result
+ethash_mining_result_t GetWorkServer::VerifySubmit(const string& nonce,
+                                                   const string& header,
+                                                   const string& mixdigest,
+                                                   const string& boundary) {
+  uint64_t winning_nonce = {0};
+
+  // convert
+  if (HexStringToUint64(nonce, &winning_nonce)) {
```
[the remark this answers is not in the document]
deepgully: got it

src/libServer/GetworkServer.cpp
ShengguangXiao: HexStringToUint64 can implemented as a more general template function HexStringToIngeter? If it is too hard, then just leave it.

src/libServer/GetworkServer.cpp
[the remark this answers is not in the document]
ShengguangXiao: If ethminer never submit pow result, it will wait for the result forever. Because only StopMining only called after it received DS block

src/libZilliqa/Zilliqa.cpp
```diff
@@ -290,6 +291,21 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, bool loadConfig,
     if (!LOOKUP_NODE_MODE) {
       LOG_GENERAL(INFO, "I am a normal node.");
 
+      if (GETWORK_SERVER_MINE) {
+        LOG_GENERAL(INFO, "Starting GetWork Mining Server at http://"
+                              << string(peer.GetPrintableIPAddress()) << ":"
```
ansnunez: Explicit cast to `string` is not necessary

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  return result;
+}
+
+bool GetWorkServer::submitWork(const string& _nonce, const string& _header,
+                               const string& _mixdigest,
+                               const string& _boundary,
+                               const string& _miner_wallet,
+                               const string& _worker) {
+  (void)_miner_wallet;
```
ansnunez: Unused parameters can be tagged with `[[gnu::unused]]` instead of doing this

src/libServer/GetworkServer.cpp
```diff
@@ -0,0 +1,286 @@
...
+      !DataConversion::NormalizeHexString(header) ||
+      !DataConversion::NormalizeHexString(mixdigest) ||
+      !DataConversion::NormalizeHexString(boundary)) {
+    LOG_GENERAL(WARNING, "Invalid input parameters");
+    return false;
+  }
+
+  auto result = VerifySubmit(nonce, header, mixdigest, boundary);
+
+  bool accepted = UpdateCurrentResult(result);
```
ansnunez: Sometimes our clang complains of unnecessary temp variables, I wonder why it didn't complain here.  But still you can directly return `UpdateCurrentResult` instead of going through `accepted`

src/libServer/GetworkServer.h
```diff
@@ -0,0 +1,166 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#ifndef __POW_GETWORK_SERVER_H__
```
ansnunez: To follow our practice, please remove `POW` so that it matches the name of the header file

src/libNode/PoWProcessing.cpp
```diff
@@ -170,6 +170,9 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
                                   fixedDSBlockDistributionDelayTime +
                                   extraWaitTime)) == cv_status::timeout) {
         lock_guard<mutex> g(m_mutexDSBlock);
+
+        POW::GetInstance().StopMining();
```
ansnunez: @ShengguangXiao is this move ok? It looks like it's a bug fix (stopmining would not have been called previously if DS block was processed during time out).

PR #1167, opened 2019-01-10
Make the seed node port dynamic and functionality for fetching micro block

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libPersistence/BlockStorage.cpp
+++ b/src/libPersistence/BlockStorage.cpp
@@ -123,7 +123,6 @@ bool BlockStorage::GetTxnFromHistoricalDB(const dev::h256& key,
                                           TxBodySharedPtr& body) {
   std::string bodyString;
 
-
   bodyString = m_txnHistoricalDB->Lookup(key);
 
   if (bodyString.empty()) {
--- a/src/libPersistence/BlockStorage.cpp
+++ b/src/libPersistence/BlockStorage.cpp
@@ -122,7 +122,7 @@ bool BlockStorage::GetTxnFromHistoricalDB(const dev::h256& key,
                                           TxBodySharedPtr& body) {
   std::string bodyString;
 
-  bodyString = m_txBodyDB->Lookup(key);
+  bodyString = m_historicalDB->Lookup(key);
 
   if (bodyString.empty()) {
     return false;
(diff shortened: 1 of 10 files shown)


src/libServer/Server.cpp
```diff
@@ -1106,3 +1106,49 @@ string Server::GetNumTxnsDSEpoch() {
...
+  if (txBlock == TxBlock()) {
+    throw JsonRpcException(RPC_INVALID_PARAMS, "Tx Block does not exist");
+  }
+
+  auto microBlockInfos = txBlock.GetMicroBlockInfos();
+
+  for (auto const& mbInfo : microBlockInfos) {
+    MicroBlockSharedPtr mbptr;
+    if (mbInfo.m_txnRootHash == TxnHash() && mbInfo.m_shardId == shardID) {
+      throw JsonRpcException(RPC_INVALID_PARAMETER, "Shard has no txns empty");
```
ansnunez: Seems like `RPC_MISC_ERROR` is a better code? Also, maybe can just say "Shard has no txns"

src/libPersistence/BlockStorage.cpp
```diff
@@ -133,6 +133,20 @@ bool BlockStorage::GetTxnFromHistoricalDB(const dev::h256& key,
   return true;
 }
 
+bool BlockStorage::GetHistoricalMicroBlock(const BlockHash& blockhash,
```
ansnunez: Is there supposed to be a `PutHistoricalMicroBlock`, or any code that stores into `m_MBHistoricalDB`?

src/libZilliqa/Zilliqa.cpp
```diff
@@ -281,6 +281,8 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, bool loadConfig,
       case SyncType::DB_VERIF:
         LOG_GENERAL(INFO, "Intitialize DB verification");
         m_n.ValidateDB();
+        std::this_thread::sleep_for(std::chrono::seconds(10));
+        raise(SIGKILL);
```
ansnunez: Doesn't `ValidateDB` already call `raise(SIGKILL)`?

KaustubhShamshery: no I removed it from there as the function might return false and then it is not invoked

src/libServer/Server.cpp
KaustubhShamshery: yeah i think i will replace it with database error

src/libPersistence/BlockStorage.cpp
KaustubhShamshery: actually that is done by backend using s3

PR #1171, opened 2019-01-10
Implement BN_generate_dsa_nonce

Gnnng:
## Description

The main part of the function is ported from OpenSSL 1.1.1a [`BN_generate_dsa_nonce`](https://github.com/openssl/openssl/blob/7835e97b6ff5cd94a10c5aeac439f4aa145a77b2/crypto/bn/bn_rand.c#L205). There're a few changes in the function to make it work, which is denoted by `FIXME` comment.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libCrypto/generate_dsa_nonce.c
+++ b/src/libCrypto/generate_dsa_nonce.c
@@ -70,7 +70,6 @@ int BN_generate_dsa_nonce(BIGNUM *out, const BIGNUM *range,
     unsigned char *k_bytes;
     int ret = 0;
 
-    //FIXME: should remove this conversion from (void *) to (unsiged char *)
     k_bytes = (unsigned char *)OPENSSL_malloc(num_k_bytes);
     if (k_bytes == NULL)
         goto err;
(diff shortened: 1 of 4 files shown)


src/libCrypto/generate_dsa_nonce.c
```diff
@@ -0,0 +1,125 @@
...
+    unsigned char random_bytes[64];
+    unsigned char digest[SHA512_DIGEST_LENGTH];
+    unsigned done, todo;
+    /* We generate |range|+8 bytes of random output. */
+    const unsigned num_k_bytes = BN_num_bytes(range) + 8;
+    unsigned char private_bytes[96];
+    unsigned char *k_bytes;
+    int ret = 0;
+
+    //FIXME: should remove this conversion from (void *) to (unsiged char *)
```
ansnunez: I think the explicit casting is OK, no need to add FIXME comment. C-style allocation methods like `OPENSSL_malloc` are intentionally generic anyway and not tied to a specific type.

Gnnng: it's removed.

PR #1173, opened 2019-01-10
Verify shard guards in sharding validator

ansnunez:
## Description

During testnet run, I noticed some DS backups had fewer PoW submissions than the DS leader.

This situation is ok, because we have code inside `VerifyPoWOrdering` that can accept additional nodes for sharding, as long as the PoWs from those extra nodes are verified successfully.

However, in my run, those extra nodes that the DS leader proposed were guard nodes, so during PoW verify, the backup complains with "PoW solution doesn't meet difficulty requirement" and aborts further processing of the sharding structure.

This PR makes sure to use shard guard diff in the event that those nodes are guard nodes.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -454,9 +454,14 @@ bool DirectoryService::VerifyPoWOrdering(
               m_mediator.m_dsBlockRand, m_mediator.m_txBlockRand,
               peer.m_ipAddress, toFind, powSoln.lookupId, powSoln.gasPrice);
 
-          auto difficulty = m_mediator.m_dsBlockChain.GetLastBlock()
-                                .GetHeader()
-                                .GetDifficulty();
+          // If it's a shard guard, use difficulty=1 only
+          auto difficulty =
+              (Guard::GetInstance().IsNodeInShardGuardList(pubKeyToPoW->first))
+                  ? 1
```
ShengguangXiao: Suggest define a constant value GUARD_NODE_DIFFICULTY instead of 1 for guard node.

PR #1174, opened 2019-01-10
Add checks for raise pow

bb111189:
## Description

Raise PoW message is an empty message. There is no signature check. Any node will be able to trigger a thread in lookup that will sleep for a certain amount of seconds. 

This PR will change the message to be consist of 
1. msg type 
2. block number
3. pubkey
4. signature

The signature will be signed over (msg, type, block number). This is to prevent a replay attack where one can use a signed message of (block number) of other message type and replay it. 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud test (commit: 3d19c95559c3d194f0b816f1979a8bccd6718130)
- [x] medium-scale cloud test (commit: 75dc27211c321ba7a0a0cd278f4cb9536aa299dd)

### DIFF
--- a/src/libMessage/Messenger.cpp
+++ b/src/libMessage/Messenger.cpp
@@ -5366,12 +5366,11 @@ bool Messenger::GetLookupSetRaiseStartPoW(const bytes& src,
     return false;
   }
 
-  msgType = result.msgtype();
-  blockNumber = result.blocknumber();
+  msgType = result.data().msgtype();
+  blockNumber = result.data().blocknumber();
 
-  bytes tmp;
-  NumberToArray<uint8_t, sizeof(uint8_t)>(msgType, tmp, 0);
-  NumberToArray<uint64_t, sizeof(uint64_t)>(blockNumber, tmp, sizeof(uint8_t));
+  bytes tmp(result.data().ByteSize());
+  result.data().SerializeToArray(tmp.data(), tmp.size());
 
   ProtobufByteArrayToSerializable(result.pubkey(), dsPubKey);
   Signature signature;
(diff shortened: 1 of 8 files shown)


src/libNode/Node.cpp
```diff
@@ -2089,6 +2089,46 @@ std::string Node::GetActionString(Action action) const {
   return true;
 }
 
+/*static*/ bool Node::GetDSLeaderPubkey(const BlockLink& lastBlockLink,
```
ShengguangXiao: The functionality of this function is the same as GetDSLeaderPeer, can combine into one? Or let one call the other?

bb111189: GetDSLeaderPeer return a peer object and not pubkey

src/libMessage/Messenger.cpp
```diff
@@ -5350,6 +5350,73 @@ bool Messenger::GetLookupSetOfflineLookups(const bytes& src,
...
+                                          uint64_t& blockNumber,
+                                          PubKey& dsPubKey) {
+  LOG_MARKER();
+
+  LookupRaiseStartPoW result;
+
+  result.ParseFromArray(src.data() + offset, src.size() - offset);
+
+  if (!result.IsInitialized()) {
+    LOG_GENERAL(WARNING, "LookupGetStartPoWFromSeed initialization failed.");
```
ansnunez: Message doesn't match function

src/libMessage/ZilliqaMessage.proto
```diff
@@ -662,7 +662,10 @@ message LookupSetOfflineLookups
 
 message LookupRaiseStartPoW
 {
-    // Empty
+    required uint32 msgtype      = 1; // Only LSB used
```
ansnunez: Might be easier to wrap the inputs to the signature in a nested message, so that we don't have to individually serialize them for the input to signing

src/libMessage/Messenger.cpp
```diff
@@ -5350,6 +5350,73 @@ bool Messenger::GetLookupSetOfflineLookups(const bytes& src,
...
+  Signature signature;
+  if (!Schnorr::GetInstance().Sign(tmp, dsKey.first, dsKey.second, signature)) {
+    LOG_GENERAL(WARNING, "Failed to sign start PoW message.");
+    return false;
+  }
+
+  SerializableToProtobufByteArray(signature, *result.mutable_signature());
+
+  if (!result.IsInitialized()) {
+    LOG_GENERAL(WARNING, "LookupSetStartPoWFromSeed initialization failed.");
```
ansnunez: Message doesn't match class name

PR #1187, opened 2019-01-13
Sig check, anti-replay and additional pubkey check for Set lookup online/offline

bb111189:
## Description 

Set lookup online and offline does not have sufficient checks. There is no signature check. The content of the message contains only port number, which can be susceptible to replay attacks. There is also no check whether the message is malformed or not when calling `ComposeGetLookupOnlineMessage`

This PR will change the message to be consist of
- msg type
- block number
- pubkey
- signature

The signature will be signed over (msg, type, block number). This is to prevent a replay attack where one can use a signed message of (port number) of other message type and replay it.

Pubkey check have been additional the original check which consists of only `Peer` checks.

An additional check was additional to check whether the output of both `ComposeGetLookupOnlineMessage` and `ComposeGetOfflineLookupNodes` is of size 0 or not. If true, the node will not proceeds to send the message and instead return false.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud test (commit: f0744338af23e0dd6928ecb4e99faf734ff375a9)
- [x] medium-scale cloud retest (commit: 89bd45ebb10a3174fd72942b35ada29316939f8c) - Failed
- [x] medium-scale cloud retest (Commit: 683a902f536463da48693dbff4634d165c06834a) - Passed

### DIFF
--- a/src/libMessage/Messenger.cpp
+++ b/src/libMessage/Messenger.cpp
@@ -5364,8 +5364,6 @@ bool Messenger::SetLookupSetLookupOffline(bytes& dst, const unsigned int offset,
     return false;
   }
 
-  SerializableToProtobufByteArray(signature, *result.mutable_signature());
-
   if (!result.IsInitialized()) {
     LOG_GENERAL(WARNING, "LookupSetLookupOffline initialization failed.");
     return false;
@@ -5435,8 +5433,6 @@ bool Messenger::SetLookupSetLookupOnline(bytes& dst, const unsigned int offset,
     LOG_GENERAL(WARNING, "LookupSetLookupOnline.Data initialization failed.");
     return false;
   }
-
-  SerializableToProtobufByteArray(signature, *result.mutable_signature());
 
   if (!result.IsInitialized()) {
     LOG_GENERAL(WARNING, "LookupSetLookupOnline initialization failed.");


src/libMessage/Messenger.cpp
```diff
@@ -5337,12 +5337,34 @@ bool Messenger::GetLookupSetStateFromSeed(const bytes& src,
...
+      LOG_GENERAL(WARNING, "Failed to sign set lookup offline message.");
+      return false;
+    }
+    SerializableToProtobufByteArray(signature, *result.mutable_signature());
+  } else {
+    LOG_GENERAL(WARNING, "LookupSetLookupOffline.Data initialization failed.");
+    return false;
+  }
+
+  SerializableToProtobufByteArray(signature, *result.mutable_signature());
```
ansnunez: Duplicate call

src/libMessage/Messenger.cpp
```diff
@@ -5366,20 +5390,53 @@ bool Messenger::GetLookupSetLookupOffline(const bytes& src,
...
+      LOG_GENERAL(WARNING, "Failed to sign set lookup online message.");
+      return false;
+    }
+    SerializableToProtobufByteArray(signature, *result.mutable_signature());
+  } else {
+    LOG_GENERAL(WARNING, "LookupSetLookupOnline.Data initialization failed.");
+    return false;
+  }
+
+  SerializableToProtobufByteArray(signature, *result.mutable_signature());
```
ansnunez: Duplicate call

PR #1198, opened 2019-01-14
Add signature and DS block number to ProcessGetStartPoWFromSeed

ansnunez:
## Description

To prevent abuse of `GETSTARTPOWFROMSEED`, which can be sent by any new node, this PR adds a signature to it, with the signature generated over both the `listenport` and the current DS block number, to mitigate message replay.

Extra Fix: I noticed `SetLookupSetRaiseStartPoW` protobuf-ed the `signature` twice so I removed the duplicate call.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libMessage/Messenger.cpp
+++ b/src/libMessage/Messenger.cpp
@@ -5581,7 +5581,6 @@ bool Messenger::SetLookupSetRaiseStartPoW(bytes& dst, const unsigned int offset,
       LOG_GENERAL(WARNING, "Failed to sign raise start PoW message.");
       return false;
     }
-    SerializableToProtobufByteArray(signature, *result.mutable_signature());
   } else {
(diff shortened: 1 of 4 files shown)


src/libMessage/Messenger.cpp
```diff
@@ -5616,19 +5636,34 @@ bool Messenger::SetLookupGetStartPoWFromSeed(bytes& dst,
 
 bool Messenger::GetLookupGetStartPoWFromSeed(const bytes& src,
                                              const unsigned int offset,
-                                             uint32_t& listenPort) {
+                                             uint32_t& listenPort,
```
bb111189: need to pass pubkey to the caller function. The caller function need to ensure the pubkey is known.

PR #1205, opened 2019-01-15
Add retrieval of only latest ds epoch

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -1064,7 +1064,11 @@ void Lookup::RetrieveTxBlocks(vector<TxBlock>& txBlocks, uint64_t& lowBlockNum,
                               uint64_t& highBlockNum) {
   lock_guard<mutex> g(m_mediator.m_node->m_mutexFinalBlock);
 
-  if (lowBlockNum == 1) {
+  if ((lowBlockNum == 1 ||
```
ansnunez: Maybe we put the `else if (lowblockNum == 0)` branch first, that way we no longer need to check `lowBlockNum != 0`

PR #1211, opened 2019-01-15
Put version field in Account and AccountDelta

kaikawaliu:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libMessage/Messenger.cpp
```diff
@@ -341,6 +371,11 @@ bool ProtobufToAccount(const ProtoAccount& protoAccount, Account& account,
        tmpStorageRoot.asArray().begin());
 
   if (protoAccount.code().size() > 0) {
```
ansnunez: I think here you have to add `protoAccount.has_code() &&`

src/libMessage/Messenger.cpp
```diff
@@ -493,6 +536,11 @@ bool ProtobufToAccountDelta(const ProtoAccount& protoAccount, Account& account,
     bool doInitContract = false;
 
     if (fullCopy) {
+      if (!CheckRequiredFieldsProtoAccountContract(protoAccount) ||
```
ansnunez: Maybe we split the `if` and `LOG_GENERAL` for these two, so it will be easier for us to know which check failed

PR #1214, opened 2019-01-15
Protect consensus message buffer, prevent unlimited message

ShengguangXiao:
## Description

Protect consensus message buffer, prevent unlimited message

Check the sender id and ip address is in my committee member list, and check if the same message already saved in the buffer before add it.

From security audit report issue 32 and 33.
`Node::ProcessMicroBlockConsensus`
    - an attacker can cause an out-of-memory condition by sending bogus Node micro block consensus messages. These are buffered in an unbounded map `m_microBlockConsensusBuffer`
`Node::ProcessMicroBlockConsensusCore`
    - an attacker can cause an arbitrary number of threads to get temporarily stuck this method. This is limited to certain phases of the consensus protocol

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/FinalBlockPostProcessing.cpp
+++ b/src/libDirectoryService/FinalBlockPostProcessing.cpp
@@ -294,8 +294,10 @@ bool DirectoryService::ProcessFinalBlockConsensus(const bytes& message,
   }
 
   uint32_t consensus_id = 0;
+  PubKey senderPubKey;
 
-  if (!m_consensusObject->GetConsensusID(message, offset, from, consensus_id)) {
+  if (!m_consensusObject->GetConsensusID(message, offset, from, consensus_id,
+                                         senderPubKey)) {
     LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum, "GetConsensusID failed.");
     return false;
   }
@@ -317,7 +319,8 @@ bool DirectoryService::ProcessFinalBlockConsensus(const bytes& message,
       return false;
     }
 
-    AddToFinalBlockConsensusBuffer(consensus_id, message, offset, from);
+    AddToFinalBlockConsensusBuffer(consensus_id, message, offset, from,
+                                   senderPubKey);
 
     LOG_EPOCH(INFO, m_mediator.m_currentEpochNum,
               "Process final block arrived early, saved to buffer");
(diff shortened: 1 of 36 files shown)


ansnunez:
@ShengguangXiao I think you misinterpreted the purpose of `GetConsensusID`.  It's supposed to be the ID of the ongoing consensus, not the ID of any of the nodes (leader or backup).  So changing it to `GetLeaderConsensusID` and `GetBackupConsensusID` is not in line with the purpose.

ShengguangXiao:
> GetConsensusID
I use GetLeaderConsensusID means to get the leader's consensus ID, not the ID of leader, for backup is same. Seperate GetConsensusID into two functions is to use different checking.

ansnunez:
> > GetConsensusID
> > I use GetLeaderConsensusID means to get the leader's consensus ID, not the ID of leader, for backup is same. Seperate GetConsensusID into two functions is to use different checking.

Ah understood.  You still get consensus ID as before, but also check the leader/backup ID.

src/libNode/MicroBlockPostProcessing.cpp
```diff
@@ -143,6 +137,35 @@ void Node::CommitMicroBlockConsensusBuffer() {
   }
 }
 
+void Node::AddToMicroBlockConsensusBuffer(uint32_t consensusId,
+                                          const bytes& message,
+                                          unsigned int offset,
+                                          const Peer& peer) {
+  lock_guard<mutex> h(m_mutexMicroBlockConsensusBuffer);
+  auto& vecPeerMsg = m_microBlockConsensusBuffer[consensusId];
+  const auto consensusMsgType = message[offset];
```
ansnunez: Same comment here, add size check

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -370,7 +362,36 @@ void DirectoryService::CommitFinalBlockConsensusBuffer() {
...
 }
 
-void DirectoryService::CleanFinalblockConsensusBuffer() {
+void DirectoryService::AddToFinalBlockConsensusBuffer(uint32_t consensusId,
+                                                      const bytes& message,
+                                                      unsigned int offset,
+                                                      const Peer& peer) {
+  lock_guard<mutex> h(m_mutexFinalBlockConsensusBuffer);
+  auto& vecPeerMsg = m_finalBlockConsensusBuffer[consensusId];
+  const auto consensusMsgType = message[offset];
```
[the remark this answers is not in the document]
ShengguangXiao: In GetConsensID function, the size of the message checked already, but here can do a double check i think

src/libNode/MicroBlockPostProcessing.cpp
ShengguangXiao: Added check already

src/libMessage/Messenger.h
```diff
@@ -650,6 +660,81 @@ class Messenger {
...
+                               << leaderId
+                               << " is out of my committee member size "
+                               << commNodes.size());
+      return false;
+    }
+
+    if (commNodes.end() ==
+        std::find_if(commNodes.begin(), commNodes.end(),
+                     [from](const PairOfNode& pairOfNode) {
+                       return from.GetIpAddress() ==
```
sandipbhoir: There could be different sender using same ip. Can use pubkey ?

PR #1215, opened 2019-01-15
Store codecache and initdata into db

kaikawaliu:
## Description

- Clean all `Account::m_codeCache` for newly created contract at the end of every vacuous epoch, to reduce memory occupation 
- Removed `Account::m_initData` and `m_initValJson`, the `InitData` will only be fetched from ContractStorage from now on
- Removed `Account::m_createBlockNum` and `m_scillaVersion`, which will be embeded in initData once contract get created
- Removed the Trie DB for account contract state totally, same for the `Account::m_prevRoot`, `Commit()` and `RollBack()` functions

- Largely rewrote the functions for InitData, including `ExportContractFiles` and so on.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libData/AccountData/Account.cpp
```diff
@@ -66,59 +66,27 @@ void Account::InitStorage() {
   }
 }
 
-bool Account::InitContract(const bytes& data, const Address& addr) {
-  SetInitData(data);
-  if (!InitContract(addr)) {
+bool Account::InitContract(const bytes& initData, const Address& addr) {
+  m_address = addr;
```
ansnunez: I think originally this was placed at the end of the function to prevent prematurely setting `m_address` before all the checks pass, to avoid setting account to invalid state.  Any reason why we can't maintain the code?

Also, if you agree to put it back at the end, maybe we can also set `m_initData.clear();` if `InitContract` failed?
But anyway if the change was necessary, you can ignore this suggestion.

ansnunez: Oh and we can use `CleanInitData` before returning false also?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -175,45 +175,74 @@ bool AccountStore::MoveUpdatesToDisk() {
...
+          initData_batch)) {
+    LOG_GENERAL(WARNING, "PutContractInitDataBatch failed");
+    ret = false;
+  }
+
+  if (ret) {
+    if (!HASHMAP_CONTRACT_STATE_DB) {
+      ContractStorage::GetContractStorage().GetStateDB().commit();
+
+      for (auto i : *m_addressToAccount) {
```
ansnunez: I think this might need to be `auto&` to avoid copy-constructing another object in `i`

src/libData/AccountData/Account.h
```diff
@@ -43,25 +43,23 @@
 /// DB storing trie storage information for all accounts.
 // static OverlayDB contractStatesDB("contractStates");
 
+static uint32_t scilla_version_place_holder;
```
ShengguangXiao: This variable used without initialize?

kaikawaliu: this is used as place holder, simply used as default reference argument, don't think we need to initialize it

src/libPersistence/ContractStorage.cpp
```diff
@@ -265,8 +277,16 @@ Json::Value ContractStorage::GetContractStateJson(const dev::h160& address) {
...
     string tValue = std::get<VALUE>(entry);
 
-    if (!tMutable) {
-      continue;
+    if (!hasScillaVersion && tVname == "_scilla_version" && tType == "Uint32" &&
+        !tMutable) {
+      try {
+        scilla_version = boost::lexical_cast<uint32_t>(tValue);
+      } catch (...) {
+        LOG_GENERAL(WARNING, "_scilla_version is not a number");
```
ansnunez: Better to log `tValue` as well

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -485,24 +493,29 @@ bool AccountStoreSC<MAP>::ExportCallContractFiles(
...
-  root_w_version = SCILLA_ROOT;
+    const uint32_t& scilla_version) {
+  m_root_w_version = SCILLA_ROOT;
   if (ENABLE_SCILLA_MULTI_VERSION) {
-    root_w_version += '/' + std::to_string(scilla_version);
+    m_root_w_version += '/' + std::to_string(scilla_version);
   }
 
-  if (!boost::filesystem::exists(root_w_version)) {
+  if (!boost::filesystem::exists(m_root_w_version)) {
```
ansnunez: Better to log `m_root_w_version` also when it doesn't exist

PR #1220, opened 2019-01-16
Make log messages easier to read for debugging

ansnunez:
## Description

This PR changes the log message format as follows:
~~1. Date/Time format.  Before = `19-01-16T08:33:19.697`.  Now = `190116 08:33:19.697`.~~
2. Added ~~line number~~ filename and line number after the time
~~3. Added filename after the line number~~
4. Shortened the function name from 30 to 25
5. Shortened BEGIN to BEG
6. For LOGSTATE, added singe space before the epoch num section
7. Reduced `MAX_BYTES_TO_DISPLAY` from 100 to 64

Sample Before:
```
[INFO][ 7520][18-12-24T13:48:35.045][ProcessSetPrimary             ][Epoch 1] My node ID for this PoW consensus is 0
```

Sample Now:
```
[INFO][10770][19-01-16T10:42:31.415][ctoryService.cpp:267][ProcessSetPrimary        ] [Epoch 1] My node ID for this PoW consensus is 0
```

Notes:
1. Overall, each line length has increased by 18 characters.
2. For `LOG_MARKER`, the line number shown for END message is same as for BEG.

Other updates:
~~1. Updated profile_time.py for the new time format~~
2. Updated some log messages inside PoWProcessing.cpp and Lookup.cpp

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/profile_time.py
+++ b/scripts/profile_time.py
@@ -105,8 +105,8 @@ def find_files(directory, pattern):
 				yield filename
 
 def get_time(line):
-	# The format is [ 181207 09:47:21.817 ], what need to get is 09:47:21.817
-	return line[9:22]
+	# The format is [ 18-12-07T09:47:21.817 ], what need to get is 09:47:21.817
+	return line[11:24]
 
 def get_block_number(line):
 	m = re.search(r'\[[0-9 ]+\]', line)
--- a/scripts/profile_time.py
+++ b/scripts/profile_time.py
@@ -105,8 +105,8 @@ def find_files(directory, pattern):
 				yield filename
 
 def get_time(line):
-	# The format is [ 18-12-07T09:47:21.817 ], what need to get is 09:47:21.817
-	return line[11:24]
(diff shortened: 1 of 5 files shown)


ansnunez:
>     1. Why change the timestamp format? Now we already followed the standard ISO 8601. (https://en.wikipedia.org/wiki/ISO_8601)
> 
>     2. Is it possible to merge [FileName:LineNum] into one field, just like gdb?

1. Mostly to reduce size improve readability, but OK will revert
2. Ok will merge

Gnnng:
Maybe it's more common to use filename and line number like this

```
[src/libMessage/Messenger.cpp:234]
```

or cut the leading characters

```
[ssage/Messenger.cpp:234]
```

ansnunez:
> Maybe it's more common to use filename and line number like this
> 
> ```
> [src/libMessage/Messenger.cpp:234]
> ```
> 
> or cut the leading characters
> 
> ```
> [ssage/Messenger.cpp:234]
> ```

Done

PR #1227, opened 2019-01-17
Log the VC blocks in lookup state log

ansnunez:
## Description

This PR adds log message to state log in lookup node, so we can quickly see when view change happens.

```
[ 19-01-17T07:55:25.473 ][VCBLK] DS = 2 Tx = 1
[ 19-01-17T07:55:25.473 ][VCBLK] Leader = <127.0.0.1:5004>
[ 19-01-17T07:55:25.473 ][VCBLK] Faulty = <127.0.0.1:5016>
```

Other minor fix: `GetVieWChangeDSEpochNo` -> `GetViewChangeDSEpochNo` (changed the `W`)

Tested locally by doing `vc3`.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/ViewChangeBlockProcessing.cpp
+++ b/src/libNode/ViewChangeBlockProcessing.cpp
@@ -164,7 +164,7 @@ bool Node::ProcessVCBlockCore(const VCBlock& vcblock) {
 
   // Check is block latest
   if (!m_mediator.CheckWhetherBlockIsLatest(
-          vcblock.GetHeader().GetVieWChangeDSEpochNo(),
+          vcblock.GetHeader().GetViewChangeDSEpochNo(),
           vcblock.GetHeader().GetViewChangeEpochNo())) {
     LOG_GENERAL(WARNING, "ProcessVCBlockCore CheckWhetherBlockIsLatest failed");
     return false;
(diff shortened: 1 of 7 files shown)


src/libNode/ViewChangeBlockProcessing.cpp
```diff
@@ -238,6 +238,18 @@ bool Node::ProcessVCBlockCore(const VCBlock& vcblock) {
 
   UpdateDSCommiteeCompositionAfterVC(vcblock, *m_mediator.m_DSCommittee);
 
+  if (LOOKUP_NODE_MODE) {
+    LOG_STATE("[VCBLK] DS = " << vcblock.GetHeader().GetViewChangeDSEpochNo()
+                              << " Tx = "
+                              << vcblock.GetHeader().GetViewChangeEpochNo());
+    LOG_STATE("[VCBLK] Leader = " << vcblock.GetHeader()
+                                         .GetCandidateLeaderNetworkInfo()
+                                         .GetPrintableIPAddress());
```
bb111189: No need GetPrintableIPAddress(). Just print the Peer object directly. It will be <IP:PORT>

src/libNode/ViewChangeBlockProcessing.cpp
```diff
@@ -238,6 +238,18 @@ bool Node::ProcessVCBlockCore(const VCBlock& vcblock) {
...
+    for (const auto& faulty : vcblock.GetHeader().GetFaultyLeaders()) {
+      LOG_STATE("[VCBLK] Faulty = " << faulty.second.GetPrintableIPAddress());
```
bb111189: No need GetPrintableIPAddress(). Just print the Peer object directly. It will be <IP:PORT>

PR #1235, opened 2019-01-17
Feature/guard network info sig

jendis:
## Description
Add signature for NodeGetGuardNodeNetworkInfoUpdate message 

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -3553,14 +3553,22 @@ bool Lookup::ProcessGetDSGuardNetworkInfo(const bytes& message,
...
 
   if (!Messenger::GetLookupGetNewDSGuardNetworkInfoFromLookup(
-          message, offset, portNo, dsEpochNo)) {
+          message, offset, portNo, dsEpochNo, senderPubKey)) {
     LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
               "Messenger::GetLookupGetNewDSGuardNetworkInfoFromLookup failed.");
     return false;
   }
 
+  if (!IsLookupNode(senderPubKey)) {
```
bb111189: The sender of the message is actually from shard node -> lookup. So this check is not correct.

src/libMessage/ZilliqaMessage.proto
```diff
@@ -948,8 +948,14 @@ message DSLookupSetDSGuardNetworkInfoUpdate
 // To update shard node of guard node's network info change
 message NodeGetGuardNodeNetworkInfoUpdate
 {
-    required uint32 portno   = 1;
-    required uint64 dsepochnumber   = 2;
+	message Data
+	{
+    	required uint32 portno          = 1;
+    	required uint64 dsepochnumber   = 2;
+    }
```
bb111189: indentation of } and content inside {}

PR #1238, opened 2019-01-18
Fix protobuf signature signing

bb111189:
## Description

**LookupSetDSBlockFromSeed**
- sig should be sig(lowBlockNum, highBlockNum, dsBlocks) instead of sig (dsBlocks)

**LookupSetTxBlockFromSeed**
- sig should be sig(lowBlockNum, highBlockNum, txblocks) instead of sig(txblocks)

**LookupSetStateDeltaFromSeed**
- sig should be sig(blockNum, stateDelta) instead of sig(stateDelta)

**Refactoring of VCNodeSetDSTxBlockFromSeed**

### Implementation

- [x] **ready for review**

- [ ] ~local machine test~

- [x] medium-scale cloud test (commit: 0b467b79b60dfa7d6d028c017e9ff59445148ee9)

### DIFF
--- a/src/libMessage/Messenger.cpp
+++ b/src/libMessage/Messenger.cpp
@@ -4838,28 +4838,30 @@ bool Messenger::SetLookupSetDSBlockFromSeed(bytes& dst,
 
   LookupSetDSBlockFromSeed result;
 
-  result.set_lowblocknum(lowBlockNum);
-  result.set_highblocknum(highBlockNum);
+  result.mutable_data()->set_lowblocknum(lowBlockNum);
+  result.mutable_data()->set_highblocknum(highBlockNum);
 
   SerializableToProtobufByteArray(lookupKey.second, *result.mutable_pubkey());
 
   for (const auto& dsblock : dsBlocks) {
-    DSBlockToProtobuf(dsblock, *result.add_dsblocks());
+    DSBlockToProtobuf(dsblock, *result.mutable_data()->add_dsblocks());
   }
 
   Signature signature;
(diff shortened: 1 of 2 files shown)


ShengguangXiao:
@bb111189 I make a change in this PR, directly return false if protobuf data is not initialized, please sync your code and do a double check

src/libMessage/Messenger.cpp
```diff
@@ -3326,22 +3325,22 @@ bool Messenger::SetDSPoWSubmission(
...
-      return false;
-    }
+  bytes tmp(result.data().ByteSize());
+  result.data().SerializeToArray(tmp.data(), tmp.size());
 
-    SerializableToProtobufByteArray(signature, *result.mutable_signature());
-  } else {
-    LOG_GENERAL(WARNING, "DSPoWSubmission.Data initialization failed.");
+  Signature signature;
+  if (!MultiSig::GetInstance().SignKey(tmp, submitterKey, signature)) {
```
bb111189: To look into this

ShengguangXiao: @sandipbhoir Why it use MultiSig here? It is not using aggregated pubkey.

sandipbhoir: I didn’t use multisig. As with my commit-   af700ad , Schnorr was used. It must have modified in following pr’s for some reason.

ansnunez: This one is okay, please refer to #1097 description for the explanation why we use Multisig

PR #1243, opened 2019-01-19
Daemon would detect file SUSPEND_LAUNCH instead of delay time

ckyang:
## Description

For recover entire network, there is some flow change.
To be more flexible flow control, after detecting Zilliqa process dead, the daemon should monitor a file "SUSPEND_LAUNCH" to see if it's time to re-launch a new Zilliqa process.
By default (no such file existed), the daemon should re-launch a new Zilliqa process immediately.
If there is this file existed, daemon will suspend the re-launch until this file is disappeared.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/daemon/ZilliqaDaemon_AWS.cpp
+++ b/daemon/ZilliqaDaemon_AWS.cpp
@@ -45,13 +45,12 @@ const string PRIVKEY_OPT = "--privk";
 const string PUBKEY_OPT = "--pubk";
 const string IP_OPT = "--address";
 const string PORT_OPT = "--port";
+const string SUSPEND_LAUNCH = "SUSPEND_LAUNCH";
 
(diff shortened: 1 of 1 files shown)


daemon/ZilliqaDaemon_AWS.cpp
```diff
@@ -244,6 +243,13 @@ void StartNewProcess(const string pubKey, const string privKey,
   pid_t pid;
 
   if (0 == (pid = fork())) {
+    while (ifstream(SUSPEND_LAUNCH).good()) {
+      log << "Temporarily suspend launch new zilliqa process, please wait "
```
sandipbhoir: May be we can print this message only once?

PR #1245, opened 2019-01-20
Use Peer IP instead of PubKey to prevent unlimited consensus buffer messages

ansnunez:
**Update**
This PR needs to be revisited.  For now, checking of IP or PubKey has been removed.  See https://github.com/Zilliqa/Issues/issues/419.

---

This PR adjusts the fix made in #1214.

The issue with #1214 is that it needs to access committee to retrieve senderPubKey.  But at the time of receipt of this consensus message, it is possible that the consensus object is not yet generated.

So instead of using senderPubKey, the best we can do at this point is use the sender peer's IP address.

https://github.com/Zilliqa/Issues/issues/419

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNetwork/ShardStruct.h
+++ b/src/libNetwork/ShardStruct.h
@@ -39,7 +39,7 @@ using DequeOfNode = std::deque<PairOfNode>;
 
 enum NodeMessage { NODE_PUBKEY, NODE_PEER, NODE_MSG };
 
-using NodeMsg = std::tuple<PubKey, Peer, bytes>;
+using NodeMsg = std::pair<Peer, bytes>;
 using VectorOfNodeMsg = std::vector<NodeMsg>;
 
 #endif /*__SHARD_STRUCT__*/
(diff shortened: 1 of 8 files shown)


src/libNetwork/ShardStruct.h
```diff
@@ -39,7 +39,7 @@ using DequeOfNode = std::deque<PairOfNode>;
 
 enum NodeMessage { NODE_PUBKEY, NODE_PEER, NODE_MSG };
 
-using NodeMsg = std::tuple<PubKey, Peer, bytes>;
+using NodeMsg = std::pair<Peer, bytes>;
 using VectorOfNodeMsg = std::vector<NodeMsg>;
```
ShengguangXiao: If remove the tuple, the enum type NodeMessage can also remove

ShengguangXiao:
Please don't merge this PR first, this change will remove all the checking on the message, it will have security issue, i am working on another PR to fix the issue Zilliqa/Issues#419

PR #1246, opened 2019-01-20
Fix bug for retrieving sharding structure crash

ckyang:
## Description

When retrieving sharding structure without ip-mapping file (using the same pod), fetching will cause coredump. This PR fixed this issue, and also fixed validateBackupDB problem caused by Zilliqa command foramt change.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/scripts/validateBackupDB.sh
+++ b/scripts/validateBackupDB.sh
@@ -15,36 +15,17 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 # [MUST BE FILLED IN] User configuration settings
-aws_access_key_id=""
-aws_secret_access_key=""
-
+backupDBName=""
 
 # Validate input argument
-if [ "$#" -ne 1 ]; then
-    echo -e "\n\n\033[0;31mUsage: source ./scripts/validateBackupDB.sh backupDBName\033[0m\n"
-    return 0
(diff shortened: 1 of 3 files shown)


scripts/validateBackupDB.sh
```diff
@@ -15,36 +15,17 @@
...
-
 # Download persistence from Amazon S3 database
-backupDBName="$1"
-export AWS_ACCESS_KEY_ID=${aws_access_key_id}
-export AWS_SECRET_ACCESS_KEY=${aws_secret_access_key}
-pip install awscli
-aws s3 cp s3://zilliqa-persistence/${backupDBName}.tar.gz ${backupDBName}.tar.gz
 rm -rf persistence
-tar xzvf ${backupDBName}.tar.gz
-echo -e "\n\033[0;32mDownload ${backupDBName}.tar.gz from Amazon S3 database successfully.\033[0m\n"
```
sandipbhoir: Why remove logic of downloading from s3. Is it moved out ??

ckyang: The validation should focus on the database correctness, and the removal of S3-related logic could give this script more flexibility. That is, once user have a persistence in hand, he can use this script immediately without uploading the database to S3 first.

PR #1263, opened 2019-01-21
Regularly clean m_addressToAccount

kaikawaliu:
## Description

- Added protobuf function for serializing AccountStore from merkle Trie. Thus m_addressToAccount can be cleaned regularly to avoid taking too much memory.

- Also added a sleep for a short period to make sure the notify_timeout conditional variable will be initiated before the ProcessTxn function finished

- There might also be situation that a contract account will be included in the delta due to change of balance, but the contract states was not changed. in this case the check for storageroothash in protobuf msg should be skipped.

- New node will store states into persistence upon fetching of account states from lookup during joining phase.

### Implementation

- [x] **ready for review**

- [x] local machine test (new node joining)

- [ ] small-scale cloud test


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -287,6 +287,8 @@ void Node::ProcessTransactionWhenShardLeader() {
 
   DetachedFunction(1, txnProcTimer);
 
+  this_thread::sleep_for(chrono::milliseconds(100));
```
ckyang: Maybe make a definition of delay inside constant file?

PR #1273, opened 2019-01-23
Fix recovery issue for node not joining network

ckyang:
## Description

For the nodes not joining network yet, some database would be missing. Then the recovery/upgrading would be failed. For this kind of situation, we always need to copy database from lookup node before applying recovery/upgrading.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -442,8 +442,10 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
 
   uint16_t ds_consensusLeaderID = 0;
 
-  BlockStorage::GetBlockStorage().GetDSCommittee(m_mediator.m_DSCommittee,
-                                                 ds_consensusLeaderID);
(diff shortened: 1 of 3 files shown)


src/libNode/Node.cpp
```diff
@@ -442,8 +442,10 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
 
   uint16_t ds_consensusLeaderID = 0;
 
-  BlockStorage::GetBlockStorage().GetDSCommittee(m_mediator.m_DSCommittee,
-                                                 ds_consensusLeaderID);
+  if (!BlockStorage::GetBlockStorage().GetDSCommittee(m_mediator.m_DSCommittee,
```
KaustubhShamshery: Add a log here ?

ckyang: Ok

PR #1275, opened 2019-01-23
Fix No Transaction Forwarded From DS After View Change

kaikawaliu:
## Description

- So the root cause is after RemoveDSMicroBlock called during ViewChange, we no longer process the DSMicroBlock because of some flags. So the DS nodes won't know they have transaction to send as the microblock was already erased, and txn will always be forwarded together with microblock.

Now the solution is to clean the flags and re-run the DSMicroBlock from a clean state ( which shall be the same as the end of MicroBlock submission). Now we requires the AccountStoreTemp to be reverted (by calling DeserializeDelta with m_stateDeltaFromShards). And also revert the AccountStore (by fixing and adopting RevertCommitTemp). Then the DS committee can rerun the DSMicroBlock after the view change, which can happen after the missing transaction was fetched.

- Several reversibility functions was implemented for ContractStorage in this PR also to make sure the reverting can work.

- For testing purpose, two build flags dm7 & dm8 was also introduced. for detail can review echo in the build.sh

### Implementation

- [x] **ready for review**

- [x] local machine test (contract execution)

- [x] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStore.h
+++ b/src/libData/AccountData/AccountStore.h
@@ -106,6 +106,8 @@ class AccountStore
 
   bool SerializeDelta();
 
+  AccountStoreTemp& GetAccountStoreTemp() { return *m_accountStoreTemp; }
+
   void GetSerializedDelta(bytes& dst);
 
   bool DeserializeDelta(const bytes& src, unsigned int offset,
@@ -140,6 +142,7 @@ class AccountStore
 
   void AddAccountDuringDeserialization(const Address& address,
                                        const Account& account,
+                                       const Account& oriAccount,
                                        const bool fullCopy = false,
                                        const bool reversible = false) {
     (*m_addressToAccount)[address] = account;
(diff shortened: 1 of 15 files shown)


src/libData/AccountData/AccountStore.h
```diff
@@ -106,6 +106,8 @@ class AccountStore
 
   bool SerializeDelta();
 
+  AccountStoreTemp& GetAccountStoreTemp() { return *m_accountStoreTemp; }
```
ShengguangXiao: Add const at the end of function for get function

kaikawaliu: This func is meant for changing the reference that it returns

kaikawaliu:
Tested with 20 aws nodes with DEX contract executed even with view change because of (1. Backup missing txn. 2. DS leader and half of the ds committee are missing txn)

PR #1276, opened 2019-01-23
Make UpdateProcessedTxn called early for DSCommittee

kaikawaliu:
## Description

Thanks for @sandipbhoir 's finding, we found:

Some ds node may finish the finalblock consensus late,
if a txn packet from a gossip neighbor is reached between the StoreFinalBlock and UpdateProcessedTxn, it will process it and insert into m_createdTxn, because we only check the EpochNum and is the packet from lookup when receiving the packet, then decides whether should buffer or process.
But when the UpdateProcessedTxn function was called, it may clean the m_createdTxn, then the txn just received will be cleaned.
And duplicate gossip is ignored, so the missed txn was always missed. 
Thus Txn Missing happened. 

So, to put UpdateProcessedTxn early before the FinalBlock storing, should solve this problem.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/FinalBlockPostProcessing.cpp
+++ b/src/libDirectoryService/FinalBlockPostProcessing.cpp
@@ -139,6 +139,10 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone() {
 
   bool isVacuousEpoch = m_mediator.GetIsVacuousEpoch();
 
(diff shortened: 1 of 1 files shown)


ShengguangXiao:
How long has this UpdateProcessedTransactions been called earlier?

PR #1278, opened 2019-01-23
Change GetTransaction and CreateTransction funcs

KaustubhShamshery:
## Description
* Change GetTransaction to return the contract type
* Change CreateTransaction to include another json value telling whether to send it to DS.
* Add a boolean in txn data type to indicate force send to ds.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


ansnunez:
#1302 has the final implementation

PR #1279, opened 2019-01-23
Remove stale TODOs

ansnunez:
## Description

This PR removes some TODO comments that are no longer relevant or no longer need to be addressed.

Comments are added to each change below to explain the removal.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libConsensus/ConsensusBackup.cpp
+++ b/src/libConsensus/ConsensusBackup.cpp
@@ -32,17 +32,11 @@ bool ConsensusBackup::CheckState(Action action) {
        {COMMIT_DONE, PROCESS_CHALLENGE},
        {COMMIT_DONE, PROCESS_COLLECTIVESIG},
        {COMMIT_DONE,
-        PROCESS_FINALCOLLECTIVESIG},  // TODO: check this logic again. Issue #43
-                                      // Node cannot proceed if
-                                      // finalcollectivesig arrived earlier (and
-                                      // get ignored by the node)
+        PROCESS_FINALCOLLECTIVESIG},
        {RESPONSE_DONE, PROCESS_CHALLENGE},
        {RESPONSE_DONE, PROCESS_COLLECTIVESIG},
        {RESPONSE_DONE,
-        PROCESS_FINALCOLLECTIVESIG},  // TODO: check this logic again. Issue #43
-                                      // Node cannot proceed if
-                                      // finalcollectivesig arrived earlier (and
-                                      // get ignored by the node)
+        PROCESS_FINALCOLLECTIVESIG},
        {FINALCOMMIT_DONE, PROCESS_FINALCHALLENGE},
        {FINALCOMMIT_DONE, PROCESS_FINALCOLLECTIVESIG},
        {FINALRESPONSE_DONE, PROCESS_FINALCHALLENGE},
--- a/src/libConsensus/ConsensusLeader.h
+++ b/src/libConsensus/ConsensusLeader.h
@@ -62,7 +62,6 @@ class ConsensusLeader : public ConsensusCommon {
   std::condition_variable cv_scheduleSubsetConsensus;
   bool m_allCommitsReceived;
 
-  // TODO: the vectors should be replaced by more space efficient DS
   std::vector<bool> m_commitMap;
   std::vector<CommitPoint>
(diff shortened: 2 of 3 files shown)


src/libConsensus/ConsensusBackup.cpp
```diff
@@ -32,17 +32,11 @@ bool ConsensusBackup::CheckState(Action action) {
        {COMMIT_DONE, PROCESS_CHALLENGE},
        {COMMIT_DONE, PROCESS_COLLECTIVESIG},
        {COMMIT_DONE,
-        PROCESS_FINALCOLLECTIVESIG},  // TODO: check this logic again. Issue #43
```
ansnunez: This fix has worked well enough the past year, and should be sufficient already.

src/libConsensus/ConsensusBackup.cpp
```diff
@@ -32,17 +32,11 @@ bool ConsensusBackup::CheckState(Action action) {
...
-                                      // Node cannot proceed if
-                                      // finalcollectivesig arrived earlier (and
-                                      // get ignored by the node)
+        PROCESS_FINALCOLLECTIVESIG},
        {RESPONSE_DONE, PROCESS_CHALLENGE},
        {RESPONSE_DONE, PROCESS_COLLECTIVESIG},
        {RESPONSE_DONE,
-        PROCESS_FINALCOLLECTIVESIG},  // TODO: check this logic again. Issue #43
```
ansnunez: This fix has worked well enough the past year, and should be sufficient already.

src/libConsensus/ConsensusLeader.h
```diff
@@ -62,7 +62,6 @@ class ConsensusLeader : public ConsensusCommon {
   std::condition_variable cv_scheduleSubsetConsensus;
   bool m_allCommitsReceived;
 
-  // TODO: the vectors should be replaced by more space efficient DS
```
ansnunez: This was more of an issue back when we could not forecast the final size of a shard.  At 600 nodes (or even a few hundred more beyond), the use of vectors here should be ok, given that the size of the data and the lifetime of a consensus are both not very large.

src/libConsensus/ConsensusLeader.h
```diff
@@ -80,7 +79,6 @@ class ConsensusLeader : public ConsensusCommon {
   std::map<unsigned int, bytes> m_commitFailureMap;
 
   // Tracking data for each consensus subset
-  // TODO: the vectors should be replaced by more space efficient DS
```
ansnunez: This was more of an issue back when we could not forecast the final size of a shard.  At 600 nodes (or even a few hundred more beyond), the use of vectors here should be ok, given that the size of the data and the lifetime of a consensus are both not very large.

src/libLookup/Lookup.cpp
```diff
@@ -416,11 +416,6 @@ uint128_t Lookup::TryGettingResolvedIP(const Peer& peer) const {
 void Lookup::SendMessageToLookupNodes(const bytes& message) const {
   LOG_MARKER();
 
-  // LOG_GENERAL(INFO, "i am here " <<
-  // to_string(m_mediator.m_currentEpochNum).c_str())
-
-  // TODO: provide interface in P2PComm instead of repopulating the lookup into
```
ansnunez: The idea was to just call `SendBroadcastMessage` directly on `m_lookupNodes`.  That was before we added more logic in the for loop below, so now it is not simply copying m_lookupNodes into another vector.  Also, we could not use `m_lookupNodes` directly anyway without having to hold on to `m_mutexLookupNodes` the whole time.  So, having a separate vector is still preferred.

src/libNetwork/DataSender.cpp
```diff
@@ -35,8 +35,6 @@ void SendDataToLookupNodesDefault(const VectorOfNode& lookups,
   }
   LOG_MARKER();
 
-  // TODO: provide interface in P2PComm instead of repopulating the lookup into
```
ansnunez: The idea was to just call `SendBroadcastMessage` directly on `lookups`.  That was before we added more logic in the for loop below, so now it is not simply copying m_lookupNodes into another vector.

PR #1280, opened 2019-01-23
Debian Stretch compile fix

bmatthewshea:
Fix: "apt install libssl1.0-dev"

## Description
Fixes compile warnings for Debian Stretch (clean VM/new install) that appear as:

`libcrypto.so.1.0.2, needed by /usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/libcurl.so, may conflict with libcrypto.so.1.1`

## Review Suggestion
I would see if you can just use `libssl1.0-dev` instead of `libssl-dev` on Ubuntu and others?

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ To run Zilliqa, we recommend the following minimum system requirements:
         libssl-dev libleveldb-dev libjsoncpp-dev libsnappy-dev cmake libmicrohttpd-dev \
         libjsonrpccpp-dev build-essential pkg-config libevent-dev libminiupnpc-dev \
         libprotobuf-dev protobuf-compiler libcurl4-openssl-dev libboost-program-options-dev \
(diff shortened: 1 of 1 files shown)


bb111189:
Unfortunately, this package is only available in Ubuntu bionic (18.04) onwards. As we are supporting and developing on Ubuntu 16.04, we are still not able to use `libssl1.0-dev`.

bb111189:
I am closing this PR for now.

bmatthewshea:
Understood @bb111189 . thanks for looking at it, though.
And maybe something to keep in mind.. seems to be the only real issue I've seen compiling under newer Debian/Ubuntu..

PR #1290, opened 2019-01-24
Log cleanup

ansnunez:
## Description

Some log improvements:
1. Reduced usage of `LOG_EPOCH`
2. Shortened some of the more frequent messages, such as for p2p and gossip
3. Reduced payload lengths
4. More consistent display of committee/shard/peer lists
5. Overall, tried to make logs more concise and able to show in single line
6. Also added temp var before calling `LIMIT_RIGHT` in logger, to avoid repeated string creation inside the macro

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -234,9 +234,9 @@ void DirectoryService::InjectPoWForDSNode(VectorOfPoWSoln& sortedPoWSolns,
   for (unsigned int i = 0; i < numOfProposedDSMembers; i++) {
     // TODO: Revise this as this is rather ad hoc. Currently, it is SHA2(PubK)
     // to act as the PoW soln
-    PubKey nodePubKey =
-        m_mediator.m_DSCommittee->at(m_mediator.m_DSCommittee->size() - 1 - i)
-            .first;
+    const unsigned int dsCommitteeIndex =
+        m_mediator.m_DSCommittee->size() - 1 - i;
+    PubKey nodePubKey = m_mediator.m_DSCommittee->at(dsCommitteeIndex).first;
     nodePubKey.Serialize(serializedPubK, 0);
     sha2.Update(serializedPubK);
     bytes PubKeyHash;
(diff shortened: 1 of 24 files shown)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -272,17 +272,19 @@ void DirectoryService::InjectPoWForDSNode(VectorOfPoWSoln& sortedPoWSolns,
...
       m_allPoWConns.emplace(m_mediator.m_DSCommittee->at(
           m_mediator.m_DSCommittee->size() - 1 - i));
-      LOG_GENERAL(INFO, m_mediator.m_DSCommittee
-                            ->at(m_mediator.m_DSCommittee->size() - 1 - i)
-                            .second);
+      LOG_GENERAL(INFO,
+                  "Injecting into PoW connections "
+                      << m_mediator.m_DSCommittee
+                             ->at(m_mediator.m_DSCommittee->size() - 1 - i)
+                             .second);
```
ShengguangXiao: Can assign "m_mediator.m_DSCommittee->size() - 1 - i" to a temp variable?

src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -316,19 +312,17 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
...
     // Accept slightly late entries as we need to multicast the DSBLOCK to
     // everyone if ((m_state != POW_SUBMISSION) && (m_state !=
     // DSBLOCK_CONSENSUS_PREP))
     if (!CheckState(VERIFYPOW)) {
-      LOG_EPOCH(INFO, m_mediator.m_currentEpochNum,
-                "Too late - current state is " << m_state);
+      LOG_GENERAL(INFO, "Too late. State = " << GetStateString());
     } else {
-      LOG_EPOCH(INFO, m_mediator.m_currentEpochNum, "POW verification passed");
+      LOG_GENERAL(INFO, "Verified OK");
```
ShengguangXiao: Log as "POW verified OK" will be more clear

ansnunez: Won't be necessary, since you can already see from the function name that it is for checking PoW submission:

```
[INFO][15306][19-01-24T14:17:43.918][oWProcessing.cpp:325][ProcessPoWSubmission] Verified OK
```

I tried to reduce the logs as much as possible, avoid being too wordy and have too much info.

ansnunez: BTW here is a fun fact, if we add something as small as "PoW ":

"PoW " = 4 bytes
4 bytes x 1800 submissions/epoch x 12 epochs/day x 365 days/year = 30MB per year!

PR #1295, opened 2019-01-25
Revert Contract State If Chain Call Failed

kaikawaliu:
## Description

Added a buffer in ContractStorage for storing the t_map before the execution.
When there is a chain call, and accidentally failed in the middle, 
the contract can use the buffer to revert back to the state when the txn was not executed.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/Account.cpp
+++ b/src/libData/AccountData/Account.cpp
@@ -223,13 +223,13 @@ bool Account::SetStorage(const vector<StateEntry>& state_entries, bool temp) {
 
 bool Account::SetStorage(const Address& addr,
                          const vector<pair<dev::h256, bytes>>& entries,
-                         bool temp, bool reversible) {
+                         bool temp, bool revertible) {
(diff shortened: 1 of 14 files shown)


src/libData/AccountData/Account.cpp
```diff
@@ -212,24 +212,24 @@ bool Account::DeserializeBase(const bytes& src, unsigned int offset) {
...
 
-bool Account::SetStorage(const vector<StateEntry>& state_entries, bool temp) {
+bool Account::SetStorage(const vector<StateEntry>& state_entries) {
   if (!isContract()) {
     return false;
   }
 
   return ContractStorage::GetContractStorage().PutContractState(
-      m_address, state_entries, m_storageRoot, temp);
+      m_address, state_entries, m_storageRoot, true);
```
kaikawaliu: true because this func is only called during txn processing

PR #1297, opened 2019-01-25
Add priority in JSONRPC

KaustubhShamshery:
* Add the var priority (by default false)
* Change logic for seed node forwarding if the flag is true

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test (40 nodes tested on #1302)

### DIFF
--- a/src/libMessage/ZilliqaMessage.proto
+++ b/src/libMessage/ZilliqaMessage.proto
@@ -359,6 +359,7 @@ message ProtoTransactionArray
     repeated ProtoTransaction transactions = 1;
 }
 
+
 message ProtoTransactionReceipt
 {
     optional bytes receipt    = 1;
@@ -754,6 +755,13 @@ message LookupSetMicroBlockFromLookup
     required ByteArray pubkey            = 2;
     required ByteArray signature         = 3;
 }
+
+message LookupForwardTxnsFromSeed
+{
+    optional ProtoTransactionArray shardtransactions = 1;
+    optional ProtoTransactionArray dstransactions    = 2;
+}
+
 
 message LookupGetTxnsFromLookup
 {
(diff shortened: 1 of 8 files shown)


src/libMessage/ZilliqaMessage.proto
```diff
@@ -755,6 +756,13 @@ message LookupSetMicroBlockFromLookup
     required ByteArray signature         = 3;
 }
 
+message LookupForwardTxnsFromSeed
+{
+    optional ProtoTransactionArray shardtransactions = 1;
+    optional ProtoTransactionArray dstransactions    = 2;
+}
```
ShengguangXiao: Message need to add pubkey and signature to verify lookup authenticity

src/libServer/Server.cpp
```diff
@@ -193,69 +212,85 @@ Json::Value Server::CreateTransaction(const Json::Value& _json) {
...
+          throw JsonRpcException(RPC_INVALID_ADDRESS_OR_KEY,
+                                 "Code is empty and To addr is null");
+          break;
+        default:
+          throw JsonRpcException(RPC_MISC_ERROR, "Txn type unexpected");
       }
     } else {
-      LOG_GENERAL(INFO, "No shards yet");
-      throw JsonRpcException(RPC_IN_WARMUP, "Could not create Transaction");
+      throw JsonRpcException(RPC_IN_WARMUP, "No Shards yet");
```
ShengguangXiao: Where is this exception captured?

src/libMessage/ZilliqaMessage.proto
KaustubhShamshery: message is from seed not lookup. https://github.com/Zilliqa/Issues/issues/425

src/libServer/Server.cpp
KaustubhShamshery: captured by default by JsonRpcCpp

iantanwx:
This has been tested and works as expected.

src/libMessage/ZilliqaMessage.proto
ShengguangXiao: Will this issue be fixed in this PR or seperately?

PR #1298, opened 2019-01-25
Fix _amount transfer during chained call

kaikawaliu:
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -1052,6 +1052,7 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
   }
 
   Address t_address = m_curContractAddr;
+  m_curSenderAddr = m_curContractAddr;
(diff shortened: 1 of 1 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -929,6 +936,15 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
     return false;
   }
 
+  try {
+    m_curAmount = boost::lexical_cast<boost::multiprecision::uint128_t>(
+        _json["message"]["_amount"].asString());
+  } catch (...) {
```
ShengguangXiao: Is there a race condition to m_curAmount?  Later it will be used in TransferBalanceAtomic, in between it may change?

kaikawaliu: The whole funcs in UpdateAccounts it's in one thread. TransferBalanceAtomic won't modify it also

PR #1302, opened 2019-01-25
merge/ds-contract-fixes

iantanwx:
## Description

This is a mega merge of #1288, #1295, #1297, #1298.
### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -894,68 +934,110 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
...
+    }
+
+    recipient = Address(_json["message"]["_recipient"].asString());
+    if (recipient == Address()) {
+      LOG_GENERAL(WARNING, "The recipient can't be null address");
+      receipt.AddError(RECEIPT_IS_NULL);
+      return false;
+    }
+
+    delete account;
```
sandipbhoir: I think we don’t use account*,  but just delete it.

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -894,68 +934,110 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(const Json::Value& _json,
...
+    account = this->GetAccount(recipient);
+
+    if (account == nullptr) {
+      AccountStoreBase<MAP>::AddAccount(recipient, {0, 0});
+    }
+
+    // Recipient is non-contract
+    if (!account->isContract()) {
```
sandipbhoir: Can potentially crash if account == nulltr at line 991

src/libData/AccountData/AccountStoreSC.tpp
kaikawaliu: This is for skipping the uninitialised warning during compilation

src/libData/AccountData/AccountStoreSC.tpp
kaikawaliu: GetAccount can give you a nullptr if the account was not created yet, so just Add a new Account obj for it shall be the right approach

src/libServer/Server.cpp
```diff
@@ -106,31 +106,50 @@ bool Server::StartCollectorThread() {
...
+      auto upperLayerNodes = m_mediator.m_lookup->GetAboveLayer();
+      auto upperLayerNode = upperLayerNodes.at(rand() % upperLayerNodes.size());
+
+      if (!Messenger::SetForwardTxnBlockFromSeed(
+              msg, MessageOffset::BODY,
+              m_mediator.m_lookup
+                  ->m_txnShardMap[SEND_TYPE::ARCHIVAL_SEND_SHARD],
+              m_mediator.m_lookup
+                  ->m_txnShardMap[SEND_TYPE::ARCHIVAL_SEND_DS])) {
+        continue;
```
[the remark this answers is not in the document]
kaikawaliu: @Kaustubh

src/libData/AccountData/AccountStoreSC.tpp
sandipbhoir: ok, may be we can initialize it as line 988 as `Account* account = nullptr`

src/libData/AccountData/AccountStoreSC.tpp
sandipbhoir: Is `AccountStoreBase<MAP>::AddAccount(recipient, {0, 0});` going to update local `account*` ?

src/libServer/Server.cpp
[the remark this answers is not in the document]
KaustubhShamshery: it is empty already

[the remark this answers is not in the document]
sandipbhoir: so we assume that if serialization in SetForwardTxnBlockFromSeed fails, its becoz m_txnShardMap was empty?

constants_local.xml
```diff
@@ -85,7 +85,7 @@
         <LEGAL_GAS_PRICE_IP>127.0.0.1</LEGAL_GAS_PRICE_IP>
     </gas>
     <gossip>
-        <BROADCAST_GOSSIP_MODE>true</BROADCAST_GOSSIP_MODE>
+        <BROADCAST_GOSSIP_MODE>false</BROADCAST_GOSSIP_MODE>
```
Gnnng: One question here: does this mean that for local testing we will disable gossip from now on?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -362,6 +362,8 @@ void Node::ScheduleMicroBlockConsensus() {
 
   auto main_func = [this]() mutable -> void { RunConsensusOnMicroBlock(); };
 
+  SetState(MICROBLOCK_CONSENSUS_PREP);
```
sandipbhoir: Already we set the state in RunConsensusOnMicrkock at start. Infact AFAIK this redundant SetState was removed in prev Pr

src/libMessage/Messenger.cpp
```diff
@@ -5921,6 +5910,47 @@ bool Messenger::GetLookupSetStartPoWFromSeed(const bytes& src,
...
+  LookupForwardTxnsFromSeed result;
+
+  result.ParseFromArray(src.data() + offset, src.size() - offset);
+
+  if (!result.IsInitialized()) {
+    LOG_GENERAL(WARNING, "LookupForwardTxnsFromSeed initialization failed.");
+    return false;
+  }
+
+  ProtobufToTransactionArray(result.shardtransactions(), shardTransactions);
```
ansnunez: @KaustubhShamshery should be good to check the result of these calls here

src/libNode/FinalBlockProcessing.cpp
ansnunez: This is already set inside `RunConsensusOnMicroBlock` and I sort of remember me and @sandipbhoir noticing this duplicate before and removing it.

src/libData/AccountData/Account.cpp
```diff
@@ -164,11 +164,7 @@ bool Account::InitContract(const bytes& code, const bytes& initData,
     string vname = v["vname"].asString();
     string type = v["type"].asString();
 
-    Json::StreamWriterBuilder writeBuilder;
-    std::unique_ptr<Json::StreamWriter> writer(writeBuilder.newStreamWriter());
-    ostringstream oss;
-    writer->write(v["value"], &oss);
-    string value = oss.str();
+    string value = JSONUtils::convertJsontoStr(v["value"]);
```
Gnnng: This just got my attention as the `Json::StreamWriterBuilder` and `Json::StreamWriter` are created and destroyed on the heap for every `InitContract` call. This remains true after replacing the entire block with `JSONUtils:: convertJsontoStr`. Are we sure this is an efficient way of using these two functions? There're some discussion [here](https://github.com/open-source-parsers/jsoncpp/issues/658#issuecomment-333073996) about the performance.

src/libNode/Node.cpp
```diff
@@ -1264,8 +1264,13 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
     }
   }
 
-  if (m_mediator.m_lookup->IsLookupNode(from) &&
-      from.GetPrintableIPAddress() != "127.0.0.1") {
+  if ((m_mediator.m_lookup->IsLookupNode(from) &&
```
ansnunez: Perhaps we can remove `IsLookupNode` now here since we already do a check for `Lookup::VerifySenderNode` a few lines earlier.

constants_local.xml
ansnunez: @Gnnng also if we don't use `-t large` when bootstrapping, does the cloud test take the settings from this xml?

src/libServer/Server.cpp
```diff
@@ -277,23 +327,38 @@ Json::Value Server::GetTransaction(const string& transactionHash) {
...
+          _json["Info"] = "Contract Creation transaction";
+          break;
+        case CONTRACT_CALL:
+          _json["Type"] = CONTRACT_CALL;
+          _json["Info"] = "Contract Call transaction";
+          break;
+        case ERROR:
+        default:
+          throw JsonRpcException(RPC_MISC_ERROR, "Unknown txn type");
+      }*/
```
Gnnng: Perhaps needs some cleanup on the dead code

src/libServer/Server.cpp
```diff
@@ -193,69 +212,85 @@ Json::Value Server::CreateTransaction(const Json::Value& _json) {
...
           } else {
-            m_mediator.m_lookup->AddToTxnShardMap(tx, 0);
+            m_mediator.m_lookup->AddToTxnShardMap(
+                tx, SEND_TYPE::ARCHIVAL_SEND_SHARD);
           }
           ret["Info"] = "Non-contract txn, sent to shard";
           ret["TranID"] = tx.GetTranID().hex();
-        } else if (!tx.GetCode().empty() && tx.GetToAddr() == NullAddress) {
+          return ret;
+          break;
```
Gnnng: The break-after-return pattern repeats several times in this switch statement. Generally speaking, it's not clean as `break` won't be executed anyway. If `break` is only used to remind people, then perhaps we should not mix the use of break and return.

src/libServer/Server.cpp
```diff
@@ -193,69 +212,85 @@ Json::Value Server::CreateTransaction(const Json::Value& _json) {
...
+          throw JsonRpcException(RPC_INVALID_ADDRESS_OR_KEY,
+                                 "Code is empty and To addr is null");
+          break;
+        default:
+          throw JsonRpcException(RPC_MISC_ERROR, "Txn type unexpected");
       }
     } else {
-      LOG_GENERAL(INFO, "No shards yet");
-      throw JsonRpcException(RPC_IN_WARMUP, "Could not create Transaction");
+      throw JsonRpcException(RPC_IN_WARMUP, "No Shards yet");
```
Gnnng: consider move this case `num_shards <= 0` before and save some indention for the big chunk of code above.

```
if (num_shards <= 0) {
  throw ...;
}

unsigned int shard = Transaction::GetShardIndex(fromAddr, num_shards);
switch (GetTransactionType(tx)) {
  ...
}
```

src/libServer/Server.cpp
```diff
@@ -267,6 +300,21 @@ Json::Value Server::CreateTransaction(const Json::Value& _json) {
...
+    } else if (!tx.GetCode().empty() && tx.GetToAddr() == NullAddress) {
+      return CONTRACT_CREATION;
+    } else {
+      return ERROR;
+    }
+  } else {
+    return CONTRACT_CALL;
+  }
+}
+
```
Gnnng: How about this

```
  if (!tx.GetData().empty() && tx.GetToAddr() != NullAddress) { 
    return CONTRACT_CALL;
  }

  if (!tx.GetCode().empty() && tx.GetToAddr() == NullAddress) {
    return CONTRACT_CREATION;
  }

  if (tx.GetData().empty() && tx.GetToAddr() != NullAddress && tx.Getcode().emtpy()) {
    return NON_CONTRACT;
  }

  return ERROR;
```

Need someone else to confirm this is equivalent

iantanwx: Your version looks equivalent to me.

ansnunez: @Gnnng looks equivalent to me too. I have pushed the change

constants_local.xml
ansnunez: Should be fine.  I think @iantanwx has frequently needed to use this setting for his small-scale cloud tests

iantanwx: Yes I usually switch it to the naive multicast for the small e2e tests.

src/libNode/Node.cpp
kaikawaliu: The verifier only check the txn origin’s PubKey. But the from Peer will change in each hop of gossip. So the earlier check can’t take the place of the IsLoopup check

PR #1303, opened 2019-01-26
Swap difficulty levels if DS diff goes below diff

ansnunez:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


ansnunez:
#1309 addresses the root cause more closely

PR #1305, opened 2019-01-26
Clean logs in ProcessPoWSubmissionFromPacket

ansnunez:
## Description

Our DS logs are **extremely** long during PoW processing.
There is a bunch of redundant information as well.
This PR makes these changes to `DirectoryService::ProcessPoWSubmissionFromPacket`:

1. Removed redundant messages after each of the 3 calls to `CheckState()`. You can tell which one caused a failure anyway.
2. Shortened the message inside `CheckState()` (removed word "Action").
3. Removed POWSTAT message and timer start/stop since a DS has to do this for **ALL** PoW submissions including duplicates.
4. Removed "Verified OK" message, since at this point of the function the PoW is already fine. It's better to just print the other cases (e.g., duplicate or replaced PoW).
5. Shortened the "Replaced harder PoW" message. Don't need to print the key again here.
6. Shortened the "Max PoW sent" message. Moved the key/peer/diff printing up so we don't need to print the key again here.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/PoWProcessing.cpp
+++ b/src/libDirectoryService/PoWProcessing.cpp
@@ -206,19 +206,16 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
     if (cv_POWSubmission.wait_for(
             cv_lk, std::chrono::seconds(POW_SUBMISSION_TIMEOUT)) ==
         std::cv_status::timeout) {
-      LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
-                "Time out while waiting for state transition ");
+      LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum, "State wait timed out");
     }
 
-    LOG_EPOCH(INFO, m_mediator.m_currentEpochNum,
-              "State transition is completed. (check for timeout)");
+    LOG_EPOCH(INFO, m_mediator.m_currentEpochNum, "State transition completed");
   }
 
   if (!CheckState(PROCESS_POWSUBMISSION)) {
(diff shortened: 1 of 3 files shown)


src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -336,17 +324,14 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
...
-                                          << "... for " << submitterPubKey);
+        // string harderSolnStr, oldSolnStr;
+        // DataConversion::charArrToHexStr(soln.result, harderSolnStr);
+        // DataConversion::charArrToHexStr(m_allPoWs[submitterPubKey].result,
+        // oldSolnStr);
+        LOG_GENERAL(INFO, "Replaced PoW");
         m_allPoWs[submitterPubKey] = soln;
       } else if (m_allPoWs[submitterPubKey].result == soln.result) {
-        LOG_GENERAL(INFO, "Ignored duplicate PoW");
+        LOG_GENERAL(INFO, "Duplicate PoW");
```
bb111189: Duplicated

src/libNode/Node.cpp
```diff
@@ -979,8 +979,8 @@ bool Node::CheckState(Action action) {
 
   if (!found) {
     LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
-              "Action " << GetActionString(action) << " not allowed in state "
-                        << GetStateString());
+              GetActionString(action)
+                  << " not allowed in state " << GetStateString());
```
bb111189: not allowed in ~state~

PR #1307, opened 2019-01-26
Check prevHash in validator and GetDSBlock

KaustubhShamshery:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libValidator/Validator.cpp
```diff
@@ -254,6 +254,8 @@ bool Validator::CheckDirBlocks(
   uint64_t totalIndex = index_num;
```
sandipbhoir: Probably, CheckDirBlock will need mutex now. Since can be called by ProcessSetDSBlockFromSeed and also ProcessSetDirBlocksFromSeed at same time

ckyang:
This PR fixed https://github.com/Zilliqa/Issues/issues/427

ckyang:
There is no `Cosig verification failed` using latest commit `54e2081` now.

PR #1311, opened 2019-01-26
Cleaned up logs for shard node

ansnunez:
## Description

One more round of log cleanup, for the messages that appear in shard node's log

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -2072,11 +2072,10 @@ std::string Node::GetActionString(Action action) const {
              : ActionStrings.at(action);
 }
 
-/*static*/ bool Node::GetDSLeader(const BlockLink& lastBlockLink,
-                                  const DSBlock& latestDSBlock,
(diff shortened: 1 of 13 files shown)


src/libNode/Node.cpp
```diff
@@ -2070,14 +2068,14 @@ map<Node::Action, string> Node::ActionStrings = {
...
-             ? "Unknown"
+             ? "UNKNOWN"
              : ActionStrings.at(action);
 }
 
 /*static*/ bool Node::GetDSLeader(const BlockLink& lastBlockLink,
                                   const DSBlock& latestDSBlock,
                                   const DequeOfNode& dsCommittee,
-                                  const uint64_t epochNumber,
+                                  [[gnu::unused]] const uint64_t epochNumber,
```
ShengguangXiao: Can remove this epochNumber parameter? Since it is unused

PR #1313, opened 2019-01-26
Check return value of ProcessStateDeltaFromFinalBlock in ProcessFinal…

ansnunez:
…Block

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -672,8 +672,10 @@ bool Node::ProcessFinalBlock(const bytes& message, unsigned int offset,
     }
   }
 
-  ProcessStateDeltaFromFinalBlock(stateDelta,
-                                  txBlock.GetHeader().GetStateDeltaHash());
+  if (!ProcessStateDeltaFromFinalBlock(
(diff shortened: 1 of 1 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -672,8 +672,10 @@ bool Node::ProcessFinalBlock(const bytes& message, unsigned int offset,
     }
   }
 
-  ProcessStateDeltaFromFinalBlock(stateDelta,
-                                  txBlock.GetHeader().GetStateDeltaHash());
+  if (!ProcessStateDeltaFromFinalBlock(
```
sandipbhoir: I think with this check, the node won't RejoinAsNormal anymore, which is called later with CheckStateRoot

```
  if (!LOOKUP_NODE_MODE &&
      (!CheckStateRoot(txBlock) || m_doRejoinAtStateRoot)) {
    RejoinAsNormal();
    return false;
  } else if (LOOKUP_NODE_MODE && !CheckStateRoot(txBlock)) {
    return false;
  }
```

PR #1321, opened 2019-01-27
Fix Coinbase

KaustubhShamshery:
## Description
Fix the reward distribution to match the percentages. 

1. Lookup may still get lucky draw
2. Guard node dont get any type of coinbase, and the award they are suppose to get remains in the coinbase addr.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/Coinbase.cpp
+++ b/src/libDirectoryService/Coinbase.cpp
@@ -421,9 +421,10 @@ void DirectoryService::InitCoinbase() {
       base_reward, base_reward_each, lookupReward, reward_each_lookup,
       nodeReward,  reward_each,      balance_left, PubKey(PrivKey()),
       Address()};
-  StoreCoinbaseInDiagnosticDB(entry);
+
   if (nonGuard.empty()) {
     LOG_GENERAL(WARNING, "No non-guard found, skip LuckyDraw");
+    StoreCoinbaseInDiagnosticDB(entry);
     return;
   }
 
@@ -445,6 +446,8 @@ void DirectoryService::InitCoinbase() {
                           << "][" << m_mediator.m_currentEpochNum << "]["
                           << balance_left << "] lucky draw");
   }
+  entry.luckyDrawWinnerAddr = luckyAddr;
+  StoreCoinbaseInDiagnosticDB(entry);
 }
 
 void DirectoryService::StoreCoinbaseInDiagnosticDB(
--- a/src/libDirectoryService/Coinbase.cpp
+++ b/src/libDirectoryService/Coinbase.cpp
@@ -354,6 +354,7 @@ void DirectoryService::InitCoinbase() {
   }
 
   // Reward based on cosigs
+  vector<Address> nonGuard;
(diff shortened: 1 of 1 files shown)


src/libDirectoryService/Coinbase.cpp
```diff
@@ -371,6 +376,7 @@ void DirectoryService::InitCoinbase() {
           if (GUARD_MODE) {
             if (Guard::GetInstance().IsNodeInDSGuardList(pk) ||
                 Guard::GetInstance().IsNodeInShardGuardList(pk)) {
+              suc_counter++;
```
ansnunez: `suc_counter` means success counter?  If so, why do we increment it here if we don't actually credit `reward_each`?

src/libDirectoryService/Coinbase.cpp
```diff
@@ -397,19 +403,21 @@ void DirectoryService::InitCoinbase() {
   }
 
   uint128_t balance_left = total_reward - (suc_counter * reward_each) -
-                           (suc_lookup_counter * reward_each_lookup) +
-                           base_reward - (node_count * base_reward_each);
+                           (suc_lookup_counter * reward_each_lookup) -
+                           (node_count * base_reward_each);
```
ansnunez: Should `base_reward_each` also rely on some success counter like `suc_base_counter` instead of `node_count`?

src/libDirectoryService/Coinbase.cpp
kaikawaliu: because we want to divide the total rewards with the whole size of the network, and skip those guard node when do the  actual rewarding.

src/libDirectoryService/Coinbase.cpp
```diff
@@ -390,75 +398,51 @@ void DirectoryService::InitCoinbase() {
...
-            break;
-          }
-        }
-      }
-      if (breakLoop) {
-        break;
-      }
+  if (nonGuard.empty()) {
+    LOG_GENERAL(WARNING, "No non-guard found, skip LuckyDraw");
+    return;
```
ansnunez: Call `StoreCoinbaseInDiagnosticDB(entry)` before return

src/libDirectoryService/Coinbase.cpp
```diff
@@ -390,75 +398,51 @@ void DirectoryService::InitCoinbase() {
...
+  // Only log reward for my self so can find out the reward of mine in state
+  // log
+  if (luckyAddr == myAddr) {
+    LOG_STATE("[REWARD][" << setw(15) << left
+                          << m_mediator.m_selfPeer.GetPrintableIPAddress()
+                          << "][" << m_mediator.m_currentEpochNum << "]["
+                          << balance_left << "] lucky draw");
   }
 
   StoreCoinbaseInDiagnosticDB(entry);
```
ansnunez: Set `entry.luckyDrawWinnerAddr = luckyAddr` here
Note that `entry.luckyDrawWinnerKey` cannot be known anymore, but should be ok

PR #1322, opened 2019-01-27
Test/dspowsolution

jendis:
## Description
Covers 100% of libData/MiningData/DSPowSolution

## Review Suggestion
In test implementation file Test_DSPowSolution.cpp serves the line 39 for a linker satisfaction otherwise undefined references to several Messenger::* methods used by *BlockHeader classes used by TestUtils occures.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/Data/Test_DSPowSolution.cpp
```diff
@@ -0,0 +1,85 @@
...
+using namespace std;
+
+BOOST_AUTO_TEST_SUITE(dspowsolutiontest)
+
+BOOST_AUTO_TEST_CASE(testDSPowSolutionClass) {
+  INIT_STDOUT_LOGGER();
+  LOG_MARKER();
+
+  // Satisfy linker
+  Account acc1(0, 0);
```
ShengguangXiao: This variable is useless

tests/Data/CMakeLists.txt
```diff
@@ -52,6 +52,11 @@ target_include_directories(Test_EIP PUBLIC ${CMAKE_SOURCE_DIR}/src)
 target_link_libraries(Test_EIP PUBLIC AccountData Crypto Trie Utils Persistence)
 add_test(NAME Test_EIP COMMAND Test_EIP)
 
+add_executable(Test_DSPowSolution Test_DSPowSolution.cpp)
```
ShengguangXiao: Put this section together with Test_POW

tests/Data/Test_DSPowSolution.cpp
ansnunez: It seems like this was added to fix a linker issue with Messenger. I think we can keep it in the meantime.

tests/Data/CMakeLists.txt
jendis: I believe in context this could make sense, but I think that until now the tests were organized according to the libraries, so you exactly knew where to find the tests. So what I would like to suggest is to even put the _Test_DSPowSolution.cpp_ into tests/Data/MiningData/Test_DSPowSolution.cpp.
Thank you for consideration.

tests/Data/Test_DSPowSolution.cpp
jendis: Unfortunately @ansnunez is right - I was not able to find any appropriate solution for linker satisfaction - still not clear to me why this is happening.

ShengguangXiao: I see

PR #1327, opened 2019-01-28
 Limit size of txn packet

kaikawaliu:
### Implementation

- [x] **ready for review**

- [x] local machine test (gentxn + contract)

- [x] small-scale cloud test


src/libMessage/Messenger.cpp
```diff
@@ -4349,14 +4349,39 @@ bool Messenger::SetNodeForwardTxnBlock(
...
+    ProtoTransaction* protoTxn = new ProtoTransaction();
+    TransactionToProtobuf(*iter, *protoTxn);
+    unsigned txn_size = protoTxn->ByteSize();
+    if ((msg_size + txn_size) > PACKET_BYTESIZE_LIMIT &&
+        txn_size >= SMALL_TXN_SIZE) {
+      continue;
+    }
+    *result.add_transactions() = *protoTxn;
     txnsCurrentCount++;
+    msg_size += protoTxn->ByteSize();
```
KaustubhShamshery: is this safe as u use the same iterator to iterate

src/libMessage/Messenger.cpp
```diff
@@ -4349,14 +4349,39 @@ bool Messenger::SetNodeForwardTxnBlock(
...
 
   for (const auto& txn : txnsGenerated) {
-    TransactionToProtobuf(txn, *result.add_transactions());
+    if (msg_size >= PACKET_BYTESIZE_LIMIT) {
+      break;
+    }
+    ProtoTransaction* protoTxn = new ProtoTransaction();
+    TransactionToProtobuf(txn, *protoTxn);
+    unsigned txn_size = protoTxn->ByteSize();
+    if ((msg_size + txn_size) > PACKET_BYTESIZE_LIMIT &&
```
KaustubhShamshery: What is the logic here?

kaikawaliu: first check is msg_size exceeded or not.
then we take a txn, to see if msg_size + this txn exceeded the limit,
if exceeded, if it's a very huge txn, we ignore it.
if it's a small txn, we accept it.
This can be used to prevent somebody intently put a very huge size txn to block the packet formation

PR #1334, opened 2019-01-28
Change Guard data storage structure  to set

KaustubhShamshery:
## Description
For efficiency, we can use set instead of vector to improve the searching effeciency

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libCrypto/Schnorr.h
+++ b/src/libCrypto/Schnorr.h
@@ -167,6 +167,22 @@ class PubKey : public Serializable {
   }
 };
 
+// hash for using PubKey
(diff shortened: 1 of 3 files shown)


kaikawaliu:
Set doesn't guarantee higher performance with large num of elements, why not use unordered_set

KaustubhShamshery:
> Set doesn't guarantee higher performance with large num of elements, why not use unordered_set

Because need to use hash <> for the PubKey, otherwise it does not compile well.

PR #1335, opened 2019-01-28
Verify and store pow from leader to local pow list

ShengguangXiao:
## Description

Verify and store pow from leader to local pow list

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -408,6 +419,10 @@ bool DirectoryService::VerifyPoWWinner(
         }
         // Insert the DS pow to my DS pow list so later can calcualte DS
         // difficulty
+        {
+          std::lock_guard<mutex> lock(m_mutexAllPOW);
+          m_allPoWs[DSPowWinner.first] = dsPowSoln;
```
ansnunez: should be m_allDSPoWs?

PR #1339, opened 2019-01-29
Read hostname for multiplier

sandipbhoir:
## Description
Read hostname for multiplier

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -161,8 +161,8 @@ void Lookup::SetLookupNodes() {
                     }) != m_lookupNodes.end()) {
           continue;
         }
-        // check for hostname only for lookup nodes
-        if (lookupType == "node.lookups") {
+        // check for hostname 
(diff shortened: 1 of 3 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -161,8 +161,8 @@ void Lookup::SetLookupNodes() {
                     }) != m_lookupNodes.end()) {
           continue;
         }
-        // check for hostname only for lookup nodes
-        if (lookupType == "node.lookups") {
+        // check for hostname 
+        if (lookupType == "node.lookups" || lookupType == "node.multiplier") {
```
ansnunez: Is it "node.multiplier" or "node.multipliers"?  A few lines later below it is the latter.

sandipbhoir: Had pushed the fix already :)

PR #1348, opened 2019-01-29
Fix receipt cumulative gas when deployment failed

kaikawaliu:
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -270,7 +270,7 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       if (!ret_checker) {
         receipt.AddError(CHECKER_FAILED);
       }
-      receipt.SetCumGas(CONTRACT_CREATE_GAS);
+      receipt.SetCumGas(transaction.GetGasLimit() - gasRemained);
       receipt.update();
 
       this->IncreaseNonce(fromAddr);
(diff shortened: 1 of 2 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -270,7 +270,7 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       if (!ret_checker) {
         receipt.AddError(CHECKER_FAILED);
       }
-      receipt.SetCumGas(CONTRACT_CREATE_GAS);
+      receipt.SetCumGas(transaction.GetGasLimit() - gasRemained);
```
bb111189: safe math if neccessary

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -738,7 +738,14 @@ bool AccountStoreSC<MAP>::ParseCreateContractJsonOutput(
     receipt.AddError(NO_GAS_REMAINING_FOUND);
     return false;
   }
-  gasRemained = atoi(_json["gas_remaining"].asString().c_str());
+  try {
+    gasRemained =
+        boost::lexical_cast<uint64_t>(_json["gas_remaining"].asString());
```
bb111189: ensure when casting string to uint, it won't have any underflow/overflow issue if neccessary

src/libData/AccountData/AccountStoreSC.tpp
kaikawaliu: gasRemained is initialized with `transaction.GetGasLimit()`, and the deduction afterward are all wrapped with safemath, unless we change the c++ code, it won't underflow

PR #1372, opened 2019-02-01
Fix argument processing for the NAT address option.

nnamon:
## Description

A simple fix to allow the `--address NAT` to work as expected. The current behaviour complains that 'NAT' is not a valid IP address because of a spurious IP format check. This simply moves the check to inside the `if address != "NAT"` block where it makes sense to check for the IP address format.


bb111189:
Thanks amon

PR #1377, opened 2019-02-03
Test/blockchain

jendis:
## Description
Maximum achievable coverage of libData/BlochChainData.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


tests/Data/BlockChainData/Test_BlockChain.cpp
```diff
@@ -0,0 +1,251 @@
...
+void appendBlockLinkAndChain_v(BlockLinkChain& blc, vector<BlockLink>& bl_v,
+                               const uint8_t& min, const uint8_t& max) {
+  if (max < min) {
+    throw "Invalid range, to < from + 1";
+  }
+
+  uint8_t lastIndex = bl_v.size();
+
+  for (uint8_t i = lastIndex;
+       i < lastIndex + TestUtils::RandomIntInRng<uint8_t>(min, max); i++) {
```
ansnunez: Will `RandomIntInRng` be re-evaluated for every iteration?  Might be better to move out the expression `lastIndex + RandomIntInRng` as a constant or variable?

tests/Data/BlockChainData/Test_BlockChain.cpp
```diff
@@ -0,0 +1,251 @@
...
+  // Add BlockLink with lower index than the latest
+  uint64_t index_old = 1;
+  BOOST_CHECK_MESSAGE(
+      blc.AddBlockLink(index_old, 1, DS, BlockHash::random()) == false,
+      "Can add BlockLink with index " + to_string(index_old) +
+          " lower then the latest index " + to_string(blc.GetLatestIndex()) +
+          ".\n");
+}
+
+DSBlockHeader createDSBlockHeader(const uint64_t& blockNum) {
```
ansnunez: If we can reuse this, perhaps we can move it to libTestUtils?
We have an existing `GenerateRandomDSBlockHeader()` there, but not one with a `blockNum` parameter.
You can probably extend that one.

tests/Data/BlockChainData/Test_BlockChain.cpp
```diff
@@ -0,0 +1,251 @@
...
+  return DSBlockHeader(TestUtils::DistUint8(), TestUtils::DistUint8(),
+                       TestUtils::GenerateRandomPubKey(), blockNum,
+                       TestUtils::DistUint64(), TestUtils::DistUint128(),
+                       SWInfo(), map<PubKey, Peer>(), DSBlockHashSet(),
+                       TestUtils::DistUint32(), CommitteeHash(), BlockHash());
+}
+
+TxBlockHeader createTxBlockHeader(const uint64_t& blockNum) {
```
ansnunez: Same here

tests/Data/BlockChainData/Test_BlockChain.cpp
jendis: Ou, of course it will, thanks and fixed!

tests/Data/BlockChainData/Test_BlockChain.cpp
jendis: Done

tests/Data/BlockChainData/Test_BlockChain.cpp
jendis: Done

PR #1390, opened 2019-02-07
Use subsets with dsguardonly subset having least priority for DS consensus

sandipbhoir:
## Description
This PR covers following:

1. We use 2 subsets for DS level consensus.
    1.a. Subset 0 - Includes commits from DS-guards only. However, if no enough commits are received
                            from dsguards, we consider the commits from the rest of committee.
    1.b. Subset 1 - Includes commits from all.

Note: subsets in general can be any number. First one is solely for ds-guard. Rest of subsets are random combinations.  subset 0 will be last to be started. 

2. Normal shard consensus use only 1 subset and behaviour is same as before.

Addresses issue - https://github.com/Zilliqa/Issues/issues/443

### Implementation

- [x] **ready for review**

- [x] local machine test

 - [x] small-scale cloud test (30 nodes, ds - 20, --guard 14/5, commit: c70789f ) 
 - [x] small-scale cloud test (200 nodes, ds- 160, commit: 78c2789)


constants.xml
```diff
@@ -29,10 +29,13 @@
         <SEED_TXN_COLLECTION_TIME_IN_SEC>5</SEED_TXN_COLLECTION_TIME_IN_SEC>
     </seed>
     <consensus>
-        <COMMIT_WINDOW_IN_SECONDS>5</COMMIT_WINDOW_IN_SECONDS>
+        <COMMIT_WINDOW_IN_SECONDS>10</COMMIT_WINDOW_IN_SECONDS>
         <CONSENSUS_MSG_ORDER_BLOCK_WINDOW>10</CONSENSUS_MSG_ORDER_BLOCK_WINDOW>
         <CONSENSUS_OBJECT_TIMEOUT>20</CONSENSUS_OBJECT_TIMEOUT>
-        <NUM_CONSENSUS_SUBSETS>1</NUM_CONSENSUS_SUBSETS>
+        <NUM_CONSENSUS_SUBSETS>2</NUM_CONSENSUS_SUBSETS>
```
KaustubhShamshery: looks  like this var is not useful

sandipbhoir: yup. removed.

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -139,11 +139,54 @@ void ConsensusLeader::GenerateConsensusSubsets() {
...
-    for (unsigned int j = 0; j < m_numForConsensus - 1; j++) {
-      unsigned int index = peersWhoCommitted.at(j);
-      subset.commitPointMap.at(index) = m_commitPointMap.at(index);
-      subset.commitPoints.emplace_back(m_commitPointMap.at(index));
-      subset.commitMap.at(index) = true;
+    // If DS consensus, then first subset should be of dsguard commits only.
+    // Fill in from rest if commits from dsguards < m_numForConsensus
+    if (m_DS && GUARD_MODE && (i == 0)) {
+      unsigned int subsetPeers = 1;  // myself
+      vector<int> nondsguardIndexes;
```
bb111189: make this unsigned int

sandipbhoir: Fixed

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -107,13 +108,12 @@ void ConsensusLeader::GenerateConsensusSubsets() {
...
-  // Generate NUM_CONSENSUS_SUBSETS lists (= subsets of peersWhoCommitted)
+  // Generate m_numOfSubsets lists (= subsets of peersWhoCommitted)
   // If we have exactly the minimum num required for consensus, no point making
   // more than 1 subset
 
-  const unsigned int numSubsets = (peersWhoCommitted.size() < m_numForConsensus)
-                                      ? 1
-                                      : NUM_CONSENSUS_SUBSETS;
+  const unsigned int numSubsets =
+      (peersWhoCommitted.size() < m_numForConsensus) ? 1 : m_numOfSubsets;
```
ansnunez: Now that I think about it, the check here should be `<=`, not `<`

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -139,11 +139,54 @@ void ConsensusLeader::GenerateConsensusSubsets() {
...
+      if (subsetPeers < m_numForConsensus) {
+        // Add from rest of nondsguards commits
+        LOG_GENERAL(
+            WARNING,
+            "[SubsetID: "
+                << i << "] Did'nt got all commits from ds-guards. Expected = "
+                << m_numForConsensus << ", Received = " << subsetPeers);
+        LOG_GENERAL(WARNING, "Will consider commits from non-dsguard nodes");
+
+        for (auto index : nondsguardIndexes) {
```
ansnunez: Shouldn't you continue to increment `subsetPeers` here and break this `for` loop once you reach `m_numForConsensus`?

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -139,11 +139,54 @@ void ConsensusLeader::GenerateConsensusSubsets() {
...
+
+      // check if we fall short of commits from dsguards
+      if (subsetPeers < m_numForConsensus) {
+        // Add from rest of nondsguards commits
+        LOG_GENERAL(
+            WARNING,
+            "[SubsetID: "
+                << i << "] Did'nt got all commits from ds-guards. Expected = "
+                << m_numForConsensus << ", Received = " << subsetPeers);
+        LOG_GENERAL(WARNING, "Will consider commits from non-dsguard nodes");
```
ansnunez: Maybe we can remove this message as it doesn't add much new info

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -139,11 +139,54 @@ void ConsensusLeader::GenerateConsensusSubsets() {
...
+        }
+      }
+
+      // check if we fall short of commits from dsguards
+      if (subsetPeers < m_numForConsensus) {
+        // Add from rest of nondsguards commits
+        LOG_GENERAL(
+            WARNING,
+            "[SubsetID: "
+                << i << "] Did'nt got all commits from ds-guards. Expected = "
```
ansnunez: "Did'nt got" -> "Didn't get"

Or I suggest to move this log message to before the `if` statement above and just display something shorter like "[SubsetID: X] Guards = X, Non-guards = X"

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -177,20 +220,29 @@ void ConsensusLeader::StartConsensusSubsets() {
   }
 
   m_numSubsetsRunning = m_consensusSubsets.size();
-  for (unsigned int index = 0; index < m_consensusSubsets.size(); index++) {
+  // subset 0 last to be started. giving community nodes the advantage
+  for (unsigned index = m_consensusSubsets.size(); index > 0; index--) {
```
ansnunez: Change `unsigned` to `unsigned int` to be more explicit

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: Wanted to log this explicitly only when number of dsguards commit received are short of expectation. So is added inside if statement.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: I think it adds explictly info that we are considering commits from non-dsguards. We should see this log only in WARNING situations, since inside if statement.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: Agree.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: No. since nondsguardIndexes only contains the remaining diff in missing commits.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: ok.

src/libConsensus/ConsensusLeader.cpp
ansnunez: Ok, can keep here, but I think a shorter and more informative message like the one proposed can suffice.  Also `m_numForConsensus` is already logged a few lines earlier anyway, so no need to display it again here.

And if you keep the log here (inside the if), and display the exact number of guards and non-guards, then that next log line below is already implied, so no need for it.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: i got your point.

src/libConsensus/ConsensusLeader.cpp
sandipbhoir: ok

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -177,20 +220,29 @@ void ConsensusLeader::StartConsensusSubsets() {
...
     // FINALCHALLENGE_DONE then it means consensus has ended and there's no
     // point in starting another subset
     if (m_state != CHALLENGE_DONE && m_state != FINALCHALLENGE_DONE) {
       break;
     }
-    ConsensusSubset& subset = m_consensusSubsets.at(index);
+
+    // delay starting every subset to avoid network congestion
+    if (index < m_consensusSubsets.size()) {
+      LOG_GENERAL(INFO, "Waiting for "
```
ansnunez: Can make it shorter, like "[SubsetID: X] Waiting X seconds"?  Recently I am a fan of shorter messages.

sandipbhoir: Sure :)

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -817,16 +818,27 @@ ConsensusLeader::ConsensusLeader(
...
+
+  m_sufficientCommitsNumForSubsets =
+      ceil((double)m_committee.size() * COMMIT_TOLERANCE_PERCENT / 100);
+
   LOG_GENERAL(INFO, "Consensus ID      = " << m_consensusID);
   LOG_GENERAL(INFO, "Leader/My ID      = " << m_myID);
   LOG_GENERAL(INFO, "Committee size    = " << committee.size());
   LOG_GENERAL(INFO, "Num for consensus = " << m_numForConsensus);
   LOG_GENERAL(INFO, "Num for failure   = " << m_numForConsensusFailure);
+  LOG_GENERAL(INFO, "Num for commits before subset-creation = "
```
ansnunez: Maybe we log this only for m_DS == true?

sandipbhoir: ok

src/libConsensus/ConsensusLeader.cpp
```diff
@@ -817,16 +818,27 @@ ConsensusLeader::ConsensusLeader(
   m_commitPointMap.at(m_myID) = *m_commitPoint;
   m_commitCounter = 1;
 
-  m_allCommitsReceived = false;
+  m_sufficientCommitsReceived = false;
   m_commitRedundantCounter = 0;
   m_commitFailureCounter = 0;
   m_numSubsetsRunning = 0;
 
+  m_txnProcessingTimeout = std::max(
```
ansnunez: Please add a comment here to explain that we took this value from `Node::NotifyTimeout`, and that it is meant to give backups enough time to finish processing transactions and send back the commits

sandipbhoir: Fixed.

PR #1393, opened 2019-02-08
Use nonblocking socket for handling connection issue

sandipbhoir:
## Description
This PR will address issue - https://github.com/Zilliqa/Issues/issues/439

### Implementation

- [x] **ready for review**

- [x] local machine test 
- Checked by Mimicking connection timeout by Gossip RoundTime - 10ms.
- Checked connection refused.

- [x] small-scale cloud test


src/libNetwork/P2PComm.cpp
```diff
@@ -275,20 +336,26 @@ void SendJob::SendMessageCore(const Peer& peer, const bytes message,
...
       Blacklist::GetInstance().Add(peer.m_ipAddress);
       return;
+    }*/
+
+    if (Blacklist::GetInstance().Exist(peer.m_ipAddress)) {
+      return;
     }
 
+    retry_counter++;
+    LOG_GENERAL(WARNING, "Socket connect failed " << retry_counter << "/"
```
bb111189: shall we move the logging to before `retry_counter++;`. I think it is more accurate that way.

constants.xml
```diff
@@ -146,6 +146,7 @@
         <MAX_GOSSIP_MSG_SIZE_IN_BYTES>5000000</MAX_GOSSIP_MSG_SIZE_IN_BYTES>
         <MIN_READ_WATERMARK_IN_BYTES>0</MIN_READ_WATERMARK_IN_BYTES>
         <MAX_READ_WATERMARK_IN_BYTES>20000000</MAX_READ_WATERMARK_IN_BYTES>
+        <CONNECTION_TIMEOUT_IN_SECONDS>1</CONNECTION_TIMEOUT_IN_SECONDS>
```
bb111189: Is 1s timeout too aggressive or it is fine?

src/libNetwork/P2PComm.cpp
sandipbhoir: yeah.

constants.xml
sandipbhoir: I think it should be fine. considering we have gossip ROUND_TIME of 1 sec. However, once we do test with large scale (600 shard size) we can re-think of increasing to 2 sec. But i think 1 sec is good enough.

src/libNetwork/P2PComm.cpp
```diff
@@ -204,19 +210,74 @@ bool SendJob::SendMessageSocketCore(const Peer& peer, const bytes& message,
...
+                  "could not set TCP listening socket to be non-blocking...");
+      return false;
+    }
+
+    int status;
+    bool connectStat = true;
+    if ((status = connect(cli_sock, (struct sockaddr*)&serv_addr,
+                          sizeof(serv_addr))) < 0) {
+      if (errno != EINPROGRESS) {
+        LOG_GENERAL(WARNING, "Error connecting(1), Cancelling connection!");
```
ansnunez: Since we anticipate getting a lot of these messages, I suggest to make them very short.  Anyway, our log includes the line number, so it should be easy to know which failure was encountered.

sandipbhoir: ok

PR #1398, opened 2019-02-09
Daemon should not launch process with --loadconfig

ckyang:
## Description

Daemon should not launch process with --loadconfig

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


KaustubhShamshery:
As discussed before, loadConfig might be needed if ds node restart in 1 epoch

ckyang:
@KaustubhShamshery I think it's fine because daemon will always happened after epoch 1

PR #1413, opened 2019-02-11
Complete removal of PeerManager and Broadcastable

ansnunez:
## Description

#1388 was a quick fix to avoid the rebroadcasting issue.

This PR completely removes `Broadcastable` as it has stopped being part of our methods for communicating between nodes ever since gossip was introduced.

This PR also removes `PeerManager`, which, apart from loading the config.xml into `PeerStore` and setting the log level, is also no longer really used.

Removal of `PeerManager` includes these steps:
1. Move `SetupLogLevel` into another class that is called upon startup (I chose `Mediator`)
2. Delete `PeerStore`
3. Re-code `DirectoryService::ProcessSetPrimary` now that `PeerStore` is removed, to maintain the current behavior
4. Update the local testnet scripts to remove loadconfig option

Some other updates:
1. Also removed `ConsensusUser`, another long-unused class
2. In `test_zilliqa_lookup.py`, `config.xml` should be taken from `config_normal.xml`, not `config_lookup.xml`. In fact, a follow-up cleanup can include removing the creation `config_lookup.xml`, if confirmed totally meaningless for local testnet.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -722,7 +717,14 @@ bool Lookup::SetDSCommitteInfo() {
       struct in_addr ip_addr;
       inet_pton(AF_INET, v.second.get<string>("ip").c_str(), &ip_addr);
       Peer peer((uint128_t)ip_addr.s_addr, v.second.get<unsigned int>("port"));
-      m_mediator.m_DSCommittee->emplace_back(make_pair(key, peer));
+
+      if (replaceMyPeerWithDefault && (key == m_mediator.m_selfKey.second)) {
```
sandipbhoir: Can put replaceMyPeerWithDefault as second condition instead of first?

ansnunez: Since the first condition is faster to execute (just a bool check), I think it's better to keep it first.  We don't want to unnecessarily do a key comparison every time first.

PR #1414, opened 2019-02-11
Add asan and tsan to make image script

bb111189:
## Description

Add `asan` and `tsan` to `make_image_for_test.sh` script. Add logging to cmake. 

In the long run, we should merge `make_image.sh` and `make_image_for_test.sh`

### Implementation

- [ ] **ready for review**

- [ ] ~local machine test~

- [x] small-scale cloud test


ansnunez:
Should we have a separate `make_image` for just the sanitizer builds, so that Jenkins won't have to build all the VC test builds every time as well?

bb111189:
> Should we have a separate `make_image` for just the sanitizer builds, so that Jenkins won't have to build all the VC test builds every time as well?

That was done on the Jenkins end. Push to `sanitize/*`

PR #1418, opened 2019-02-13
Avoid massive memory consumption during state retrieval

kaikawaliu:
## Description

we don't have to refill the m_addressToAccount during the retrieval, and this may cause huge memory consumption. It should be fine to get the account from the m_state when it is needed after it been initialized.

### Implementation

- [x] **ready for review**

~- [ ] local machine test~

- [x] small-scale cloud test (tested with recovery)


ckyang:
@kaikawaliu Can you have a try using "recovery on separated testnet"? Because it's the real script we will use on mainnet.

PR #1422, opened 2019-02-13
Check if user using latest sw version to join mainnet

ShengguangXiao:
## Description

Check if user using latest sw version to join mainnet

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/SWInfo.cpp
+++ b/src/libUtils/SWInfo.cpp
@@ -16,6 +16,8 @@
  */
 
 #include "SWInfo.h"
+#include <curl/curl.h>
+#include <json/json.h>
 #include "libMessage/MessengerSWInfo.h"
 #include "libUtils/Logger.h"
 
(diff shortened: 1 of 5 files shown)


src/libUtils/SWInfo.h
```diff
@@ -22,8 +22,9 @@
 #include <iostream>
 #include "common/Serializable.h"
 
-const std::string ZILLIQA_BRAND =
-    "Copyright (C) Zilliqa. Version 4.0.2. <https://www.zilliqa.com/> ";
+const std::string VERSION_TAG = "v4.0.2";
```
ckyang: Since the version place changed, maybe the release document also need to be updated.

src/libUtils/SWInfo.cpp
```diff
@@ -172,27 +174,91 @@ bool SWInfo::operator!=(const SWInfo& r) const { return !(*this == r); }
...
+  ((string*)userp)->append((char*)contents, size * nmemb);
+  return size * nmemb;
+}
+
+bool SWInfo::IsLatestVersion() {
+  string curlRes;
+  auto curl = curl_easy_init();
+  curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);
+  curl_easy_setopt(curl, CURLOPT_URL,
+                   "https://api.github.com/repos/Zilliqa/Zilliqa/tags");
```
ckyang: Suggest to give this URL a meaningful definition, in case we need to change it in the future.

PR #1426, opened 2019-02-14
Shifting ProcessTxn before receiving announcement for backup node

kaikawaliu:
## Description

the way that backups verifying the txn order composed by the leader is letting themselves also process the txn from the txn-pool basing on the multi-index map.
previously is to start processing them when receiving the announcement, however this can be started earlier

so now the backup will generate an order locally before the announcement was received, and use this order to match the one from the leader.
so the leader and backup can share the same timewindow for processing txns, which utilize the time for the consensus

it's still following the way of verification the announcement. the only difference is previously the order is calculated in real time, now is to make it calculated in advance.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (with vc tests and contract tests)


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -676,7 +673,8 @@ bool Node::RunConsensusOnMicroBlockWhenShardLeader() {
 
   if (m_mediator.m_ds->m_mode == DirectoryService::Mode::IDLE &&
       !m_mediator.GetIsVacuousEpoch()) {
-    std::this_thread::sleep_for(chrono::milliseconds(TX_DISTRIBUTE_TIME_IN_MS));
+    std::this_thread::sleep_for(chrono::milliseconds(TX_DISTRIBUTE_TIME_IN_MS +
```
ansnunez: Why do we need to add the extra delay here for shard consensus?

ansnunez: Ah I mean, shouldn't this be for both shard MB and DS MB consensus?

kaikawaliu: the DS leader already have an extra sleep in RunConsensusOnFinalBlockWhenPrimary

kaikawaliu: The purpose of this sleep is to make sure the backup already finished the txn processing before the announcement was received

PR #1430, opened 2019-02-14
provide issue-template

lazaridiscom:
Minimal issue-template to relay users to the relevant issue-trackers.


bb111189:
https://github.com/Zilliqa/docs is not meant for public contribution. Hence, creating issue there will be inappropriate.  It is used by us to host resources for public viewing only.

PR #1433, opened 2019-02-15
Two-subset consensus

ansnunez:
## Description

Alternative solution to #1390

Challenge and Response protobuf messages are changed to receive one or more challenges/responses depending on the number of subsets.  This means leader and backup only have to talk to each other once, regardless of how many subsets the backup belongs to.

Communication method is preserved from current implementation where leader multicasts challenge instead of gossiping it.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libConsensus/ConsensusBackup.h
```diff
@@ -71,7 +68,7 @@ class ConsensusBackup : public ConsensusCommon {
                                    State nextstate);
   bool ProcessMessageChallenge(const bytes& challenge, unsigned int offset);
   bool GenerateResponseMessage(bytes& response, unsigned int offset,
-                               uint16_t subsetID);
+                               std::vector<ResponseSubsetInfo> subsetInfo);
```
sandipbhoir: can be declared as const &

ansnunez: ok will change

PR #1436, opened 2019-02-15
Use a non class variable when in DS consensus is not finalized

bb111189:
## Description

Copy `m_allDSPoWs` to a local variable to avoid polluting and inconsistency to the class variable when ds consensus is not yet completed. 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -747,7 +747,7 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary() {
     allPoWs = m_allPoWs;
   }
 
-    {
+  {
     std::lock_guard<std::mutex> g(m_mutexAllDSPOWs);
     allDSPoWs = m_allDSPoWs;
   }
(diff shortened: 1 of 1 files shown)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -764,16 +770,16 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary() {
         LOG_GENERAL(INFO,
                     "Node " << pubKeyPoW.first
                             << " failed to join because priority not enough.");
-        if (m_allDSPoWs.find(pubKeyPoW.first) != m_allDSPoWs.end()) {
-          m_allDSPoWs.erase(pubKeyPoW.first);
+        if (allDSPoWs.find(pubKeyPoW.first) != allDSPoWs.end()) {
+          allDSPoWs.erase(pubKeyPoW.first);
```
sandipbhoir: Do we need to reassign allDSPoWs back to m_allDSPoWs?

bb111189: no need. The variable was not used after DSBlockPre. It can be re-used if there is a view change.

Issue #1434, opened 2019-02-15
[Remote Mining] the timeout of DS PoW is wrong 


deepgully:
The timeout of PoW work is fixed "POW_WINDOW_IN_SECONDS" currently, which is not correct for DS PoW.
When the DS PoW send to proxy, the Shard PoW had taken up some time of PoW window, we should subtract the time taken by Shard PoW.

https://github.com/Zilliqa/Zilliqa/blob/0af01d4cb1333ed00d1177219946db0478d9901d/src/libPOW/pow.cpp#L372-L377

deepgully:
@ShengguangXiao could you help to fix?

ShengguangXiao:
@deepgully I will handle it

PR #1439, opened 2019-02-18
Fix DS pow time window not accurate problem

ShengguangXiao:
## Description

Fix DS pow time window not accurate problem

The DS pow time window is total pow time window minus shard pow time

https://github.com/Zilliqa/Zilliqa/issues/1434

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -183,7 +183,7 @@ bool POW::EthashConfigureClient(uint64_t block_number, bool fullDataset) {
 
 ethash_mining_result_t POW::MineGetWork(uint64_t blockNum,
                                         ethash_hash256 const& headerHash,
-                                        uint8_t difficulty) {
+                                        uint8_t difficulty, int timeWindow) {
   LOG_MARKER();
   int ethash_epoch = ethash::get_epoch_number(blockNum);
   std::string seed = BlockhashToHexString(ethash::calculate_seed(ethash_epoch));
@@ -193,15 +193,14 @@ ethash_mining_result_t POW::MineGetWork(uint64_t blockNum,
   PoWWorkPackage work = {headerStr, seed, boundary, blockNum, difficulty};
 
   GetWorkServer::GetInstance().StartMining(work);
-  // -1: wait for the 1st accept result
-  auto result = GetWorkServer::GetInstance().GetResult(-1);
+  auto result = GetWorkServer::GetInstance().GetResult(timeWindow);
   GetWorkServer::GetInstance().StopMining();
   return result;
 }
 
 ethash_mining_result_t POW::MineLight(ethash_hash256 const& headerHash,
                                       ethash_hash256 const& boundary,
-                                      uint64_t startNonce) {
+                                      uint64_t startNonce, int timeWindow) {
   uint64_t nonce = startNonce;
   auto startTime = std::chrono::high_resolution_clock::now();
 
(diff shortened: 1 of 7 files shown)


src/libPOW/pow.cpp
```diff
@@ -193,15 +193,14 @@ ethash_mining_result_t POW::MineGetWork(uint64_t blockNum,
   PoWWorkPackage work = {headerStr, seed, boundary, blockNum, difficulty};
 
   GetWorkServer::GetInstance().StartMining(work);
-  // -1: wait for the 1st accept result
-  auto result = GetWorkServer::GetInstance().GetResult(-1);
+  auto result = GetWorkServer::GetInstance().GetResult(timeWindow);
```
deepgully: the getwork server will sleep 'timeWindow' and return the last result, no chance for DS pow.

deepgully: we should change here
https://github.com/Zilliqa/Zilliqa/blob/4c192f3d61b7bd480aab0356ce29f0cb0d9639c2/src/libServer/GetWorkServer.cpp#L132-L136

ShengguangXiao: Yes, it is changed already. The node need to finish shard pow first then start to do DS pow. If somebody want to directly do DS pow, then he need to change the code himself

deepgully: https://github.com/Zilliqa/Zilliqa/blob/4c192f3d61b7bd480aab0356ce29f0cb0d9639c2/src/libServer/GetWorkServer.cpp#L118-L124

remove this sleep and use wait_ms in m_cvGotResult.wait_for?

ShengguangXiao: Yes, already modified like that

src/libServer/GetWorkServer.cpp
```diff
@@ -113,26 +113,11 @@ int GetWorkServer::GetSecondsToNextPoW() {
...
-// if wait_ms > 0: wait until timeout, return the last result
-ethash_mining_result_t GetWorkServer::GetResult(const int& wait_ms) {
-  if (wait_ms >= 0) {
-    std::this_thread::sleep_for(std::chrono::milliseconds(wait_ms));
-    return m_curResult;
-  }
-
-  // wait_ms < 0
-  // wait until the first accept result
-  if (!m_isMining || m_curResult.success) {
```
deepgully: should we keep this for these 2 conditions?
1. the pow was already done before call GetResult
2. the pow was stopped before call GetResult

src/libServer/GetWorkServer.h
```diff
@@ -147,7 +147,7 @@ class GetWorkServer : public AbstractStubServer {
   // if wait_ms < 0: wait until the first accept result
```
deepgully: the comments should be changed

src/libServer/GetWorkServer.cpp
ShengguangXiao: I think only need to check m_isMining, because if m_curResult.success is true, the conditional variable must be triggered, then it can return. I don't check m_curResult here because it is not protected by mutex

src/libServer/GetWorkServer.h
ShengguangXiao: Yes, this one need to change

src/libServer/GetWorkServer.cpp
ShengguangXiao: Ok, can add this check

PR #1444, opened 2019-02-19
Buffer and Process received final blocks after seed is syncedup

sandipbhoir:
## Description
This PR address issue - https://github.com/Zilliqa/Issues/issues/449

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test :  Scenario where in buffered txnblocks are rejected since stale.
- [x] small-scale cloud test :  Scenario where in buffered txnblocks are accepted ( Test image with -DSJ_TEST_SJ_TXNBLKS_PROCESS_SLOW , commit: dae6bf4 image: dae6bf4-sj1 ).

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -558,15 +558,14 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
       // Buffer the Final Block
       lock_guard<mutex> g(m_mutexSeedTxnBlksBuffer);
       m_seedTxnBlksBuffer.push_back(message);
-      LOG_GENERAL(INFO, "Seed not yet synced, so buffered this final block");
+      LOG_GENERAL(INFO, "Seed not synced, buffered this FBLK");
       return false;
     } else {
       // If seed node is synced and have buffered txn blocks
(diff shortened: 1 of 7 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -533,6 +533,13 @@ void Node::PrepareGoodStateForFinalBlock() {
 bool Node::ProcessFinalBlock(const bytes& message, unsigned int offset,
                              [[gnu::unused]] const Peer& from) {
   LOG_MARKER();
+  return ProcessFinalBlockCore(message, offset, from);
+}
+
+bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
```
kaikawaliu: is it necessary to split out a `Core` function for this case? I feel simply modify the `ProcessFinalBlock` shall be enough

sandipbhoir: Yeah, Initially i had done same. But our instrucionHandler expects only 3 parameters. And i have to use 4th optionally.
https://github.com/Zilliqa/Zilliqa/blob/631a49e407081b8e9eb2c0ecc4d2777ec3e3940f/src/libNode/Node.cpp#L1973

PR #1447, opened 2019-02-20
Add Checks for return from IncreaseBalance

kaikawaliu:
### Implementation

- [x] **ready for review**

~- [ ] local machine test~

- [x] small-scale cloud test (gentxn, contract)

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -281,7 +281,9 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       this->m_addressToAccount->erase(toAddr);
       return false;
     }
-    this->IncreaseBalance(fromAddr, gasRefund);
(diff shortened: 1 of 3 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -446,7 +448,9 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       return false;
     }
 
-    this->IncreaseBalance(fromAddr, gasRefund);
+    if (!this->IncreaseBalance(fromAddr, gasRefund)) {
+      LOG_GENERAL(WARNING, "IncreaseBalance failed for gasRefund");
```
ansnunez: What is the criterion for WARNING or FATAL?

kaikawaliu: Comparing to the others which are impossible to happen, this one is possible. If during a series of  chain call in one transaction, another contract transfers its balance to this sender, which may cause the balance being larger than the moment when the gasDeposit was not deducted, such may cause overflow issue.

PR #1479, opened 2019-02-25
Add exception checking inside convertStrtoJson

ansnunez:
## Description

`try-catch` now added to `convertStrtoJson`.

So, we also don't need to add `try-catch` wherever `convertStrtoJson` is called.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/JsonUtils.h
+++ b/src/libUtils/JsonUtils.h
@@ -51,10 +51,10 @@ class JSONUtils {
 
   /// Convert a string to Json object
   bool convertStrtoJson(const std::string& str, Json::Value& dstObj) {
-    std::lock_guard<std::mutex> g(m_mutexReader);
     bool result = true;
     try {
       std::string errors;
+      std::lock_guard<std::mutex> g(m_mutexReader);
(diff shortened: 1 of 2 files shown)


src/libUtils/JsonUtils.h
```diff
@@ -51,14 +51,20 @@ class JSONUtils {
...
-    std::string errors;
     std::lock_guard<std::mutex> g(m_mutexReader);
-    if (!m_reader->parse(str.c_str(), str.c_str() + str.size(), &dstObj,
-                         &errors)) {
-      LOG_GENERAL(WARNING, "Corrupted JSON: " << errors);
-      return false;
+    bool result = true;
+    try {
+      std::string errors;
+      if (!m_reader->parse(str.c_str(), str.c_str() + str.size(), &dstObj,
```
ShengguangXiao: Is this m_reader thread safe?

ShengguangXiao: Can put std::lock_guard<std::mutex> g(m_mutexReader); on top of m_reader->parse....?
So can easily find it is protected by mutex

ansnunez: Done

PR #1482, opened 2019-02-26
Fix m_syncType may have data race issue

ShengguangXiao:
## Description

Fix m_syncType may have data race issue

Change m_syncType to atomic

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


ShengguangXiao:
45 nodes test run to block 200 without problem

PR #1486, opened 2019-02-26
Add state check before entering WhenDone funcs for all consensus

kaikawaliu:
## Description

Fix race condition that may allow consensus message in the same round entering `WhenDone` function twice 

### Implementation

- [x] **ready for review**

~- [ ] local machine test~

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -1230,7 +1230,9 @@ void DirectoryService::RunConsensusOnDSBlock(bool isRejoin) {
     RejoinAsDS();
   }
 
-  SetState(DSBLOCK_CONSENSUS_PREP);
+  if (m_state != DSBLOCK_CONSENSUS_PREP) {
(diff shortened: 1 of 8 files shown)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -1230,7 +1230,9 @@ void DirectoryService::RunConsensusOnDSBlock(bool isRejoin) {
     RejoinAsDS();
   }
 
-  SetState(DSBLOCK_CONSENSUS_PREP);
+  if (m_state != DSBLOCK_CONSENSUS_PREP) {
```
ansnunez: Why do we need to check `m_state` here before setting?  Same comment for other change below.

kaikawaliu: It's just about the tidy of the log, if it harms I'll remove the if check

ansnunez: Hmm it's OK, we can keep

PR #1489, opened 2019-02-26
Improve messaging

ansnunez:
### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -1147,10 +1147,11 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
   uint32_t shardId = 0;
   PubKey lookupPubKey;
   vector<Transaction> transactions;
+  Signature signature;
 
   if (!Messenger::GetNodeForwardTxnBlock(message, offset, epochNumber,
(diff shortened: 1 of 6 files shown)


src/libNode/Node.cpp
```diff
@@ -1165,6 +1166,17 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
 
   LOG_GENERAL(INFO, "Received from " << from);
 
+  // Avoid using the original message for broadcasting in case it contains
+  // excess data beyond the TxnPacket
```
kaikawaliu: Can you give me some backgroudn on why is it bothering us, any issue caused by not doing so?

src/libNode/Node.cpp
```diff
@@ -1165,6 +1166,17 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
...
+  bytes message2 = {MessageType::NODE, NodeInstructionType::FORWARDTXNPACKET};
```
kaikawaliu: seems this `message2` is no longer used after serialization.

PR #1494, opened 2019-02-27
Adjust blacklist exclusion list

bb111189:
## Description

This PR adds DS guards to the blacklist exclusion list. The changes are as follows:
1. **[Completed]** Add function `RemoveExclude()` in `Blacklist.cpp\h` to remove IP from the exclusion list
2. **[Completed]** Add function `AddDSGuardToBlacklistExcludeList()` in `Guard.cpp\h` to remove IP from the exclusion 
3. **[Tested]** After recovery, DS guards are re-added to the exclusion list for 1) ds comm 2) node 3) lookup
4. **[Tested]** When a DS guard network information changes, old IP is removed from exclusion list and new IP is added to the exclusion list at 1) DS Committee and 2) Node/Lookup
5. **[Tested]** After the New node joined, add ds guard in ds committee to the exclusion list   
6. **[Tested]** During bootstrap, add ds guard to the exclusion list for 1) ds comm 2) node 3) lookup

Other fixes:
- Fixed a missing ds guard in local test

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test (50 nodes, guard mode enabled)


src/libDirectoryService/DirectoryService.cpp
```diff
@@ -257,6 +257,10 @@ bool DirectoryService::ProcessSetPrimary(const bytes& message,
     m_consensusMyID++;
   }
 
+  // Add ds guard to exclude list for ds comm at bootstrap
+  Guard::GetInstance().AddDSGuardToBlacklistExcludeList(
+      *m_mediator.m_DSCommittee);
```
ckyang: Maybe this piece of code should also be added inside Node::StartRetrieveHistory(), after setting the dscommittee from persistence? (https://github.com/Zilliqa/Zilliqa/blob/master/src/libNode/Node.cpp#L485)

bb111189: already exists in the branch

ckyang: I saw the code now

src/libNetwork/Guard.cpp
```diff
@@ -221,6 +222,28 @@ void Guard::AddToExclusionList(const uint128_t& ft, const uint128_t& sd) {
   }
 }
 
+void Guard::AddDSGuardToBlacklistExcludeList(const DequeOfNode& dsComm) {
+  if (GUARD_MODE) {
+    unsigned int dsIndex = 0;
+    for (const auto& i : dsComm) {
+      if (dsIndex < GetNumOfDSGuard()) {
```
sandipbhoir: may be we don't need check with `dsIndex`. we already check it below.

bb111189: the else should be in the outer `if`. Will make the change

bb111189: I guess it just let it break out of the loop earlier

PR #1497, opened 2019-02-28
Avoid excessive key searches in coinbase rewarding algorithm

ansnunez:
## Description

Current way is inefficient because of repeated pubkey-guard list checking:
```
for each block
   for each pubkey in block
      check pubkey against guard list
```

This PR keeps track of which pubkeys have already been identified as guards.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


KaustubhShamshery:
Overall, fine with the PR. only thing remains is to have a concrete check on the percentages.

PR #1498, opened 2019-02-28
Fix AccountStore Lock in MacOS

kaikawaliu:
## Description

in MacOS, the handling the previous way of declaring scoped lock is different from Linux, such resulted in stucking at mix usage of `lock_guard` and `shared_lock` or `unique_lock` in scoped lock.
The revised version passed the Test_AccountStore in both linux and macos system.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStore.cpp
+++ b/src/libData/AccountData/AccountStore.cpp
@@ -178,9 +178,9 @@ void AccountStore::MoveRootToDisk(const h256& root) {
 bool AccountStore::MoveUpdatesToDisk() {
   LOG_MARKER();
 
-  lock(m_mutexPrimary, m_mutexDB);
-  unique_lock<shared_timed_mutex> g(m_mutexPrimary, adopt_lock);
-  lock_guard<mutex> g2(m_mutexDB, adopt_lock);
+  unique_lock<shared_timed_mutex> g(m_mutexPrimary, defer_lock);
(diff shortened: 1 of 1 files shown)


Gnnng:
The following is a minimal reproducible example of the hanging issue.

It's because the `std::lock` will call the exclusive locking function on the mutex whereas the `shared_lock` will only involves shared locking and unlocking. 

In the `serialize()` function, the mutex `m1` got exclusived locked through `lock(m1, m2);` while the `shared_lock` only calls shared unlocking when the function exits. Therefore, when `deserialize` is called, it would immediately be blocked if `m1` was not exclusively unlocked.

```
// g++ --std=c++14
#include<iostream>
#include<mutex>
#include<shared_mutex>

using namespace std;

shared_timed_mutex m1;
mutex m2;

void serialize() {
    lock(m1, m2); // should not use std::lock on shared_timed_mutex
    shared_lock<shared_timed_mutex> shared(m1, adopt_lock);
    lock_guard<mutex> lk(m2, adopt_lock);
}

void deserialize() {
    // m1.unlock(); // without exclusive unlocking the program will hang on macOS
    lock(m1, m2);
    unique_lock<shared_timed_mutex> unique(m1, adopt_lock);
    lock_guard<mutex> lk(m2, adopt_lock);
}

int main() {
    serialize();
    deserialize();
}
```

PR #1501, opened 2019-03-01
Repopulate Merkle Trie When MoveUpdatesToAccountStore

kaikawaliu:
## Description

Merkle Trie is an immutable incremental data structure. Every updating of the value will introduce a new node being inserting into the trie. So the trie size will increase linearly with the times of account being updated, as a result, it may take extremely large space to store all the old data nodes. 

As Ethereum may need it for states reverting as it's PoW based consensus which doesn't provide instant finality, the old data is needed (referring to https://blog.ethereum.org/2015/06/26/state-tree-pruning/). But Zilliqa actually provides finality in every epoch, so the old nodes are no longer useful in our situation.

The solution is repopulating the state trie database regularly, such as "per 10 DS epoch". We will detach a thread when calling `MoveUpdatesToDisk` so that it won't block the normal consensus process (normally the writing could be long with enormous accounts in the trie). Once we repopulate the state trie, the old data will be removed thus save space.

However, the repopulating may require extra cpu usage during the PoW phase, if user is willing to sacrifice disk space for spending more power on PoW, it's configurable to disable it by modifying the constants `ENABLE_REPOPULATE`.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNode/FallbackBlockProcessing.cpp
```diff
@@ -290,7 +290,24 @@ bool Node::ProcessFallbackBlock(const bytes& message, unsigned int cur_offset,
                                      *m_mediator.m_DSCommittee,
                                      m_mediator.m_ds->m_shards);
     }
-    StoreState();
+
+    auto writeStateToDisk = [this]() mutable -> void {
```
ShengguangXiao: writeStateToDisk has been defined 4 times, can be optimized?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -235,6 +238,90 @@ bool AccountStore::MoveUpdatesToDisk() {
...
 
+bool AccountStore::RepopulateStateTrie() {
+  LOG_MARKER();
+
+  unsigned int counter = 0;
+  bool batched_once = false;
+
+  vector<pair<Address, bytes>> entries;
+
+  for (const auto& i : m_state) {
```
ShengguangXiao: Is m_state thread safe? Now RepopulateStateTrie is running in a seprate thread

kaikawaliu: Yes, the mutex was declared in the function where this was called

src/libNode/FallbackBlockProcessing.cpp
kaikawaliu: There are some slight difference among them, so I think better to keep it customized and keep on stack. In the later enhancement I may change this part totally.

constants.xml
```diff
@@ -228,6 +228,10 @@
         <PACKET_EPOCH_LATE_ALLOW>1</PACKET_EPOCH_LATE_ALLOW>
         <PACKET_BYTESIZE_LIMIT>1572864</PACKET_BYTESIZE_LIMIT>
         <SMALL_TXN_SIZE>1024</SMALL_TXN_SIZE>
+        <ACCOUNT_IO_BATCH_SIZE>2000000</ACCOUNT_IO_BATCH_SIZE>
+        <ENABLE_REPOPULATE>true</ENABLE_REPOPULATE>
+        <REPOPULATE_STATE_IN_DS>5</REPOPULATE_STATE_IN_DS>
```
KaustubhShamshery: Put comment indicating REPOPULATE_STATE_IN_DS cannot be greater than REPOPULATE_STATE_PER_N_D

src/libData/AccountData/AccountStore.cpp
```diff
@@ -235,6 +238,90 @@ bool AccountStore::MoveUpdatesToDisk() {
...
+        batched_once = true;
+      }
+    }
+
+    Address address(i.first);
+
+    if (!batched_once) {
+      auto iter = this->m_addressToAccount->find(address);
+
+      if (iter != this->m_addressToAccount->end()) {
```
ckyang: Can simplify like this, to reduce the temporarily variable (iter) usage.

       if (this->m_addressToAccount->find(address) != this->m_addressToAccount->end()) {

src/libData/AccountData/AccountStore.cpp
```diff
@@ -288,7 +375,9 @@ bool AccountStore::UpdateAccountsTemp(const uint64_t& blockNum,
                                       TransactionReceipt& receipt) {
   // LOG_MARKER();
 
-  lock_guard<mutex> g(m_mutexDelta);
+  unique_lock<shared_timed_mutex> g(m_mutexPrimary, defer_lock);
+  unique_lock<mutex> g2(m_mutexDelta, defer_lock);
```
ckyang: Why perform 2 different kinds of mutex here?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -721,6 +716,13 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
     return false;
   }
 
+  auto resumeBlackList = []() mutable -> void {
+    this_thread::sleep_for(chrono::seconds(RESUME_BLACKLIST_DELAY_IN_SECONDS));
```
ckyang: Why need a constant-defined delay (In general it's sort of experience value) here to resume blacklist? Is it possible to enable blacklist back in more accurate place?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -235,6 +238,90 @@ bool AccountStore::MoveUpdatesToDisk() {
   return true;
 }
 
+bool AccountStore::RepopulateStateTrie() {
+  LOG_MARKER();
+
+  unsigned int counter = 0;
+  bool batched_once = false;
+
+  vector<pair<Address, bytes>> entries;
```
ansnunez: `entries` seems unused?

src/libLookup/Lookup.cpp
```diff
@@ -1880,11 +1880,6 @@ bool Lookup::ProcessSetStateDeltaFromSeed(const bytes& message,
     return false;
   }
 
-  if (!AccountStore::GetInstance().MoveUpdatesToDisk()) {
```
ansnunez: Can you explain the removal here?

src/libPersistence/BlockStorage.cpp
```diff
@@ -204,6 +204,56 @@ bool BlockStorage::GetRangeMicroBlocks(const uint64_t lowEpochNum,
...
+  lock_guard<mutex> g(m_mutexTempState);
+
+  if (iter == nullptr) {
+    iter = m_tempStateDB->GetDB()->NewIterator(leveldb::ReadOptions());
+    iter->SeekToFirst();
+  }
+
+  unsigned int counter = 0;
+
+  for (; iter->Valid() && counter < ACCOUNT_IO_BATCH_SIZE;
```
ansnunez: For consistency with `RepopulateStateTrie`, maybe change to `counter <=`?

src/libNode/FinalBlockProcessing.cpp
ansnunez: Actually it is just existing thread that has been moved up.  Any reason why it is moved earlier now?

constants.xml
kaikawaliu: Thanks, good point

src/libData/AccountData/AccountStore.cpp
kaikawaliu: Got it

src/libData/AccountData/AccountStore.cpp
kaikawaliu: Because the m_mutexPrimary protects the state trie, the m_mutexDelta protects the AccountStoreTemp for state delta

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: Because the `MoveUpdatesToDisk` may take longer time than before. We can detach this function which has a sleep to finish it earlier

src/libLookup/Lookup.cpp
kaikawaliu: Shouldn't store the states when received the delta while syncing, this was done unintentionally before.

src/libPersistence/BlockStorage.cpp
kaikawaliu: `RepopulateStateTrie` is using `>=`, so in the case of consistency it's better to use `<`

constants.xml
KaustubhShamshery: @kaikawaliu have u put the comment?

PR #1505, opened 2019-03-04
[PoW] fix invalid signature for REMOTE_MINE

deepgully:
## Description
when network upgrade to v4.2.0, the proxy rejects all pow requests from node because of "invalid signature".

The root cause is node send request with "timeWindow" but sign with "POW_WINDOW_IN_SECONDS".
https://github.com/Zilliqa/Zilliqa/blob/3e3339f0944218f1e6fe182ec730c7aeb5b192c5/src/libPOW/pow.cpp#L375-L384

A hotfix added in proxy code already, verify signature with fixed POW_WINDOW_IN_SECONDS.
https://github.com/DurianStallSingapore/Zilliqa-Mining-Proxy/commit/3d6534a950325e6488671ac684ac2cbe489b031a

### Implementation
- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -378,8 +378,7 @@ bool POW::SendWorkToProxy(const PairOfKey& pairOfKey, uint64_t blockNum,
           timeWindow);
   jsonValue[4] = "0x" + strPoWTime;
   auto powTimeBytes =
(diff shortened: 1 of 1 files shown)


deepgully:
@ShengguangXiao 
could you help to take a look?

PR #1508, opened 2019-03-05
Adjustment to shard guards

bb111189:
## Description

This PR 
- allows the adjustment of % of shard guards in the network via the constants file
- reduces the number of shard guards from 2/3 to 1/3
- minor refinement of logging
- adds new function to parse double in constants file

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud and upgrading test from `v4.2.0` to `adfe321903b5e15397beadbce5ed37a09f77490b`


constants_local.xml
```diff
@@ -295,8 +296,5 @@
         <SHARDPUBKEY>02F650040BC0F3158B5D1A870EFC0ABE84FF02A4021A222FC49F9AD070ACFA2DA7</SHARDPUBKEY>
         <SHARDPUBKEY>030DFD4A0CFD68016DF6EFC4A6BF0B67BA42ABEAF8D9AD65C25E97B9DA90CA4DB1</SHARDPUBKEY>
         <SHARDPUBKEY>03129EAEF8A136355FB1485941A593B4BEC4DFC5504D1114138A6D92332005DC59</SHARDPUBKEY>
-        <SHARDPUBKEY>037D49C420B04B9862BC1F0660544FF27F1D81EA0E5E7C161F1647FBF239F8780E</SHARDPUBKEY>
```
ansnunez: Just want to confirm if we need to remove these keys

bb111189: Yes. It is or local test. This is so that the local node is no longer a shard guard.

PR #1513, opened 2019-03-06
Adjust and refactor of code for shard operation

bb111189:
## Description

This PR implements a best effort approach for shard leader selection. It will randomly re-pick a candidate as a shard leader if the current choice of shard leader does not match the condition for selection. This number of iteration for this process is determined by a newly added constant `SHARD_LEADER_SELECT_TOL`

## Backward Compatibility

- [x] Undecided
### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] medium-scale cloud upgrading test
- [x] medium-scale cloud upgrading test with new node reduction and new node joining
- [x] (RE-TEST) medium-scale cloud upgrading test with new node reduction and new node joining


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -335,7 +335,8 @@ void Node::UpdateStateForNextConsensusRound() {
       m_consensusLeaderID =
           lastBlockHash % Guard::GetInstance().GetNumOfDSGuard();
     } else {
-      m_consensusLeaderID = lastBlockHash % m_myShardMembers->size();
+      m_consensusLeaderID =
+          CalculateShardLeader(lastBlockHash, m_myShardMembers->size());
```
kaikawaliu: Shall we also apply this function to other places such as:
https://github.com/Zilliqa/Zilliqa/blob/4c3213b75cc1a5998a4cdde69c9b59b73ae4fa8b/src/libNode/DSBlockProcessing.cpp#L266
https://github.com/Zilliqa/Zilliqa/blob/2de1bf88ac5b9035e4e62eaa7fa0d305e8c1b138/src/libLookup/Lookup.cpp#L3364

bb111189: For DSBlockProcessing, that code is already applied right?

KaustubhShamshery: I can safely say u have already changed there.

kaikawaliu:
It is a break change from my opinion

PR #1515, opened 2019-03-08
Enable server on non-lookup nodes. 

KaustubhShamshery:
## Description
1. Enable `GetCurrentMiniEpoch` and `GetCurrentDSEpoch` on normal nodes
2. add a script for miners to easily query such data
3. add a api for querying the type of node
4. add API to query for state of the node
5. upgrade to version v0.7.0

## Backward Compatibility

## Review Suggestion
NOTE: This adds a TCP server in the shard nodes. Hence, zilliqa-js library wont be compatible for it.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


Snuff:
Are you going to include this feature into 4.3.0 release, @KaustubhShamshery?

nnamon:
Would it be possible to make the SERVER_PORT constant be a configurable option like the GETWORK_SERVER_PORT?

https://github.com/Zilliqa/Zilliqa/blob/master/src/common/Constants.h#L66

src/libServer/Server.cpp
```diff
@@ -72,7 +72,13 @@ Server::~Server(){
     // destructor
 };
 
-string Server::GetNetworkId() { return to_string(CHAIN_ID); }
+string Server::GetNetworkId() {
+  if (!LOOKUP_NODE_MODE) {
+    throw JsonRpcException(RPC_INVALID_REQUEST, "Sent to a non-lookup");
+    ;
```
Gnnng: consider removing this?
```suggestion
```

nnamon:
> @KaustubhShamshery Can you check if there's a way to implement listening on localhost only for non-lookup nodes? I'm afraid that the current way of `StartListening` will possibly accept all the incoming traffic regardless where it originates from. We would like to provide an option to let the user select if the server will take only local traffic or public traffic.

To generalise this further, could the option allow the user to select exactly which IP to bind too as well? Would help with segregating access to the API by having it bind to a virtual network interface in Docker  as opposed to the host-network interface that is required to be used by the node.

scripts/miner_info.py
```diff
@@ -0,0 +1,90 @@
...
+	parser.add_argument("--address","-a",help="URL for querying, default: localhost", default="127.0.0.1")
+	parser.add_argument("option",help="input option for the query", choices=options)
+	args = parser.parse_args()
+	return args
+
+def make_options_dictionary(options_dict):
+	options_dict["epoch"] = "GetCurrentMiniEpoch"
+	options_dict["dsepoch"] = "GetCurrentDSEpoch"
+	options_dict["type"] = "GetNodeType"
+	options_dict["ds"] = "GetDSCommitee"
```
ansnunez: Method typo `GetDSCommitee`

PR #1518, opened 2019-03-10
Refactor and fix calculate shard leader

bb111189:
## Description

Refactor `CalculateShardLeader` into `CalculateShardLeaderFromDequeOfNode` and `CalculateShardLeaderFromShard`
- `CalculateShardLeaderFromDequeOfNode` accepts `DequeOfNode` which is `vector<pair<PubKey, Peer>>`
- `CalculateShardLeaderFromShard` accepts `Shard` which is `vector<std::tuple<PubKey, Peer, uint16_t>>`
- Both function do not works on member variables directly.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] medium-scale cloud test
- [x] all vc cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -995,7 +995,8 @@ uint32_t Node::CalculateShardLeaderFromShard(uint16_t lastBlockHash,
 
     unsigned int iterationCount = 0;
     while (!Guard::GetInstance().IsNodeInShardGuardList(
-               std::get<0>(shardMembers.at(consensusLeaderIndex))) &&
+               std::get<SHARD_NODE_PUBKEY>(
+                   shardMembers.at(consensusLeaderIndex))) &&
            (iterationCount < SHARD_LEADER_SELECT_TOL)) {
       LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
                 "consensusLeaderIndex " << consensusLeaderIndex
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -20,6 +20,7 @@
 #include <chrono>
 #include <functional>
 #include <thread>
+#include <tuple>
 
 #include <boost/filesystem.hpp>
 #include <boost/property_tree/ptree.hpp>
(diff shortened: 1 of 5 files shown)


src/libNode/Node.cpp
```diff
@@ -958,15 +959,43 @@ void Node::StartSynchronization() {
...
+uint32_t Node::CalculateShardLeaderFromShard(uint16_t lastBlockHash,
+                                             uint32_t sizeOfShard,
+                                             const Shard& shardMembers) {
+  LOG_MARKER();
+  if (GUARD_MODE) {
+    uint32_t consensusLeaderIndex = lastBlockHash % sizeOfShard;
+
+    unsigned int iterationCount = 0;
+    while (!Guard::GetInstance().IsNodeInShardGuardList(
+               std::get<0>(shardMembers.at(consensusLeaderIndex))) &&
```
ansnunez: Can use `SHARD_NODE_PUBKEY` instead of `0`

src/libNode/Node.h
```diff
@@ -563,7 +563,12 @@ class Node : public Executable {
   bool IsShardNode(const PubKey& pubKey);
   bool IsShardNode(const Peer& peerInfo);
 
-  uint32_t CalculateShardLeader(uint16_t lastBlockHash, uint32_t sizeOfShard);
+  uint32_t CalculateShardLeaderFromDequeOfNode(uint16_t lastBlockHash,
+                                               uint32_t sizeOfShard,
+                                               const DequeOfNode& shardMembers);
+  uint32_t CalculateShardLeaderFromShard(uint16_t lastBlockHash,
```
KaustubhShamshery: if it is not too much effort, can make a template container and merge the two functions

bb111189: Tried that approach but it can be tricky due to the differences in data structure
```
vector<pair<PubKey, Peer>>
vector<std::tuple<PubKey, Peer, uint16_t>>
```

KaustubhShamshery: I think `std::get` would work on both.

PR #1522, opened 2019-03-12
Force multicast to shard guards after upgrading

kaikawaliu:
## Description

Once we call recovery, turn the flag `m_forceMulticast` to true so that before the next DS epoch consensus, the finalblock will be sent to shards via multicast. And if in guard mode, the receiver will be reloaded by the shard guards so that in the new PoW phase those nodes can keep in sync.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -529,6 +529,8 @@ void DirectoryService::ProcessDSBlockConsensusWhenDone() {
 
   m_mediator.UpdateDSBlockRand();
 
+  m_forceMulticast = false;
+
   // Now we can update the sharding structure and transaction sharing
   // assignments
   if (m_mode == BACKUP_DS) {
@@ -577,7 +579,7 @@ void DirectoryService::ProcessDSBlockConsensusWhenDone() {
         *m_pendingDSBlock, *(m_mediator.m_DSCommittee), m_shards, {},
         m_mediator.m_lookup->GetLookupNodes(),
         m_mediator.m_txBlockChain.GetLastBlock().GetBlockHash(),
-        m_consensusMyID, composeDSBlockMessageForSender,
+        m_consensusMyID, composeDSBlockMessageForSender, false,
         sendDSBlockToLookupNodesAndNewDSMembers, sendDSBlockToShardNodes);
   }
 
--- a/src/libDirectoryService/DirectoryService.h
+++ b/src/libDirectoryService/DirectoryService.h
@@ -525,7 +525,7 @@ class DirectoryService : public Executable {
   bool m_doRejoinAtFinalConsensus = false;
 
   /// Force multicast when sending block to shard
-  bool m_forceMulticast = false;
+  std::atomic<bool> m_forceMulticast;
 
   /// Constructor. Requires mediator reference to access Node and other global
   /// members.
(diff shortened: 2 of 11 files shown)


src/libDirectoryService/DirectoryService.h
```diff
@@ -522,6 +524,9 @@ class DirectoryService : public Executable {
   bool m_doRejoinAtDSConsensus = false;
   bool m_doRejoinAtFinalConsensus = false;
 
+  /// Force multicast when sending block to shard
+  bool m_forceMulticast = false;
```
bb111189: atomic<bool>

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -180,11 +181,17 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone() {
...
 
+  DequeOfShard t_shards;
+  if (m_forceMulticast && GUARD_MODE) {
+    ReloadGuardedShards(t_shards);
+  }
+
   DataSender::GetInstance().SendDataToOthers(
-      *m_finalBlock, *m_mediator.m_DSCommittee, m_shards, t_microBlocks,
+      *m_finalBlock, *m_mediator.m_DSCommittee,
+      (m_forceMulticast && GUARD_MODE) ? t_shards : m_shards, t_microBlocks,
```
ckyang: Maybe can be changed like following, to avoid multiple comparison

   DequeOfShard t_shards;
  if (m_forceMulticast && GUARD_MODE) {
    ReloadGuardedShards(t_shards);
  }
  else {
    t_shards = m_shards;
  }

   DataSender::GetInstance().SendDataToOthers(
      *m_finalBlock, *m_mediator.m_DSCommittee, m_shards, t_microBlocks,

src/libDirectoryService/ViewChangePostProcessing.cpp
```diff
@@ -322,11 +322,18 @@ void DirectoryService::ProcessViewChangeConsensusWhenDone() {
...
 
+    DequeOfShard t_shards;
+    if (m_forceMulticast && GUARD_MODE) {
+      ReloadGuardedShards(t_shards);
+    }
+
     DataSender::GetInstance().SendDataToOthers(
-        *m_pendingVCBlock, tmpDSCommittee, m_shards, t_microBlocks,
+        *m_pendingVCBlock, tmpDSCommittee,
+        (m_forceMulticast && GUARD_MODE) ? t_shards : m_shards, t_microBlocks,
```
ckyang: Maybe can be changed like following, to avoid multiple comparison

   DequeOfShard t_shards;
  if (m_forceMulticast && GUARD_MODE) {
    ReloadGuardedShards(t_shards);
  }
  else {
    t_shards = m_shards;
  }

   DataSender::GetInstance().SendDataToOthers(
      *m_finalBlock, *m_mediator.m_DSCommittee, m_shards, t_microBlocks,

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -577,7 +579,7 @@ void DirectoryService::ProcessDSBlockConsensusWhenDone() {
         *m_pendingDSBlock, *(m_mediator.m_DSCommittee), m_shards, {},
         m_mediator.m_lookup->GetLookupNodes(),
         m_mediator.m_txBlockChain.GetLastBlock().GetBlockHash(),
-        m_consensusMyID, composeDSBlockMessageForSender,
+        m_consensusMyID, composeDSBlockMessageForSender, false,
```
ckyang: How about use m_forceMulticast directly here?

src/libDirectoryService/DirectoryService.h
kaikawaliu: updated

PR #1528, opened 2019-03-13
Remove old upgrading mechanism

ckyang:
## Description

Remove old upgrading mechanism because it never be used anymore.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -128,9 +128,7 @@ bool Node::Install(const SyncType syncType, const bool toRetrieveHistory) {
   }
 
   if (toRetrieveHistory) {
-    bool wakeupForUpgrade = false;
-
-    if (!StartRetrieveHistory(syncType, wakeupForUpgrade)) {
+    if (!StartRetrieveHistory(syncType)) {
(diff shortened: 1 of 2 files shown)


src/libNode/Node.cpp
```diff
@@ -522,23 +504,22 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
...
+      m_retriever->RetrieveTxBlocks(RECOVERY_TRIM_INCOMPLETED_BLOCK &&
+                                    SyncType::RECOVERY_ALL_SYNC == syncType);
 
   if (!tx_result) {
     return false;
   }
 
   /// Retrieve lacked Tx blocks from lookup nodes
   if (SyncType::NO_SYNC == m_mediator.m_lookup->GetSyncType() &&
-      !(LOOKUP_NODE_MODE && wakeupForUpgrade) &&
```
Gnnng: Just to confirm with you: is removing `LOOKUP_NODE_MODE` the desired change?

ckyang: Yes, removing LOOKUP_NODE_MODE here is desired to change

PR #1531, opened 2019-03-15
IncrementalDB for new node and seed node joining

sandipbhoir:
## Description:
**Pre-launch of zilliqa process:** (Not part of this PR)
 - downloadIncrDB.py will fetch the following from S3:
    - Entire persistence from S3 
    - StateDeltas from S3 ( All state deltas from last 10 DS epoch if available )

**This PR covers following:**

**Launch of zilliqa process:**
 - New node and seed node will be started with --recovery option.
 - This will enable retrieve from existing local persistence downloaded pre-launch of zilliqa where states    will be regenerated for last 10 DS epoch (or less depending on available state-deltas ).
 - After this, both will start syncronization from lookup. However, instead of asking for state from lookup, it will ask for state-deltas to generate latest state.
  

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


ckyang:
For incremental DB things, it's better to verify like that:

1. Launch a small-scale (ex. 33 nodes) testnet with current master
2. Use this PR + https://github.com/Zilliqa/testnet/pull/426 to upgrade (1) to a separated testnet
    (doc: https://docs.google.com/spreadsheets/d/1J5wVeY3o4q7SIu38HbvUdHkdV3FAyCMWgSNg8wnuOjs/edit#gid=2001092950)
3. See if anything works fine in the separated testnet

src/libLookup/Lookup.cpp
```diff
@@ -1825,6 +1974,21 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
...
+  if (m_syncType == SyncType::NEW_SYNC ||
+      m_syncType == SyncType::NEW_LOOKUP_SYNC) {  // only for new node joining
+    // Get the state-delta for all txBlocks from random lookup nodes
+    GetStateDeltasFromSeedNodes(lowBlockNum, highBlockNum);
+
+    std::unique_lock<std::mutex> cv_lk(m_mutexSetStateDeltaFromSeed);
+    if (cv_setStateDeltasFromSeed.wait_for(
+            cv_lk, std::chrono::seconds(GETSTATEDELTAS_TIMEOUT_IN_SECONDS)) ==
+        std::cv_status::timeout) {
+      LOG_GENERAL(WARNING, "Didn't receive statedeltas!");
```
kaikawaliu: If the node missed the first time for fetching delta, will the querying in the next try includes the deltas that it failed to fetch previously?

src/libLookup/Lookup.cpp
```diff
@@ -1901,6 +2085,69 @@ bool Lookup::ProcessSetStateDeltaFromSeed(const bytes& message,
...
+
+  bool isAnyTxBlkFromVacuousEpoch = false;
+  for (auto i = lowBlockNum; i <= highBlockNum; i++) {
+    if ((i + 1) % NUM_FINAL_BLOCK_PER_POW == 0) {
+      isAnyTxBlkFromVacuousEpoch = true;
+      break;
+    }
+  }
+
+  if (isAnyTxBlkFromVacuousEpoch &&
```
kaikawaliu: This judge is a bit weird. E.g. a node trying to sync at epoch 21 (10 tx epoch / ds), the whole states it fetched is at epoch 10 (very likely to happen as the states uploading may not finish). Then the state delta is from epoch 11 ~ 21. Which includes an vacuous epoch. Then in this case, we will result in calling `MoveUpdatesToDisk` for the state in epoch 21, this is what we don't want.
So the change I suggest is to call `MoveUpdatesToDisk` only when the highBlockNum being the vacuous block.

src/libMessage/Messenger.cpp
```diff
@@ -5432,6 +5518,45 @@ bool Messenger::GetLookupSetStateDeltaFromSeed(const bytes& src,
...
+
+  result.ParseFromArray(src.data() + offset, src.size() - offset);
+
+  if (!result.IsInitialized()) {
+    LOG_GENERAL(WARNING, "LookupSetStateDeltasFromSeed initialization failed");
+    return false;
+  }
+
+  lowBlockNum = result.data().lowblocknum();
+  highBlockNum = result.data().highblocknum();
```
kaikawaliu: It will be better to have a check on the size of delta and the (highBlockNum - lowBlockNum) is needed

src/libLookup/Lookup.cpp
sandipbhoir: Agree. Will try to keep track of missing statedeltas and Try fetching them next time.

src/libLookup/Lookup.cpp
sandipbhoir: Agree.

src/libLookup/Lookup.cpp
```diff
@@ -1901,6 +2085,69 @@ bool Lookup::ProcessSetStateDeltaFromSeed(const bytes& message,
...
+                  << lookupPubKey << " is not in my lookup node list.");
+    return false;
+  }
+
+  LOG_EPOCH(INFO, m_mediator.m_currentEpochNum,
+            "ProcessSetStateDeltasFromSeed sent by "
+                << from << " for blocks: " << lowBlockNum << " to "
+                << highBlockNum);
+
+  for (const auto& delta : stateDeltas) {
```
kaikawaliu: I think we need to verify the state delta against the state delta hash in the corresponding txblock. One way to locate the txBlock is to incrementally using the count in the vector and the lowBlockNum. Another way can be modifying the message structure to indicate the block number of each delta, which may require extra code change however but it's better for future maintenance.

src/libNode/Node.cpp
```diff
@@ -134,6 +134,11 @@ bool Node::Install(const SyncType syncType, const bool toRetrieveHistory) {
       return false;
     }
 
+    if (SyncType::NEW_SYNC == syncType ||
+        SyncType::NEW_LOOKUP_SYNC == syncType) {
+      return true;
+    }
```
kaikawaliu: If we don't skip what will happen?

src/libNode/Node.cpp
```diff
@@ -1855,26 +1863,9 @@ bool Node::ProcessDSGuardNetworkInfoUpdate(const bytes& message,
   LOG_GENERAL(INFO, "Received from lookup " << from);
 
   {
+    // Process and update ds committee network info
     lock_guard<mutex> lock(m_mediator.m_mutexDSCommittee);
     for (const auto& dsguardupdate : vecOfDSGuardUpdateStruct) {
-      // Remove old ds guard IP info from blacklist exclude list
```
kaikawaliu: Are these removals related to this PR?

src/libPersistence/Retriever.cpp
```diff
@@ -50,26 +47,105 @@ bool Retriever::RetrieveTxBlocks(bool trimIncompletedBlocks) {
...
-    m_mediator.m_node->AddBlock(*block);
-  }
-
   /// Retrieve final block state delta from last DS epoch to
-  /// current TX epoch
+  /// current TX epoch and buffer the statedelta for each.
   for (const auto& block : blocks) {
     if (block->GetHeader().GetBlockNum() >= lastBlockNum + 1 - extra_txblocks) {
       bytes stateDelta;
       BlockStorage::GetBlockStorage().GetStateDelta(
```
kaikawaliu: Just notice that, maybe we should handle the boolean return of this function.

src/libPersistence/Retriever.cpp
```diff
@@ -50,26 +47,105 @@ bool Retriever::RetrieveTxBlocks(bool trimIncompletedBlocks) {
...
+  }
+
+  if ((lastBlockNum - extra_txblocks + 1) %
+          (INCRDB_DSNUMS_WITH_STATEDELTAS * NUM_FINAL_BLOCK_PER_POW) ==
+      0) {
+    // we must have latest state currently. so need not recreate states
+    LOG_GENERAL(INFO, "Current state is up-to-date until txblk :"
+                          << lastBlockNum - extra_txblocks);
+  } else {
+    // create states from last INCRDB_DSNUMS_WITH_STATEDELTAS *
```
kaikawaliu: Hmmm sorry to say but actually this part of code is not needed. You can just keep DeserializeDelta from the last state checkpoint until the latest epoch.

src/libNode/Node.cpp
sandipbhoir: Haven't tried.. Just that we don't need the code below for new nodes joining.

src/libLookup/Lookup.cpp
sandipbhoir: Fixed. Handled it differently.

src/libLookup/Lookup.cpp
sandipbhoir: Since, we check on the count of deltas received against the request low/hignum and also verify the sender (lookup), additional check on state-delta hash can be considered for future improvements. If you agree.

src/libMessage/Messenger.cpp
sandipbhoir: Fixed

src/libNode/Node.cpp
sandipbhoir: Fixed

src/libPersistence/Retriever.cpp
sandipbhoir: Fixed. If any of extra statedelta missing in BlockStorage, simple trim the extra tx blocks. we can then fetch the rest  from lookup.

src/libLookup/Lookup.cpp
kaikawaliu: Yes, we can do that enhancement in the future

src/libNode/Node.cpp
kaikawaliu: Didn't find the code change. Do we really need to remove this part in this PR?

kaikawaliu:
We can try the testing again with state trie repopulating enabled. To make it happen in every DS epoch. You can set `REPOPULATE_STATE_IN_DS` to 0, and `REPOPULATE_STATE_PER_N_DS` to 1

src/libLookup/Lookup.cpp
```diff
@@ -1812,9 +1898,72 @@ bool Lookup::ProcessSetTxBlockFromSeed(const bytes& message,
...
+    return;
+  }
+
+  m_mediator.m_lookup->SendMessageToRandomSeedNode(getpowsubmission_message);
+}
+
 void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
   LOG_GENERAL(INFO, "[TxBlockVerif]"
                         << "Success");
+  uint64_t lowBlockNum = txBlocks.front().GetHeader().GetBlockNum();
```
KaustubhShamshery: unused?

src/libLookup/Lookup.cpp
```diff
@@ -2142,7 +2142,8 @@ bool Lookup::ProcessSetStateDeltasFromSeed(const bytes& message,
         return false;
       }
       BlockStorage::GetBlockStorage().PutStateDelta(txBlkNum, delta);
-      if (txBlkNum + 1 % NUM_FINAL_BLOCK_PER_POW == 0) {
+      if (((txBlkNum + 1) % NUM_FINAL_BLOCK_PER_POW == 0) &&
+          (txBlkNum + NUM_FINAL_BLOCK_PER_POW < highBlockNum)) {
         if (!AccountStore::GetInstance().MoveUpdatesToDisk()) {
```
kaikawaliu: No param passing to MoveUpdatesToDisk means don't do repopulate. Aren't you going to repopulate at this point?

sandipbhoir: You are correct. This change is already made in PR#1540.

PR #1537, opened 2019-03-18
Fix DS backup cannot find pow in sorted solutions

ShengguangXiao:
## Description

Fix DS backup cannot find pow in sorted solutions

When there is a lot of pow submissions, backup also choose the pows from high priority nodes and then do pow sorting and verifying.

https://github.com/Zilliqa/Issues/issues/464

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/PoWProcessing.cpp
+++ b/src/libDirectoryService/PoWProcessing.cpp
@@ -314,8 +314,17 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
       return false;
     }
   } else {
-    if (difficultyLevel != expectedDSDiff && difficultyLevel != expectedDiff &&
-        difficultyLevel != expectedShardGuardDiff) {
+    bool difficultyCorrect = true;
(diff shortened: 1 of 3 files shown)


src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -314,8 +314,17 @@ bool DirectoryService::ProcessPoWSubmissionFromPacket(
       return false;
     }
   } else {
-    if (difficultyLevel != expectedDSDiff && difficultyLevel != expectedDiff &&
-        difficultyLevel != expectedShardGuardDiff) {
+    bool difficultyCorrect = true;
+    if (Guard::GetInstance().IsNodeInShardGuardList(submitterPubKey)) {
```
ansnunez: Can change to `if (GUARD_MODE && IsNodeInShardGuardList)` so that we don't get flooded with `Not in Guard mode` messages.

ShengguangXiao: On top already checked GUARD_MODE

PR #1540, opened 2019-03-18
Rejoining as Normal/DS from S3

sandipbhoir:
## Description

This PR will cover:
- Rejoining Normal/DS from S3

Also, additionally, new node/ new lookup node will keep try to RetrieveFromHistory until succeed.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -1157,9 +1157,7 @@ void Lookup::RetrieveTxBlocks(vector<TxBlock>& txBlocks, uint64_t& lowBlockNum,
     // give all the blocks till now in blockchain
     lowBlockNum = 1;
 
-  } else if (lowBlockNum <= m_mediator.m_dsBlockChain.GetLastBlock()
-                                .GetHeader()
-                                .GetEpochNum()) {
+  } else if (lowBlockNum == 1) {
```
kaikawaliu: The condition here is to avoid the lookup return too many txblocks here. e.g. the lowBlockNum == 2, it will still be quite large. Shall we have a limitation on the maximum blocks can be fetched such as `lowBlockNum <= m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetEpochNum() - MAXIMUM_NUM_TX_BLOCK_CAN_BE_FETCHED`? So that can fetch more blocks but also have a control on it

sandipbhoir: Yes, this was intentionally changed to == 1.
Because when seed nodes ask for say txblknum range 198-0 (current txepoch is 202), it was returning txblknonly onward 200 thereby missing 198-199.
However, with == 1 change, the normal nodes which always happen to ask for range 1-0 and will still return TxBlocks from last dsepoch. 
This hack is to avoid the breaking change otherwise required to identify if request is from seed node or normal node to handle more elegantly.

sandipbhoir: Also, rejoining of normal nodes will ask for exact range of TxBlocks from lookup. Because we need to statedelta for exact those txblks to calculate state.

KaustubhShamshery: Why does a normal node need such old blocks namely 198, 199. It does not need it right, that was the logic for removing it

sandipbhoir: To get the statedelta for each such txblks and thereby have correct uptodate state. They don’t get entire state from lookup anymore.

KaustubhShamshery: OK i get the point, however is there still a way to upper bound this, like 10 ds ?

sandipbhoir: We can have but it’s anyways not going to help if we limit the txblks. The asking node will eventually have stateroot mismatch.
With download from S3 incrb, new node/seed node won’t have so many missing txblks.. won’t be lagging more than few txblks. Definitely less than 100 txblks :)

But yes, putting limit to last 10 DS block is not harm aswell. Will make a change, if you agree?

KaustubhShamshery: yeah, i agree no harm as well

sandipbhoir: Fixed

ckyang:
Have done following tests and both worked fine.
The branch of `feature/new-node-join-incrdb` looks good for me.

1. Launch a testnet with `v4.3.1`, and upgrading on separated testnet using `e902278`
2. Launch a testnet with `e902278`, and upgrading on separated testnet using `e902278`

PR #1543, opened 2019-03-19
[POW] Divided difficulty adjustment

deepgully:
## Description
The current difficulty adjustment is log2 based, the minimum step is *2 or 1/2.
The difficulty increment/decrement is not corresponding to the real hashrate change, sometime the network dropped a shard because the difficulty was doubled but the real hashrate was not.

This PR implemented a new difficulty adjustment method based on current data structure.
We divide one log2 difficulty level to N_DEVIDED(default is 8), that means we can adjust difficulty by  12.5% of current step. With the smaller step size, we can adjust difficulty more accurately based on 
 the percentage of extra pow submissions.

With current uint8 difficulty value and N_DEVIDED 8, the max hashrate we can reach is 1.02 Eh.

There is a demo:
https://github.com/deepgully/jupyter/blob/master/Zilliqa/difficulty.ipynb

example:
Current method: Diff 32->33->34 ...-> 50  Hashrate 4.29 Gh -> 8.59 Gh -> 17.18 Gh ...-> 1.13 Ph
Divided method: Diff 32->33->34 ...-> 50 Hashrate 4.29 Gh -> 4.58 Gh -> 4.91 Gh ...-> 19.63 Gh

the blue line is current method, the purple one is proposed.
image

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


deepgully:
@ShengguangXiao

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -1057,8 +1061,10 @@ uint8_t DirectoryService::CalculateNewDifficultyCore(uint8_t currentDifficulty,
     adjustment = -MAX_ADJUST_STEP;
   }
 
-  return std::max((uint8_t)(adjustment + currentDifficulty),
-                  (uint8_t)(minDifficulty));
+  uint16_t newDifficulty = (uint16_t)adjustment + (uint16_t)currentDifficulty;
```
bb111189: (uint16_t)adjustment shouldm't be uint?

sandipbhoir: Also, adjustment can be -ve `adjustment = -MAX_ADJUST_STEP;`
 `(uint16_t)adjustment` can give unexpected result.

ansnunez: Any issue with just keeping the original code for the return value?

ShengguangXiao: The original code cannot check maximum value, i will handle this case

PR #1549, opened 2019-03-21
Add option to run test with pre-generated accounts  

kaikawaliu:
## Description

With this pr, we can load dummy accounts into the network to enlarge the size of accounts trie. And in each epoch will choose part of the accounts to be updated in the state delta. Such we can test the performance of Account Store system in a more realistic scenario. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -657,6 +665,22 @@ void Node::ProcessTransactionWhenShardBackup() {
   }
 }
 
+void Node::UpdateBalanceForPreGeneratedAccounts() {
+  LOG_MARKER();
+  int counter = 0;
+  for (unsigned int i = 0; i < m_populated_addresses.size(); i++) {
+    if ((i % (m_mediator.m_ds->m_shards.size() + 1) == m_myshardId) &&
+        (i % NUM_FINAL_BLOCK_PER_POW ==
```
ansnunez: Should we exclude vacuous epoch?

src/libNode/Node.h
```diff
@@ -84,6 +84,9 @@ class Node : public Executable {
   // Sharding information
   std::atomic<uint32_t> m_numShards;
 
+  // pre-generated addresses
+  std::vector<Address> m_populated_addresses;
```
ansnunez: For coding consistency, can you rename to `m_populatedAddresses`?

src/libNode/Node.cpp
```diff
@@ -66,7 +66,26 @@ const unsigned int MIN_CHILD_CLUSTER_SIZE = 2;
...
+    while (getline(keys_file, line)) {
+      vector<string> key_pair;  // pub/priv
+      boost::algorithm::split(key_pair, line, boost::algorithm::is_any_of(" "));
+      Address t_addr = Account::GetAddressFromPublicKey(
+          PubKey::GetPubKeyFromString(key_pair[0]));
+      AccountStore::GetInstance().AddAccount(t_addr, {0, 0});
+      m_populated_addresses.emplace_back(t_addr);
+    }
+  } catch (std::exception& e) {
+    std::cerr << "Problem occured when processing keys on line: "
```
ansnunez: Should we use `LOG_GENERAL` instead, or do both `cerr` and `LOG_GENERAL`?

PR #1552, opened 2019-03-26
Add status RPC

KaustubhShamshery:
## Description
1. Upgrades libjson-rpc-cpp version to 0.7
2. Adds APIs for shard server
3. Add TCP socket server for IP binding

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


constants.xml
```diff
@@ -200,6 +200,11 @@
         <RESUME_BLACKLIST_DELAY_IN_SECONDS>30</RESUME_BLACKLIST_DELAY_IN_SECONDS>
         <INCRDB_DSNUMS_WITH_STATEDELTAS>10</INCRDB_DSNUMS_WITH_STATEDELTAS>
     </recovery>
+    <server>
```
Gnnng: The name `server` is a bit vague. Can it be just called `jsonrpc`?

Also, the option `ENABLE_SHARD_SERVER`. Are we sure about the name `SHARD_SERVER`?

KaustubhShamshery: okay, how about `jsonrpc` and `ENABLE_STATUS_RPC`

constants.xml
```diff
@@ -204,6 +204,11 @@
...
+        <IP_TO_BIND>127.0.0.1</IP_TO_BIND>
+        <SERVER_PORT>4201</SERVER_PORT>
```
bb111189: rename `SERVER_PORT` to something suitable if we are changing the name of `SHARD_SERVER`

PR #1556, opened 2019-03-27
Pregen Accounts In Batch and Toggle Pregened Accounts Updating

kaikawaliu:
## Description

With this PR, we can separate the accounts pre-populating into different ds epochs. Thus avoid throttling the instance for too long. And also added a constants to toggle the updating of pre-populated accounts. 

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libNode/Node.cpp
```diff
@@ -68,22 +68,44 @@ const unsigned int MIN_CHILD_CLUSTER_SIZE = 2;
 
 #define IP_MAPPING_FILE_NAME "ipMapping.xml"
 
-void Node::PopulateAccounts() {
+void Node::PopulateAccounts(bool temp) {
+  if (!ENABLE_ACCOUNTS_POPULATING) {
+    return;
```
ShengguangXiao: Add a log here?

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -170,6 +170,10 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone() {
         << m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() +
                1
         << "] FINISH WRITE STATE TO DISK");
+    if (m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() <
```
ansnunez: Should we add a check for `ENABLE_ACCOUNTS_POPULATING` to make this part more efficient?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -758,6 +758,10 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
         << m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum() +
                1
         << "] FINISH WRITE STATE TO DISK");
+    if (m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() <
```
ansnunez: Should we add a check for `ENABLE_ACCOUNTS_POPULATING` to make this part more efficient?

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -170,6 +170,10 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone() {
...
+        PREGEN_ACCOUNT_TIMES) {
```
ShengguangXiao: This variable means PREGEN accounts before this block number? Maybe should change to PREGEN_ACCOUNT_BLOCK_NUM?

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: It's checked inside the function in the first line, should not be an issue to the efficiency

src/libDirectoryService/FinalBlockPostProcessing.cpp
kaikawaliu: It means the time pregen has been called. Block num is just used as a reference here

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -761,6 +762,11 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
                                         .GetBlockNum() +
                                     1
                              << "] FINISH WRITE STATE TO DISK");
+        if (m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() <
```
ansnunez: Putting `ENABLE_ACCOUNTS_POPULATING` check first will be even more efficient.
So we avoid stuff like `GetLastBlock()` having to acquire mutex, etc...

kaikawaliu: Fixed for both places

PR #1561, opened 2019-03-28
LatestEpochStatesUpdated/DetachMoveUpdatesToDisk/MutexInBlockStorage

kaikawaliu:
## Description

- Update LatestEpochStatesUpdated after running MoveUpdatesToDisk
- Detach MoveUpdatesToDisk into new thread
- Added Json API for querying LatestEpochStatesUpdated

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


bb111189:
On hold temporary due to breaking change. Will not be part of v440

src/libPersistence/BlockStorage.cpp
```diff
@@ -680,6 +688,29 @@ bool BlockStorage::GetStateRoot(bytes& data) {
...
+    epochNumStr = m_stateRootDB->Lookup(LATEST_EPOCH_STATES_UPDATED);
+  }
+
+  if (epochNumStr.empty()) {
+    LOG_GENERAL(INFO, "No Latest Epoch State Updated get");
+    return false;
+  }
+
+  try {
+    epochNum = stoul(epochNumStr);
```
ckyang: `stoull` should be used here instead of `stoul`, because the epochNum is `uint64_t`.
And, if possible, please also change the other wrong usage inside the functions: `GetDiagnosticDataNodes` and `GetDiagnosticDataCoinbase`.
Thanks!

kaikawaliu: Fixed

PR #1577, opened 2019-03-30
Fix RejoinAsNormal after any failures 

sandipbhoir:
## Description
Fix RejoinAsNormal to handle any failures

- Fix for Rejoining of node after recovery and that not in shard is extended to all other rejoining cases that can happen for `RejoinAsNormal`
- RejoinAsNormal -> Install -> StartRetrieveFromHistory -> will skip checking of node if in shard or not.
 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -173,8 +173,7 @@ bool Node::Install(const SyncType syncType, const bool toRetrieveHistory,
 
     if (SyncType::NEW_SYNC == syncType ||
         SyncType::NEW_LOOKUP_SYNC == syncType ||
-        (rejoiningAfterRecover && (SyncType::NORMAL_SYNC == syncType ||
-                                   SyncType::DS_SYNC == syncType))) {
+        (rejoiningAfterRecover && (SyncType::NORMAL_SYNC == syncType))) {
       return true;
     }
 
@@ -563,7 +562,8 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
   }
 
   if (SyncType::NEW_SYNC == syncType || SyncType::NEW_LOOKUP_SYNC == syncType ||
-      (rejoiningAfterRecover && (SyncType::NORMAL_SYNC == syncType))) {
+      (rejoiningAfterRecover &&
+       (SyncType::NORMAL_SYNC == syncType || SyncType::DS_SYNC == syncType))) {
     return true;
   }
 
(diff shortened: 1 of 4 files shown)


src/libPersistence/BlockStorage.cpp
```diff
@@ -536,6 +536,7 @@ bool BlockStorage::GetAllTxBlocks(std::list<TxBlockSharedPtr>& blocks) {
 
   leveldb::Iterator* it =
       m_txBlockchainDB->GetDB()->NewIterator(leveldb::ReadOptions());
+  uint64_t count = 0;
```
KaustubhShamshery: uint64_t seems an overkill

sandipbhoir: number of TxBlks?

KaustubhShamshery: it is just a counter right, uint32_t should be enough, unless u wanna count more than 4294967295

sandipbhoir: yes we may need it...
we use uint64_t for all epoch, blocknums..

src/libNode/Node.cpp
```diff
@@ -1655,7 +1656,7 @@ void Node::RejoinAsNormal(bool rejoiningAfterRecover) {
         }
         BlockStorage::GetBlockStorage().RefreshAll();
         AccountStore::GetInstance().RefreshDB();
-        if (this->Install(SyncType::NORMAL_SYNC, true, rejoiningAfterRecover)) {
+        if (this->Install(SyncType::NORMAL_SYNC, true, true)) {
```
ShengguangXiao: Why rejoiningAfterRecover becomes always true here?

sandipbhoir: Because we don’t want it to do further check of whether this node was part of shard or is lagging behind. We return immediately from StartRetrieveFromHistory. It’s false only for the case in RECOVERY_ALL_SYNC when Zilliqa::Zilliqa() calls m_n.Install with false option.

jiayaoqijia: @ckyang Any side effect by replacing rejoiningAfterRecover?

sandipbhoir: it was added for supporting rejoining of node that is not in shard after recovery from another testnet.

PR #1580, opened 2019-04-01
Find solution in pow list before verify pow from leader

ShengguangXiao:
## Description

Find solution in pow list before verify pow from leader

When DS backup find some different nodes in priority list with same priority, the verify pow ordering will using different nodes, then it cannot find pow in sorted list, it need to check pow in local pow list first, because it is faster, if from local cannot find it, then check the pow from ds leader.

https://github.com/Zilliqa/Issues/issues/477

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -442,64 +442,27 @@ bool DirectoryService::VerifyPoWOrdering(
         LOG_GENERAL(WARNING, "Failed to find key in the PoW ordering "
                                  << toFind << " " << sortedPoWSolns.size());
 
-        LOG_GENERAL(INFO,
-                    "Checking for the key and PoW in the announcement...");
-
-        auto pubKeyToPoW = allPoWsFromLeader.find(toFind);
-        if (pubKeyToPoW != allPoWsFromLeader.end()) {
-          const auto& peer = std::get<SHARD_NODE_PEER>(shardNode);
-          const auto& powSoln = pubKeyToPoW->second;
-          auto headerHash = POW::GenHeaderHash(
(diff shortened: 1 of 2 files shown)


bb111189:
Do we need to add m_mutexAllPOW mutex to RunConsensusOnDSBlockWhenDSPrimary at line 676

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -564,6 +527,52 @@ bool DirectoryService::VerifyPoWOrdering(
...
+
+  if (!POW::GetInstance().PoWVerify(m_pendingDSBlock->GetHeader().GetBlockNum(),
+                                    difficulty, headerHash, powSoln.nonce,
+                                    resultStr, mixHashStr)) {
+    LOG_GENERAL(WARNING, "Failed to verify PoW solution from leader for node: "
+                             << pubKey);
+    return false;
+  }
+
+  m_allPoWs[pubKey] = powSoln;
```
bb111189: let's use at() ?

ShengguangXiao:
> Do we need to add `m_mutexAllPOW` and `m_mutexAllDSPOWs` mutex to RunConsensusOnDSBlockWhenDSPrimary at line 747

I think no need, because when run consensus, we don't receive pow already.

src/libDirectoryService/DSBlockPreProcessing.cpp
ShengguangXiao: This is an insert operation, i will use insert instead

ShengguangXiao: This is an insert operation, cannot use at

PR #1583, opened 2019-04-03
Fix the DNS resolve issue

sandipbhoir:
## Description
This PR fixes the issue : https://github.com/Zilliqa/Issues/issues/478

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libServer/Server.cpp
+++ b/src/libServer/Server.cpp
@@ -135,9 +135,6 @@ bool Server::StartCollectorThread() {
 
       bytes msg = {MessageType::LOOKUP, LookupInstructionType::FORWARDTXN};
 
-      auto upperLayerNodes = m_mediator.m_lookup->GetAboveLayer();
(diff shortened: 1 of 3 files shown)


src/libServer/Server.cpp
```diff
@@ -147,9 +144,7 @@ bool Server::StartCollectorThread() {
         continue;
       }
 
-      LOG_GENERAL(INFO, "Sent to " << upperLayerNode);
-
-      P2PComm::GetInstance().SendMessage(upperLayerNode, msg);
+      m_mediator.m_lookup->SendMessageToRandomSeedNode(msg);
```
KaustubhShamshery: I think we can just remove the func `GetAboveLayer`

sandipbhoir: Done

PR #1587, opened 2019-04-08
Implement the DS Reputation System.

nnamon:
## Description

- Implements the DS Reputation proposal as written here: https://github.com/nnamon/zilliqa-research/blob/master/ds_reputation/proposal.md.
- Diverges from the proposal by adding a separate constant for the max number of Byzantine nodes to remove to allow for node turnover eventuality.
- Diverges from the proposal by deriving the performance from the coinbase rewards rather than tracking it separately.
- Add two new constants: 
  - `DS_PERFORMANCE_THRESHOLD_PERCENT`: the ratio for the minimum number of co-sigs for a DS node to be considered non-Byzantine (0.0 - 1.0).
  - `NUM_DS_BYZANTINE_REMOVED`: the number of Byzantine DS nodes to remove. Must be less than or equal to `NUM_DS_ELECTION`.
- Diverges from the proposal to use the powWinners field and add a new field to the DS Block Header structure: `m_removeDSNodePubkeys` 
  - Update the protobuf definition.
  - Update the Messenger to serialize and deserialize the new field.
  - Increment `DSBLOCK_VERSION` to 2 in constants.xml.
  - Update existing tests to incorporate the new change to the DS Block header.
- Add new tests for the following components:
  - `UpdateDSComposition`
  - `SaveDSPerformance`
  - `DetermineByzantineNodes`

## Backward Compatibility

The `DirectoryService::VerifyRemovedByzantineNodes` will not pass on old DS Blocks when validating the history.

## Review Suggestion

When testing with a local testnet, it should be noted that nodes that are removed for non-performance and invoke the `RejoinAsNormal` call will be stuck because there is no S3 bucket or script to pull the latest state from.

### Implementation

- [ ] **check that `RejoinAsNormal` works after a node is removed for non-performance.**
- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -233,38 +229,48 @@ void DirectoryService::UpdateMyDSModeAndConsensusId() {
...
-      m_mediator.m_DSCommittee->size()) {
+
+  // Find my new consensus ID.
+  DequeOfNode::iterator it;
+  bool isDropout = true;
+  for (it = m_mediator.m_DSCommittee->begin();
+       it != m_mediator.m_DSCommittee->end(); ++it) {
+    // Look for my public key.
+    if (m_mediator.m_selfKey.second == it->first) {
+      m_consensusMyID = it - m_mediator.m_DSCommittee->begin();
```
KaustubhShamshery: better use the `distance` func

nnamon: Thanks, resolved in https://github.com/Zilliqa/Zilliqa/pull/1587/commits/95906a7d75a2be4474f3fd02b09ad55977828c08.

constants.xml
```diff
@@ -145,6 +145,11 @@
         <!-- Shard size will be automatically calculated if COMM_SIZE = 0 -->
         <COMM_SIZE>200</COMM_SIZE>
         <NUM_DS_ELECTION>5</NUM_DS_ELECTION>
+        <!-- DS Node must meet this minimum threshold to remain in the committee -->
+        <DS_PERFORMANCE_THRESHOLD_PERCENT>0.05</DS_PERFORMANCE_THRESHOLD_PERCENT>
```
ShengguangXiao: DS_PERFORMANCE_THRESHOLD_PERCENT and NUM_DS_BYZANTINE_REMOVED looks never used in the code

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -233,38 +229,48 @@ void DirectoryService::UpdateMyDSModeAndConsensusId() {
                                             .GetHashForRandom()
                                             .asBytes());
   }
-  // Check if I am the oldest backup DS (I will no longer be part of the DS
-  // committee)
-  if ((uint32_t)(m_consensusMyID + numOfIncomingDs) >=
-      m_mediator.m_DSCommittee->size()) {
+
+  // Find my new consensus ID.
+  DequeOfNode::iterator it;
```
ShengguangXiao: Put the definition into the for loop, can use auto it = m_mediator.m_DSCommittee->begin(), and use the m_mutexDSCommittee to lock the m_DSCommittee before use it.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -407,8 +541,14 @@ bool DirectoryService::VerifyPoWOrdering(
                                    << MAX_MISORDER_NODE << " nodes.");
 
   auto sortedPoWSolns = SortPoWSoln(priorityNodePoWs, true);
-  InjectPoWForDSNode(sortedPoWSolns,
-                     m_pendingDSBlock->GetHeader().GetDSPoWWinners().size());
+
+  // Count the number of actual PoW winners and removed nodes.
+  std::map<PubKey, Peer> dsPoWWinners =
+      m_pendingDSBlock->GetHeader().GetDSPoWWinners();
```
ShengguangXiao: Use auto

constants.xml
ansnunez: @ShengguangXiao DSBlockPreProcessing.cpp

nnamon: Used here: https://github.com/Zilliqa/Zilliqa/pull/1587/files#diff-76372bc310da6631dec75919a39cb4e2R1426

src/libDirectoryService/DSBlockPostProcessing.cpp
nnamon: Thanks, relevant iterators are now using auto instead of explicitly declaring iterator type in this commit: https://github.com/Zilliqa/Zilliqa/pull/1587/commits/83971d8f35d9b5f2eb51d743d8cf3957a46c8e62

src/libDirectoryService/DSBlockPreProcessing.cpp
nnamon: Thanks, changed in https://github.com/Zilliqa/Zilliqa/pull/1587/commits/91acff0ca0f2c1d7140c6c7d3eb05d79ec3f12eb.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -1216,6 +1366,124 @@ bool DirectoryService::ProcessShardingStructure(
...
+    if (score < threshold) {
+      // Only add the node to be removed if there is still capacity.
+      if (numByzantine < numToRemove) {
+        // Check if the public key that was found to a Byzantine node is mine
+        // because the network information for mine is zeroed out.
+        if (it->first == m_mediator.m_selfKey.second) {
+          powDSWinners[it->first] = Peer();
+        } else {
+          powDSWinners[it->first] = it->second;
+        }
```
ShengguangXiao: The under performanced DS node should kick out, if put it into powDSWinner, it will come back again?

src/libUtils/DSComposition.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  // Get the map of all pow winners from the DS Block
+  const auto& NewDSMembers = dsblock.GetHeader().GetDSPoWWinners();
+  DequeOfNode::iterator it;
+  unsigned int NumWinners = 0;
+
+  for (const auto& DSPowWinner : NewDSMembers) {
+    // Check if the current pow candidate is an existing DS Committee member.
+    // ('loser') and find its index.
+    for (it = dsComm.begin(); it != dsComm.end(); ++it) {
```
ShengguangXiao: use std::find instead of for loop

nnamon: Can't use `std::find` since `dsComm` is a `DequeOfNodes (Pair<PubKey, Peer>)` and `it` may have a zeroed out Peer and `it->first` is a `PubKey`. This causes `no match for ‘operator==’` errors because the type does not match for comparison. `std::find_if` should be used instead but this introduces lambdas which looks even less clear.

nnamon: I made the change to `std::find` initially but reverted it because it breaks the build.

ShengguangXiao: I see

nnamon:
Added the unit tests for libUtils/DSComposition.cpp.

src/libUtils/DSComposition.h
```diff
@@ -0,0 +1,32 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#ifndef __DSCOMPOSITION_H__
+#define __DSCOMPOSITION_H__
+
+#include "Logger.h"
```
ansnunez: Minor comments (no need to change for now)
1. I prefer just including the minimum required headers in a header file.  Logger, for example, is not needed here
2. We can probably keep this function inside libDirectoryService instead of libUtils?

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -460,18 +446,30 @@ void DirectoryService::StartFirstTxEpoch() {
       }
     }
 
+    // If I cannot find myself in the sharding structure, it means I must have
+    // been a non-performant node and must rejoin as normal.
     if (!found) {
       LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
-                "WARNING: Oldest DS node not in any of the new shards!");
+                "WARNING: My DS node is not in any of the new shards! I must "
```
ansnunez: Please change to:
`LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum, "My DS node signed insufficient blocks. Kicked out and invoking RejoinAsNormal now.");`

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -209,27 +209,90 @@ void DirectoryService::ComputeSharding(const VectorOfPoWSoln& sortedPoWSolns) {
...
     LOG_GENERAL(WARNING,
                 "FATAL: number of proposed ds member is larger than current ds "
                 "committee. numOfProposedDSMembers: "
-                    << numOfProposedDSMembers << " m_DSCommittee size: "
+                    << powDSWinners.size() << " m_DSCommittee size: "
                     << m_mediator.m_DSCommittee->size());
     return;
   }
 
+  if (numOfRemovedMembers > powDSWinners.size()) {
```
ansnunez: Is this check necessary?

nnamon: Not really, but it helps with being paranoid and not making so many assumptions about the incoming data.

src/libUtils/DSComposition.h
nnamon: 1. I've removed the unnecessary '#include "Logger.h"', thanks.
2. I can move it to libDirectoryService. The original rationale was that it was split out because libNode had an exact same copy of the code in it's own DSBlockProcessing.cpp in the current codebase and I did not want to introduce even more coupling. I moved the file to libDirectoryService.

nnamon:
Will be making changes to incorporate @ShengguangXiao's suggestion to introduce a new field in the DSBlock instead of using powWinners.

src/libDirectoryService/DSBlockPostProcessing.cpp
nnamon: Thanks for the clearer message suggestion. It has been changed.

nnamon:
Updated the PR to include the bare minimum to start working on the new field. Fixed all the tests and updated the core messaging code/ds block header to handle it. Incremented the `DSBLOCK_VERSION` constant too.

@ShengguangXiao, one question: Should there be anything to document that this field is introduced in DS Block Version 2 or do I just use the release version like in your example?

https://github.com/Zilliqa/Zilliqa/blob/bc0bc4ce725ade1285a0f0438cbac02645518ea7/src/libMessage/ZilliqaMessage.proto#L83

Logic to use the new field is in the next commit(s).

nnamon:
Logic implemented. @ShengguangXiao wanted to check if https://github.com/Zilliqa/Zilliqa/pull/1587/files#diff-76372bc310da6631dec75919a39cb4e2R1193 is acceptable to allow the validation to pass on old blocks. It will only verify the `m_removeDSNodePubkeys` on DS block version 2 and up.

src/libDirectoryService/DSBlockPreProcessing.cpp
nnamon: Removed the unnecessary checks with the newly added field in the latest commit.

nnamon:
Ready for review.

nnamon:
Ready for review.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -233,38 +230,47 @@ void DirectoryService::UpdateMyDSModeAndConsensusId() {
                                             .GetHashForRandom()
                                             .asBytes());
   }
-  // Check if I am the oldest backup DS (I will no longer be part of the DS
-  // committee)
-  if ((uint32_t)(m_consensusMyID + numOfIncomingDs) >=
-      m_mediator.m_DSCommittee->size()) {
+
+  // Find my new consensus ID.
+  bool isDropout = true;
```
bb111189: To check for Dropout, I think now we can make use of `m_removeDSNodePubkeys`

src/libMessage/ZilliqaMessage.proto
```diff
@@ -78,8 +78,9 @@ message ProtoDSBlock
             optional ByteArray val      = 2;  // Added in: v1.0, Deprecated in: N/A
             // Add new members here
         }
-        repeated PowDSWinners dswinners = 10;  // Added in: v1.0, Deprecated in: N/A
+        repeated PowDSWinners dswinners = 10; // Added in: v1.0, Deprecated in: N/A
         optional DSBlockHashSet hash    = 11; // Added in: v1.0, Deprecated in: N/A
+        repeated ByteArray dsremoved = 12; // Added in: v4.5, Deprecated in: N/A
```
bb111189: Change to `added in: v2.0` since the version is from block version

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -222,14 +230,41 @@ void DirectoryService::InjectPoWForDSNode(VectorOfPoWSoln& sortedPoWSolns,
     return;
   }
 
+  if (numOfRemovedMembers > numOfProposedDSMembers) {
```
bb111189: while the previous implementation is `>`, I think it may make more sense it we use `!=` here

src/libDirectoryService/DSBlockPostProcessing.cpp
nnamon: `dropout` is on two possibilities:

1. is a removed node
2. is an expired node.

This covers both possibilities so I think handling them separately is not the best.

src/libDirectoryService/DSBlockPreProcessing.cpp
nnamon: No, `numOfRemovedMembers` must be strictly less than or equals to `numOfProposedDSMembers` because `numOfRemovedMembers` contains the members to be removed for non-performance only, not expiry.

src/libDirectoryService/DSBlockPostProcessing.cpp
nnamon: To clarify, the `m_removeDSNodePubkeys` field contains only removals for non-performance as suggested in the @ShengguangXiao's proposal to add the new field. The expired nodes are calculated in `UpdateDSCommitteeComposition`. I think this is better than putting expired nodes in the field because it allows some distinction deterministically when just looking at the DS block alone.

src/libMessage/ZilliqaMessage.proto
nnamon: Thanks, made the change in https://github.com/Zilliqa/Zilliqa/pull/1587/commits/5f896af587a3653bb5cc0aac0d518253d0ba6819

src/libDirectoryService/DSBlockPreProcessing.cpp
bb111189: Ok then this will make sense.

src/libDirectoryService/DSBlockPostProcessing.cpp
bb111189: I see. Then it does make sense. Maybe we can add an inline comment for the possibilities of dropout
```
is a removed node
is an expired node
```

nnamon: Thanks for the suggestion. Included more comments to make it clearer.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -287,44 +293,20 @@ void DirectoryService::UpdateMyDSModeAndConsensusId() {
   }
 }
 
-void DirectoryService::UpdateDSCommiteeComposition() {
+void DirectoryService::UpdateDSCommitteeComposition() {
```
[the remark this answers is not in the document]
ShengguangXiao: Yes, it is better to make it protected

src/libMessage/ZilliqaMessage.proto
```diff
@@ -78,8 +78,9 @@ message ProtoDSBlock
             optional ByteArray val      = 2;  // Added in: v1.0, Deprecated in: N/A
             // Add new members here
         }
-        repeated PowDSWinners dswinners = 10;  // Added in: v1.0, Deprecated in: N/A
+        repeated PowDSWinners dswinners = 10; // Added in: v1.0, Deprecated in: N/A
         optional DSBlockHashSet hash    = 11; // Added in: v1.0, Deprecated in: N/A
+        repeated ByteArray dsremoved = 12; // Added in: v2.0, Deprecated in: N/A
```
ShengguangXiao: Please align the number and the comment

nnamon: Thanks, done.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -209,10 +212,15 @@ void DirectoryService::ComputeSharding(const VectorOfPoWSoln& sortedPoWSolns) {
   }
 }
 
-void DirectoryService::InjectPoWForDSNode(VectorOfPoWSoln& sortedPoWSolns,
-                                          unsigned int numOfProposedDSMembers) {
-  // Add the oldest n DS committee member to m_allPoWs and m_allPoWConns so it
-  // gets included in sharding structure
+void DirectoryService::InjectPoWForDSNode(
+    VectorOfPoWSoln& sortedPoWSolns, unsigned int numOfProposedDSMembers,
+    std::vector<PubKey>& removeDSNodePubkeys) {
```
ShengguangXiao: use const infront of std::vector<PubKey>&

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -43,6 +44,8 @@
 using namespace std;
 using namespace boost::multiprecision;
 
+#define REMOVED_FIELD_DSBLOCK_VERSION 2
```
ShengguangXiao: use const uint32_t instead of #define

src/libDirectoryService/DSBlockPreProcessing.cpp
nnamon: Thanks, made the change.

src/libDirectoryService/DSBlockPreProcessing.cpp
nnamon: Thanks, made the change.

src/libDirectoryService/DirectoryService.h
```diff
@@ -631,6 +645,20 @@ class DirectoryService : public Executable {
...
+          coinbaseRewardees,
+      std::map<PubKey, uint32_t>& dsMemberPerformance, DequeOfNode& dsComm,
+      uint64_t currentEpochNum, unsigned int numOfFinalBlock,
+      int finalblockRewardID);
+  static unsigned int DetermineByzantineNodesCore(
+      unsigned int numOfProposedDSMembers,
+      std::vector<PubKey>& removeDSNodePubkeys, uint64_t currentEpochNum,
+      unsigned int numOfFinalBlock, double performanceThreshold,
+      unsigned int maxByzantineRemoved, DequeOfNode& dsComm,
+      std::map<PubKey, uint32_t>& dsMemberPerformance);
```
ShengguangXiao: use const for dsMemberPerformance

ShengguangXiao:
@nnamon Can you do a rebase to sync with the latest master, then i can start to do cloud test for this change.

src/libDirectoryService/DirectoryService.h
nnamon: Thanks, done.

nnamon:
@ShengguangXiao the branch is synced to master.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -68,9 +67,6 @@ unsigned int DirectoryService::ComputeDSBlockParameters(
     }
     powDSWinners[submitter.second] = m_allPoWConns[submitter.second];
     dsWinnerPoWs[submitter.second] = m_allDSPoWs[submitter.second];
-    sortedPoWSolns.erase(
```
nnamon: @ShengguangXiao to re-order the SortPoWSoln calls, I need to remove the dependency of `sortedPoWSolns` in `DirectoryService::ComputeDSBlockParameters`.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -509,6 +503,22 @@ bool DirectoryService::VerifyPoWOrdering(
   std::vector<PubKey> removeDSNodePubkeys =
       m_pendingDSBlock->GetHeader().GetDSRemovePubKeys();
 
+  // Sort and trim the PoW solutions
+  auto sortedPoWSolns =
+      SortPoWSoln(priorityNodePoWs, true, removeDSNodePubkeys.size());
+
+  // Remove the DS solutions from the PoW solutions.
```
nnamon: Need to check with you on this portion, if it's needed. I just made sure the same computation was done for the Backup nodes like in the Leader.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -744,10 +755,15 @@ VectorOfPoWSoln DirectoryService::SortPoWSoln(const MapOfPubKeyPoW& mapOfPoWs,
...
   if (trimBeyondCommSize) {
     const uint32_t numNodesTotal = PoWOrderSorter.size();
+
+    // Number of Nodes to Trim. Account for the removed Byzantine nodes that do
+    // not get injected as a shard solution.
     const uint32_t numNodesAfterTrim =
-        ShardSizeCalculator::GetTrimmedShardCount(
-            m_mediator.GetShardSize(false), SHARD_SIZE_TOLERANCE_LO,
-            SHARD_SIZE_TOLERANCE_HI, numNodesTotal);
+        std::min(ShardSizeCalculator::GetTrimmedShardCount(
```
nnamon: Allowance for extra solutions to be left in is calculated here.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -913,12 +928,28 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary() {
...
 
   // Determine the losers from the performance.
-  DetermineByzantineNodes(numOfProposedDSMembers, removeDSNodePubkeys);
+  unsigned int numByzantine =
+      DetermineByzantineNodes(numOfProposedDSMembers, removeDSNodePubkeys);
+
+  // Sort and trim the PoW solutions.
+  auto sortedPoWSolns = SortPoWSoln(allPoWs, true, numByzantine);
+
+  // Remove the DS solutions from the PoW solutions.
```
nnamon: I move the removing of DS solutions from the PoW solutions here from `ComputeDSBlockParameters`, after the DSBlock is computed and the sorted PoW solutions are generated.

nnamon:
@ShengguangXiao The requested changes for the rolling upgrade has been made. Do you want me to sync to master too?

ShengguangXiao:
> @ShengguangXiao The requested changes for the rolling upgrade has been made. Do you want me to sync to master too?

Yes, please sync to latest master then i can test again @nnamon

nnamon:
@ShengguangXiao Synced to master. Build should pass now.

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -1071,6 +1217,20 @@ bool DirectoryService::DSBlockValidator(
     return false;
   }
 
+  // Check if the current block version to be validated requires removed nodes
+  // validation.
+  const uint32_t REMOVED_FIELD_DSBLOCK_VERSION = 2;
+  if (m_pendingDSBlock->GetHeader().GetVersion() >=
+          REMOVED_FIELD_DSBLOCK_VERSION &&
+      m_mediator.m_currentEpochNum >= ENABLE_NEW_FEATURE_DS_NUM) {
```
ShengguangXiao: Change to m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() >=
      ENABLE_NEW_FEATURE_DS_NUM

src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -797,21 +917,47 @@ bool DirectoryService::RunConsensusOnDSBlockWhenDSPrimary() {
...
-      sortedDSPoWSolns, sortedPoWSolns, powDSWinners, dsWinnerPoWs,
-      dsDifficulty, difficulty, blockNum, prevHash);
+  // Determine the DS PoW winners.
+  unsigned int numOfProposedDSMembers =
+      ComputeDSBlockParameters(sortedDSPoWSolns, powDSWinners, dsWinnerPoWs,
+                               dsDifficulty, difficulty, blockNum, prevHash);
+
+  // Determine the losers from the performance.
+  unsigned int numByzantine = 0;
+  if (m_mediator.m_currentEpochNum >= ENABLE_NEW_FEATURE_DS_NUM) {
```
ShengguangXiao: Change to m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() >=
      ENABLE_NEW_FEATURE_DS_NUM

ShengguangXiao:
@nnamon The PR is almost ready, just need to change if (m_mediator.m_currentEpochNum >= ENABLE_NEW_FEATURE_DS_NUM) to if (m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() >=
      ENABLE_NEW_FEATURE_DS_NUM), because we need to use DS block number to do the check instead of txn block number

nnamon:
@ShengguangXiao thanks, change has been made.

ShengguangXiao:
> @ShengguangXiao thanks, change has been made.

Good, i think this PR is ready, will merge after the smart contract upgrade is done

ShengguangXiao:
@nnamon @ckyang There is already one field UPGRADE_TARGET_DS_NUM in constants.xml reserved for upgrading breaking change, can you remove ENABLE_NEW_FEATURE_DS_NUM and use UPGRADE_TARGET_DS_NUM instead?

nnamon:
@ShengguangXiao changes made, thanks.

bb111189:
Tagging the issue that comes with this PR
https://github.com/Zilliqa/Zilliqa/issues/1517

PR #1588, opened 2019-04-09
DS committee exclude ds guard nodes from blacklist

ShengguangXiao:
## Description

DS committee exclude ds guard nodes from blacklist

https://github.com/Zilliqa/Issues/issues/483

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -1609,73 +1609,69 @@ bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
     return false;
   }
 
-  else {
-    bool isVerif = true;
+  bool isVerif = true;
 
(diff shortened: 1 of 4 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -1609,75 +1609,71 @@ bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
     return false;
   }
 
-  else {
-    bool isVerif = true;
+  bool isVerif = true;
```
ansnunez: @ShengguangXiao just to check: without the `else`, the remainder of this function will also be executed for non-lookup every time the node receives `SETDSINFOFROMSEED`. Is this OK?

ckyang: Actually there's no logic change here, because in line 1612 `return false` over there.

PR #1589, opened 2019-04-10
Check status of a txn in a normal node's mempool

KaustubhShamshery:
## Description
1. Add API for normal nodes for querying txn status
2. Add API for lookup nodes for getting information about some of the shard nodes
3. Modify `miner_info` to support

Example output

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/MicroBlockPreProcessing.cpp
+++ b/src/libNode/MicroBlockPreProcessing.cpp
@@ -432,15 +432,7 @@ void Node::ProcessTransactionWhenShardLeader() {
 
   cv_TxnProcFinished.notify_all();
   // Put txns in map back into pool
-  for (const auto& kv : t_addrNonceTxnMap) {
-    for (const auto& nonceTxn : kv.second) {
-      t_createdTxns.insert(nonceTxn.second);
-    }
-  }
-
-  for (const auto& t : gasLimitExceededTxnBuffer) {
-    t_createdTxns.insert(t);
-  }
+  ReinstateMemPool(t_addrNonceTxnMap, gasLimitExceededTxnBuffer);
 }
 
 bool Node::VerifyTxnsOrdering(const vector<TxnHash>& tranHashes,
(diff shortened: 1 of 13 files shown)


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -653,16 +645,40 @@ void Node::ProcessTransactionWhenShardBackup() {
...
+  }
+}
+
+PoolTxnStatus Node::IsTxnInMemPool(const TxnHash& txhash) const {
+  shared_lock<shared_timed_mutex> g(m_unconfirmedTxnsMutex, defer_lock);
+  // Try to lock for 100 ms
+  if (!g.try_lock_for(chrono::milliseconds(100))) {
+    return PoolTxnStatus::ERROR;
+  }
+  if (m_unconfirmedTxns.find(txhash) == m_unconfirmedTxns.end()) {
```
ansnunez: To avoid redundant search, you can put the result of `find` in a variable and access that for the last `return` statement

src/libServer/Server.h
```diff
@@ -434,6 +454,8 @@ class AbstractZServer : public jsonrpc::AbstractServer<AbstractZServer> {
   virtual std::string GetNodeType() = 0;
   virtual Json::Value GetDSCommittee() = 0;
   virtual std::string GetNodeState() = 0;
+  virtual Json::Value IsTxnInMemPool(const std::string& param01) = 0;
+  virtual Json::Value GetShardMembers(uint32_t shardID) = 0;
```
ansnunez: For naming consistency, should we also call this `uint32_t param01`?

src/libServer/Server.cpp
```diff
@@ -1376,4 +1422,46 @@ string Server::GetNodeState() {
...
+  shuffle(v.begin(), v.end(), eng);
+  vector<uint> newVec(v.begin(), v.begin() + num);
+  return newVec;
+}
+
+Json::Value Server::GetShardMembers(uint32_t shardID) {
+  if (!LOOKUP_NODE_MODE) {
+    throw JsonRpcException(RPC_INVALID_REQUEST, "Sent to a non-lookup");
+  }
+  const auto& num_shards = m_mediator.m_lookup->GetShardPeers().size();
```
ansnunez: I think each of the two calls to `GetShardPeers()` in this function creates a temporary anonymous `DequeOfShard` object, which can be costly. Also, the call itself needs to acquire mutex each time, and maybe in the second time the shard structure has changed (e.g., shard dropped).  Perhaps store it to a local after the first call.

src/libNode/MicroBlockPreProcessing.cpp
KaustubhShamshery: done

src/libServer/Server.h
KaustubhShamshery: done

src/libServer/Server.cpp
KaustubhShamshery: done

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -653,16 +645,41 @@ void Node::ProcessTransactionWhenShardBackup() {
...
     for (const auto& nonceTxn : kv.second) {
       t_createdTxns.insert(nonceTxn.second);
+      m_unconfirmedTxns.emplace(nonceTxn.second.GetTranID(),
+                                PoolTxnStatus::PRESENT_NONCE_HIGH);
     }
   }
 
   for (const auto& t : gasLimitExceededTxnBuffer) {
     t_createdTxns.insert(t);
+    m_unconfirmedTxns.emplace(t.GetTranID(),
```
ShengguangXiao: Is t_createdTxns thread safe here? Need m_mutexCreatedTransactions to protect?

PR #1591, opened 2019-04-11
Recovery of dsguard

sandipbhoir:
## Description
This PR fixed following issue w.r.t recovery of dsguard:

- https://github.com/Zilliqa/Issues/issues/485
- https://github.com/Zilliqa/Issues/issues/486
- Handle case of GUARD_DS_SYNC - syncType 7. 
- Handle CleanVariable

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libZilliqa/Zilliqa.cpp
+++ b/src/libZilliqa/Zilliqa.cpp
@@ -289,7 +289,8 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
       case SyncType::GUARD_DS_SYNC:
         LOG_GENERAL(INFO, "Sync as a ds guard node");
         m_ds.m_awaitingToSubmitNetworkInfoUpdate = true;
-        // m_ds.StartSynchronization();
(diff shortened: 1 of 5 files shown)


src/libZilliqa/Zilliqa.cpp
```diff
@@ -289,7 +289,9 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
       case SyncType::GUARD_DS_SYNC:
         LOG_GENERAL(INFO, "Sync as a ds guard node");
         m_ds.m_awaitingToSubmitNetworkInfoUpdate = true;
-        m_ds.StartSynchronization();
+        // m_ds.StartSynchronization();
```
bb111189: Remove this line and leave a comment on the logic of the following 2 lines. This is so that it is easier to understand the logic in the future

sandipbhoir: Added.

PR #1592, opened 2019-04-11
Fix Json Conversion For Handling Corner Case Of Contract Txn

kaikawaliu:
## Description

Fixed the issue when the data field in transaction is parsed into a Json stringObject, which caused undefined behavior when modifying the object.

And also remove the non-empty check for Microblock TxnHash during Forwarding Txn and MicroBlock from shard to lookup. Otherwise failed transaction will always remain in the txn pool before vacuous epoch.
 
Besides some try-catch was added for safety reason.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -283,13 +283,10 @@ bool AccountStoreSC<MAP>::UpdateAccounts(
                       "Txn processing timeout! Interrupt current contract "
                       "deployment, pid: "
                           << pid);
-          try {
-            if (pid >= 0) {
-              kill(pid, SIGKILL);
-            }
-          } catch (const std::exception& e) {
-            LOG_GENERAL(WARNING, "Exception caught in kill pid: " << e.what());
+          if (pid >= 0) {
+            kill(pid, SIGKILL);
           }
+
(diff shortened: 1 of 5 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -159,21 +159,28 @@ bool AccountStoreSC<MAP>::UpdateAccounts(
...
+        LOG_GENERAL(WARNING, "ExportCreateContractFiles failed");
+        init = false;
+      }
 
-    if (init && !this->DecreaseBalance(fromAddr, gasDeposit)) {
+      if (init && !this->DecreaseBalance(fromAddr, gasDeposit)) {
+        init = false;
+      }
+    } catch (const std::exception& e) {
+      LOG_GENERAL(WARNING,
```
KaustubhShamshery: It might be beneficial to add more logging as to what inpur caused the exception

kaikawaliu: Yeah agree, but it needs more refined exception handling structure. The current one is more likely to serve for avoid crashes so its scope is a bit wide, which makes it difficult to log down everything inside the `try` field.

PR #1595, opened 2019-04-12
Put a limit on storage of lookup

KaustubhShamshery:
## Description

This PR adds an upper bound TXN_STORAGE_LIMIT to the lookup. It also disallows sending a txn with same txn hash if it already has this txn.
Also, contains a fix for a possible infinite loop case https://github.com/Zilliqa/Zilliqa/blob/419eb5bde7c00ea8996524ef185b34e0d9cf826b/src/libMessage/Messenger.cpp#L4360 .

## Backward Compatibility

## Review Suggestion
The CreateTransaction function has been broken to look more cleaner. Reviewers please see that the function is same as functionally before

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -3542,7 +3547,21 @@ bool Lookup::AddToTxnShardMap(const Transaction& tx, uint32_t shardId) {
...
+    LOG_GENERAL(INFO, "Number of txns exceeded limit");
+    return false;
+  }
+
+  // case where txn already exist
+  if (find_if(m_txnShardMap.first[shardId].begin(),
+              m_txnShardMap.first[shardId].end(), [tx](const Transaction& txn) {
+                return tx.GetTranID() == txn.GetTranID();
+              }) != m_txnShardMap.first[shardId].end()) {
+    return false;
```
bb111189: shall we add a log warning here?

src/libLookup/Lookup.cpp
```diff
@@ -3542,7 +3547,21 @@ bool Lookup::AddToTxnShardMap(const Transaction& tx, uint32_t shardId) {
 
   lock_guard<mutex> g(m_txnShardMapMutex);
 
-  m_txnShardMap[shardId].push_back(tx);
+  if (m_txnShardMap.second >= TXN_STORAGE_LIMIT) {
+    LOG_GENERAL(INFO, "Number of txns exceeded limit");
```
bb111189: shall we use log warning here

src/libLookup/Lookup.h
```diff
@@ -181,7 +189,9 @@ class Lookup : public Executable {
   VectorOfNode GetSeedNodes() const;
 
   std::mutex m_txnShardMapMutex;
```
bb111189: move mutex to line 133

KaustubhShamshery: cannot, needs to be public mutex

PR #1596, opened 2019-04-12
 Fix a peer can connect to another peer with unlimited connections

ShengguangXiao:
## Description

Fix a peer can connect to another peer with unlimited connections

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNetwork/P2PComm.h
```diff
@@ -99,6 +99,9 @@ class P2PComm {
   Peer m_selfPeer;
   PairOfKey m_selfKey;
 
+  static std::mutex m_mutexPeerConnectionCount;
+  static std::map<uint128_t, uint16_t> m_peerConnectionCount;
```
ckyang: For header files, it's better to have as less coupling with other module as possible. Suggest to use `boost::multiprecision::uint128_t` here instead of `uint128_t`, just like what other header files did. Then in `src/common/BaseType.h` neither the extra including header file, nor the `using uint128_t` declaration are needed also.

ShengguangXiao: Later i will do a refactor to replace all the boost::multiprecision::uint128_t with uint128_t, then the code will looks more clear to read.

PR #1597, opened 2019-04-12
Fix wrong usage of updating DSINCOMPLETED for lookup node

kaikawaliu:
## Description

Let lookup only update MetaType::DSINCOMPLETED only when both of states storing and microblock receiving were done.  

## Backward Compatibility

### Implementation

- [x] **ready for review**

~- [ ] local machine test~

- [x] small-scale cloud test


kaikawaliu:
Refer to issue: https://github.com/Zilliqa/Issues/issues/488

PR #1598, opened 2019-04-12
Add RPC for addition to whitelist and removal

KaustubhShamshery:
## Description
Add RPC for removal  and addition of nodes from whitelist (blacklist exclusion list)

Example output:
```
python scripts/miner_info.py -p 4301 whitelist_add -pm 142.23.2.2 abcd 278.55.1.12
Parameter: 142.23.2.2	True
Parameter: abcd	{u'message': u'IP Address provided not valid', u'code': -8, u'data': None}
Parameter: 278.55.1.12	{u'message': u'IP Address provided not valid', u'code': -8, u'data': None}
```

## Backward Compatibility

## Review Suggestion
use `python scripts/miner_info.py whitelist_add --pm [IP]` to add 
use `python scripts/miner_info.py whitelist_remove --pm [IP]` to remove 

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libServer/Server.cpp
+++ b/src/libServer/Server.cpp
@@ -35,6 +35,7 @@
 #include "libData/AccountData/Transaction.h"
 #include "libMediator/Mediator.h"
 #include "libMessage/Messenger.h"
+#include "libNetwork/Blacklist.h"
 #include "libNetwork/P2PComm.h"
 #include "libNetwork/Peer.h"
 #include "libPersistence/BlockStorage.h"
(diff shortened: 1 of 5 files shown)


src/libServer/Server.cpp
```diff
@@ -1376,4 +1377,61 @@ string Server::GetNodeState() {
...
+
+    if (!IPConverter::ToNumericalIPFromStr(ipAddr, numIP)) {
+      throw JsonRpcException(RPC_INVALID_PARAMETER,
+                             "IP Address provided not valid");
+    }
+
+    if (!Blacklist::GetInstance().Exclude(numIP)) {
+      throw JsonRpcException(
+          RPC_INVALID_PARAMETER,
+          "Could not add IP Address in exclusion list, already present");
```
bb111189: There is another case. i.e blacklist is disabled. 
Maybe good to refine the message

PR #1599, opened 2019-04-14
Handle VC Precheck cornercase

bb111189:
## Description

When view change precheck failed on a node, `cv_ViewChangeVCBlock` was not notified. As such, when the node rejoins as DS, there is a possibility of view change happening again (on that node), disrupting that particular node ability to participate in normal consensus. 

This PR fixes the aboved mention issue. This PR also refactoring the post view change cleaning up into one function.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


bb111189:
Will test together with the upcoming release

bb111189:
> I roughly remember there could be case where ds member may join back with RejoinAsNormal. May be we need to cleanupviewchange if any if such case exist.
> Otherwise it looks good.

That will be triggered under `FinishRejoinAsDS()`. So that case is covered.

PR #1600, opened 2019-04-15
Add txn status api

KaustubhShamshery:
## Description
this PR is same as of 1589, except changes in miner_info.py

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libServer/Server.cpp
```diff
@@ -1352,7 +1397,8 @@ string Server::GetNodeType() {
   } else if (m_mediator.m_ds->m_mode != DirectoryService::Mode::IDLE) {
     return "DS Node";
   } else {
-    return "Shard Node";
+    return string("Shard Node of shard") +
```
sandipbhoir: still space missing

PR #1609, opened 2019-04-18
Auto recovery of seed nodes if lagging due to missing FB and DS blocks

sandipbhoir:
## Description
This PR addresses issue -> https://github.com/Zilliqa/Issues/issues/495

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3095,6 +3095,33 @@ bool Lookup::GetMyLookupOnline(bool fromRecovery) {
   return true;
 }
 
+void Lookup::RejoinAsNewLookup() {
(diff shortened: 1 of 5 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3095,6 +3096,39 @@ bool Lookup::GetMyLookupOnline(bool fromRecovery) {
   return true;
 }
 
+void Lookup::RejoinAsNewLookup() {
+  if (!LOOKUP_NODE_MODE && !ARCHIVAL_LOOKUP) {
```
ansnunez: Should be `||`?

sandipbhoir: Fixed

PR #1612, opened 2019-04-22
Fix corner case of safemath

ShengguangXiao:
## Description

Fix corner case of safemath

An signed integer substract, for example, type int8_t, when a = 0, b = -128, the substract is overflowed, but it is not handled. Now it has been fixed.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/Utils/Test_SafeMath.cpp
+++ b/tests/Utils/Test_SafeMath.cpp
@@ -542,6 +542,23 @@ BOOST_AUTO_TEST_CASE(test_int8_t) {
   BOOST_CHECK_MESSAGE(false == SafeMath<int8_t>::mul(num1, num2, mulRes6),
                       "Test mul underflow failed!");
 
+  int8_t subRes;
+  BOOST_CHECK_MESSAGE(
+      false ==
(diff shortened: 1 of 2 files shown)


tests/Utils/Test_SafeMath.cpp
```diff
@@ -792,6 +860,23 @@ BOOST_AUTO_TEST_CASE(test_boost_int128_t) {
...
+  BOOST_REQUIRE(std::numeric_limits<boost::multiprecision::int128_t>::max() ==
+                subRes);
+
+  BOOST_CHECK_MESSAGE(
+      false == SafeMath<boost::multiprecision::int128_t>::sub(
+                   1,
+                   std::numeric_limits<boost::multiprecision::int128_t>::min(),
+                   subRes),
+      "Test signed sub overflow failed!");
+
```
ckyang: May we need to check `-1-min` for `int128_t`, `int256_t`, `int512_t`, and `int1024_t`?

ShengguangXiao: I change the test case because for boost int128_t the negative range is same as positive range

PR #1615, opened 2019-04-23
Fix notification of CV and deprecate of FetchDSBlocks

KaustubhShamshery:
## Description
If the node is in SYNC state, it might get the ds block through setDirectoryBlocks, which should also trigger the CV 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3415,8 +3415,7 @@ bool Lookup::ProcessSetDirectoryBlocksFromSeed(
       m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum();
 
   if (dsblocknumafter > dsblocknumbefore) {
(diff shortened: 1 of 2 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3289,6 +3289,10 @@ bool Lookup::ProcessGetDirectoryBlocksFromSeed(const bytes& message,
     return false;
   }
 
+  lock_guard<mutex> g(m_mediator.m_node->m_mutexDSBlock);
+
+  LOG_GENERAL(INFO, "Index Num: " << index_num);
```
ShengguangXiao: LOG_GENERAL(INFO, "DS block link chain Index Num: " << index_num);

sandipbhoir:
Do we need this PR? @KaustubhShamshery

PR #1617, opened 2019-04-23
Add size check

ShengguangXiao:
## Description

This PR add size check to various part of the code
Details as below
https://github.com/Zilliqa/Issues/issues/497

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -1140,7 +1140,7 @@ bool Node::ProcessSubmitMissingTxn(const bytes& message, unsigned int offset,
     LOG_GENERAL(WARNING, "Invalid txn message, message size: "
                              << message.size()
                              << ", txn data offset: " << offset);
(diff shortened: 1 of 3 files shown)


src/libNode/Node.cpp
```diff
@@ -1136,6 +1136,14 @@ bool Node::ProcessSubmitMissingTxn(const bytes& message, unsigned int offset,
     return true;
   }
 
+  if (offset >= message.size()) {
```
bb111189: should this be `>` or `>=` ?

ShengguangXiao: Originally i using >=, but it will break a lot of unit test case in Account

ShengguangXiao: Now changed to >=, please review again @bb111189

PR #1620, opened 2019-04-24
Update ProcessSetDSInfoFromSeed

ansnunez:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -1571,7 +1571,13 @@ void Lookup::SendGetMicroBlockFromLookup(const vector<BlockHash>& mbHashes) {
 
 bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
                                       const Peer& from) {
-  //#ifndef IS_LOOKUP_NODE
+  if (LOOKUP_NODE_MODE && (m_mediator.m_currentEpochNum > 1)) {
+    LOG_GENERAL(
+        WARNING,
+        "Lookup::ProcessSetDSInfoFromSeed not expected to be called from "
+        "LookUp node.");
+    return true;
+  }
 
   LOG_MARKER();
 
@@ -1599,6 +1605,13 @@ bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
       LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
                 "The message sender pubkey: "
                     << senderPubKey << " is not in my lookup node list.");
+      return false;
+    }
+  } else {
+    if (GUARD_MODE && !Guard::GetInstance().IsNodeInDSGuardList(senderPubKey)) {
+      LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
+                "The message sender pubkey: " << senderPubKey
+                                              << " is not in DS guard list.");
       return false;
     }
   }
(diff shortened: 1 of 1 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -1571,7 +1571,13 @@ void Lookup::SendGetMicroBlockFromLookup(const vector<BlockHash>& mbHashes) {
 
 bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
                                       const Peer& from) {
-  //#ifndef IS_LOOKUP_NODE
+  if (LOOKUP_NODE_MODE && (m_mediator.m_currentEpochNum > 1)) {
```
KaustubhShamshery: actually it can call this while syncing, only it should not be sent by DS nose

src/libLookup/Lookup.cpp
```diff
@@ -1653,17 +1657,18 @@ bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
...
+                     << m_mediator.m_blocklinkchain.GetBuiltDSComm().size()
+                     << " " << dsNodes.size() << " does not match");
+    return false;
   }
 
+  bool isVerif = true;
+
   for (i = 0; i < m_mediator.m_blocklinkchain.GetBuiltDSComm().size(); i++) {
-    if (!(m_mediator.m_DSCommittee->at(i).first ==
+    if (!(dsNodes.at(i).first ==
```
ShengguangXiao: Change ! == to !=

ansnunez: We don't have `!=` implemented for pubkey, so that would require a change to libCrypto as well.  But yes would be good to have that soon in another PR.

src/libLookup/Lookup.cpp
```diff
@@ -1601,6 +1599,14 @@ bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
                     << senderPubKey << " is not in my lookup node list.");
       return false;
     }
+  } else {
+    if ((m_mediator.m_currentEpochNum <= 1) && GUARD_MODE &&
```
KaustubhShamshery: (m_mediator.m_currentEpochNum <= 1) || (GUARD_MODE && !Guard::GetInstance().IsNodeInDSGuardList(senderPubKey))

src/libLookup/Lookup.cpp
```diff
@@ -1601,6 +1599,14 @@ bool Lookup::ProcessSetDSInfoFromSeed(const bytes& message, unsigned int offset,
...
+        !Guard::GetInstance().IsNodeInDSGuardList(senderPubKey)) {
+      LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
+                "The message sender pubkey: " << senderPubKey
```
bb111189: should log that `m_currentEpochNum <= 1`

ansnunez: Since we use `LOG_EPOCH`, then maybe it's sufficient?  Also, this is just during bootstrap anyway

src/libLookup/Lookup.cpp
ansnunez: Implemented but reverted, because we only want to check pubkey for initial epoch

PR #1621, opened 2019-04-24
DS store and send DSMicroBlock at every epoch

kaikawaliu:
## Description

Before this PR:
1. DS never store DSMB
2. DS only send DSMB to lookup when 1. non-vacuous 2. TxRootHash not null
3. Lookup load every microblocks involved in the TxBlock

After this PR:
1. DS store DSMB when TxnRootHash not null
2. DS send DSMB whenever TxRootHash not null
3. Lookup won't load DSMB if TxRootHash is null

@KaustubhShamshery can check if there is anything need to be changed on the historical mb verification, basically can skip ds mb if the TxnRootHash is empty
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -708,11 +708,12 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
 
   DetachedFunction(1, resumeBlackList);
 
+  if (!LoadUnavailableMicroBlockHashes(
+          txBlock, txBlock.GetHeader().GetBlockNum(), toSendTxnToLookup)) {
+    return false;
+  }
+
   if (!isVacuousEpoch) {
-    if (!LoadUnavailableMicroBlockHashes(
-            txBlock, txBlock.GetHeader().GetBlockNum(), toSendTxnToLookup)) {
-      return false;
-    }
     StoreFinalBlock(txBlock);
   } else {
     LOG_GENERAL(INFO, "isVacuousEpoch now");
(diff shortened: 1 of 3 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -123,8 +123,13 @@ bool Node::LoadUnavailableMicroBlockHashes(const TxBlock& finalBlock,
                             << info.m_microBlockHash << " [TxnRootHash] "
                             << info.m_txnRootHash << " shardID "
                             << info.m_shardId);
-      m_unavailableMicroBlocks[blocknum].push_back(
-          {info.m_microBlockHash, info.m_txnRootHash});
+      if (info.m_shardId == m_mediator.m_ds->m_shards.size() &&
```
KaustubhShamshery: i would prefer if(!(predicate)

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -708,11 +713,12 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
 
   DetachedFunction(1, resumeBlackList);
 
+  if (!LoadUnavailableMicroBlockHashes(
+          txBlock, txBlock.GetHeader().GetBlockNum(), toSendTxnToLookup)) {
+    return false;
+  }
+
   if (!isVacuousEpoch) {
```
KaustubhShamshery: Why bring this outside?

kaikawaliu: vacuous also load. so as for securing the completeness of coinbase info

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: !

PR #1622, opened 2019-04-24
Minor fixes for AccountStoreSC

kaikawaliu:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -842,8 +842,8 @@ bool AccountStoreSC<MAP>::ParseCreateContractJsonOutput(
     return false;
   }
   try {
-    gasRemained =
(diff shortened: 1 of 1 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -1128,6 +1128,7 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(
     if (ret) {
       if (!contractAccount->SetStorage(state_entries, temp)) {
         LOG_GENERAL(WARNING, "SetStorage failed");
+        return false;
```
vaivaswatha: Similar to this, is a `return false` needed on line 1144 as well? (GitHub doesn't let me comment on that line directly).

PR #1624, opened 2019-04-25
Limit TxBlocks on lookup

sandipbhoir:
## Description
This PR covers:
- limits number of txBlocks to be sent back by lookup.
- Also, fix verification of DirectoryBlock against upper seed list instead of lookup list.
- Multiple replacement to use `*FromSeed` instead of `*FromLookup`
- Fixes StateDelta issue introduced as side-effect of   PR#1617

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -1115,13 +1115,13 @@ bool Lookup::ProcessGetTxBlockFromSeed(const bytes& message,
     return false;
   }
 
-  vector<TxBlock> txBlocks;
-  RetrieveTxBlocks(txBlocks, lowBlockNum, highBlockNum);
-
   LOG_EPOCH(INFO, m_mediator.m_currentEpochNum,
             "ProcessGetTxBlockFromSeed requested by " << from << " for blocks "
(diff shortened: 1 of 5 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -1148,21 +1149,19 @@ void Lookup::RetrieveTxBlocks(vector<TxBlock>& txBlocks, uint64_t& lowBlockNum,
...
-                                  : 0;
-    if (lowBlockNum <= lowestLimitNum) {
-      // Limit the number of txn blocks upto last N ds epochs
-      lowBlockNum = lowestLimitNum;
-    }
+  }
+
+  uint64_t lowestLimitNum =
+      (m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() - 1) *
+      NUM_FINAL_BLOCK_PER_POW;
```
kaikawaliu: It's better to use `GetEpochNum()` in case of any change of NUM_FINAL_BLOCK_PER_ROW in the future

PR #1625, opened 2019-04-26
Reshuffle the txn-shard map after each ds epoch

KaustubhShamshery:
## Description
If the shading structure has changed , the lookup would re-shuffle the txn-shard map
Also, add unit test for the same.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3623,7 +3623,7 @@ bool Lookup::DeleteTxnShardMap(uint32_t shardId) {
   return true;
 }
 
-void Lookup::SenderTxnBatchThread() {
+void Lookup::SenderTxnBatchThread(const uint32_t oldShardSize) {
   if (!LOOKUP_NODE_MODE) {
     LOG_GENERAL(WARNING,
                 "Lookup::SenderTxnBatchThread not expected to be called from "
@@ -3638,20 +3638,17 @@ void Lookup::SenderTxnBatchThread() {
     return;
   }
 
-  auto main_func = [this]() mutable -> void {
+  auto main_func = [this, oldShardSize]() mutable -> void {
     m_startedTxnBatchThread = true;
     uint32_t numShards = 0;
     while (true) {
       if (!m_mediator.GetIsVacuousEpoch()) {
-        {
-          lock_guard<mutex> g(m_mediator.m_ds->m_mutexShards);
-          numShards = m_mediator.m_ds->m_shards.size();
-        }
+        numShards = m_mediator.GetShardNum();
         if (numShards == 0) {
           this_thread::sleep_for(chrono::milliseconds(1000));
           continue;
         }
-        SendTxnPacketToNodes(numShards);
+        SendTxnPacketToNodes(oldShardSize, numShards);
       }
       break;
     }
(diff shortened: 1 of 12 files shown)


tests/Lookup/Test_txn_send.cpp
```diff
@@ -0,0 +1,72 @@
...
+      LOG_GENERAL(WARNING, "Failed to fetch txns");
+      return;
+    }
+    for (uint j = rangeOfShards.first; j <= rangeOfShards.second; j++) {
+      for (const auto& shard : mp) {
+        for (const auto& tx : shard.second) {
+          lk.AddToTxnShardMap(tx, shard.first);
+        }
+      }
+      lk.RectifyTxnSharMap(i, j);
```
ansnunez: `RectifyTxnSharMap` is misspelled. No compile error?

ansnunez: Do you want to add the test to CMakeLists?

src/libLookup/Lookup.cpp
```diff
@@ -3672,13 +3710,22 @@ void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
     return;
   }
 
+  const uint32_t& numShards = newShardSize;
```
ansnunez: Reference not necessary

src/libLookup/Lookup.cpp
```diff
@@ -3640,20 +3640,17 @@ void Lookup::SenderTxnBatchThread() {
...
-          lock_guard<mutex> g(m_mediator.m_ds->m_mutexShards);
-          numShards = m_mediator.m_ds->m_shards.size();
-        }
+        numShards = m_mediator.GetShardNum();
         if (numShards == 0) {
           this_thread::sleep_for(chrono::milliseconds(1000));
           continue;
         }
-        SendTxnPacketToNodes(numShards);
+        SendTxnPacketToNodes(oldShardSize, numShards);
```
ansnunez: Perhaps `oldNumShards` or `numShardsPrevEpoch` would be clearer

src/libLookup/Lookup.cpp
```diff
@@ -3662,7 +3659,48 @@ void Lookup::SenderTxnBatchThread() {
   DetachedFunction(1, main_func);
 }
 
-void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
+void Lookup::RectifyTxnShardMap(const uint32_t oldShardSize,
+                                const uint32_t newShardSize) {
+  LOG_MARKER();
+
+  auto t_start = std::chrono::high_resolution_clock::now();
```
ansnunez: Should we put the performance measuring code inside preprocessor flag?

src/libLookup/Lookup.cpp
```diff
@@ -3662,7 +3659,48 @@ void Lookup::SenderTxnBatchThread() {
   DetachedFunction(1, main_func);
 }
 
-void Lookup::SendTxnPacketToNodes(uint32_t numShards) {
+void Lookup::RectifyTxnShardMap(const uint32_t oldShardSize,
```
ansnunez: For consistency with other function, might be better to say `oldNumShards` and `newNumShards`

src/libLookup/Lookup.cpp
```diff
@@ -3662,7 +3659,48 @@ void Lookup::SenderTxnBatchThread() {
...
+  LOG_GENERAL(INFO, "Shard dropped or gained, shuffling txn shard map");
+  LOG_GENERAL(INFO, "New Shard Size: " << newShardSize
+                                       << "Old Shard Size: " << oldShardSize);
+  for (const auto& shard : m_txnShardMap) {
+    if (shard.first == oldShardSize) {
+      // ds txns
+      continue;
+    }
+    for (const auto& tx : shard.second) {
+      const PubKey& senderPubKey = tx.GetSenderPubKey();
```
ansnunez: Can just call `tx.GetSenderAddr()`

src/libLookup/Lookup.cpp
```diff
@@ -3662,7 +3659,48 @@ void Lookup::SenderTxnBatchThread() {
...
+      Address fromAddr = Account::GetAddressFromPublicKey(senderPubKey);
+      unsigned int correct_shard =
+          Transaction::GetShardIndex(fromAddr, newShardSize);
+      tempTxnShardMap[correct_shard].push_back(tx);
```
ansnunez: Maybe `emplace_back` can give some performance gain

src/libMediator/Mediator.cpp
```diff
@@ -157,6 +157,12 @@ bool Mediator::GetIsVacuousEpoch(const uint64_t& epochNum) {
   return ((epochNum + NUM_VACUOUS_EPOCHS) % NUM_FINAL_BLOCK_PER_POW) == 0;
 }
 
+uint32_t Mediator::GetShardNum() const {
```
ansnunez: Probably better to move this function to DS?  A bit weird also for `Mediator` to access a mutex owned by `DS`

ansnunez: Also, between `shardSize`, `shardNum`, and `numShards`, I think the last one is more easily understood

tests/Lookup/Test_txn_send.cpp
KaustubhShamshery: No these were deliberately kept out. as this one will slow the tests down.

src/libLookup/Lookup.cpp
KaustubhShamshery: This process should not be invoked very frequently and only on lookup, so it would be nice if we could have real time stats

tests/Lookup/Test_txn_send.cpp
kaikawaliu: It's fine to slow by several seconds or you can reduce the num of txn in total

KaustubhShamshery: there is also no check for correctness, need to manually verify

src/libLookup/Lookup.cpp
```diff
@@ -3662,7 +3659,48 @@ void Lookup::SenderTxnBatchThread() {
...
+  for (const auto& shard : m_txnShardMap) {
+    if (shard.first == oldShardSize) {
+      // ds txns
+      continue;
+    }
+    for (const auto& tx : shard.second) {
+      const PubKey& senderPubKey = tx.GetSenderPubKey();
+      Address fromAddr = Account::GetAddressFromPublicKey(senderPubKey);
+      unsigned int correct_shard =
+          Transaction::GetShardIndex(fromAddr, newShardSize);
```
kaikawaliu: It would be much preferred if you can keep the shard deciding logic in a reusable API. So whenever you try to get the shard to assign to just use that API instead of drafting the algo again. Which serves for consistency among the code and also easier for future maintenance

KaustubhShamshery: Okay will do in separate PR

PR #1628, opened 2019-04-29
Consolidate shard calculation

KaustubhShamshery:
## Description
Construct a function for reusing the code for calculating the shard index.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


bb111189:
To only merge after V4.6 is released

PR #1631, opened 2019-04-30
Daemon support for community seed

sandipbhoir:
## Description
This PR will cover core part of issue: https://github.com/Zilliqa/Issues/issues/501

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/daemon/ZilliqaDaemon_AWS.cpp
+++ b/daemon/ZilliqaDaemon_AWS.cpp
@@ -42,7 +42,6 @@ namespace po = boost::program_options;
 
 #define SUCCESS 0
 #define ERROR_IN_COMMAND_LINE -1
-#define ERROR_UNHANDLED_EXCEPTION -2
 
 const vector<string> programName = {"zilliqa"};
 const string restart_zilliqa =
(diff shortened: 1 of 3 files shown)


daemon/ZilliqaDaemon_AWS.cpp
```diff
@@ -25,16 +25,24 @@
...
+#include <thread>
 #include <unordered_map>
 #include <vector>
 
 using namespace std;
+namespace po = boost::program_options;
+
+#define SUCCESS 0
+#define ERROR_IN_COMMAND_LINE -1
+#define ERROR_UNHANDLED_EXCEPTION -2
```
ckyang: Seems `ERROR_UNHANDLED_EXCEPTION` never be used. Is it exception need to be handled?

KaustubhShamshery: Prefer inline constexpr functions to function macros. (from coding guideline)

sandipbhoir: Yeah. Fixed.

KaustubhShamshery: still see macros.

PR #1633, opened 2019-05-02
Persistence Validation Script

KaustubhShamshery:
## Description
For usage in testnet ( in /run/zilliqa ) : `bash /zilliqa/scripts/validatePersistence.sh` 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -307,10 +307,8 @@ void Node::AddGenesisInfo(SyncType syncType) {
   }
 }
 
-bool Node::ValidateDB() {
+bool Node::CheckIntegrity() {
   DequeOfNode dsComm;
-  const string lookupIp = "127.0.0.1";
-  const unsigned int port = SEED_PORT;
 
   for (const auto& dsKey : *m_mediator.m_initialDSCommittee) {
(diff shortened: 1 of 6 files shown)


src/libNode/Node.cpp
```diff
@@ -384,8 +382,7 @@ bool Node::ValidateDB() {
       dirBlocks.emplace_back(*fallbackwshardingstruct);
     }
   }
-
-  if (!m_mediator.m_validator->CheckDirBlocks(dirBlocks, dsComm, 0, dsComm)) {
+  if (!m_mediator.m_validator->CheckDirBlocks(dirBlocks, dsComm, 1, dsComm)) {
```
ansnunez: @KaustubhShamshery what does the change in index from 0 to 1 mean?

ansnunez: @KaustubhShamshery what does the change in index from 0 to 1 mean?

KaustubhShamshery: The genesis blocks are synthesized at beginning of the binary, so no need to verify

PR #1638, opened 2019-05-07
Fix remap for contract call txns

KaustubhShamshery:
## Description
- Send the unmatched contract call txn to ds

## Backward Compatibility

## Review Suggestion
[After #1628 ]

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3676,8 +3676,20 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
       continue;
     }
     for (const auto& tx : shard.second) {
-      unsigned int correct_shard = tx.GetShardIndex(newNumShards);
-      tempTxnShardMap[correct_shard].emplace_back(tx);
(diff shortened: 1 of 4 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3658,8 +3658,20 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
       continue;
     }
     for (const auto& tx : shard.second) {
-      unsigned int correct_shard = tx.GetShardIndex(newNumShards);
-      tempTxnShardMap[correct_shard].emplace_back(tx);
+      unsigned int fromShard = tx.GetShardIndex(newNumShards);
```
ansnunez: @KaustubhShamshery For the case of the old DS shard (few lines above this), should we skip the transactions there, or do another check to see if the `toShard` and `fromShard` for those don't match?

KaustubhShamshery: @ansnunez they might have priority set to true. so sending them to new ds is appropriate,

PR #1641, opened 2019-05-07
Properer handling for failures of BlockStorage

kaikawaliu:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -2006,8 +2006,11 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
     // Store Tx Block to disk
     bytes serializedTxBlock;
     txBlock.Serialize(serializedTxBlock, 0);
-    BlockStorage::GetBlockStorage().PutTxBlock(
-        txBlock.GetHeader().GetBlockNum(), serializedTxBlock);
+    if (!BlockStorage::GetBlockStorage().PutTxBlock(
+            txBlock.GetHeader().GetBlockNum(), serializedTxBlock)) {
+      LOG_GENERAL(WARNING, "BlockStorage::PutTxBlock failed");
+      return;
+    }
   }
 
   m_mediator.m_currentEpochNum =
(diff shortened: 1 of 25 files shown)


ckyang:
I've launched a small-scale testnet with `v4.6.1` (current mainnet) and try to rolling-upgrade dsguard using `912ab7d` in this PR. It works fine, thus this PR looks good to me for not being a breaking change.

src/libLookup/Lookup.cpp
```diff
@@ -2006,8 +2006,11 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
     // Store Tx Block to disk
     bytes serializedTxBlock;
     txBlock.Serialize(serializedTxBlock, 0);
-    BlockStorage::GetBlockStorage().PutTxBlock(
-        txBlock.GetHeader().GetBlockNum(), serializedTxBlock);
+    if (!BlockStorage::GetBlockStorage().PutTxBlock(
+            txBlock.GetHeader().GetBlockNum(), serializedTxBlock)) {
+      LOG_GENERAL(WARNING, "BlockStorage::PutTxBlock failed " << txBlock);
```
ckyang: How about move the error message into `PutTxBlock` then we can handle the debug message in one place for easier maintenance. For caller here, it would be simpler to check the function result only.

src/libNode/FallbackBlockProcessing.cpp
```diff
@@ -292,8 +293,12 @@ bool Node::ProcessFallbackBlock(const bytes& message, unsigned int cur_offset,
         LOG_GENERAL(WARNING, "MoveUpdatesToDisk failed, what to do?");
         return;
       }
-      BlockStorage::GetBlockStorage().PutMetadata(MetaType::DSINCOMPLETED,
-                                                  {'0'});
+      if (!BlockStorage::GetBlockStorage().PutMetadata(MetaType::DSINCOMPLETED,
```
ckyang: How about move the error message into `PutMetadata` then we can handle the debug message in one place for easier maintenance. For caller here, it would be simpler to check the function result only.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -725,7 +727,10 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
     // Remove because shard nodes will be shuffled in next epoch.
     CleanMicroblockConsensusBuffer();
 
-    StoreFinalBlock(txBlock);
+    if (!StoreFinalBlock(txBlock)) {
+      LOG_GENERAL(WARNING, "StoreFinalBlock failed!");
```
ckyang: How about move the error message into `StoreFinalBlock` then we can handle the debug message in one place for easier maintenance. For caller here, it would be simpler to check the function result only.

src/libPersistence/Retriever.cpp
```diff
@@ -138,9 +145,17 @@ bool Retriever::RetrieveTxBlocks(bool trimIncompletedBlocks) {
...
+          if (!AccountStore::GetInstance().MoveUpdatesToDisk()) {
+            LOG_GENERAL(WARNING, "AccountStore::MoveUpdatesToDisk failed");
+            return false;
+            ;
+          }
           // clear the stateDelta db
-          BlockStorage::GetBlockStorage().ResetDB(BlockStorage::STATE_DELTA);
+          if (!BlockStorage::GetBlockStorage().ResetDB(
+                  BlockStorage::STATE_DELTA)) {
+            LOG_GENERAL(WARNING, "BlockStorage::ResetDB (STATE_DELTA) failed");
```
ckyang: How about move the error message into `ResetDB` then we can handle the debug message in one place for easier maintenance. For caller here, it would be simpler to check the function result only.

src/libLookup/Lookup.cpp
ckyang: Yes it make sense to let user know who calls this function. But the point what I mentioned is, the responsibility of telling user what's wrong should be the `callee` instead of `caller`. It's no harm that `caller` give extra information like here, but the `callee` should always takes the responsibility to talk enough about what's going on inside itself. Anyway it should not a roadblock for this PR so I suggest to raise another issue for the log refactoring.

PR #1642, opened 2019-05-07
Add exhaustive test for SafeMath

KaustubhShamshery:
## Description
Check each and every combination of given type and examine if the underflow and overflow is handled correctly. 

Also, throw exception when power fails 
```
terminate called after throwing an instance of 'std::runtime_error'
what():  [Critical] SafeMath::power failed
```

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/SafeMath.tpp
+++ b/src/libUtils/SafeMath.tpp
@@ -122,6 +122,9 @@ T SafeMath<T>::power(const T& base, const T& exponent, bool isCritical) {
   if (!SafeMath::power_core(base, exponent, ret)) {
     LOG_GENERAL(isCritical ? FATAL : WARNING,
                 "SafeMath::power failed ret: " << ret << " base " << base);
+    if (isCritical) {
+      throw std::runtime_error("[Critical] SafeMath::power failed");
+    }
     return ret;
   }
   return ret;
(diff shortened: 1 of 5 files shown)


src/libUtils/SafeMath.tpp
```diff
@@ -118,26 +118,12 @@ bool SafeMath<T>::power(const T& base, const T& exponent, T& result) {
...
-  while (count > 0) {
-    if (!SafeMath::mul(ret, base, ret)) {
-      LOG_GENERAL(isCritical ? FATAL : WARNING,
-                  "SafeMath::pow failed ret: " << ret << " base " << base);
-      return ret;
-    }
-    --count;
+  T ret{};
+  if (!SafeMath::power_core(base, exponent, ret)) {
+    LOG_GENERAL(isCritical ? FATAL : WARNING,
```
ShengguangXiao: When will be isCritical enabled?

KaustubhShamshery: This is from previous version of code, if u see, I also dont know why it is there.

ShengguangXiao: I think can be removed, all the other safe math function is log warning

Issue #1643, opened 2019-05-08
Support customized log path


Gnnng:
The log files are directly written into the working directory where the process is running. There's a need to store the logs into a separate directory, when

1. A system-wide log path is preferred, e.g., `/var/log`
2. A user would like to use sidecar container to rotate the log.

To avoid introducing breaking change when supporting this, we can keep the current behavior as the default while allowing a user to set a customized path.

bb111189:
Already implemented. Issue closed

PR #1651, opened 2019-05-09
Support customized log path

ckyang:
## Description

As https://github.com/Zilliqa/Zilliqa/issues/1643 described, we should support the customized log path, in order to support the non-root ziliqa in the future.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/cmd/main.cpp
+++ b/src/cmd/main.cpp
@@ -49,6 +49,7 @@ int main(int argc, const char* argv[]) {
     PrivKey privkey;
     PubKey pubkey;
     string address;
+    string logpath("./");
     int port = -1;
     unique_ptr<NAT> nt;
     uint128_t ip;
@@ -70,7 +71,8 @@ int main(int argc, const char* argv[]) {
         "Specifies port to bind to, if not specified in address")(
         "loadconfig,l", "Loads configuration if set (deprecated)")(
         "synctype,s", po::value<unsigned int>(&synctype), synctype_descr)(
-        "recovery,r", "Runs in recovery mode if set");
+        "recovery,r", "Runs in recovery mode if set")(
+        "logpath,g", po::value<string>(&logpath), "customized log path");
 
     po::variables_map vm;
     try {
@@ -132,9 +134,9 @@ int main(int argc, const char* argv[]) {
       return ERROR_IN_COMMAND_LINE;
     }
 
-    INIT_FILE_LOGGER("zilliqa");
-    INIT_STATE_LOGGER("state");
-    INIT_EPOCHINFO_LOGGER("epochinfo");
+    INIT_FILE_LOGGER("zilliqa", logpath.c_str());
+    INIT_STATE_LOGGER("state", logpath.c_str());
+    INIT_EPOCHINFO_LOGGER("epochinfo", logpath.c_str());
 
     LOG_GENERAL(INFO, ZILLIQA_BRAND);
 
--- a/src/cmd/main.cpp
+++ b/src/cmd/main.cpp
@@ -49,7 +49,7 @@ int main(int argc, const char* argv[]) {
     PrivKey privkey;
     PubKey pubkey;
     string address;
-    string logpath("./");
+    string logpath(boost::filesystem::absolute("./").string());
     int port = -1;
     unique_ptr<NAT> nt;
     uint128_t ip;
--- a/src/cmd/main.cpp
+++ b/src/cmd/main.cpp
@@ -72,7 +72,9 @@ int main(int argc, const char* argv[]) {
         "loadconfig,l", "Loads configuration if set (deprecated)")(
         "synctype,s", po::value<unsigned int>(&synctype), synctype_descr)(
         "recovery,r", "Runs in recovery mode if set")(
-        "logpath,g", po::value<string>(&logpath), "customized log path");
+        "logpath,g", po::value<string>(&logpath),
+        "customized log path, could be relative path (ex. \"./logs/\"), or "
+        "absolute path (ex. \"/usr/local/test/logs/\")");
 
     po::variables_map vm;
     try {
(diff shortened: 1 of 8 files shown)


constants.xml
```diff
@@ -10,6 +10,7 @@
         <!-- testnet -->
         <GENESIS_PUBKEY>03B70CF2ABEAE4E86DAEF1A36243E44CD61138B89055099C0D220B58FB86FF588A</GENESIS_PUBKEY>
         <UPGRADE_TARGET_DS_NUM>1</UPGRADE_TARGET_DS_NUM>
+        <LOG_PATH>./</LOG_PATH>
```
Gnnng: can we make a command-line option?

ckyang: ok, will do that

src/libUtils/Logger.cpp
```diff
@@ -81,17 +83,21 @@ void Logger::newLog() {
            m_seqNum);
   m_fileName = m_fileNamePrefix + buf;
 
+  if (!boost::filesystem::exists(LOG_PATH)) {
```
ansnunez: I think we can move this to the constructor, so that it doesn't have to be checked every time the log rolls over to a new file.

ckyang: ok

src/libUtils/Logger.cpp
```diff
@@ -52,6 +54,10 @@ Logger::Logger(const char* prefix, bool log_to_file, streampos max_file_size) {
   this->m_logToFile = log_to_file;
   this->m_maxFileSize = max_file_size;
 
+  if (!boost::filesystem::exists(LOG_PATH)) {
```
chitowncrispy: This check shouldn't be needed since `boost::filesystem::create_directory` already does a check to see if the path already exists. This is from the documentation of the method "Creation failure because p resolves to an existing directory shall not be treated as an error.". Thus if the path already exists `boost::filesystem::create_directory` will return false and no error will be reported.

ckyang: Nice catch, I'll remove the `exists` checking here. Thanks!

src/libUtils/Logger.h
```diff
@@ -160,25 +164,26 @@ class ScopeMarker {
...
-        << PAD(get_ms(cur), 3, '0') << " ]" << msg;                   \
-    Logger::GetStateLogger(NULL, true).LogState(oss.str().c_str());   \
+#define LOG_STATE(msg)                                                    \
+  {                                                                       \
+    std::ostringstream oss;                                               \
+    auto cur = std::chrono::system_clock::now();                          \
+    auto cur_time_t = std::chrono::system_clock::to_time_t(cur);          \
+    oss << "[ " << std::put_time(gmtime(&cur_time_t), "%y-%m-%dT%T.")     \
+        << PAD(get_ms(cur), 3, '0') << " ]" << msg;                       \
+    Logger::GetStateLogger(NULL, true, "./").LogState(oss.str().c_str()); \
```
chitowncrispy: All of these loggers that pass in `"./"` will be logging to a different path than the one passed in as an argument it seems. Is that the desired functionality or do you want all loggers to be logging to the custom path?

ckyang: Actually inside `GetStateLogger()` there is a static construction of `Logger` (singleton), while the fist construction call of `Logger` would be called by `INIT_STATE_LOGGER` within customized log path beforehand. The reason we still put `./` here is just in case the false usage forgetting to call `INIT_STATE_LOGGER` beforehand, then the default log path `./` should be used instead.

src/libUtils/Logger.cpp
```diff
@@ -48,9 +49,13 @@ inline pid_t getCurrentPid() {
...
     1024 * 1024 * 100;  // 100MB per log file
 
-Logger::Logger(const char* prefix, bool log_to_file, streampos max_file_size) {
+Logger::Logger(const char* prefix, bool log_to_file, const char* logpath,
+               streampos max_file_size) {
   this->m_logToFile = log_to_file;
   this->m_maxFileSize = max_file_size;
+  this->m_logPath = logpath;
+
+  boost::filesystem::create_directory(logpath);
```
Gnnng: The directory creation could fail so some error handling multiple be needed here.

ckyang:
Need to support both absolute path and relative path in the these filepath options.

src/cmd/main.cpp
```diff
@@ -70,7 +71,8 @@ int main(int argc, const char* argv[]) {
         "Specifies port to bind to, if not specified in address")(
         "loadconfig,l", "Loads configuration if set (deprecated)")(
         "synctype,s", po::value<unsigned int>(&synctype), synctype_descr)(
-        "recovery,r", "Runs in recovery mode if set");
+        "recovery,r", "Runs in recovery mode if set")(
+        "logpath,g", po::value<string>(&logpath), "customized log path");
```
Gnnng: Can we make it clear in the help message what kinds of paths are supported (relative or absolute or both)

src/cmd/main.cpp
```diff
@@ -70,7 +71,10 @@ int main(int argc, const char* argv[]) {
         "Specifies port to bind to, if not specified in address")(
         "loadconfig,l", "Loads configuration if set (deprecated)")(
         "synctype,s", po::value<unsigned int>(&synctype), synctype_descr)(
-        "recovery,r", "Runs in recovery mode if set");
+        "recovery,r", "Runs in recovery mode if set")(
+        "logpath,g", po::value<string>(&logpath),
+        "customized log path, could be relative path (ex. \"./logs/\"), or "
```
Gnnng:
```suggestion
        "customized log path, could be relative path (e.g., \"./logs/\"), or "
```

src/cmd/main.cpp
```diff
@@ -70,7 +71,10 @@ int main(int argc, const char* argv[]) {
...
+        "absolute path (ex. \"/usr/local/test/logs/\")");
```
Gnnng:
```suggestion
        "absolute path (e.g., \"/usr/local/test/logs/\")");
```

PR #1656, opened 2019-05-10
Use EPOCHFIN to replace redundant recovery related conditions

kaikawaliu:
## Description

Previously we are using `DSINCOMPLETED` or `LATEST_EPOCH_STATES_UPDATED` for different purpose, which are quite difficult to manage.
By introducing the `EPOCHFIN` which represents the real complement of an epoch, which stands for:
-- Both the states (if in vacuous epoch) and microblocks forwarding (for every ordinary epoch) were completely processed from the `lookup's perspective`.
Thus we can detach from the jumble declaration of the two constants mentioned above.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libPersistence/Retriever.cpp
```diff
@@ -236,11 +236,25 @@ bool Retriever::RetrieveBlockLink(bool trimIncompletedBlocks) {
...
+    }
+    if (isDSIncompleted[0] == '1') {
+      /// Removing incompleted DS for upgrading protocol
+      /// Keeping incompleted DS for node recovery
+      if (trimIncompletedBlocks) {
+        LOG_GENERAL(INFO, "Has incompleted DS Block, remove it");
+        toDelete = true;
+      }
+    }
   }
```
sandipbhoir: do we need to handle `else` here, return false?

kaikawaliu: If `else` then we will just ignore, no special action needed

PR #1657, opened 2019-05-10
Standardize Include Guard Check

KaustubhShamshery:
## Description
Standardize header guards.

Reference : https://google.github.io/styleguide/cppguide.html#The__define_Guard

- For our case, it is `ZILLIQA_<path_to_file>_H_`

- to generate header guard put, 
```
#ifndef A_H_ 
#define A_H_
....
.... 
#endif  // A_H_
```  
and run `./scripts/replace_header_guard.py` and the statements will be replaced by correct header guard

## Backward Compatibility

### Implementation

- [x] Check License 
- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,11 +44,11 @@ addons:
     - clang-format-5.0
     - clang-tidy-5.0
     - curl
+    - gawk
     - git
     - lcov
     - libxml2-utils
     - python-pip
-    - gawk
 
 script:
   - ./scripts/ci_build.sh;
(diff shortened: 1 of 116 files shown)


.travis.yml
```diff
@@ -48,6 +48,7 @@ addons:
     - lcov
     - libxml2-utils
     - python-pip
+    - gawk
```
Gnnng: mind the order of the names. check the comment above.

PR #1658, opened 2019-05-10
Handle connection timeout

sandipbhoir:
## Description
This PR addresses issue - https://github.com/Zilliqa/Issues/issues/439

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test ( 33 node )
- [x] large-scale cloud test ( 1200 nodes )


kaikawaliu:
How about using epoll in our case, I feel it may bring better performance for our use case. Ref: https://devarea.com/linux-io-multiplexing-select-vs-poll-vs-epoll/#.XNj-z3VuZhE

kaikawaliu:
and also https://jvns.ca/blog/2017/06/03/async-io-on-linux--select--poll--and-epoll/

sandipbhoir:
epoll may not fit in this context ( for solving connection timeout of individual client connection )
epoll will be useful when we create multiple client connection before hand and likely to wait on it until fd are ready or timeout.

So i think poll fits here well because with current design we open new connection when we have something to be sent to specific client.

Unless, we redesign complete networking to have open connection to all peers in the list which has its own challenge because sending port has to be dynamic since we need multiple connection between same sender and receiver in parallel.

bb111189:
While this will prevent incoming connection from fully consume all FD and unable to get new FD, I think another angle we want to look at leaves some buffer for FD, else other parts of Zilliqa may not work as intended (eg. reading of JSON from smart contract, writing to log file). 

Here is what I found for checking the FD limit . 

> For the current process count, you can use getrlimit to get the file descriptor limit, then iterate over all integers from 0 to that limit and try calling fcntl with the F_GETFD command. It will succeed only on the file descriptors which are actually open, letting you count them.


https://stackoverflow.com/questions/7976769/getting-count-of-current-used-file-descriptors-from-c-code

sandipbhoir:
Bit confused. I don't see this PR preventing any fd's used by incoming connection. It is only for outgoing connection.

bb111189:
> While this will prevent incoming connection from fully consume all FD and unable to get new FD, I think another angle we want to look at leaves some buffer for FD, else other parts of Zilliqa may not work as intended (eg. reading of JSON from smart contract, writing to log file).
> 
> Here is what I found for checking the FD limit .
> 
> > For the current process count, you can use getrlimit to get the file descriptor limit, then iterate over all integers from 0 to that limit and try calling fcntl with the F_GETFD command. It will succeed only on the file descriptors which are actually open, letting you count them.
> 
> https://stackoverflow.com/questions/7976769/getting-count-of-current-used-file-descriptors-from-c-code

Resolved. sorry for the confusion

src/libNetwork/P2PComm.cpp
```diff
@@ -204,22 +209,93 @@ bool SendJob::SendMessageSocketCore(const Peer& peer, const bytes& message,
...
+    }
+    arg |= O_NONBLOCK;
+    if (fcntl(cli_sock, F_SETFL, arg) < 0) {
+      LOG_GENERAL(WARNING, "couldn't set socket as non-blocking");
+      return false;
+    }
+
+    int status;
+    bool connectStat = true;
+    if ((status = connect(cli_sock, (struct sockaddr*)&serv_addr,
```
ansnunez: It looks like if `connect()` fails then `connectStat` remains `true`.  Is the flow for that case OK?

PR #1660, opened 2019-05-13
Fix some uninitialized params & Remove explicit copy constructor for std::move 

kaikawaliu:
## Description

- [x] Fix some uninitialized parameters.
- [x] Solving issue https://github.com/Zilliqa/Issues/issues/511 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.h
+++ b/src/libData/AccountData/AccountStoreSC.h
@@ -62,26 +62,26 @@ class AccountStoreSC : public AccountStoreBase<MAP> {
   Address m_curSenderAddr;
 
   /// the transfer amount while executing each txn
-  uint128_t m_curAmount;
+  uint128_t m_curAmount{0};
 
   /// the gas limit while executing each txn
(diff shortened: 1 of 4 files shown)


KaustubhShamshery:
Can u use `{}` instead for uniform initialization syntax

src/libData/AccountData/AccountStoreSC.h
```diff
@@ -53,7 +53,7 @@ class AccountStoreSC : public AccountStoreBase<MAP> {
   std::mutex m_mutexUpdateAccounts;
 
   /// the blocknum while executing each txn
-  uint64_t m_curBlockNum;
+  uint64_t m_curBlockNum = 0;
```
chitowncrispy: Is there a reason why this assignment wasn't changed to be uniform for initialization?

KaustubhShamshery: Nice catch, thanks

PR #1666, opened 2019-05-22
Restore.

KaustubhShamshery:
## Description
binary to restore the zilliqa states

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/cmd/restore.cpp
```diff
@@ -0,0 +1,398 @@
...
+    cout << "Failed to put ds committee" << endl;
+    return false;
+  }
+  return true;
+}
+
+bool PutStateDeltaInLocalPersistence(uint32_t lastBlockNum,
+                                     const list<TxBlockSharedPtr>& blocks) {
+  unsigned int extra_txblocks = 0;
+  if ((lastBlockNum - extra_txblocks + 1) %
```
sandipbhoir: may be get rid of `extra_txblocks` since its always 0 in this context

PR #1669, opened 2019-05-23
[Release 4.6] Custom storage path

sandipbhoir:
## Description
This PR addresses issue https://github.com/Zilliqa/Issues/issues/515

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/depends/libDatabase/LevelDB.cpp
```diff
@@ -88,7 +88,7 @@ LevelDB::LevelDB(const std::string & dbName, const std::string& subdirectory, bo
     // that might not be the current directory (case when 'diagnostic' is true).
     // Its default value is false, and the non-diagnostic path is preserved
     // from the original code.
-    string db_path = diagnostic ? (m_subdirectory + PERSISTENCE_PATH) : ("./" + PERSISTENCE_PATH + (m_subdirectory.empty() ? "" : "/" + m_subdirectory));
+    string db_path = diagnostic ? (m_subdirectory + "/persistence") : (STORAGE_PATH + "/persistence" + (m_subdirectory.empty() ? "" : "/" + m_subdirectory));
```
chitowncrispy: Would it make more sense to add `const std::string PERSISTENCE_PATH = "/persistence";` into this file so that way any edits to the path at which the DB is located can be easily updated?

src/libPersistence/Retriever.cpp
```diff
@@ -152,11 +153,12 @@ bool Retriever::RetrieveTxBlocks(bool trimIncompletedBlocks) {
     }
   }
 
-  if (boost::filesystem::exists("StateDeltaFromS3")) {
+  if (boost::filesystem::exists(STORAGE_PATH + "/StateDeltaFromS3")) {
```
chitowncrispy: Again I would recommend adding `const std::string STATE_DELTA_S3_PATH = "/StateDeltaFromS3";` to this file so the path can be easily updated at a later date.

src/depends/libDatabase/LevelDB.cpp
sandipbhoir: yeah. will add it.

src/depends/libDatabase/LevelDB.cpp
```diff
@@ -88,7 +88,7 @@ LevelDB::LevelDB(const std::string & dbName, const std::string& subdirectory, bo
     // that might not be the current directory (case when 'diagnostic' is true).
     // Its default value is false, and the non-diagnostic path is preserved
     // from the original code.
-    string db_path = diagnostic ? (m_subdirectory + PERSISTENCE_PATH) : ("./" + PERSISTENCE_PATH + (m_subdirectory.empty() ? "" : "/" + m_subdirectory));
+    string db_path = diagnostic ? (m_subdirectory + PERSISTENCE_PATH) : (STORAGE_PATH + "/" + PERSISTENCE_PATH + (m_subdirectory.empty() ? "" : "/" + m_subdirectory));
```
chitowncrispy: I believe this line has an extra `/` in it when resolving the path. `(STORAGE_PATH + "/" + PERSISTENCE_PATH` should be changed to `(STORAGE_PATH + PERSISTENCE_PATH` as far as I can tell. Please check and test that.

PR #1672, opened 2019-05-24
Custom storage path

sandipbhoir:
## Description
This PR addresses issue Zilliqa/Issues#515

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/depends/libDatabase/LevelDB.cpp
```diff
@@ -427,7 +427,7 @@ bool LevelDB::RefreshDB()
 
     leveldb::DB* db;
 
-    leveldb::Status status = leveldb::DB::Open(options, "./" + PERSISTENCE_PATH + "/" + this->m_dbName, &db);
+    leveldb::Status status = leveldb::DB::Open(options, STORAGE_PATH + PERSISTENCE_PATH + "/" + this->m_dbName, &db);
```
KaustubhShamshery: wouldn't STORAGE_PATH + PERSISTENCE_PATH evaluate to `.//persistence`

sandipbhoir: Yes.. but doesn’t matter.

KaustubhShamshery: I think need to fix a convention atleast, put it in only one.

sandipbhoir: Intentionally have it that ways. Storage path is any path user want to have.
And persistence path is more of constant if we may want to change in future.

KaustubhShamshery: i see no need of `/` in the storage path. Either way cannot have it in both.

sandipbhoir: Storage path is also accepted from user via launch.sh.
They can give with one ending with / or without /
And should work in either case. So better to have PERSISTENCE_PATH starting with /

KaustubhShamshery: OK sure, then can the `STORAGE_PATH` be without the `\` ?

sandipbhoir: I see.. you want to set <STORAGE_PATH> to be . 
Not sure existing function SetSeedNodeHistoricalDB still will work in that case.

sandipbhoir: I will test it later.

chitowncrispy: The best solution is to have the custom paths either always end or start with a `/` but never both. As long as that can be made uniform then a change like this will work without adding extra `/`s to any of the paths.

src/libPersistence/Retriever.cpp
```diff
@@ -167,11 +168,12 @@ bool Retriever::RetrieveTxBlocks(bool trimIncompletedBlocks) {
     }
   }
 
-  if (boost::filesystem::exists("StateDeltaFromS3")) {
+  if (boost::filesystem::exists(STORAGE_PATH + "/StateDeltaFromS3")) {
```
chitowncrispy: Can the constant `STATEDELTAFROMS3_PATH` be used here?

sandipbhoir: Fixed

constants_local.xml
```diff
@@ -9,6 +9,8 @@
         <CHAIN_ID>2</CHAIN_ID>
         <GENESIS_PUBKEY>02AAE728127EB5A30B07D798D5236251808AD2C8BA3F18B230449D0C938969B552</GENESIS_PUBKEY>
         <UPGRADE_TARGET_DS_NUM>1</UPGRADE_TARGET_DS_NUM>
+        <!-- Do not add a trailing "/" in the path-->
```
Gnnng: @sandipbhoir Is this warning for users? Unfortunately, all the comments got stripped out when the file was processed before it's published on joining URLs. If this is important, we may need some checks in the code instead of a warning in the constants.xml

PR #1674, opened 2019-05-28
Add proto3 dependency

kaikawaliu:
## Backward Compatibility

### Implementation

- [x] **ready for review**

~- [ ] local machine test~

- [x] small-scale cloud test

### DIFF
--- a/src/libPersistence/ScillaMessage.proto
+++ b/src/libPersistence/ScillaMessage.proto
@@ -1,0 +1,19 @@
+syntax = "proto3";
+
+message ProtoScillaVal {
+	message Map {
+		map<string, ProtoScillaVal> m = 1;
+	}
+
+    oneof ValType {
+        bytes bval = 1;
+        Map mval = 2;
+    }
+}
+
+message ProtoScillaQuery {
+    bytes address =1;
+    string name = 2;
+    uint32 mapdepth = 3;
+    repeated bytes indices = 4;
+}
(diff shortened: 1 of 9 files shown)


src/libPersistence/ScillaMessage.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+message ProtoScillaVal {
+	message Map {
+		map<string, ProtoScillaVal> m = 1;
```
kaikawaliu: bytes cannot be used as key in proto map, only string can.

Gnnng:
Several pieces of information seem to be missing:
* Install instructions: How a user can install the new proto3? (users cannot directly run `scripts/download_proto3.sh` because some steps there requires root permissions but running the entire script in root is not safe)
* Compatibility clarification: Will installing proto3 be conflicting with existing proto2 on the user machine?

Also, please kindly check all the places where the apt packages need to be updated.

ansnunez:
> Several pieces of information seem to be missing:
> 
>     * Install instructions: How a user can install the new proto3? (users cannot directly run `scripts/download_proto3.sh` because some steps there requires root permissions but running the entire script in root is not safe)
> 
>     * Compatibility clarification: Will installing proto3 be conflicting with existing proto2 on the user machine?
> 
> 
> Also, please kindly check all the places where the apt packages need to be updated.

@Gnnng for Install instructions, will it suffice to add the call to our README after the other root-level commands? https://github.com/Zilliqa/Zilliqa#build-dependencies

ansnunez:
Fixed in #1778

PR #1676, opened 2019-05-29
Remove redundant pow test case

ShengguangXiao:
## Description

Remove redundant pow test case

Some time this test case will fail on travis.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/POW/test_POW.cpp
+++ b/tests/POW/test_POW.cpp
@@ -395,29 +395,6 @@ BOOST_AUTO_TEST_CASE(mining_and_verification_full) {
   BOOST_REQUIRE(!verifyWinningNonce);
 }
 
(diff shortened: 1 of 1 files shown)


tests/POW/test_POW.cpp
```diff
@@ -395,29 +395,6 @@ BOOST_AUTO_TEST_CASE(mining_and_verification_full) {
   BOOST_REQUIRE(!verifyWinningNonce);
 }
 
-BOOST_AUTO_TEST_CASE(mining_low_diffculty_time_out) {
```
chitowncrispy: Which test is this a duplicate of?

ShengguangXiao: It is to test the time out mechanism, the mining_high_diffculty_time_out test case is enough

PR #1677, opened 2019-05-29
Stop lookup api server when it is synchronizing

ShengguangXiao:
## Description

Stop lookup api server when it is synchronizing

https://github.com/Zilliqa/Issues/issues/519

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -2052,7 +2052,12 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
       if (!m_currDSExpired) {
         SetSyncType(SyncType::NO_SYNC);
         m_isFirstLoop = true;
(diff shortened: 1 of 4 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3118,6 +3125,9 @@ void Lookup::RejoinAsNewLookup() {
 
   LOG_MARKER();
   if (m_mediator.m_lookup->GetSyncType() == SyncType::NO_SYNC) {
+    m_lookupServer->StopListening();
+    LOG_GENERAL(INFO, "API Server stopped listen for syncing");
```
chitowncrispy: Would `"API Server stopped listening to sync client"` be more descriptive?

PR #1679, opened 2019-05-29
Fetch constant file from dsguard instead of normal node

ckyang:
## Description

`release.sh` will fetch constant files from running testnet. For dsguard/normal nodes, currently we fetch from `normal-0`. It would be safer to fetch from `dsguard-0` instead of `normal-0`, especially now `normal-0` ~ `normal-179` are not available anymore.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


chitowncrispy:
What is the difference between copying from a dsguard node and a normal node?

PR #1682, opened 2019-06-04
Check Lookup Server for default block

KaustubhShamshery:
## Description
Check the listing apis for default block.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -1052,6 +1052,10 @@ Json::Value LookupServer::DSBlockListing(unsigned int page) {
 
   uint maxPages = (currBlockNum / PAGE_SIZE) + 1;
 
+  if (currBlockNum == INIT_BLOCK_NUMBER) {
+    throw JsonRpcException(RPC_IN_WARMUP, "No DS Blocks");
+  }
+
   _json["maxPages"] = maxPages;
 
   lock_guard<mutex> g(m_mutexDSBlockCache);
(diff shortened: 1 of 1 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -1143,6 +1147,10 @@ Json::Value LookupServer::TxBlockListing(unsigned int page) {
       m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum();
   Json::Value _json;
 
+  if (currBlockNum == INIT_BLOCK_NUMBER) {
+    throw JsonRpcException(RPC_IN_WARMUP, "No tx blocks");
```
Gnnng: Not sure if it should be capitalized.
```suggestion
    throw JsonRpcException(RPC_IN_WARMUP, "No TX blocks");
```

KaustubhShamshery: Think `Tx` would be better, thanks

src/libServer/LookupServer.cpp
```diff
@@ -1052,6 +1059,10 @@ Json::Value LookupServer::DSBlockListing(unsigned int page) {
 
   uint maxPages = (currBlockNum / PAGE_SIZE) + 1;
 
+  if (currBlockNum == INIT_BLOCK_NUMBER) {
+    throw JsonRpcException(RPC_IN_WARMUP, "No DS Blocks");
```
ansnunez: `Blocks` here vs `blocks` elsewhere

Issue #1681, opened 2019-06-03
Repeat failed HTTP request during PoW window


Snuff:
Hello,

I have plenty of nodes configured to mine remotely through a proxy. Currently, If HTTP request to the proxy fails, due to any reasons, PoW round will be stopped. That leads to losing valid shares, mining revenue, and network instability. Here I want to review all request types, define impact, and propose a solution.

Example:

```
[SendWorkToProxy     ] Exception captured in jsonrpc api zil_requestWork,
 exception: Exception -32003 : Client connector error: libcurl error:
 28 -> Operation timed out
```

There are three types of request to proxy:
1. Send work to the proxy
2. Check for solution
3. Confirm verification of the solution

### Send work to the proxy

**Impact:** the node will stand idle till next round, a proxy might won't be able to provide enough works for miners.

**Solution:** repeat request till the end of the round every N seconds, considering lost time in work duration.

### Check for solution

**Impact:** the node will stand idle till next round, lost of mining revenue and losing a shard sometimes.

**Solution:** Don't stop PoW round.

### Confirm verification of the solution

**Impact:** UI glitch.

**Solution:** Repeat request 2-3 times.

Please, review my solutions and implement them.

Swaggadan:
I seem to be having the same issue. 

`[WARN][  225][19-05-31T12:21:47.599][c/libPOW/pow.cpp:493][Sen                                                                  dWorkToProxy     ] Exception captured in jsonrpc api zil_reque                                                                  stWork, exception: Exception -32003 : Client connector error:                                                                   libcurl error: 7 -> Could not connect to http://zilpool:5000/`                                                                  api

reb0rn21:
Same issue all nodes broken from today :(
zero reward even if the proxy find and verifie for miner, the node receive zero and stop submiting POW

I have only one DS node that work fine atm till it last without POW

ShengguangXiao:
1. The zil_requestWork can have a retry mechanism
2. The mining proxy need to handle timeout exception, because it may affect process PoW request.

Snuff:
Hi @ShengguangXiao, What do you mean by `mining proxy`? Software like https://github.com/DurianStallSingapore/Zilliqa-Mining-Proxy?

ShengguangXiao:
> Hi @ShengguangXiao, What do you mean by `mining proxy`? Software like https://github.com/DurianStallSingapore/Zilliqa-Mining-Proxy?

@Snuff Yes, you are correct

[...]

PR #1683, opened 2019-06-06
Add retry mechanism for send pow to proxy

ShengguangXiao:
## Description

Some times because network issue, the pow request failed to send to proxy, have a retry mechanism will have more chance to success.

https://github.com/Zilliqa/Zilliqa/issues/1681

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPOW/pow.cpp
+++ b/src/libPOW/pow.cpp
@@ -387,8 +387,7 @@ ethash_mining_result_t POW::RemoteMine(const PairOfKey& pairOfKey,
   m_shouldMine = true;
 
   ethash_mining_result_t miningResult{"", "", 0, false};
-  const int MAX_RETRY_SEND_TIME = 5;
-  int retryTime = 0;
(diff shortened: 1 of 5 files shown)


src/libPOW/pow.cpp
```diff
@@ -387,7 +387,20 @@ ethash_mining_result_t POW::RemoteMine(const PairOfKey& pairOfKey,
   m_shouldMine = true;
 
   ethash_mining_result_t miningResult{"", "", 0, false};
-  if (!SendWorkToProxy(pairOfKey, blockNum, headerHash, boundary, timeWindow)) {
+  const int MAX_RETRY_SEND_TIME = 5;
```
ansnunez: Should we move this to constants.xml?

ShengguangXiao: I think can

PR #1685, opened 2019-06-10
[Release-4.6] Check Server.cpp for Default block

KaustubhShamshery:
## Backward Compatibility

## Status

#1682 
### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -1059,13 +1059,12 @@ Json::Value LookupServer::DSBlockListing(unsigned int page) {
   }
   uint64_t currBlockNum =
       m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum();
-  Json::Value _json;
-
-  uint maxPages = (currBlockNum / PAGE_SIZE) + 1;
-
   if (currBlockNum == INIT_BLOCK_NUMBER) {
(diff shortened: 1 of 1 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -1056,6 +1063,10 @@ Json::Value LookupServer::DSBlockListing(unsigned int page) {
 
   uint maxPages = (currBlockNum / PAGE_SIZE) + 1;
 
+  if (currBlockNum == INIT_BLOCK_NUMBER) {
```
ckyang: How about move this error checking to line 1062, just in back of the parameter assignment? This can avoid the potential misuse of this variable in the future.

src/libServer/LookupServer.cpp
```diff
@@ -1147,6 +1158,10 @@ Json::Value LookupServer::TxBlockListing(unsigned int page) {
       m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum();
   Json::Value _json;
 
+  if (currBlockNum == INIT_BLOCK_NUMBER) {
```
ckyang: How about move this error checking to line 1159, just in back of the parameter assignment? This can avoid the potential misuse of this variable in the future.

PR #1693, opened 2019-06-17
move scripts from testnet to core repo

sandipbhoir:
## Description
This PR addresses https://github.com/Zilliqa/testnet/issues/405

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


jiayaoqijia:
fetchHistorical.py
fetchHistorical.sh
launch_docker.sh
launch.sh

sandipbhoir:
`launch.sh and launch_docker.sh` are more of interface and has to be there in testnet repo.
`fetchHistorical.py, fetchHistorical.sh, download_and_verify.sh` are dormant and will be cleaned.

ckyang:
Also `auto_back_up.py` will be clean from `testnet` repo

sandipbhoir:
Moved `template/automateBackup.py` to core repo.

scripts/auto_back_up.py
```diff
@@ -0,0 +1,128 @@
...
+import subprocess
+import json
+import socket
+import tarfile
+import xml.etree.cElementTree as xtree
+
+TAG_NUM_FINAL_BLOCK_PER_POW = "NUM_FINAL_BLOCK_PER_POW"
+TESTNET_NAME= "TEST_NET_NAME"
+BUCKET_NAME='BUCKET_NAME'
+AWS_PERSISTENCE_LOCATION= "s3://"+BUCKET_NAME+"/persistence/"+TESTNET_NAME
```
Gnnng: can we convert all these into command-line options? This would provide a robust interface for users and avoid on-the-fly patching of the scripts.

scripts/uploadIncrDB.py
```diff
@@ -0,0 +1,392 @@
...
+
+PERSISTENCE_SNAPSHOT_NAME='incremental'
+STATEDELTA_DIFF_NAME='statedelta'
+BUCKET_NAME='BUCKET_NAME'
+NUM_TXBLOCK = 1
+NUM_DSBLOCK= "PUT_INCRDB_DSNUMS_WITH_STATEDELTAS_HERE"
+NUM_FINAL_BLOCK_PER_POW= "PUT_NUM_FINAL_BLOCK_PER_POW_HERE"
+SOURCE = './'
+TESTNET_NAME= "TEST_NET_NAME"
+SYNC_INTERVAL = 1
```
Gnnng: Can we convert some of the variables here (e.g., `BUCKET_NAME`, `TESTNET_NAME`) into command-line options? Unlike the real constants that are not subject to change, they are actually variables that need meaningful values before running.

scripts/auto_back_up.py
sandipbhoir: I agree to this to extend. It might work well enough for scripts like auto_back_up.py and uploadIncrDB.py which runs in our controlled testnet network and we can modify the invocation of these script everywhere with required parameters somehow.

however, for downloadIncrDB.py which is occassionally called by zilliqa process  instead of launch.sh/launch_docker.sh we need process to be aware of testnet specific parameters either through `constants.xml` or having `zilliqa` process itself take these as command line arguments and make then available to script like downloadIncrDB.py at runtime.
We need more work on this.

Gnnng: Thanks for the response. Let's perhaps plan this in the future.

PR #1709, opened 2019-07-01
Fix epochfin for lookup when no shard formed

kaikawaliu:
## Description

Let the lookup put epochfin during non-vacuous epoch if no microblock is loaded from finalblock.
For fixing issue of PutEpochFin didn't get triggered when no shard is formed:
https://github.com/Zilliqa/Issues/issues/535
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -716,7 +716,11 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
       LOG_GENERAL(WARNING, "StoreFinalBlock failed!");
       return false;
     }
-    if (!LOOKUP_NODE_MODE) {
+    // if lookup and loaded microblocks, then skip
+    lock_guard<mutex> g(m_mutexUnavailableMicroBlocks);
+    if (!(LOOKUP_NODE_MODE &&
+          m_unavailableMicroBlocks.find(txBlock.GetHeader().GetBlockNum()) !=
+              m_unavailableMicroBlocks.end())) {
       if (!BlockStorage::GetBlockStorage().PutEpochFin(
               m_mediator.m_currentEpochNum)) {
         LOG_GENERAL(WARNING, "BlockStorage::PutEpochFin failed "


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -716,7 +716,10 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
       LOG_GENERAL(WARNING, "StoreFinalBlock failed!");
       return false;
     }
-    if (!LOOKUP_NODE_MODE) {
+    // if lookup and loaded microblocks, then skip
```
ansnunez: Do you need to acquire `m_mutexUnavailableMicroBlocks`?

kaikawaliu: fixed

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -716,7 +716,11 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
...
+    lock_guard<mutex> g(m_mutexUnavailableMicroBlocks);
+    if (!(LOOKUP_NODE_MODE &&
+          m_unavailableMicroBlocks.find(txBlock.GetHeader().GetBlockNum()) !=
+              m_unavailableMicroBlocks.end())) {
```
sandipbhoir: would this mean that we store new epochfin even if any of the microblocks are not received from shard?

kaikawaliu: nope. this means we will store new epochfin if the txblock doesn't contain any microblock hash for the lookup to wait for.

ansnunez:
@kaikawaliu any similar fix needed here? https://github.com/Zilliqa/Zilliqa/blob/f3b96ea73f665d0d4af2d78992866d09ef8aa964/src/libNode/FinalBlockProcessing.cpp#L760

ansnunez:
> @kaikawaliu any similar fix needed here?
> 
> https://github.com/Zilliqa/Zilliqa/blob/f3b96ea73f665d0d4af2d78992866d09ef8aa964/src/libNode/FinalBlockProcessing.cpp#L760

After discussion with @kaikawaliu he says no change is needed here or elsewhere

PR #1717, opened 2019-07-09
Fix potential concurrency issue for AccountStore::m_state access

kaikawaliu:
## Description

Fixing issue: https://github.com/Zilliqa/Issues/issues/539

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libPersistence/BlockStorage.cpp
+++ b/src/libPersistence/BlockStorage.cpp
@@ -675,8 +675,10 @@ bool BlockStorage::PutEpochFin(const uint64_t& epochNum) {
       DataConversion::StringToCharArray(to_string(epochNum)));
 }
 
-bool BlockStorage::GetMetadata(MetaType type, bytes& data) {
-  LOG_MARKER();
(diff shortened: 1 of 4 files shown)


src/libPersistence/BlockStorage.cpp
```diff
@@ -675,8 +675,10 @@ bool BlockStorage::PutEpochFin(const uint64_t& epochNum) {
       DataConversion::StringToCharArray(to_string(epochNum)));
 }
 
-bool BlockStorage::GetMetadata(MetaType type, bytes& data) {
-  LOG_MARKER();
+bool BlockStorage::GetMetadata(MetaType type, bytes& data, bool muteLog) {
+  if (!muteLog) {
```
ShengguangXiao: Why need muteLog? g3log should be thread safe

kaikawaliu: It's not due to safety issue. If we don't mute it for GetEpochFin specifically, the log will be flooded with this this.

PR #1718, opened 2019-07-11
Automatically remove ipMapping.xml after using it

ckyang:
## Description

The `ipMapping.xml` will ONLY be used in recovery the entire network. So for more user-friendly, the zilliqa process should automatically remove this file after using it.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -832,7 +832,7 @@ void Node::RemoveIpMapping() {
       LOG_GENERAL(INFO,
                   IP_MAPPING_FILE_NAME << " has been removed successfully.");
     } else {
-      LOG_GENERAL(WARNING, IP_MAPPING_FILE_NAME << " cannot been removed!");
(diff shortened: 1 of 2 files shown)


src/libNode/Node.cpp
```diff
@@ -822,6 +824,21 @@ void Node::GetIpMapping(unordered_map<string, Peer>& ipMapping) {
...
 
+void Node::RemoveIpMapping() {
+  LOG_MARKER();
+
+  if (boost::filesystem::exists(IP_MAPPING_FILE_NAME)) {
+    if (boost::filesystem::remove(IP_MAPPING_FILE_NAME)) {
+      LOG_GENERAL(INFO,
+                  IP_MAPPING_FILE_NAME << " has been removed successfully.");
+    } else {
+      LOG_GENERAL(WARNING, IP_MAPPING_FILE_NAME << " cannot been removed!");
```
ShengguangXiao: " cannot be remove!"?

PR #1723, opened 2019-07-15
Fix handling of unavailable microblocks

sandipbhoir:
## Description
This PR addresses issue https://github.com/Zilliqa/Issues/issues/546

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


sandipbhoir:
Found another corner case where newlookup node is marked synced but yet to receive the sharing structure, in which case if it receives fb before sharding struct  it will skip mb hash for shard-0 because shard.size() = 0.

PR #1725, opened 2019-07-17
Create ScillaIPCServer

advaypal:
## Description

To create a jsonrpc server that will be used for communication with the Scilla executable

This is part of the larger IPC project for more efficient state communication

See branch feature/contractmap

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


kaikawaliu:
You may need to run `./build.sh style` to pass the style check in travis

vaivaswatha:
I have cleaned up and updated the code to mimic the standalone tested toy implementation of the server (and also the client) in https://github.com/Zilliqa/ipc-experiment.

@kaikawaliu This PR is now ready to be reviewed and merged to your branch. Most of the white-space changes are due to running `./build.sh style`.

@advaypal if you have the option, please make this "ready to review" rather than WIP.

src/libServer/ScillaIPCServer.cpp
```diff
@@ -15,64 +15,78 @@
...
+  }
 
-  this->contract_address = contract_address;
+  // Prepare the result and finish.
+  response.clear();
+  response.append(Json::Value(found));
+  response.append(Json::Value(value));
 }
 
-void ScillaIPCServer::setContractAddress(dev::h160 &address) {
```
advaypal: You might want to check with @kaikawaliu whether it is ok to remove this method, because it was after discussing with him that I had put it there.

src/libServer/ScillaIPCServer.cpp
```diff
@@ -15,64 +15,78 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-#include "ScillaIPCServer.h"
+#include <jsonrpccpp/server/connectors/unixdomainsocketserver.h>
+
 #include "libPersistence/ContractStorage2.h"
 #include "libUtils/DataConversion.h"
 
-using namespace jsonrpc;
```
advaypal: I understand the benefits of not polluting the global namespace, but given the number of times we have to use `jsonrpc::__` in the code I'm not sure it was worth removing this?

src/libServer/ScillaIPCServer.cpp
```diff
@@ -15,64 +15,78 @@
...
   }
+
+  string value_new = DataConversion::CharArrayToString(destination);
+  value.swap(value_new);
+  return true;
 }
 
-bool ScillaIPCServer::updateStateValue(const string &query,
-                                       const string &value) {
+bool ScillaIPCServer::ScillaIPCServer::updateStateValue(const string &query,
```
advaypal: Why are there two `ScillaIPCServer`s here?

src/libServer/ScillaIPCServer.cpp
```diff
@@ -15,64 +15,78 @@
...
+  }
 
-dev::h160 ScillaIPCServer::getContractAddress() {
-  return this->contract_address;
+  // Dummy response because a return value is needed for JSON-RPC.
+  response = "";
 }
 
-string ScillaIPCServer::fetchStateValue(const string &query) {
+bool ScillaIPCServer::ScillaIPCServer::fetchStateValue(const string &query,
```
advaypal: Similar to comment on line 87

src/libServer/ScillaIPCServer.cpp
```diff
@@ -15,64 +15,78 @@
...
-
-  this->bindAndAddMethod(
-      Procedure("updateStateValue", PARAMS_BY_NAME, JSON_BOOLEAN, "query",
-                JSON_STRING, "value", JSON_STRING, NULL),
+ScillaIPCServer::ScillaIPCServer(const dev::h160 &contrAddr,
+                                 jsonrpc::UnixDomainSocketServer &conn,
+                                 jsonrpc::serverVersion_t type)
+    : jsonrpc::AbstractServer<ScillaIPCServer>(conn, type),
+      m_contrAddr(contrAddr) {
+  // These signatures matche that of the actual functions below.
```
advaypal: Minor typo here.

src/libServer/ScillaIPCServer.cpp
vaivaswatha: `m_contrAddr` is a `const` private member. I can't imagine what advantage having an accessor for it would provide. If @kaikawaliu insists, I'll add it.

src/libServer/ScillaIPCServer.cpp
vaivaswatha: I don't have a strong opinion either way on this. I'll update to add the `using namespace jsonrpc` back.

src/libServer/ScillaIPCServer.cpp
vaivaswatha: Typo. Will fix right now.

src/libServer/ScillaIPCServer.cpp
vaivaswatha: Typo. Will fix right now.

vaivaswatha:
@advaypal your review comments have been addressed.

src/libServer/ScillaIPCServer.cpp
advaypal: Not an accessor, the ability to set it.

vaivaswatha: Would there be a situation where it needs to be set _after_ the server object is instantiated?

advaypal: That's what needs to be checked with @kaikawaliu

kaikawaliu: Yes that's needed, as long as the scilla query doesn't include the address itself and we need the address to compose the index (you can see the param is needed to be fed to the ContractStorage2 functions), we have to somewhere provide the address to the ipc somewhere.

kaikawaliu: the addr needed to be change frequently, so the set function for the address is needed here

vaivaswatha: Ah that makes sense. I thought once the server is initialized, it will be for the same address. I'll make the change. Thank you.

src/libServer/ScillaIPCServer.cpp
```diff
@@ -15,64 +15,78 @@
...
-                                   JSON_STRING, "query", JSON_STRING, NULL),
-                         &ScillaIPCServer::fetchStateValueI);
-
-  this->bindAndAddMethod(
-      Procedure("updateStateValue", PARAMS_BY_NAME, JSON_BOOLEAN, "query",
-                JSON_STRING, "value", JSON_STRING, NULL),
+ScillaIPCServer::ScillaIPCServer(const dev::h160 &contrAddr,
+                                 jsonrpc::UnixDomainSocketServer &conn,
+                                 jsonrpc::serverVersion_t type)
+    : jsonrpc::AbstractServer<ScillaIPCServer>(conn, type),
```
advaypal: Also, I just realized this, but in the client code we explicitly say the protocol being used is jsonrpc v2. Since we're adding the ability to use a different version here, should we perhaps also change the client? (or alternatively mandate that its v2)

src/libServer/ScillaIPCServer.cpp
vaivaswatha: @kaikawaliu this is now done.

src/libServer/ScillaIPCServer.cpp
vaivaswatha: Thank you @advaypal . Good catch. I've mandated V2 now.

PR #1726, opened 2019-07-18
Sync/rejoin lagging-behind lookup nodes after recovery

ckyang:
## Description

As https://github.com/Zilliqa/Issues/issues/542 mentioned, we need to synchronize after recovery, with the larger and larger size of persistence database.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -1460,106 +1467,114 @@ bool Lookup::AddMicroBlockToStorage(const MicroBlock& microblock) {
...
-  //                         << "Request for microBlockHash " << mbhash);
-  //   shared_ptr<MicroBlock> mbptr;
-  //   if (!BlockStorage::GetBlockStorage().GetMicroBlock(mbhash, mbptr)) {
-  //     LOG_GENERAL(WARNING, "Failed to fetch micro block Hash " << mbhash);
-  //     continue;
-  //   } else {
-  //     retMicroBlocks.push_back(*mbptr);
-  //   }
-  // }
+  LOG_GENERAL(INFO, "Reques for " << microBlockHashes.size() << " blocks");
```
sandipbhoir: spell mistake `Reques`

src/libLookup/Lookup.cpp
```diff
@@ -1460,106 +1467,114 @@ bool Lookup::AddMicroBlockToStorage(const MicroBlock& microblock) {
...
+    shared_ptr<MicroBlock> mbptr;
+    int retryCount = 5;
 
-  // if (!Messenger::SetLookupSetMicroBlockFromLookup(
-  //         retMsg, MessageOffset::BODY, m_mediator.m_selfKey, retMicroBlocks))
-  //         {
-  //   LOG_GENERAL(WARNING, "Failed to Process ");
-  //   return false;
-  // }
+    while (retryCount-- > 0) {
```
sandipbhoir: Do we really need retry on lookup while responding to request for mbs?

src/libLookup/Lookup.cpp
```diff
@@ -2372,82 +2408,91 @@ bool Lookup::ProcessSetStateFromSeed(const bytes& message, unsigned int offset,
...
                                       [[gnu::unused]] unsigned int offset,
                                       [[gnu::unused]] const Peer& from) {
-  LOG_GENERAL(WARNING, "Function not in used");
-  return false;
+  LOG_MARKER();
 
-  // vector<TxnHash> txnhashes;
-  // txnhashes.clear();
+  vector<TxnHash> txnhashes;
+  txnhashes.clear();
```
sandipbhoir: may not need to clear

src/libLookup/Lookup.cpp
```diff
@@ -2372,82 +2408,91 @@ bool Lookup::ProcessSetStateFromSeed(const bytes& message, unsigned int offset,
...
-  // uint128_t ipAddr = from.m_ipAddress;
-  // Peer requestingNode(ipAddr, portNo);
+  if (txnhashes.size() == 0) {
+    LOG_GENERAL(INFO, "No txn requested");
+    return true;
+  }
 
-  // bytes setTxnMsg = {MessageType::LOOKUP,
-  //                    LookupInstructionType::SETTXNFROMLOOKUP};
+  vector<TransactionWithReceipt> txnvector;
```
sandipbhoir: can rename it to `txns` ??

src/libLookup/Lookup.cpp
ckyang: Yes we need it because it's a kind of late-request, and in worst case some previous mb even will be received after the later mb.

PR #1727, opened 2019-07-18
 Fetch blocks from multiple seed nodes

ShengguangXiao:
## Description

Fetch blocks from multiple seed nodes. It can help when one of seed node has problem, the normal node still can get latest information from other seed nodes.

https://github.com/Zilliqa/Issues/issues/549

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


sandipbhoir:
Need to make note that this can potentially result in overwhelming of level2lookup's  (seed) twice.
Is it possible to try fetching once and if fails to receive response, retry with other one.

ShengguangXiao:
> Need to make note that this can potentially result in overwhelming of level2lookup's (seed) twice.
> Is it possible to try fetching once and if fails to receive response, retry with other one.

Right now we have retry mechanism already, but another retry is 80 seconds late, the node may missed pow window already, you can check the log in the issue.

ShengguangXiao:
> Need to make note that this can potentially result in overwhelming of level2lookup's (seed) twice.
> Is it possible to try fetching once and if fails to receive response, retry with other one.

It is possible the level2lookup is overwhelmed, so we may need powerful level2lookup

ShengguangXiao:
Close this PR, because there is a better solution for this issue.

PR #1729, opened 2019-07-19
Pending Txns API

KaustubhShamshery:
## Description

Add an API in the lookup nodes to tell whether a transaction might be pending or not on the shard nodes.

1. N shard nodes + leader send txn status to lookup every epoch
2. Lookup validates it and saves to local buffer for API query

Such a packet is sent every epoch and it is important to remember that only the latest view counts.

Testing :  
Using DM test achieved

1. Shard could not form MB (as DS refused to accept them) hence pending txns in the pool.
2. Sent the DS txns of higher nonce, hence these txns were reported by API

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -2143,6 +2144,11 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
       return;
     }
   }
+  if (LOOKUP_NODE_MODE) {
+    std::unique_lock<shared_timed_mutex> lock(
```
ansnunez: For easier and safer usage, can you create a function to do both the mutex acquisition and list clearing?

src/libMessage/Messenger.cpp
```diff
@@ -4629,6 +4629,99 @@ bool Messenger::SetNodeMBnForwardTransaction(
...
+  Signature signature;
+  if (!Schnorr::Sign(tmp, key.first, key.second, signature)) {
+    LOG_GENERAL(WARNING, "Failed to sign NodePendingTxn");
+    return false;
+  }
+
+  SerializableToProtobufByteArray(signature, *result.mutable_signature());
+
+  if (!result.IsInitialized()) {
+    LOG_GENERAL(WARNING, "PMHello initialization failed");
```
ansnunez:
```suggestion
    LOG_GENERAL(WARNING, "NodePendingTxn initialization failed");
```

src/libMessage/Messenger.cpp
```diff
@@ -4629,6 +4629,99 @@ bool Messenger::SetNodeMBnForwardTransaction(
...
+                             << src.size() << ", offset " << offset);
+    return false;
+  }
+
+  NodePendingTxn result;
+
+  result.ParseFromArray(src.data() + offset, src.size() - offset);
+
+  if (!result.IsInitialized() || !result.data().IsInitialized()) {
+    LOG_GENERAL(WARNING, "PMHello initialization failed");
```
ansnunez:
```suggestion
    LOG_GENERAL(WARNING, "NodePendingTxn initialization failed");
```

src/libMessage/Messenger.cpp
```diff
@@ -4629,6 +4629,99 @@ bool Messenger::SetNodeMBnForwardTransaction(
...
+
+  PROTOBUFBYTEARRAYTOSERIALIZABLE(result.data().pubkey(), pubKey);
+  Signature signature;
+  PROTOBUFBYTEARRAYTOSERIALIZABLE(result.signature(), signature);
+
+  bytes tmp(result.data().ByteSize());
+  result.data().SerializeToArray(tmp.data(), tmp.size());
+
+  if (!Schnorr::Verify(tmp, 0, tmp.size(), signature, pubKey)) {
+    LOG_GENERAL(WARNING, "PMHello signature wrong");
```
ansnunez:
```suggestion
    LOG_GENERAL(WARNING, "NodePendingTxn signature wrong");
```

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -692,6 +700,8 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
 
   bool toSendTxnToLookup = false;
 
+  bool toSendPendingTxn = !(GetUnconfirmedTxns().empty());
```
ansnunez: The construction of return object from `GetUnconfirmedTxns` can be quite expensive.  Maybe you can create a function that returns the size of `m_unconfirmedTxns`, since the check for empty is done in more than one place.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1118,6 +1138,109 @@ bool Node::ProcessMBnForwardTransaction(const bytes& message,
...
+  return true;
+}
+
+bool Node::SendPendingTxnToLookup() {
+  if (LOOKUP_NODE_MODE) {
+    LOG_GENERAL(WARNING, "SendPendingTxnToLookup called from lookup");
+    return false;
+  }
+
+  if (m_consensusMyID > 5 && !m_isPrimary) {
```
ansnunez: Maybe you can add this to constants.xml under `data_sharing` section

src/libDirectoryService/MicroBlockProcessing.cpp
```diff
@@ -406,13 +406,8 @@ bool DirectoryService::ProcessMicroblockSubmissionFromShard(
...
 #endif  // DM_TEST_DM_LESSMB_ONE
 
-#ifdef DM_TEST_DM_LESSMB_ALL
-  if (m_mediator.m_ds->m_mode == BACKUP_DS) {
-    LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
-              "Letting all of the backups refuse some Microblock submission "
-              "(DM_TEST_DM_LESSMB_ALL)");
-    return false;
-  }
+#if 1
```
sandipbhoir: Any reason to remove this?

KaustubhShamshery: This is just for tetsing on small scale. Will remove.

PR #1735, opened 2019-07-22
Store txns in DB

KaustubhShamshery:
## Description

Shard and DS nodes store processed txns in LevelDB

*Note*: `processedTxnTmp` stores processed txns, however, these may not be confirmed.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


sandipbhoir:
can you verify if the `persistence` with `txBodiesTmp` from normal shard node is available in recovered node after recovery?

constants.xml
```diff
@@ -262,6 +262,7 @@
         <ENABLE_REPOPULATE>true</ENABLE_REPOPULATE>
         <REPOPULATE_STATE_IN_DS>0</REPOPULATE_STATE_IN_DS>
         <REPOPULATE_STATE_PER_N_DS>10</REPOPULATE_STATE_PER_N_DS>
+        <NUM_STORE_TX_BODIES>5</NUM_STORE_TX_BODIES>
```
ShengguangXiao: Change NUM_STORE_TX_BODIES to STORE_TX_BODIES_BLK_INTERVAL will be more understandable

src/libPersistence/BlockStorage.cpp
```diff
@@ -1505,11 +1514,12 @@ bool BlockStorage::ResetAll() {
 bool BlockStorage::RefreshAll() {
   if (!LOOKUP_NODE_MODE) {
     return RefreshDB(META) & RefreshDB(DS_BLOCK) & RefreshDB(TX_BLOCK) &
-           RefreshDB(MICROBLOCK) & RefreshDB(DS_COMMITTEE) &
-           RefreshDB(VC_BLOCK) & RefreshDB(FB_BLOCK) & RefreshDB(BLOCKLINK) &
-           RefreshDB(SHARD_STRUCTURE) & RefreshDB(STATE_DELTA) &
-           RefreshDB(TEMP_STATE) & RefreshDB(DIAGNOSTIC_NODES) &
-           RefreshDB(DIAGNOSTIC_COINBASE) & RefreshDB(STATE_ROOT) &
+           RefreshDB(TX_BODY) & RefreshDB(MICROBLOCK) &
```
sandipbhoir: Should this be for `TX_BODY_TMP` instead?

KaustubhShamshery: yes, thanks

sandipbhoir:
what would behavior if same processed  txns are already commited to `txBodies` db and after recovery of shard node we have them in both `txBodies` and `txBodyTmp`.

KaustubhShamshery:
> what would behavior if same processed txns are already commited to `txBodies` db and after recovery of shard node we have them in both `txBodies` and `txBodyTmp`.

They would not affect each other, they are different level DBs.

sandipbhoir:
> > what would behavior if same processed txns are already commited to `txBodies` db and after recovery of shard node we have them in both `txBodies` and `txBodyTmp`.
> 
> They would not affect each other, they are different level DBs.
 
Ofcorse, but normally processed txns are removed from Tmp after moving to txBodies. In this case, we have it in both. Just curious if that breaks any logic later. If its ok, then all good.

KaustubhShamshery:
> > > what would behavior if same processed txns are already commited to `txBodies` db and after recovery of shard node we have them in both `txBodies` and `txBodyTmp`.
> > 
> > 
> > They would not affect each other, they are different level DBs.
> 
> Ofcorse, but normally processed txns are removed from Tmp after moving to txBodies. In this case, we have it in both. Just curious if that breaks any logic later. If its ok, then all good.

That is in case of lookup nodes, both of these werent being used in shard nodes before.

kaikawaliu:
TxnBodyTmp is used for cleaning the transactions processed in the latest DS epoch. It's not used for storage purpose. I suggest making another db instance for this specific usage.

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -485,9 +485,20 @@ void Node::UpdateProcessedTransactions() {
     m_createdTxns = move(t_createdTxns);
```
sandipbhoir: Looks like `UpdateProcessedTransactions` will be triggered only after receiving FinalBlock. `ProcessFinalBlockCore`->`LoadUnavailableMicroBlockHashes`->`UpdateProcessedTransactions`

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -641,9 +646,22 @@ void Node::ProcessTransactionWhenShardBackup() {
 
   cv_TxnProcFinished.notify_all();
 
+  PutTxnsInTempDataBase(t_processedTransactions);
```
sandipbhoir: We need to consider the fact that the txns in leveldb for backup may not be confirmed transaction.

PR #1737, opened 2019-07-24
Optimize DS leader verify fail log

ShengguangXiao:
## Description

Optimize DS leader verify fail log

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -2803,8 +2803,9 @@ bool Lookup::ProcessRaiseStartPoW(const bytes& message, unsigned int offset,
     return false;
   }
 
-  if (!(expectedDSLeader.first == dspubkey)) {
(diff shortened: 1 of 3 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -2803,8 +2803,9 @@ bool Lookup::ProcessRaiseStartPoW(const bytes& message, unsigned int offset,
     return false;
   }
 
-  if (!(expectedDSLeader.first == dspubkey)) {
-    LOG_GENERAL(WARNING, "Message does not comes from DS leader");
+  if (expectedDSLeader.first != dspubkey) {
+    LOG_CHECK_FAIL("Message does not comes from DS leader", dspubkey,
```
ansnunez:
```suggestion
    LOG_CHECK_FAIL("DS leader key", dspubkey,
```

PR #1739, opened 2019-07-29
Don't fetch info from blacklisted seed nodes

ShengguangXiao:
## Description

Don't fetch info from blacklisted seed nodes, in case some of the seed nodes are problematic, the node still can sync with other good seed nodes without delay. It will be a better solution if compared with another approach in https://github.com/Zilliqa/Zilliqa/pull/1727.

https://github.com/Zilliqa/Issues/issues/549

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -189,7 +189,7 @@ void DirectoryService::SendDSBlockToShardNodes(
 
     if (BROADCAST_TREEBASED_CLUSTER_MODE) {
       // Choose N other Shard nodes to be recipient of DS block
-      std::vector<Peer> shardDSBlockReceivers;
+      VectorOfPeer shardDSBlockReceivers;
       unsigned int numOfDSBlockReceivers =
           NUM_FORWARDED_BLOCK_RECEIVERS_PER_SHARD;
       if (numOfDSBlockReceivers <= NUM_DS_ELECTION) {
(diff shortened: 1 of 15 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -919,19 +919,35 @@ bool Lookup::ProcessGetDSInfoFromSeed(const bytes& message, unsigned int offset,
...
+    lock_guard<mutex> lock(m_mutexSeedNodes);
+    if (0 == m_seedNodes.size()) {
+      LOG_GENERAL(WARNING, "Seed nodes are empty");
+      return;
+    }
+
+    for (const auto& node : m_seedNodes) {
+      auto seedNodeIpToSend = TryGettingResolvedIP(node.second);
+      if (!Blacklist::GetInstance().Exist(seedNodeIpToSend)) {
+        notBlackListedSeedNodes.push_back(node);
```
sandipbhoir: We can store the `resolvedIP`  as well.  So that we don't need to resolve IPS for all `notBlacklistedSeedNodes`again at line 947.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -368,6 +368,7 @@ void DirectoryService::StartFirstTxEpoch() {
     Guard::GetInstance().AddDSGuardToBlacklistExcludeList(
         *m_mediator.m_DSCommittee);
   }
+  m_mediator.m_lookup->RemoveSeedNodeFromBlackList();
```
sandipbhoir: can rename it to `RemoveSeedNodesFromBlackList()`?

PR #1741, opened 2019-07-29
Avoid validateDB crashing when checking normal node

KaustubhShamshery:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -435,8 +435,8 @@ bool Node::CheckIntegrity(bool continueOnError) {
       }
       if (BlockStorage::GetBlockStorage().GetMicroBlock(mbInfo.m_microBlockHash,
                                                         mbptr)) {
-        auto tranHashes = mbptr->GetTranHashes();
         if (LOOKUP_NODE_MODE) {
+          auto tranHashes = mbptr->GetTranHashes();
           for (const auto& tranHash : tranHashes) {
             TxBodySharedPtr tx;
             if (!BlockStorage::GetBlockStorage().GetTxBody(tranHash, tx)) {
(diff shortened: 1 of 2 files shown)


src/libNode/Node.cpp
```diff
@@ -435,14 +435,16 @@ bool Node::CheckIntegrity(bool continueOnError) {
...
                                                         mbptr)) {
-        auto tranHashes = mbptr->GetTranHashes();
-        for (const auto& tranHash : tranHashes) {
-          TxBodySharedPtr tx;
-          if (!BlockStorage::GetBlockStorage().GetTxBody(tranHash, tx)) {
-            LOG_GENERAL(WARNING, "Missing Tx: " << tranHash);
-            result = false;
-            if (!continueOnError) {
-              break;
+        if (LOOKUP_NODE_MODE) {
```
ShengguangXiao: So when we validate lookup DB we need to enable the `LOOKUP_NODE_MODE` option?

KaustubhShamshery: when the node is full node (i.e. lookup l2lookup)

ansnunez: Since validatePersistence.sh also fetches the constants.xml for the specific node, this should be OK (i.e., no need to manually enable the option)

ansnunez:
@KaustubhShamshery can I piggy-back one more change into this PR:

Executing validateDB usually takes several minutes, and we have no idea what the current progress is.  Can you add a `cout` (not a `LOG_GENERAL`), say, for every `NUM_FINAL_BLOCK_PER_POW` blocks processed?

```
Checking FB 0
Checking FB 100
Checking FB 200
...
```

src/libNode/Node.cpp
ansnunez: @KaustubhShamshery this would mean a non-lookup still at least attempts to fetch all the microblocks.  That would make the resulting log file unreadable with all the "Missing MB" messages.  Perhaps we can skip looping through `txBlocks` entirely for non-lookup.

PR #1749, opened 2019-08-05
Save txns to S3

sandipbhoir:
## Description
Addresses issue - https://github.com/Zilliqa/Issues/issues/559

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -664,6 +665,52 @@ void Node::PutTxnsInTempDataBase(
...
+    size = serializedTxn.size();
+    txns_file.write(reinterpret_cast<const char*>(&size), sizeof(size_t));
+    txns_file.write(reinterpret_cast<const char*>(serializedTxn.data()), size);
+  }
+
+  txns_file.close();
+
+  // upload to S3
+  if (!SysCommand::ExecuteCmd(SysCommand::WITHOUT_OUTPUT,
+                              GetAwsS3CpString(txns_filename))) {
```
ansnunez: Do we want to keep the `/tmp/` part of the filename when we upload to S3?

sandipbhoir: /tmp/ part is used by `aws s3 cp` to only get the local path of file/object. Only the leaf object without `/tmp` part is copied over to S3.

PR #1772, opened 2019-08-15
Develop new daemon which can launch zilliqa process

ShengguangXiao:
## Description

Develop new daemon which can launch zilliqa process

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


scripts/upload_incr_DB.py
```diff
@@ -385,12 +403,13 @@ def main():
...
 	dd = int(args['dsblktime'])
 	webhook = args['webhook']
+	source = os.path.dirname(os.path.abspath(__file__)) + "/"
 
 	# create temp folder
-	if not os.path.exists(SOURCE+'temp'):
-		os.makedirs(SOURCE+'temp')
-	CleanupDir(SOURCE+'temp')
-	os.chdir(SOURCE)
+	if not os.path.exists(source+'temp'):
```
Gnnng: Can we use the [standard library](https://docs.python.org/3.7/library/tempfile.html) to handle the temporary directory? If the folder is really meant to be temporary and not meant to be permanent, it should be recycled automatically by the code or by the system (if you are using `/tmp/` path)

src/libNode/Node.cpp
```diff
@@ -155,7 +155,8 @@ bool Node::DownloadPersistenceFromS3() {
   // TBD - find better way to capture the exit status of command
   string excludembtxns = LOOKUP_NODE_MODE ? "false" : "true";
   SysCommand::ExecuteCmdWithOutput(
-      "./downloadIncrDB.py " + STORAGE_PATH + "/ " + excludembtxns, output);
+      "/run/zilliqa/download_incr_DB.py " + STORAGE_PATH + "/ " + excludembtxns,
```
Gnnng: The binary path should not be hard-coded like this. Assuming the user will put the binary in the same location as we do is not ideal unless you have told users to do so.

ckyang: ok, I'll put the binary path to `constants.xml`

scripts/upload_incr_DB.py
ckyang: It's a good suggestion and I think it's better to have a separated PR to refactor the `temp` thing. @sandipbhoir

CMakeLists.txt
```diff
@@ -255,7 +255,7 @@ if(TESTS)
 endif()
 
 # installation
-set_target_properties(zilliqa sendcmd genkeypair getpub getaddr gentxn signmultisig verifymultisig ZilliqaDaemon_AWS gensigninitialds grepperf getnetworkhistory getrewardhistory validateDB restore genTxnBodiesFromS3
+set_target_properties(zilliqa sendcmd genkeypair getpub getaddr gentxn signmultisig verifymultisig ZilliqaDaemon gensigninitialds grepperf getnetworkhistory getrewardhistory validateDB restore genTxnBodiesFromS3
```
Gnnng: Can we change the daemon binary name (`ZilliqaDaemon`) to a name consisting of lowercase letters? Following the typical naming convention for daemon, a better name is `zilliqad`

scripts/auto_backup.py
```diff
@@ -95,6 +95,7 @@ def backUp(curr_blockNum):
     return None
 
 def main():
+    os.chdir(os.path.dirname(os.path.abspath(__file__)))
     parser = argparse.ArgumentParser(description='Lookup autorecovery script')
```
Gnnng: The `description` seems to be wrong

daemon/ZilliqaDaemon.cpp
```diff
@@ -15,95 +15,158 @@
...
+#define ERROR_IN_COMMAND_LINE -1
+
+const vector<string> programName = {"zilliqa"};
+const string SYNCTYPE_OPT = "--synctype";
+const char* synctype_descr =
+    "0(default) for no, 1 for new, 2 for normal, 3 for ds, 4 for lookup, 5 "
+    "for node recovery, 6 for new lookup , 7 for ds guard node sync and 8 "
+    "for offline validation of DB";
+
+const string scripts_path = "/run/zilliqa/";
```
Gnnng: Can we try to parameterize these path constants? Otherwise, users are forced to follow the paths you hardcoded.

ckyang: Ok, will put to `constants.xml` also

daemon/ZilliqaDaemon.cpp
```diff
@@ -15,95 +15,158 @@
...
+      m_died.erase(pid);
+    }
+  }
 }
 
-void setIsDs(const pid_t& pid, ofstream& log) {
-  string s = ReadLastLine(Path[pid], log);
+string ZilliqaDaemon::CurrentTimeStamp() {
+  auto tm = chrono::system_clock::to_time_t(chrono::system_clock::now());
+  char* t = ctime(&tm);
```
Gnnng: Perhaps a better choice is `std::strftime` so the manual string operation can be avoided.

See [`std::strftime`](https://en.cppreference.com/w/cpp/chrono/c/strftime) and [`std::ctime`](https://en.cppreference.com/w/cpp/chrono/c/ctime)
> POSIX marks this function obsolete and recommends std::strftime instead.

daemon/ZilliqaDaemon.h
```diff
@@ -0,0 +1,70 @@
...
+#include <unordered_map>
+#include <vector>
+
+class ZilliqaDaemon {
+ public:
+  ZilliqaDaemon(int argc, const char* argv[], std::ofstream* log);
+  void MonitorProcess();
+
+ private:
+  std::ofstream* m_log;
```
Gnnng: A better alternative is to use reference here
```suggestion
  std::ofstream& m_log;
```
For the initialization, using member initialization list will allow you to set the initial value.

daemon/ZilliqaDaemon.cpp
```diff
@@ -113,203 +175,229 @@ vector<pid_t> getProcIdByName(const string& procName, ofstream& log) {
...
-int getRestartValue(pid_t pid, const string& prgname) {
-  if (prgname == "latezilliqa" || prgname == "zilliqa") {
-    if (isDS[pid]) {
-      return DS_SYNC;
+void ZilliqaDaemon::StartNewProcess() {
+  m_log << ZilliqaDaemon::CurrentTimeStamp().c_str()
+        << "Create new Zilliqa process..." << endl;
+  signal(SIGCHLD, SIG_IGN);
+
+  if (0 == fork()) {
```
Gnnng: Consider adding an error print when fork failed.

daemon/ZilliqaDaemon.cpp
```diff
@@ -15,95 +15,157 @@
...
+
+  return result;
+}
+
+bool ZilliqaDaemon::DownloadPersistenceFromS3() {
+  string output;
+  m_log << ZilliqaDaemon::CurrentTimeStamp().c_str()
+        << "downloading persistence from S3" << endl;
+  output = Execute("python " + m_curPath + download_incr_DB_script);
+  return (output.find("Done!") != std::string::npos);
```
Gnnng: Can we look for some support on the download script to indicate the success through the exit code? text match based on the output is not a stable solution.

daemon/ZilliqaDaemon.cpp
```diff
@@ -15,95 +15,157 @@
...
-  getline(logFile, lastLine);
+  for (const pid_t& pid : m_pids[name]) {
+    // If sig is 0 (the null signal), error checking is performed but no signal
+    // is actually sent
+    int w = kill(pid, 0);
 
-  // log<<"Line :"<<lastLine<<endl;
+    if (w < 0) {
+      if (errno == EPERM) {
+        m_log << ZilliqaDaemon::CurrentTimeStamp().c_str()
```
Gnnng: If the timestamp is needed for every logging call, why not creating a helper function to reduce the redundancy

daemon/ZilliqaDaemon.cpp
```diff
@@ -323,28 +411,23 @@ int main() {
...
-  sid = setsid();
-
-  if (sid < 0) {
+  if (setsid() < 0) {
     log << "Unable to set sid" << endl;
     exit(EXIT_FAILURE);
   }
 
   if ((chdir("..")) < 0) {
-    log << "Failed to chdir" << endl;
```
Gnnng: seems this logging is dropped accidentally. I couldn't get it why changing to the parent directory is needed. Is there any assumption about the working directory where the daemon is executing?

daemon/ZilliqaDaemon.cpp
```diff
@@ -15,95 +15,157 @@
...
+  }
+
+  return result;
+}
+
+bool ZilliqaDaemon::DownloadPersistenceFromS3() {
+  string output;
+  m_log << ZilliqaDaemon::CurrentTimeStamp().c_str()
+        << "downloading persistence from S3" << endl;
+  output = Execute("python " + m_curPath + download_incr_DB_script);
```
Gnnng: Can we update the script `download_incr_DB.py`  to support python3 interpreter?

daemon/ZilliqaDaemon.cpp
```diff
@@ -113,203 +175,229 @@ vector<pid_t> getProcIdByName(const string& procName, ofstream& log) {
...
+      m_log << "Suspend launch is " << bSuspend
+            << ", set syncType = " << strSyncType
+            << ", recovery = " << m_recovery << endl;
     }
-  } else if (prgname == "lzilliqa") {
-    return LOOKUP_SYNC;
-  }
 
-  return -1;
+    string cmdToRun = "python " + m_curPath + launch_zilliqa_script + " " +
```
Gnnng: Can we update the `launch_zilliqa.py` script to support python3 interpreter?

daemon/ZilliqaDaemon.cpp
ckyang: It's a nice suggestion and I agree to do that. Maybe we can raise a separate to refactor some design of `download_incr_DB.py`. @sandipbhoir

daemon/ZilliqaDaemon.cpp
Gnnng: Another concern is why would we still need the script `launch_zilliqa.py`?

If I'm not wrong we are adding a python wrapper around `zilliqa` binary for some additional functions that could be done in C++.

ckyang: Actually it's original design and I just keep it. But anyway, yes this script could be removed. I'll try to remove `launch_zilliqa.py` in this PR also.

daemon/ZilliqaDaemon.cpp
ckyang: After some checking, `python3` cannot be used to run `download_incr_DB.py` because of the lack of `request` package. I'll leave it to later's refactor. @sandipbhoir

src/libNode/Node.cpp
```diff
@@ -154,8 +154,10 @@ bool Node::DownloadPersistenceFromS3() {
   string output;
   // TBD - find better way to capture the exit status of command
   string excludembtxns = LOOKUP_NODE_MODE ? "false" : "true";
-  SysCommand::ExecuteCmdWithOutput(
-      "./downloadIncrDB.py " + STORAGE_PATH + "/ " + excludembtxns, output);
+  SysCommand::ExecuteCmdWithOutput(boost::filesystem::current_path().string() +
```
ShengguangXiao: boost::filesystem::current_path().string() should included zilliqa already, we should log the command the software executed

ckyang:
@Gnnng @sandipbhoir @ShengguangXiao  I've made some modification according to previous review. Could you please check again? Thanks!

scripts/upload_incr_DB.py
```diff
@@ -27,17 +27,39 @@
...
+def setup_logging():
+  logfile = os.path.dirname(os.path.abspath(__file__)) + "/upload_incr_DB-log.txt"
+  backup_count = 5
+  rotating_size = 8
+  fh = handlers.RotatingFileHandler(
+    logfile, maxBytes=rotating_size * 1024 * 1024, backupCount=backup_count
+    )
+  fh.setFormatter(FORMATTER)
+  rootLogger.addHandler(fh)
+
```
sandipbhoir: will this impact or work as expected with the default compression that happens every `10 mins` ?

src/libNode/Node.cpp
sandipbhoir: Like @ShengguangXiao suggested, can we log and verify if the download script works as expected?

daemon/ZilliqaDaemon.cpp
```diff
@@ -15,95 +15,152 @@
...
-    return "";
+  for (int i = 1; i < argc; ++i) {
+    msg += string(" ") + argv[i];
   }
 
-  bool keepLooping = true;
+  ZilliqaDaemon::LOG(m_log, msg);
+  KillProcess();
+  StartNewProcess();
+  StartScripts();
```
sandipbhoir: I think we are missing something.  One's new process is spawned after being deleted, it should kill any running scripts and start scripts again. But its not the case right now.

daemon/ZilliqaDaemon.cpp
```diff
@@ -113,238 +170,287 @@ vector<pid_t> getProcIdByName(const string& procName, ofstream& log) {
...
   close(STDERR_FILENO);
 
-  unordered_map<string, vector<pid_t>> pids;
-  unordered_map<pid_t, bool> died;
+  ZilliqaDaemon daemon(argc, argv, log);
 
-  initialize(pids, died, log);
   while (1) {
-    MonitorProcess(pids, died, log);
+    daemon.MonitorProcess();
```
sandipbhoir: Can we run MonitorProcess for each program in list.? If enable polling for more than 1 processes if needed in future.
```
   for (const auto&p : programName)
   {
        daemon.MonitorProcess(p);
   }
  ```
Also, if we can pass program names as parameter to daemon.

File: scripts/release.sh [hunk 18]
```diff
@@ -269,7 +269,7 @@ cd -
...
+    cmd="tar cfz ${testnet_to_be_upgraded}.tar.gz -C $(dirname ${pubKeyFile}) $(basename ${pubKeyFile}) -C $(realpath ./scripts) miner_info.py -C $(realpath ./scripts) auto_backup.py -C $(realpath ./scripts) download_incr_DB.py -C $(realpath ./scripts) download_and_verify.sh -C $(realpath ./scripts) fetchHistorical.py -C $(realpath ./scripts) fetchHistorical.sh -C $(realpath ./scripts) upload_incr_DB.py -C $(realpath ./scripts) automateBackup.py -C $(realpath ./${releaseDir}) $(basename ${versionFile}) -C $(dirname ${constantFile}) $(basename ${constantFile}) -C $(dirname ${constantLookupFile}) $(basename ${constantLookupFile})_lookup"
```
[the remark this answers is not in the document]
ckyang: done

daemon/ZilliqaDaemon.cpp
```diff
@@ -449,9 +507,11 @@ int main(int argc, const char* argv[]) {
 
   ZilliqaDaemon daemon(argc, argv, log);
 
-  while (1) {
-    daemon.MonitorProcess();
-    sleep(5);
+  for (const auto& name : programName) {
+    while (1) {
```
sandipbhoir: `While loop` should be outer loop for `for loop`

PR #1775, opened 2019-08-16
Missing include mbstxns flag

sandipbhoir:
## Description
This PR addresses issue  https://github.com/Zilliqa/Issues/issues/569

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -153,8 +153,9 @@ bool Node::DownloadPersistenceFromS3() {
   LOG_MARKER();
   string output;
   // TBD - find better way to capture the exit status of command
-  SysCommand::ExecuteCmdWithOutput("./downloadIncrDB.py " + STORAGE_PATH + "/",
-                                   output);
+  string excludembtxns = LOOKUP_NODE_MODE ? "false" : "true";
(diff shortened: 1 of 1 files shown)


src/libNode/Node.cpp
```diff
@@ -153,8 +153,9 @@ bool Node::DownloadPersistenceFromS3() {
   LOG_MARKER();
   string output;
   // TBD - find better way to capture the exit status of command
-  SysCommand::ExecuteCmdWithOutput("./downloadIncrDB.py " + STORAGE_PATH + "/",
-                                   output);
+  string excludembtxns = LOOKUP_NODE_MODE ? "false" : "true";
+  SysCommand::ExecuteCmdWithOutput(
+      "./downloadIncrDB.py " + STORAGE_PATH + "/ " + excludembtxns, output);
```
ckyang: Maybe can directly write the condition inside, to avoid the extra variable declaration?

sandipbhoir: I would prefer clean readability over extra variable in this case. Can change if you insist?

PR #1777, opened 2019-08-19
Generate txBodies ldb and optionally json from raw bytes from S3 files

sandipbhoir:
## Description
This PR addresses : https://github.com/Zilliqa/Issues/issues/566

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/cmd/genTxnBodiesFromS3.cpp
```diff
@@ -0,0 +1,175 @@
...
+          if (saveToJsonFormat) {
+            Json::Value v = JSONConversion::convertTxtoJson(r_tr);
+
+            // create file with filename as "<<txnhash>>.json"
+            ofstream ofile(jsonOutputPath + "/" + r_txn_hash.hex() + ".json");
+            ofile << v.toStyledString();
+            ofile.close();
+          }
+        }
+      } catch (...) {
```
KaustubhShamshery: instead of `...` can we just catch standard exception ?

sandipbhoir: basically want to make sure we catch all kind of exception and program continues.
with `std::exception` can catch only those extending std::exception but no `runtime_error`  or any other. In that case needs to handle `runtime_exception` followed by `std::exception` and `...`.
so preferred this one over it.

PR #1781, opened 2019-08-26
Fix node rejoin but cannot get DS block issue

ShengguangXiao:
## Description

There is a potential racing issue when start sync after GetLatestDSBlock function. If the server reply slow, then BlockLink haven't increased, the StartSynchronization can get DS block from seed node. If server reply fast, the shard node BlockLink already increased, when request new DS block, it cannot get anything from seed node.
So for safe run, should use back the old method clean the pensistence and rejoin as normal.

https://github.com/Zilliqa/Issues/issues/571

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/PoWProcessing.cpp
+++ b/src/libNode/PoWProcessing.cpp
@@ -183,8 +183,7 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
             // exciplitly declare in the same thread
             m_mediator.m_lookup->m_startedPoW = false;
           }
-          m_mediator.m_lookup->SetSyncType(SyncType::NORMAL_SYNC);
-          StartSynchronization();
+          RejoinAsNormal();
         } else {
           LOG_GENERAL(WARNING, "DS block not recvd, what to do ?");
         }


sandipbhoir:
`StartSynchronization` is called only after `GetLatestDSBlock` finishes, right? 
Also, looks like `GetLatestDSBlock` does not add new ds block to the blocklink since it returns 
immediately after `AlreadyJoinedNetwork` at https://github.com/Zilliqa/Zilliqa/blob/1e6172b964e8eba5b35d143eb8e892be0309eb05/src/libLookup/Lookup.cpp#L1767

ShengguangXiao:
> `StartSynchronization` is called only after `GetLatestDSBlock` finishes, right?
> Also, looks like `GetLatestDSBlock` does not add new ds block to the blocklink since it returns
> immediately after `AlreadyJoinedNetwork` at
> 
> https://github.com/Zilliqa/Zilliqa/blob/1e6172b964e8eba5b35d143eb8e892be0309eb05/src/libLookup/Lookup.cpp#L1767

I see, i think i ignored it is returned here, then this issue still need furture investigate.

ansnunez:
Fixed in #1782

PR #1783, opened 2019-08-28
Change default value for small scale test

ShengguangXiao:
## Description

Change default value for small scale test, because the right now the default parameter, the ./bootstrap.py command will have warning about the parameter not appropriate, user need to change every time before run `./testnet.sh up`

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


sandipbhoir:
can mention reason for changing these defaults?

ShengguangXiao:
> can mention reason for changing these defaults?

I updated the comment.

PR #1784, opened 2019-08-28
Fix access null pointer

ShengguangXiao:
## Description

Fix access null pointer, the m_retriever is initialized in StartRetrieveHistory function, but for a new node, StartRetrieveHistory function is not called.

https://github.com/Zilliqa/Issues/issues/573

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


ShengguangXiao:
> I'm good for the solution, however still need testing and fix of clang-format checking.

I tested and the result is good

PR #1789, opened 2019-09-02
Integrate fetching a substate of a smart contract

KaustubhShamshery:
## Description

Add an API to fetch a sub state of a contract as required by the new changes in the Scilla. 

Also, `GetSmartContract` has a breaking change.

We have changed our JSON format for more efficient accesses of field variables and map keys, representing them as JSON associative objects instead of as JSON arrays. This also makes automatic comparisons of states represented as JSONs simpler. With this change, we no longer specify the types of values and vnames in the JSON, they must be obtained and inferred from the output of `scilla-checker`

Previous Output:
```[
   {
      "type" : "Map (ByStr20) (Bool)",
      "value" : [
         {
            "key" : "0xdfa89866ae86632b36361d53b76c1373448c28fa",
            "val" : {
               "argtypes" : [],
               "arguments" : [],
               "constructor" : "True"
            }
         }
      ],
      "vname" : "admins"
   },
   {
      "type" : "Uint128",
      "value" : "0",
      "vname" : "_balance"
   }
]
```
Now
```
{
   "_balance" : "0",
   "admins" : {
      "0xdfa89866ae86632b36361d53b76c1373448c28fa" : {
         "argtypes" : [],
         "arguments" : [],
         "constructor" : "True"
      }
   }
} 
```

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


tests/Data/AccountData/CMakeLists.txt
```diff
@@ -2,10 +2,10 @@ configure_file(${CMAKE_SOURCE_DIR}/constants.xml constants.xml COPYONLY)
 
 link_directories(${CMAKE_BINARY_DIR}/lib)
 
-add_executable(Test_Account Test_Account.cpp)
-target_include_directories(Test_Account PUBLIC ${CMAKE_SOURCE_DIR}/src ${CMAKE_SOURCE_DIR}/tests)
-target_link_libraries(Test_Account PUBLIC AccountData Crypto Trie Utils Persistence TestUtils)
-add_test(NAME Test_Account COMMAND Test_Account)
+#add_executable(Test_Account Test_Account.cpp)
```
KaustubhShamshery: Is this test supposed to be re-enabled?

src/libMessage/MessengerSWInfo.cpp
```diff
@@ -35,6 +35,9 @@ bool SerializeToArray(const T& protoMessage, bytes& dst,
 }
 
 inline bool CheckRequiredFieldsProtoSWInfo(const ProtoSWInfo& protoSWInfo) {
+  return true;
+// TODO: Check if default value is acceptable for each field
+#if 0
```
KaustubhShamshery: can we just remove the code wrapped in `#if 0`

ansnunez:
Moved to #1803

PR #1791, opened 2019-09-03
Enable fetching missing mbs/txns from lookup for seed nodes

sandipbhoir:
## Description
Addresses issue - Zilliqa/Issues#547

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


constants.xml
```diff
@@ -70,6 +70,7 @@
         <GETSTATEDELTAS_TIMEOUT_IN_SECONDS>5</GETSTATEDELTAS_TIMEOUT_IN_SECONDS>
         <RETRY_REJOINING_TIMEOUT>10</RETRY_REJOINING_TIMEOUT>
         <RETRY_GETSTATEDELTAS_COUNT>3</RETRY_GETSTATEDELTAS_COUNT>
+        <MAX_FETCHMISSINGMBS_NUM>12</MAX_FETCHMISSINGMBS_NUM>
```
ShengguangXiao: Why don't use MAX_FETCH_MISSING_MBS_NUM

src/libMessage/ZilliqaMessage.proto
```diff
@@ -806,14 +806,16 @@ message LookupForwardTxnsFromSeed
 message LookupGetTxnsFromLookup
 {
     uint32 portno   = 1;
-    repeated bytes txnhashes = 2;
+    bytes mbhash   = 2;
```
ShengguangXiao: Number better to be aligned

src/libLookup/Lookup.cpp
```diff
@@ -4199,3 +4300,86 @@ bool Lookup::ProcessSetHistoricalDB(const bytes& message, unsigned int offset,
...
+                                 return e.second == TxnHash();
+                               }),
+                mbs.end());
+
+      LOG_GENERAL(INFO,
+                  "After deleting microblock bodies with no transactions, "
+                  "Unavailable count = "
+                      << mbs.size());
+
+      if (0 == mbs.size()) {
```
ShengguangXiao: Use mbs.empty instead of check size

src/libLookup/Lookup.cpp
```diff
@@ -4199,3 +4300,86 @@ bool Lookup::ProcessSetHistoricalDB(const bytes& message, unsigned int offset,
...
+      }
+      SendGetMicroBlockFromLookup(mbHashes);
+      if (limitReached) {
+        break;
+      }
+    }
+
+    // Delete the entry for those fb with no pending mbs
+    for (auto it = unavailableMBs.begin(); it != unavailableMBs.end();) {
+      if (it->second.size() == 0) {
```
ShengguangXiao: Use it->second.empty()

PR #1793, opened 2019-09-04
Force serialization of primitive proto fields for block and tx data

ansnunez:
## Description

Proto3 does not serialize fields that are set to default value ([reference](https://developers.google.com/protocol-buffers/docs/proto3))

This is a breaking change from proto2, and will be problematic for any of our code that relies on byte consistency (e.g., hashes based on serialized blocks).

To fix this, we force the serialization of these fields by wrapping each one in a `oneof`. ([reference](https://stackoverflow.com/questions/37034746/do-not-set-default-values-in-optional-protobuf-fields-to-minimize-data-sent-over))

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libMessage/ZilliqaMessage.proto
+++ b/src/libMessage/ZilliqaMessage.proto
@@ -19,10 +19,10 @@ message ByteArray
 // Used in database "blockLinks"
 message ProtoBlockLink
 {
-    uint32 version   = 1; // Added in: v1.0, Deprecated in: N/A
-    uint64 index     = 2; // Added in: v1.0, Deprecated in: N/A
-    uint64 dsindex   = 3; // Added in: v1.0, Deprecated in: N/A
-    uint32 blocktype = 4; // Added in: v1.0, Deprecated in: N/A
+    oneof oneof1 { uint32 version   = 1; } // Added in: v1.0, Deprecated in: N/A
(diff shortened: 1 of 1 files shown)


src/libMessage/ZilliqaMessage.proto
```diff
@@ -310,15 +310,15 @@ message ProtoAccountStore
 message ProtoPeer
 {
     ByteArray ipaddress    = 1;
-    uint32 listenporthost  = 2;
+    oneof oneof2 { uint32 listenporthost  = 2; }
 }
 
 message ProtoPoWSolution
 {
     uint64 nonce        = 1;
```
sandipbhoir: Any reason, we don't apply for `oneof` to `nonce` field?

ansnunez: > Any reason, we don't apply for `oneof` to `nonce` field?

I think a zero nonce for pow solution is highly unlikely?

sandipbhoir: In general, what is the reason for skipping other fields like `version` and `gaslimit`. These were optional in proto2 version previously.

KaustubhShamshery: so issue comes when the default value of these are actual values in the block (in this case the default value is 0). version starts from one and gaslimit is also non-zero.

PR #1796, opened 2019-09-04
Fix genesis DS block hash not correct problem

ShengguangXiao:
## Description

Fix genesis DS block hash not correct problem

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/common/Constants.h
+++ b/src/common/Constants.h
@@ -124,6 +124,8 @@ const std::string dsNodeFile = "dsnodes.xml";
 
 const float ONE_HUNDRED_PERCENT = 100.f;
 
+const unsigned int GENESIS_DSBLOCK_VERSION = 1;
+
 // General constants
 extern const unsigned int DEBUG_LEVEL;
 extern const bool ENABLE_DO_REJOIN;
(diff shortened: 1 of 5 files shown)


constants.xml
```diff
@@ -8,7 +8,7 @@
         <MAX_ENTRIES_FOR_DIAGNOSTIC_DATA>25</MAX_ENTRIES_FOR_DIAGNOSTIC_DATA>
         <CHAIN_ID>1</CHAIN_ID>
         <!-- testnet -->
-        <GENESIS_PUBKEY>03B70CF2ABEAE4E86DAEF1A36243E44CD61138B89055099C0D220B58FB86FF588A</GENESIS_PUBKEY>
+        <GENESIS_PUBKEY>02FB704BA6668B47F1ACD20FB9CB78E816B18C9D7F33F8D360D66402FEF42DBA5C</GENESIS_PUBKEY>
```
KaustubhShamshery: this need not be changed in current master (only mainnet)

ShengguangXiao: I see, because i add a test case to check the genesis block hash. So should check with another genesis block hash.

PR #1798, opened 2019-09-05
Fix genesis DS block hash not correct problem

ShengguangXiao:
## Description

Fix genesis DS block hash not correct problem, and also add auto test case to verify in the future in the change break block hash calculation.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


PirateGypsy:
Unsiscribe

 
Ryan Hicks  
c. 501.553.1003 |  24/hr ER. 844.333.0939

E3Response.com

PR #1800, opened 2019-09-05
Fix possible race cond

sandipbhoir:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -4322,6 +4322,8 @@ void Lookup::CheckAndFetchUnavailableMBs() {
   unsigned int maxMBSToBeFetched = MAX_FETCHMISSINGMBS_NUM;
   auto main_func = [this, maxMBSToBeFetched]() mutable -> void {
     m_startedFetchMissingMBsThread = true;
+    std::lock_guard<mutex> lock(
+        m_mediator.m_node->m_mutexUnavailableMicroBlocks);
     auto& unavailableMBs = m_mediator.m_node->GetUnavailableMicroBlocks();
```
ShengguangXiao: I suggest to put the lock inside the GetUnavailableMicroBlocks function

PR #1803, opened 2019-09-05
Integrate fetching a substate of a smart contract

ansnunez:
## Description

Cherry-picked commit, based on #1789 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/JSONConversion.cpp
+++ b/src/libServer/JSONConversion.cpp
@@ -270,7 +270,7 @@ const vector<string> JSONConversion::convertJsonArrayToVector(
                                       "Every array value should be a string");
     }
 
-    vec.push_back(ele.asString());
+    vec.emplace_back(ele.asString());
   }
   return vec;
 }
--- a/src/libServer/JSONConversion.cpp
+++ b/src/libServer/JSONConversion.cpp
@@ -255,6 +255,26 @@ bool JSONConversion::checkJsonTx(const Json::Value& _json) {
   return ret;
 }
 
+const vector<string> JSONConversion::convertJsonArrayToVector(
+    const Json::Value& _json) {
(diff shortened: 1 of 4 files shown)


src/libServer/JSONConversion.cpp
```diff
@@ -255,6 +255,26 @@ bool JSONConversion::checkJsonTx(const Json::Value& _json) {
...
 }
 
+const vector<string> JSONConversion::convertJsonArrayToVector(
+    const Json::Value& _json) {
+  if (!_json.isArray()) {
+    throw jsonrpc::JsonRpcException(Server::RPC_INVALID_PARAMETER,
+                                    "Expected Array type");
+  }
+
+  vector<string> vec;
```
ShengguangXiao: Can you add a `using VectorOfString = std::vector<std::string>` in BaseType.h?

KaustubhShamshery: This will require a major refactor as `bool Account::FetchStateJson(Json::Value& root, const string& vname,
                             const vector<string>& indices, bool temp)` is also the same.

src/libServer/JSONConversion.cpp
```diff
@@ -255,6 +255,26 @@ bool JSONConversion::checkJsonTx(const Json::Value& _json) {
...
+
+  for (const auto& ele : _json) {
+    if (!ele.isString()) {
+      throw jsonrpc::JsonRpcException(Server::RPC_INVALID_PARAMETER,
+                                      "Every array value should be a string");
+    }
+
+    vec.push_back(ele.asString());
```
kaikawaliu: Can use `emplace_back`

PR #1813, opened 2019-09-06
seed nodes and l2l to fetch any missing mb from last N epoch and Rejoin optimization

sandipbhoir:
## Description
 - On startup or after recovery, seed nodes and l2l to check and fetch any missing mb from last N epoch 
Addresses - https://github.com/Zilliqa/Issues/issues/568
and missing mb on startup or rejoining for any reasons.
- Also, lagging lookup now check and fetch missing mbs from last N txblks.

- Additionally, when seed nodes misses FB/DSBlock, it now rejoins from S3 based on following condition:
 - Misses DB block
  Always sync from S3

 - Misses Final block
```
   IF Current DS blockNum >  received final block's corresponding DS blockNum: 
          Sync from S3
   ELSE IF Current Tx blockNum > received final block's Tx blockNum :
                    IF  (newlookup or level2lookups): 
                            IF number of missing TxBlocks > NUM_FINAL_BLOCK_NUM:
                                       Sync from S3 
                            ELSE:
                                      Sync from lookup
                    ELSE:
                               Sync from lookup
```
 
- Lookup synchronization to avoid syncing until vacaous epoch. Buffer FB/MB/TXNS while in LOOKUP_SYNC and commit them later so that don't miss the MB/TXNS during recovery window.
 
Note: In `CommitTxBlk` , we also check and find any missing MB/TXNS in last N TxBlks for lookup's/ newlookup/ level2lookup nodes

- Fix issue https://github.com/Zilliqa/Issues/issues/577

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libNode/Node.cpp
```diff
@@ -611,6 +611,15 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
     return false;
   }
 
+  if (LOOKUP_NODE_MODE && ARCHIVAL_LOOKUP &&
```
bb111189: should it be `or` instead of `and` for `LOOKUP_NODE_MODE && ARCHIVAL_LOOKUP`

sandipbhoir: `LOOKUP_NODE_MODE && ARCHIVAL_LOOKUP` means `newlookup` and `level2lookup` types.
`LOOKUP_NODE_MODE && !ARCHIVAL_LOOKUP` means `lookup` types

src/libLookup/Lookup.cpp
```diff
@@ -3128,18 +3150,30 @@ void Lookup::StartSynchronization() {
...
+      while (GetSyncType() != SyncType::NO_SYNC) {
+        GetDSBlockFromSeedNodes(m_mediator.m_dsBlockChain.GetBlockCount(), 0);
+        GetTxBlockFromSeedNodes(m_mediator.m_txBlockChain.GetBlockCount(), 0);
+        this_thread::sleep_for(chrono::seconds(NEW_NODE_SYNC_INTERVAL));
+      }
+    };
+    DetachedFunction(1, func);
+  } else {
+    auto func = [this]() -> void {
+      if (!ARCHIVAL_LOOKUP) {
```
ShengguangXiao: No need to check again with !ARCHIVAL_LOOKUP, this two lamba function can use one, inside can check with ARCHIVAL_LOOKUP

sandipbhoir: Fixed.

src/libLookup/Lookup.cpp
```diff
@@ -2152,6 +2138,42 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
...
 }
 
+void Lookup::FindMissingMBsForLastNTxBlks(const uint32_t& num) {
+  LOG_MARKER();
+  uint64_t upperLimit =
+      m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum();
+  uint64_t lowerLimit = 1;
+
+  if (upperLimit > num) {
+    lowerLimit = upperLimit - num + 1;
```
bb111189: safemath here. will be required in the situation when we launch a new testnet to avoid underflow

sandipbhoir: As discussed, check on `upperLimit and num` avoids underflow.

src/libLookup/Lookup.cpp
```diff
@@ -2140,30 +2137,70 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
       }
       m_currDSExpired = false;
     }
-  } else if (m_syncType == SyncType::NEW_LOOKUP_SYNC) {
-    LOG_EPOCH(INFO, m_mediator.m_currentEpochNum,
-              "New lookup node - Already should have latest state by now.");
+  } else if (m_syncType == LOOKUP_SYNC ||
```
ShengguangXiao: Put SyncType:: infront of LOOKUP_SYNC?

PR #1815, opened 2019-09-06
Fixes for Smart Contract execution on empty map key & RevertCommitTemp

kaikawaliu:
## Description

Basically fixed:
1. Composing JSON State for JS API when encountering empty map entry key
2. Fixed revertCommitTemp with enabling commitRevertible in Test_AccountStore
3. Hide some log under LOG_SC during contract processing
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libPersistence/ContractStorage2.cpp
```diff
@@ -412,6 +412,9 @@ bool ContractStorage2::FetchContractFieldsMapDepth(const dev::h160& address,
 }
 
 void UnquoteString(string& input) {
+  if (input == "\"\"") {
```
ShengguangXiao: 1. Software will crash if input is empty
2. What will happen if input is "\"     \""

kaikawaliu: will be fine. if a space being there the key is not empty after unquote

PR #1821, opened 2019-09-09
Add option to ignore scilla checker for outdated contract

kaikawaliu:
## Description

1) Add option to ignore scilla checker for migration in case some outdated contract affecting the testnet
2) Remove extra quotes during migration to make states inconsistent
3) Enable empty map key in states json api

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/cmd/data_migrate.cpp
+++ b/src/cmd/data_migrate.cpp
@@ -15,6 +15,10 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+#include <iostream>
+
+#include <boost/program_options.hpp>
+
 #include "libData/AccountData/AccountStore.h"
 #include "libMediator/Mediator.h"
 #include "libPersistence/BlockStorage.h"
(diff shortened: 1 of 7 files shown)


src/cmd/data_migrate.cpp
```diff
@@ -25,30 +29,72 @@
...
+#define ERROR_UNHANDLED_EXCEPTION -2
+
 using namespace std;
-int main() {
+namespace po = boost::program_options;
+
+int main(int argc, const char* argv[]) {
   PairOfKey key;  // Dummy to initate mediator
   Peer peer;
+  string ignore_checker_str;
```
ShengguangXiao: ignore_checker_str default should initialize to "false"

kaikawaliu: The bool will be passed by the comparison with the string to "true". so it's unnecessary to initialize.

ShengguangXiao: It is not a bool type, it is a string type

ShengguangXiao: Ok, i find if it is empty, ignore_checker also becomes false

PR #1833, opened 2019-09-10
skip myip while sending to random lookup

sandipbhoir:
## Description
Addresses https://github.com/Zilliqa/Issues/issues/577

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -490,14 +490,15 @@ void Lookup::SendMessageToRandomLookupNode(const bytes& message) const {
     return;
   }
 
-  // To avoid sending message to multiplier
+  // To avoid sending message to multiplier and himself
   VectorOfNode tmp;
   std::copy_if(m_lookupNodes.begin(), m_lookupNodes.end(),
                std::back_inserter(tmp), [this](const PairOfNode& node) {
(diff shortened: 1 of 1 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -490,14 +490,15 @@ void Lookup::SendMessageToRandomLookupNode(const bytes& message) const {
...
   VectorOfNode tmp;
   std::copy_if(m_lookupNodes.begin(), m_lookupNodes.end(),
                std::back_inserter(tmp), [this](const PairOfNode& node) {
-                 return find_if(m_multipliers.begin(), m_multipliers.end(),
-                                [&node](const PairOfNode& mult) {
-                                  return node.second == mult.second;
-                                }) == m_multipliers.end();
+                 return (find_if(m_multipliers.begin(), m_multipliers.end(),
+                                 [&node](const PairOfNode& mult) {
+                                   return node.second == mult.second;
```
ShengguangXiao: The code logic is correct but so hard to read, can implement in a more readable way?

sandipbhoir: Can't think of better readable way than this together with preserving performance ? Please suggest.
Note: added comment though for readability of logic.

PR #1838, opened 2019-09-10
Don't update contract states every epoch & export contract addresses during migration

kaikawaliu:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStore.cpp
+++ b/src/libData/AccountData/AccountStore.cpp
@@ -551,8 +551,14 @@ void AccountStore::RevertCommitTemp() {
   ContractStorage2::GetContractStorage().RevertContractStates();
 }
 
-bool AccountStore::MigrateContractStates(bool ignoreCheckerFailure) {
+bool AccountStore::MigrateContractStates(
+    bool ignoreCheckerFailure, const string& contract_address_output_dir) {
   LOG_MARKER();
+
+  std::ofstream os;
+  if (!contract_address_output_dir.empty()) {
+    os.open(contract_address_output_dir);
+  }
 
   for (const auto& i : m_state) {
     Address address(i.first);
@@ -565,6 +571,9 @@ bool AccountStore::MigrateContractStates(bool ignoreCheckerFailure) {
     }
     if (account.isContract()) {
       account.SetAddress(address);
+      if (!contract_address_output_dir.empty()) {
+        os << address.hex() << endl;
+      }
     } else {
       this->AddAccount(address, account);
       continue;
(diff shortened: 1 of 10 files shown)


src/libPersistence/ContractStorage2.cpp
```diff
@@ -157,8 +157,18 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
...
+  auto d_found = t_indexToBeDeleted.find(key);
+  if (d_found != t_indexToBeDeleted.end()) {
+    // ignore the deleted empty placeholder
+    if ((unsigned int)query.indices().size() == query.mapdepth()) {
+      foundVal = false;
+      return true;
+    }
+  }
+
+  d_found = m_indexToBeDeleted.find(key);
```
KaustubhShamshery: no need of this statement?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -551,9 +551,20 @@ void AccountStore::RevertCommitTemp() {
...
 }
 
-bool AccountStore::MigrateContractStates(bool ignoreCheckerFailure) {
+bool AccountStore::MigrateContractStates(
+    bool ignoreCheckerFailure, const string& contract_address_output_dir,
+    const string& normal_address_output_dir) {
   LOG_MARKER();
 
+  std::ofstream os_1;
+  std::ofstream os_2;
```
ShengguangXiao: Can define as osContract and osNormal? It will be easier to read.

src/libPersistence/ContractStorage2.cpp
ansnunez: Needed I think, since `d_found` is set to something else in line 160

KaustubhShamshery: Correct, ignore

src/libPersistence/ContractStorage2.cpp
```diff
@@ -939,6 +986,7 @@ void ContractStorage2::InitTempState() {
   LOG_MARKER();
 
   t_stateDataMap.clear();
+  t_indexToBeDeleted.clear();
```
ansnunez: Also need to clear in `CommitStateDB()`?

kaikawaliu: CommitStateDB will call InitTempState()

src/libPersistence/ContractStorage2.cpp
kaikawaliu: no need, they share same data structure

src/libData/AccountData/AccountStore.cpp
kaikawaliu: okay maybe i'll change once the current image was built

PR #1845, opened 2019-09-11
Fix recovered lookup start to listening crash

ShengguangXiao:
## Description

Fix recovered lookup start to listening crash

https://github.com/Zilliqa/Issues/issues/578

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


KaustubhShamshery:
after this PR, lookup server wont be able to restart if the pointer is not initialised right?

ShengguangXiao:
> after this PR, lookup server wont be able to restart if the pointer is not initialised right?

After Install success, it will start in Zilliqa.cpp

PR #1852, opened 2019-09-11
[Release-5.0] Don't start lookup server if lookup still syncing

ShengguangXiao:
## Description

Don't start lookup server if lookup still syncing

https://github.com/Zilliqa/Issues/issues/580

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


ShengguangXiao:
Tested in recovered community testnet, after recover and syncing until next DS epoch, it will be synchronized and start listening again.

PR #1857, opened 2019-09-11
[release-5.0] Set syncType based on lookup's syncType if RECOVERY_LOOKUP_SYNC

sandipbhoir:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libZilliqa/Zilliqa.cpp
+++ b/src/libZilliqa/Zilliqa.cpp
@@ -280,6 +280,7 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
         // When doing recovery, make sure to let other lookups know I'm back
         // online
         if (LOOKUP_NODE_MODE) {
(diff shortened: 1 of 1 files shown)


src/libZilliqa/Zilliqa.cpp
```diff
@@ -274,10 +274,12 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
         break;
       case SyncType::RECOVERY_ALL_SYNC:
         LOG_GENERAL(INFO, "Recovery all nodes");
+        if (m_mediator.m_lookup->GetSyncType() == SyncType::RECOVERY_ALL_SYNC) {
```
ShengguangXiao: I think this check will never go true, because we never call SetSyncType(SyncType::RECOVERY_ALL_SYNC)

PR #1863, opened 2019-09-12
AccountStore InitSoft call missing for RejoinAsNewLookup

sandipbhoir:
## Description
Clean variables such that AccountStore::InitSoft should be called

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


sandipbhoir:
closed the PR since lagging node now continue syncing from lookup instead of S3.

PR #1884, opened 2019-09-18
Using bool operator to check address

ShengguangXiao:
## Description

Using bool operator to check address, the current compare method always need to create a temporary Address object, after compare it will be disposed, which is not efficient.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


ansnunez:
@ShengguangXiao it looks like we have `const Address NullAddress` defined in Address.h, and some code still uses `== NullAddress`.  Shall we change all parts to use one way?

ShengguangXiao:
> @ShengguangXiao it looks like we have `const Address NullAddress` defined in Address.h, and some code still uses `== NullAddress`. Shall we change all parts to use one way?

Using "== NullAddress" the readability is better, but using bool operator is more efficient, because it only access one address in memory.

PR #1885, opened 2019-09-18
Check on min. cpu and ram for mining machine

sandipbhoir:
## Description
Issue - https://github.com/Zilliqa/Issues/issues/591

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libUtils/HardwareSpecification.h
+++ b/src/libUtils/HardwareSpecification.h
@@ -25,13 +25,15 @@
 #include "common/Constants.h"
 #include "libUtils/Logger.h"
 
+namespace {
 // RAM specific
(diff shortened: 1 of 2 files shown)


src/libUtils/HardwareSpecification.h
```diff
@@ -0,0 +1,101 @@
...
+
+#include <fstream>
+#include <iostream>
+#include <string>
+
+#include "common/Constants.h"
+#include "libUtils/Logger.h"
+
+// RAM specific
+#define MEMORYINFO_SOURCE_FILE "/proc/meminfo"
```
KaustubhShamshery: this will pollute the global memory space, can we use const declaration inside a nameless namespace?

PR #1889, opened 2019-09-20
No PoW received, network should progress with DS shard 

sandipbhoir:
## Description
This PR will address https://github.com/Zilliqa/Issues/issues/592

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test
  - [x] Individual node recoveries 
  - [x] Entire network recovery from existing running network with no shard
  - [x] Entire network recovery from existing stalled network with no shard

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -408,11 +408,15 @@ void DirectoryService::StartFirstTxEpoch() {
     // m_mediator.m_node->m_myshardId = std::numeric_limits<uint32_t>::max();
     m_mediator.m_node->m_myshardId = m_shards.size();
     m_mediator.m_node->m_justDidFallback = false;
-    m_mediator.m_node->CommitTxnPacketBuffer();
     m_stateDeltaFromShards.clear();
 
     // Start sharding work
     SetState(MICROBLOCK_SUBMISSION);
+
+    auto func1 = [this]() mutable -> void {
+      m_mediator.m_node->CommitTxnPacketBuffer();
+    };
+    DetachedFunction(1, func1);
 
     LOG_STATE(
         "[MIBLKSWAIT]["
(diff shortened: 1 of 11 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -1369,13 +1371,13 @@ Json::Value LookupServer::GetShardingStructure() {
 
     unsigned int num_shards = shards.size();
 
-    if (num_shards == 0) {
+    /* if (num_shards == 0) {
```
KaustubhShamshery: I think can keep this check still?

sandipbhoir: Not really sure.. Do we want to alert user when no shard exist situation could be valid. Since dscommitte always exist. If it’s used only by explorer, we can keep this check.

KaustubhShamshery: If you see the API, it will just return an empty object if there are no shards.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -432,33 +432,46 @@ void DirectoryService::StartFirstTxEpoch() {
...
+          "No other shards. So no other microblocks expected to be received");
+      m_stopRecvNewMBSubmission = true;
+
+      auto func1 = [this]() mutable -> void {
+        m_mediator.m_node->CommitTxnPacketBuffer();
+      };
+      DetachedFunction(1, func1);
+
+      RunConsensusOnFinalBlock();
+    } else {
```
ckyang: Maybe add the `CommitTxnPacketBuffer` calling inside the else case, to be safer

src/libServer/LookupServer.cpp
sandipbhoir: @KaustubhShamshery , `LookupServer::GetBlockchainInfo()` currently fails if we raise exception for `No Shard Yet` and explorer dont update the info. So, we need to fall back where we don't raise exception because its completely valid to have no shards. Other option needs evaluation on api or explorer itself.

src/libServer/LookupServer.cpp
```diff
@@ -1369,13 +1363,10 @@ Json::Value LookupServer::GetShardingStructure() {
 
     unsigned int num_shards = shards.size();
 
-    if (num_shards == 0) {
-      throw JsonRpcException(RPC_IN_WARMUP, "No shards yet");
-    } else {
-      for (unsigned int i = 0; i < num_shards; i++) {
-        _json["NumPeers"].append(static_cast<unsigned int>(shards[i].size()));
-      }
+    for (unsigned int i = 0; i < num_shards; i++) {
```
KaustubhShamshery: As discussed, it is better to keep the check here.

sandipbhoir: Please check the previous comment.

PR #1894, opened 2019-09-30
Save consensus leader after view change

ShengguangXiao:
## Description

Save consensus leader after view change, fix if recover after view change, the DS leader verifiy fail problem.

https://github.com/Zilliqa/Issues/issues/582

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/ViewChangePostProcessing.cpp
+++ b/src/libDirectoryService/ViewChangePostProcessing.cpp
@@ -213,7 +213,7 @@ void DirectoryService::ProcessViewChangeConsensusWhenDone() {
         candidateLeaderInfo.second == m_mediator.m_selfPeer) {
       SetConsensusLeaderID(m_consensusMyID.load());
     } else {
-      DequeOfNode::iterator iterConsensusLeaderID =
+      auto iterConsensusLeaderID =
           find(m_mediator.m_DSCommittee->begin(),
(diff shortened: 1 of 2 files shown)


src/libDirectoryService/ViewChangePostProcessing.cpp
```diff
@@ -287,6 +287,12 @@ void DirectoryService::ProcessViewChangeConsensusWhenDone() {
     return;
   }
 
+  if (!BlockStorage::GetBlockStorage().PutDSCommittee(m_mediator.m_DSCommittee,
```
sandipbhoir: I think also needs to update the same on Lookup's after VC?

ShengguangXiao: Lookup and shard node will call function Node::ProcessVCDSBlocksMessage to process VCDS block, and the PutDSCommittee function is called inside it already.

ShengguangXiao: Fixed, please review again. @sandipbhoir

ShengguangXiao:
Test with recovered community testnet, kill DS leader to trigger view change, after view change success and run for some blocks, to backup persistence before next DS block, and recover another testnet from the persistence, the retrieved the DS leader is the leader after view change.

PR #1897, opened 2019-09-30
Catch potential stoull exception

ShengguangXiao:
## Description

Catch potential stoull exception

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPersistence/Retriever.cpp
+++ b/src/libPersistence/Retriever.cpp
@@ -256,8 +256,17 @@ bool Retriever::RetrieveBlockLink(bool trimIncompletedBlocks) {
         LOG_GENERAL(WARNING, "Get LatestActiveDSBlockNum failed");
         return false;
       }
-      m_mediator.m_ds->m_latestActiveDSBlockNum = std::stoull(
-          DataConversion::CharArrayToString(latestActiveDSBlockNumVec));
+
+      auto dsBlockNumStr =
+          DataConversion::CharArrayToString(latestActiveDSBlockNumVec);
+      try {
+        m_mediator.m_ds->m_latestActiveDSBlockNum = std::stoull(dsBlockNumStr);
+      } catch (const std::exception& e) {
+        LOG_GENERAL(WARNING, "Cannot convert invalid DS block number "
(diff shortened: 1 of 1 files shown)


src/libPersistence/Retriever.cpp
```diff
@@ -256,8 +256,16 @@ bool Retriever::RetrieveBlockLink(bool trimIncompletedBlocks) {
         LOG_GENERAL(WARNING, "Get LatestActiveDSBlockNum failed");
         return false;
       }
-      m_mediator.m_ds->m_latestActiveDSBlockNum = std::stoull(
-          DataConversion::CharArrayToString(latestActiveDSBlockNumVec));
+
+      auto dsBlockNumStr =
+          DataConversion::CharArrayToString(latestActiveDSBlockNumVec);
+      try {
```
KaustubhShamshery: The two exceptions thrown by `stoull` can be caught by `std::exception`. There is no need to use elipses.

ShengguangXiao: Why is the try.. catch for this stoull? I cannot find it...

src/libPersistence/Retriever.cpp
```diff
@@ -256,8 +256,16 @@ bool Retriever::RetrieveBlockLink(bool trimIncompletedBlocks) {
...
+        m_mediator.m_ds->m_latestActiveDSBlockNum = std::stoull(dsBlockNumStr);
+      } catch (...) {
```
KaustubhShamshery:
```suggestion
      } catch (const exception& e) {
```

KaustubhShamshery:
```suggestion
      } catch (const exception &e) {
      LOG_GENERAL(WARNING, "Exception when convert invalid DS block number "
                                 << dsBlockNumStr<<e.what());
```

src/libPersistence/Retriever.cpp
ShengguangXiao: Ok, i see. I fixed with rebase the commit, please review again. @KaustubhShamshery

PR #1898, opened 2019-10-02
Replace libCrypto with Zilliqa/schnorr

ansnunez:
## Description

This PR replaces `libCrypto` with the equivalent implementation residing in the [Zilliqa/schnorr](https://github.com/Zilliqa/schnorr) repository.

Notes:
1. The main motivation for moving out the Schnorr implementation is to promote reuse by our other projects, as well as to be able to independently manage its development/testing.
1. `libCrypto` is completely removed from the build, and only `Sha2.h` is retained.
1. Zilliqa/schnorr is added as a git module under `depends/Schnorr` in the same manner as `protobuf`.
1. Components directly reliant on `Crypto` now link to `Schnorr`.
1. Components directly reliant on OpenSSL (e.g., for SHA-2) now have to additionally link to `crypto` (previously this was indirectly done through linkage with `Crypto`).
1. Components directly reliant on other components such as `boost` and `Util` now have to additionally link to those as well (previously this was indirectly done through linkage with `Crypto`).
1. `Schnorr` components (e.g., `PubKey`, `Signature`) are now derived from `SerializableCrypto` instead of `Serializable`.
1. Almost all `Schnorr` and `MultiSig` functionality and code implementation are preserved inside Zilliqa/schnorr. Externally, there are just 2 breaking changes.
1. Breaking change 1: `Schnorr::GetInstance()` and `MultiSig::GetInstance()` no longer exist. The singleton instances have been removed, and functions have been changed to `static`. This means we can directly call the functions without using `GetInstance()`
1. Breaking change 2: `Serialize` and `Deserialize` now return `bool` (`true` on success), whereas before they returned either an offset or zero.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


tests/Crypto/CMakeLists.txt
```diff
@@ -2,13 +2,13 @@ link_directories(${CMAKE_BINARY_DIR}/lib)
...
+#add_executable(Test_Schnorr Test_Schnorr.cpp)
+#target_link_libraries(Test_Schnorr PUBLIC Crypto)
+#add_test(NAME Test_Schnorr COMMAND Test_Schnorr)
 
-add_executable(Test_MultiSig Test_MultiSig.cpp)
-target_link_libraries(Test_MultiSig PUBLIC Crypto)
-add_test(NAME Test_MultiSig COMMAND Test_MultiSig)
+#add_executable(Test_MultiSig Test_MultiSig.cpp)
+#target_link_libraries(Test_MultiSig PUBLIC Crypto)
+#add_test(NAME Test_MultiSig COMMAND Test_MultiSig)
```
ckyang: We skip schnorr and multisig testing?

ansnunez: > We skip schnorr and multisig testing?

Now these unit tests are moved into the Schnorr repository

PR #1900, opened 2019-10-03
Blacklist at 2 levels strict or relaxed to allow incoming msg

sandipbhoir:
## Description
Handles issue - https://github.com/Zilliqa/Issues/issues/597

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNetwork/Blacklist.h
+++ b/src/libNetwork/Blacklist.h
@@ -46,8 +46,8 @@ class Blacklist {
   std::unordered_map<uint128_t, bool>
       m_blacklistIP;  // IP <-> Strict/Relaxed
                       // Strict -> Blacklisted for both sending and incoming msg
-                      // Relaxed -> Blacklisted for incoming msg onlyy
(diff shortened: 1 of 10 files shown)


src/libNetwork/Blacklist.h
```diff
@@ -43,18 +43,22 @@ class Blacklist {
   void operator=(Blacklist const&) = delete;
 
   std::mutex m_mutexBlacklistIP;
-  std::unordered_map<uint128_t, bool> m_blacklistIP;
-  std::set<uint128_t> m_excludedIP;
+  std::unordered_map<uint128_t, bool>
```
ansnunez: It looks like before this PR we never actually used the `bool` in this map for anything.  So now we're preserving the `map` structure but just making use of the `bool` for the `strict/relaxed`, is that correct?

sandipbhoir: Correct

PR #1902, opened 2019-10-07
Support for blacklist removal message

sandipbhoir:
## Description
This PR provides with support for processing message for removal of node from its blacklist. This will enable restarted node to get back with joining in same ds epoch.

This will be building block for https://github.com/Zilliqa/Issues/issues/596

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libMessage/ZilliqaMessage.proto
```diff
@@ -1060,3 +1060,11 @@ message SeedSetHistoricalDB
     ByteArray signature        = 3;
 
 }
+
+// For removing sender from local blacklist
+message NodeRemoveFromBlacklist
+{
+    ByteArray ipAddress    = 1;
```
bb111189: Need additional information to ensure no replay attack can be done. 

For instance, sig(ipAddress, dsepoch_no)
Ignore request if it exceeds 1 ds epoch.

src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -359,6 +359,11 @@ void DirectoryService::StartFirstTxEpoch() {
   Blacklist::GetInstance().Pop(BLACKLIST_NUM_TO_POP);
   P2PComm::ClearPeerConnectionCount();
 
+  {
+    lock_guard<mutex> g(m_mediator.m_node->m_mutexWhitelistReqs);
```
ansnunez: For easier maintenance (and better encapsulation) might be better to have a function to both acquire the mutex and clear the map

PR #1905, opened 2019-10-10
Shard node try rejoin immediately after being restart

sandipbhoir:
## Description
Shard node try rejoin immediately after being restart, **if it was already part of shard**
Issue - https://github.com/Zilliqa/Issues/issues/596

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -2093,46 +2084,60 @@ void Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
...
-      if (!m_currDSExpired) {
-        if (FinishRejoinAsLookup()) {
-          SetSyncType(SyncType::NO_SYNC);
-
-          if (m_lookupServer) {
-            if (m_lookupServer->StartListening()) {
-              LOG_GENERAL(INFO, "API Server started to listen again");
-            } else {
-              LOG_GENERAL(WARNING, "API Server couldn't start");
+  if (m_syncType == SyncType::NEW_SYNC || m_syncType == SyncType::NORMAL_SYNC) {
```
KaustubhShamshery: The condition for `(m_syncType == SyncType::RECOVERY_ALL_SYNC` was removed. Just want to know what will happen in such case

sandipbhoir: Good catch. However, the check condition is dormant. we should never come across `StartSynchronization` phase with m_syncType `RECOVERY_ALL_SYNC`. Also, `GetStateFromSeedNodes` should have been dormant long back. Lookup should not server complete state.

PR #1921, opened 2019-10-31
Support for fetching coinbase info for specific txblock

sandipbhoir:
## Description
Support for fetching coinbase info for specific txblock.
This is requirement coming from DS Node to rejoin faster ( https://github.com/Zilliqa/Issues/issues/596 )

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -1590,6 +1590,99 @@ void Lookup::SendGetMicroBlockFromLookup(const vector<BlockHash>& mbHashes) {
...
+  if (!LOOKUP_NODE_MODE) {
+    LOG_GENERAL(WARNING,
+                "Function not expected to be called from non-lookup node");
+    return false;
+  }
+
+  // verify if sender is from know DS Committee
+  uint128_t ipSenderAddr = from.m_ipAddress;
+  bool isDSNode = false;
+  for (const auto& node : *m_mediator.m_DSCommittee) {
```
kaikawaliu: Since the complexity is slow here. Is it possible to make DSCommittee a set instead of deque in another PR for m_DSCommittee?

src/libLookup/Lookup.cpp
```diff
@@ -1590,6 +1590,99 @@ void Lookup::SendGetMicroBlockFromLookup(const vector<BlockHash>& mbHashes) {
...
+            << from.GetPrintableIPAddress()
+            << " is not in Present DS Committee list. Ignore the request");
+    return false;
+  }
+
+  uint64_t blockNum;
+  uint32_t portNo = 0;
+  PubKey dsPubKey;
+  if (!Messenger::GetLookupGetCosigsRewardsFromSeed(message, offset, dsPubKey,
+                                                    blockNum, portNo)) {
```
kaikawaliu: Since this only returns the cosigs for only one txblock but most of the scenario multiple txblocks are needed. Shall we make the output list of blockNum here for aggregating the queries? And perhaps a checking of blockNum is needed also  to avoid abusing the api for querying from outdated blocks.

src/libLookup/Lookup.cpp
```diff
@@ -4175,6 +4274,15 @@ bool Lookup::VerifySenderNode(const VectorOfNode& vecLookupNodes,
   return vecLookupNodes.cend() != iter;
 }
 
+bool Lookup::VerifySenderNode(const DequeOfNode& deqLookupNodes,
```
kaikawaliu: Since this function was used for veryfing DS node, shall we rename the parameter here? Maybe just use a more generic name.

src/libLookup/Lookup.cpp
sandipbhoir: Intentionally not allowing asking for multiple tx blocks because after considering all cases, node should not be asking for cosigs for multiple txblocks from any single lookup which will also help to avoid overwhelming. 
Regarding blockNum check, yes it makes sense to check if older than curr ds epoch. will make the change.

src/libLookup/Lookup.cpp
sandipbhoir: yeah. can consider. will make a note,

PR #1922, opened 2019-10-31
Make txblock check in Node::CheckIntegrity threaded

ansnunez:
## Description

The goal of this PR is to speed up the operation of `Node::CheckIntegrity`, particularly in the part where we check each TxBlock for the presence of both microblocks and transaction bodies. The idea is to perform concurrent checking across a fixed number of threads.

1. Changed `BlockStorage::GetAllTxBlocks` result from `list` to `deque` since `deque` has random element access
1. Templatized `Validator::CheckTxBlocks` so that it supports both these containers of txblocks:
   - `deque<TxBlockSharedPtr>` which we get from `BlockStorage::GetAllTxBlocks`
   - `vector<TxBlock>` which is the type of `Lookup::m_txBlockBuffer`
1. Removed `ValidatorBase` because doing step 2 above complicates the hierarchy, and also because we never actually have other derived classes besides `Validator`
1. Removed the step where the sorted `BlockStorage::GetAllTxBlocks` result is moved to a `vector` since this is no longer necessary after step 2 above
1. Added new lambda `validateTxBlock` inside `Node::CheckIntegrity` which takes in one txblock and does the checking on it
1. Modified the txblocks check inside `Node::CheckIntegrity` to basically use a job pool of size `NUMTHREADS`, with each job/thread running the lambda
1. For each txblock processed we also remove it from the `deque` to slowly gain back the memory
1. Once pool capacity is beyond `MAXJOBSLEFT`, we wait indefinitely before adding more jobs

For testing:
1. `CheckIntegrity` change needs to be tested in 2 scenarios
   - [x] `validateDB`
   - [ ] ~~Startup with sync type `DB_VERIF`~~ (no longer in use)
1. `CheckTxBlocks` using `deque` needs to be tested in 2 scenarios
   - [x] `validateDB`
   - [ ] ~~Send `SetDirectoryBlocksFromSeed` with `m_txBlockBuffer` not empty~~ (skipped)
1. `CheckTxBlocks` using `vector` needs to be tested in 1 scenario
   - [x] Send `SetTxBlockFromSeed`

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -351,18 +351,19 @@ bool Node::CheckIntegrity(bool fromIsolatedBinary) {
            std::get<BlockLinkIndex::INDEX>(b);
   });
 
-  std::list<TxBlockSharedPtr> txblocks;
-  if (!BlockStorage::GetBlockStorage().GetAllTxBlocks(txblocks)) {
+  std::deque<TxBlockSharedPtr> txBlocks;
+  if (!BlockStorage::GetBlockStorage().GetAllTxBlocks(txBlocks)) {
     LOG_GENERAL(WARNING, "Failed to get Tx Blocks");
     return false;
   }
 
-  txblocks.sort([](const TxBlockSharedPtr& a, const TxBlockSharedPtr& b) {
-    return a->GetHeader().GetBlockNum() < b->GetHeader().GetBlockNum();
-  });
+  sort(txBlocks.begin(), txBlocks.end(),
+       [](const TxBlockSharedPtr& a, const TxBlockSharedPtr& b) {
+         return a->GetHeader().GetBlockNum() < b->GetHeader().GetBlockNum();
+       });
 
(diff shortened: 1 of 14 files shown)


src/libNode/Node.cpp
```diff
@@ -446,32 +447,57 @@ bool Node::CheckIntegrity(bool fromIsolatedBinary) {
...
         if (!fromIsolatedBinary) {
           break;
         }
       }
     }
-    if (!result && !fromIsolatedBinary) {
+  };
+
+  const unsigned int NUMTHREADS = 10;
+  const int MAXJOBSLEFT = NUMTHREADS * 3;
```
KaustubhShamshery: Do not understand this 3 here? Is it arbitrary?

ansnunez: Yep, just arbitrarily defined the backlog limit to 3 times the number of threads

PR #1923, opened 2019-11-01
Zilliqa Websocket Server

kaikawaliu:
## Description

Implemented the WebsocketServer class.
reference: https://github.com/Zilliqa/dev-docs/blob/master/core/websocket-server.md

Besides, a client executable was written for local testing which works fine now.

For implementation of websocketpp, please refer to: https://github.com/zaphoyd/websocketpp/tree/master/examples

## Backward Compatibility

## Status
Tested with local client on both NewBlock and EventLog.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test (pending)


src/cmd/websocketsubscriber.cpp
```diff
@@ -0,0 +1,191 @@
...
+      std::cerr << e.what() << std::endl << std::endl;
+      return ERROR_UNHANDLED_EXCEPTION;
+    } catch (websocketpp::lib::error_code e) {
+      std::cerr << e.message() << std::endl << std::endl;
+      return ERROR_UNHANDLED_EXCEPTION;
+    } catch (...) {
+      std::cerr << "other exception" << std::endl << std::endl;
+      return ERROR_UNHANDLED_EXCEPTION;
+    }
+  } catch (std::exception& e) {
```
sandipbhoir: can catch as `const ref`

src/libServer/WebsocketServer.cpp
```diff
@@ -0,0 +1,494 @@
...
+    Address addr(log["address"].asString());
+
+    auto find = m_elsockettracker.m_addr_ip_map.find(addr);
+    if (find == m_elsockettracker.m_addr_ip_map.end()) {
+      continue;
+    }
+    Json::Value j_eventlog;
+    j_eventlog["_eventname"] = log["_eventname"];
+    j_eventlog["params"] = log["params"];
+    for (const string& id : find->second) {
```
sandipbhoir: can we rename it to `ip` instead of `id` to make it more obvious?

src/libServer/WebsocketServer.cpp
```diff
@@ -0,0 +1,494 @@
...
+      }
+      break;
+    case EVENTLOG:
+      it = m_eventlog_websockets.find(ip);
+      if (it == m_eventlog_websockets.end()) {
+        return false;
+      }
+      break;
+  }
+
```
sandipbhoir: shall we add default case to handle invalid `it` by returning `false` ?

kaikawaliu: since query only has two options it will never goes into the default case.

src/libServer/LookupServer.cpp
```diff
@@ -1459,6 +1458,16 @@ Json::Value LookupServer::GetTransactionsForTxBlock(const string& txBlockNum) {
 
   auto const& txBlock = m_mediator.m_txBlockChain.GetBlock(txNum);
 
+  return GetTransactionsForTxBlock(txBlock,
+                                   m_mediator.m_lookup->m_historicalDB);
+}
+
+Json::Value LookupServer::GetTransactionsForTxBlock(const TxBlock& txBlock,
```
ansnunez: Any reason for adding this extra level of function call (`GetTransactionsForTxBlock(txBlock) -> GetTransactionsForTxBlock(txBlock, historicalDB)`)?

ansnunez: Any reason for adding this extra level of function call (`GetTransactionsForTxBlock(txBlock) -> GetTransactionsForTxBlock(txBlock, historicalDB)`)?

src/libZilliqa/Zilliqa.cpp
```diff
@@ -336,6 +337,8 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
       m_lookupServer =
           make_shared<LookupServer>(m_mediator, *m_lookupServerConnector);
 
+      (void)WebsocketServer::GetInstance();
```
ansnunez: Should we add check for `ENABLE_WEBSOCKET` before invoking start on the server?

src/libServer/WebsocketServer.h
```diff
@@ -0,0 +1,172 @@
...
+class WebsocketServer : public Singleton<WebsocketServer> {
+ public:
+  /// Returns the singleton AccountStore instance.
+  static WebsocketServer& GetInstance() {
+    static WebsocketServer ws;
+    return ws;
+  }
+
+  /// clean in-memory data structures
+  void clean();
```
ansnunez: I guess your naming convention is to use lower case for private functions?  In that case, we can move this function to the private section perhaps.

src/libServer/WebsocketServer.h
```diff
@@ -0,0 +1,172 @@
...
+
+  /// Send string data to hdl connection
+  bool sendData(const websocketpp::connection_hdl& hdl,
```
ansnunez: I guess your naming convention is to use lower case for private functions?  In that case, we can move this function to the private section perhaps.

src/libServer/LookupServer.cpp
kaikawaliu: Making it static so that can be used as a utility function for WebsocketServer

kaikawaliu: As above

src/libZilliqa/Zilliqa.cpp
```diff
@@ -336,6 +337,10 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
       m_lookupServer =
           make_shared<LookupServer>(m_mediator, *m_lookupServerConnector);
 
+      if (ENABLE_WEBSOCKET) {
```
KaustubhShamshery: I think in a singleton , it gets initialized the first time you call it. The statement seems unneeded

sandipbhoir: I think we need it. If the server `start` fails we reset the ENABLE_WEBSOCKET.
If we rely on getinstance during calling functions, it will be too late otherwise.

PR #1925, opened 2019-11-04
State hashing optimization via buffering

kaikawaliu:
## Description

Instead of hashing whenever a transition being processed. we buffer the address need to be rehashed and doing the heavy lift in the end of transaction processing to avoid duplication. Which should significantly reduce the overhead of processing multiple transactions regarding to single contract with large state. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -1350,8 +1350,9 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(
 
 template <class MAP>
 void AccountStoreSC<MAP>::ProcessStorageRootUpdateBuffer() {
+  std::lock_guard<std::mutex> g(m_mutexUpdateAccounts);
   for (const auto& addr : m_storageRootUpdateBuffer) {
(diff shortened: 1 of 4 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -577,11 +580,18 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
...
 
-  if (Transaction::GetTransactionType(transaction) ==
-          Transaction::CONTRACT_CREATION ||
-      Transaction::GetTransactionType(transaction) ==
-          Transaction::CONTRACT_CALL) {
-    LOG_GENERAL(INFO, "Executing contract transaction finished");
+  switch (Transaction::GetTransactionType(transaction)) {
+    case Transaction::CONTRACT_CALL: {
+      /// since txn succeeded, commit the atomic buffer
+      m_storageRootUpdateBuffer.insert(m_storageRootUpdateBufferAtomic.begin(),
```
sandipbhoir: May be need to clear `m_storageRootUpdateBuffer` at some point ?

PR #1926, opened 2019-11-05
Enable ds node re-join faster not waiting until next vacuous epoch

sandipbhoir:
## Description
To allow ds node re-join faster not waiting until next vacous epoch.
https://github.com/Zilliqa/Issues/issues/596

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] small-scale cloud test
     - [x] Recovery of DSGuard at random epoch -> Recovers -> May miss FB -> VC Precheck fails -> RejoinAsDS in next epoch
     - [x] Recovery of DSGuard at POW_SUBMISSION window -> Recovers -> Wait for FB announcement -> May get in to VC -> VC Precheck fails -> RejoinAsDS in next epoch.
     - [x] DoRejoin DsGuard by sending rejoining at FB message 
     - [x] DoRejoin Non-DsGuard by sending rejoining at FB message
     - [x] DoRejoin DsGuard by sending rejoining at FB message at FirstTxEpoch
     - [x] DoRejoin Non-DsGuard by sending rejoining at FB message at FirstTxEpoch
     - [x] Scale new node which joins the DS Committe and DoRejoin by sending rejoining at FB message at FirstTxEpoch
     - [x] Scale new node which joins the DS Committe and DoRejoin by sending rejoining at FB message at any other epoch


tests/Incentives/test_coinbase.cpp
```diff
@@ -85,9 +86,11 @@ BOOST_AUTO_TEST_CASE(test_coinbase_correctness) {
 
   *mediator.m_DSCommittee = dummy_ds_comm;
   dummyDS.m_shards = dummy_shards;
+  DSBlock lastBlock =
+      DSBlock(TestUtils::createDSBlockHeader(1), CoSignatures());
+  mediator.m_dsBlockChain.AddBlock(lastBlock);
 
-  const uint num_test_epoch =
-      1;  // Can be changed to NUM_FINAL_BLOCK_PER_POW for proper testing
+  const uint num_test_epoch = NUM_FINAL_BLOCK_PER_POW;
```
KaustubhShamshery: Need to change back??

src/libLookup/Lookup.cpp
```diff
@@ -1655,7 +1655,8 @@ bool Lookup::ProcessGetCosigsRewardsFromSeed(
   uint32_t retryCount = 5;
   while (retryCount-- > 0) {
     if (!BlockStorage::GetBlockStorage().GetTxBlock(blockNum, txblkPtr)) {
-      LOG_GENERAL(WARNING, "Failed to fetch tx block, retry... " << blockNum);
+      LOG_GENERAL(WARNING,
```
KaustubhShamshery: Do we need some verification for this message, eg Checking pubkey of lookup??

tests/Incentives/test_coinbase.cpp
sandipbhoir: No. Test case fails. `InitCoinbase` is changed. Actually coinbase for epoch 1 is un-realistic.

src/libLookup/Lookup.cpp
sandipbhoir: This message is processed by lookup itself. This message is verified for sender's pubkey to be part  ds committee.

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -90,7 +97,7 @@ void DirectoryService::StartSynchronization(bool clean) {
   };
 
   auto func2 = [this]() -> void {
-    if (!m_mediator.m_lookup->GetDSInfoLoop()) {
+    while (!m_mediator.m_lookup->GetDSInfoLoop()) {
       LOG_GENERAL(WARNING, "Unable to fetch DS info");
```
ckyang: Should we need to sleep here before trying again? To avoid some kind of busy loop

src/libLookup/Lookup.cpp
```diff
@@ -1670,17 +1671,32 @@ bool Lookup::ProcessGetCosigsRewardsFromSeed(
...
   for (const auto& mbInfo : microblockInfos) {
+    if (mbInfo.m_shardId ==
+        m_mediator.m_ds->GetNumShards()) {  // ignore ds microblock
+      continue;
+    }
     MicroBlockSharedPtr mbptr;
-    if (!BlockStorage::GetBlockStorage().GetMicroBlock(mbInfo.m_microBlockHash,
-                                                       mbptr)) {
-      cout << "Could not get MicroBlock " << mbInfo.m_microBlockHash << endl;
+    retryCount = 5;
```
ckyang: For the hard-code retryCount, maybe we can make it meaningful, and also combine with previous retryCount (in line 1655) as well.

src/libDirectoryService/DirectoryService.cpp
sandipbhoir: `GetDSInfoLoop` itself wait and retries multiple time. So we don't need additional sleep time over here.

src/libLookup/Lookup.cpp
sandipbhoir: retryCount has to be used separately for txBlk and microblocks. For meaninfulness, you meant move it to local constant?

src/libDirectoryService/DirectoryService.cpp
ckyang: ok

src/libLookup/Lookup.cpp
ckyang: In my opinion, no need to put into constant file. Just give a meaningful name by using `#define` / `const variable` should be enough

PR #1932, opened 2019-11-14
Allow consensus leader to send to node in relaxed blacklist

ansnunez:
## Description

This is the change needed before we can re-enable the relaxed blacklist currently disabled in #1917 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNetwork/P2PComm.h
+++ b/src/libNetwork/P2PComm.h
@@ -51,7 +51,7 @@ class SendJob {
   unsigned char m_startbyte{};
   bytes m_message;
   bytes m_hash;
-  bool m_allowSendToRelaxedBlacklist;
+  bool m_allowSendToRelaxedBlacklist{};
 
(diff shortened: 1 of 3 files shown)


src/libNetwork/P2PComm.h
```diff
@@ -51,6 +51,7 @@ class SendJob {
   unsigned char m_startbyte{};
   bytes m_message;
   bytes m_hash;
+  bool m_allowSendToRelaxedBlacklist{};
```
bb111189: set to atomic?

sandipbhoir: I don't think we need that in this case. Its as good as other params only set once in
https://github.com/Zilliqa/Zilliqa/blob/7e37c8cfd771f85cb799226156848241c16f896c/src/libNetwork/P2PComm.cpp#L821
Thread safety is fine.

PR #1936, opened 2019-11-22
Scilla multi-message support

kaikawaliu:
## Description

- We are adopting depth first search for the processing of messages returned from Scilla Interpreter. Thus the change is basically looping the previous process of message output.
- In this way, we are removing the restriction of chain call depth since it's becoming a tree structure.
- For the new format of scilla message, please refer to an example here:
https://github.com/Zilliqa/scilla/blob/387_multiple_send_msg/tests/runner/multiple-msgs/output_1.json

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -1149,200 +1148,200 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(
...
+      if (ret) {
+        // return true;
+        continue;
+      }
 
-  if (ret) {
-    return true;
-  }
+      LOG_GENERAL(INFO, "Call another contract");
+      receipt.AddDepth();
```
KaustubhShamshery: Is this cleared when a new member in the `message` is accessed??

kaikawaliu: the depth now actually means how many times of interpreter being called.

KaustubhShamshery: Would not that lead to issues in error reporting ? anyways not a blocker

kaikawaliu: it will just tell user your error is happened in which transition, basically this line is not newly added, I just moved it from another line to here

PR #1951, opened 2019-12-02
Python Runner

KaustubhShamshery:
## Description

Call Python scripts from inside binary

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


scripts/download_incr_DB.py
```diff
@@ -219,9 +219,8 @@ def run():
 			continue
 
 	print("[" + str(datetime.datetime.now()) + "] Done!")
-	exit(0)
-
-if __name__ == "__main__":
+	return True
```
sandipbhoir: leave a line separator

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -1023,7 +1023,7 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSBackup() {
     return true;
   }
 
-#ifdef VC_TEST_VC_PRECHECK_2
+#if 1
```
sandipbhoir: assuming this will be reverted.

src/libNode/Node.cpp
```diff
@@ -151,14 +151,9 @@ Node::~Node() {}
...
   LOG_MARKER();
-  string output;
-  // TBD - find better way to capture the exit status of command
   string excludembtxns = LOOKUP_NODE_MODE ? "false" : "true";
-  SysCommand::ExecuteCmdWithOutput(boost::filesystem::current_path().string() +
-                                       "/" + "download_incr_DB.py " +
-                                       STORAGE_PATH + "/ " + excludembtxns,
-                                   output);
-  return (output.find("Done!") != std::string::npos);
+  return PythonRunner::RunPyFunc("download_incr_DB", "start",
```
sandipbhoir: Do we need to pass path to download script while calling func from it???

docker/Dockerfile.cuda
```diff
@@ -46,10 +48,13 @@ RUN apt-get update \
...
     python \
     python-pip \
+    python3-setuptools \
+    python3-pip \
     libsecp256k1-dev \
     && rm -rf /var/lib/apt/lists/* \
     && pip install setuptools \
-    && pip install request requests clint futures
+    && pip install request requests clint futures \
+    && pip3 install request requests clint futures
```
sandipbhoir: should be added to `Dockerfile` as well?

src/libZilliqa/Zilliqa.cpp
```diff
@@ -190,6 +190,21 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
...
+      this_thread::sleep_for(chrono::seconds(RETRY_REJOINING_TIMEOUT));
+    }
+    if (!BlockStorage::GetBlockStorage().RefreshAll()) {
+      LOG_GENERAL(WARNING, "BlockStorage::RefreshAll failed");
+    }
+    if (!AccountStore::GetInstance().RefreshDB()) {
+      LOG_GENERAL(WARNING, "AccountStore::RefreshDB failed");
+    }
+  }
+
```
ckyang: Why we need this logic here for rejoin?

KaustubhShamshery: New node when joins need to refresh DB after downloading persistence

PR #1969, opened 2019-12-19
New mining node to block message before its ready

sandipbhoir:
## Description
This PR will fix the corner case where new node can potentially accept any message before its ready.
Issue - https://github.com/Zilliqa/Issues/issues/608

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libZilliqa/Zilliqa.cpp
+++ b/src/libZilliqa/Zilliqa.cpp
@@ -172,6 +172,11 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
     LOG_GENERAL(FATAL, "Archvial lookup is true but not lookup ");
   }
 
+  if(SyncType::NEW_SYNC == syncType){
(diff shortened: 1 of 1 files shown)


src/libZilliqa/Zilliqa.cpp
```diff
@@ -172,6 +172,11 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
     LOG_GENERAL(FATAL, "Archvial lookup is true but not lookup ");
   }
 
+  if (SyncType::NEW_SYNC == syncType) {
+    // Setting it earliest before even p2pcomm is instantiated
+    m_n.m_runFromLate = true;
```
ansnunez: Does this mean we can remove line 274?

sandipbhoir: I had preferred to kept it safer because there are other calls in between which might change the syncType.

PR #1975, opened 2020-01-12
Improve logs around benchmarking

KaustubhShamshery:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libServer/LookupServer.cpp
```diff
@@ -258,29 +258,40 @@ bool LookupServer::StartCollectorThread() {
...
-        for (const auto& tx : txns) {
-          m_mediator.m_lookup->AddToTxnShardMap(tx, 0);
+      if (!txnsShard.empty()) {
+        for (const auto& tx : txnsShard) {
+          m_mediator.m_lookup->AddToTxnShardMap(tx,
+                                                SEND_TYPE::ARCHIVAL_SEND_SHARD);
         }
       }
-      // LOG_GENERAL(INFO, "Size of txns " << txns.size());
+      LOG_GENERAL(INFO, "Size of txns to shard" << txnsShard.size());
```
ansnunez, line 287: Can you move this line into the code block immediately above? So that in mainnet we don't keep seeing `Size of txns to shard0` (also, missing a space after "shard")

src/libServer/LookupServer.cpp
```diff
@@ -258,29 +258,40 @@ bool LookupServer::StartCollectorThread() {
...
+      if (!txnsDS.empty()) {
+        for (const auto& tx : txnsDS) {
+          m_mediator.m_lookup->AddToTxnShardMap(tx,
+                                                SEND_TYPE::ARCHIVAL_SEND_DS);
+        }
+      }
+      LOG_GENERAL(INFO, "Size of txns to DS" << txnsDS.size());
```
ansnunez, line 294: Can you move this line into the code block immediately above? So that in mainnet we don't keep seeing `Size of txns to DS0` (also, missing a space after "DS")

scripts/scan_consolidate.py
```diff
@@ -0,0 +1,273 @@
+#!/bin/bash
+# Copyright (C) 2019 Zilliqa
```
ansnunez, line 2:
```suggestion
# Copyright (C) 2020 Zilliqa
```

scripts/scan_consolidate.py
```diff
@@ -0,0 +1,273 @@
+#!/bin/bash
```
ansnunez, line 1: Did you mean python3?

src/libLookup/Lookup.cpp
```diff
@@ -325,9 +326,9 @@ bool Lookup::GenTxnToSend(size_t num_txn, vector<Transaction>& txn) {
       continue;
     }
 
-    copy(txns.begin(), txns.end(), back_inserter(txn));
+    copy(txns.begin(), txns.end(), back_inserter(DSTxn));
   }
-  return !txn.empty();
+  return !(shardTxn.empty() || DSTxn.empty());
```
saayan-zilliqa, line 331: should it be && ?
both shardTxn and DSTxn should be empty, right?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -623,6 +624,23 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
...
+  if (LOOKUP_NODE_MODE && LOG_PARAMETERS) {
+    LOG_STATE("[FBSIZE] Size: " << message.size() << " BlockNum: "
+                                << txBlock.GetHeader().GetBlockNum());
+
+    uint64_t timeDiff = txBlock.GetTimestamp() -
+                        m_mediator.m_txBlockChain.GetLastBlock().GetTimestamp();
+
+    cpp_dec_float_50 td_float(timeDiff);
+    cpp_dec_float_50 numTxns(txBlock.GetHeader().GetNumTxs());
+    td_float = td_float / 1000000.0;
```
saayan-zilliqa, line 636: Use some predefined constant (not #define) instead of magic number 1000000

src/libLookup/Lookup.cpp
ansnunez: Should be both not empty, so I think || is correct

saayan-zilliqa: Yes, that's fine then

PR #1986, opened 2020-01-17
[v6.1.0] Use Buffered address when append and add DEPTH in transition field

kaikawaliu:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


bb111189:
Need to fix format

PR #1987, opened 2020-01-17
Use Buffered address when append and add DEPTH in transition field

kaikawaliu:
## Description

- Change m_curContractAddr -> curContractAddr to avoid chain call affecting the current caller address;
- Change m_depth -> m_edge with more relevant meaning
- Add "depth" field in receipt transition field to indicate the current depth in the transition tree of the message

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


bb111189:
Need to fix format

PR #2003, opened 2020-02-20
Adding accepted field in receipt

kaikawaliu:
## Description

In order to know if a contract sent money to another contract, whether the second contract accepted the money, we have this PR.

**Notice**:
- The `accepted` field not inside transitions are regarding to the first call invoked by the transaction. 
- For a pure payment transition, which means it won't invoke scilla interpreter, there won't be `accepted` field.

**Example**:
```
"receipt": {
	"accepted" : false,
	"cumulative_gas" : "1118",
	"success" : true,
	"transitions" : 
	[
		{
			"addr" : "0xff5793e4bde5432d2206f5acfae2d6cdb19a4e38",
			"depth" : 0,
                        "accepted" : false,
			"msg" : 
			{
				"_amount" : "0",
				"_recipient" : "0x604e8b191ebdb89b762028d6a5bbfe58e8119428",
				"_tag" : "Transfer",
				"params" : 
				[
					{
						"type" : "ByStr20",
						"value" : "0x9bfec715a6bd658fcb62b0f8cc9bfa2ade71434a",
						"vname" : "to"
					},
					{
						"type" : "Uint128",
						"value" : "100",
						"vname" : "tokens"
					}
				]
			}
		},
		{
			"addr" : "0x604e8b191ebdb89b762028d6a5bbfe58e8119428",
			"depth" : 0,
			"msg" : 
			{
				"_amount" : "0",
				"_recipient" : "0x712e8b191ebdb89b762028d6a5bbfe58e8119283",
				"_tag" : "",
				"params" : []
			}
		}
	]
}
```
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/TransactionReceipt.cpp
+++ b/src/libData/AccountData/TransactionReceipt.cpp
@@ -109,6 +109,20 @@ void TransactionReceipt::AddTransition(const Address& addr,
   m_tranReceiptObj["transitions"].append(_json);
 }
 
+void TransactionReceipt::AddAccepted(bool accepted) {
+  m_tranReceiptObj["accepted"] = accepted;
+}
+
+bool TransactionReceipt::AddAcceptedForLastTransition(bool accepted) {
+  LOG_MARKER();
+  if (m_tranReceiptObj["transitions"].empty()) {
+    return false;
+  }
+  m_tranReceiptObj["transitions"][(m_tranReceiptObj["transitions"].size() - 1)]
+                  ["accepted"] = accepted;
+  return true;
+}
+
 void TransactionReceipt::RemoveAllTransitions() {
   m_tranReceiptObj.removeMember("transitions");
 }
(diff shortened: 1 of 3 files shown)


src/libData/AccountData/TransactionReceipt.cpp
```diff
@@ -109,6 +109,16 @@ void TransactionReceipt::AddTransition(const Address& addr,
   m_tranReceiptObj["transitions"].append(_json);
 }
 
+void TransactionReceipt::AddAccepted(bool accepted) {
+  m_tranReceiptObj["accepted"] = accepted;
+}
+
+void TransactionReceipt::AddAcceptedForLastTransition(bool accepted) {
+  LOG_MARKER();
+  m_tranReceiptObj["transitions"][(m_tranReceiptObj["transitions"].size() - 1)]
```
ansnunez: Is there a possibility of transitions array being empty?

kaikawaliu: This will only be called when there were `Last Transition`, so it won't be empty. But agree to add a check to be safer.

kaikawaliu: Addressed @ansnunez

PR #2013, opened 2020-03-05
Commit deleted trie nodes from mem to disk & compact LevelDB by reopening

kaikawaliu:
## Description

This PR resolves the size limit of account state trie. 

## Motivation
We used to observe an unlimited growth of disk usage of state trie DB even there is only update operation to it. The old solution is to repopulate the whole trie in the end of a DS epoch, but this may take too long to process if the size of accounts in trie is getting huge. Which is not what we want in the long run.

## Rationale
In the old dependency, the deleted trie nodes were not committed to the disk when we update a specific node in the Merkle Partria Tree. In this fix, we dump all the deleted nodes in memory and delete from LevelDB.

Besides, the deletion doesn't take effects in LevelDB until running compaction. Which is normally triggered when we reopen the DB or run compact mannualy. Since compaction won't eliminate LOG and MANIFEST files which grows unlimitedly as long as the process is running, but which will be reset when the LevelDB is reopened. So reopening by recreating instance linking to the DB in memory solves this issue.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/depends/libDatabase/LevelDB.cpp
+++ b/src/depends/libDatabase/LevelDB.cpp
@@ -40,16 +40,16 @@ LevelDB::LevelDB(const string& dbName, const string& path, const string& subdire
         return;
     }
 
-    leveldb::Options options;
-    options.max_open_files = 256;
-    options.create_if_missing = true;
+    m_options.max_open_files = 256;
+    m_options.create_if_missing = true;
 
     leveldb::DB* db;
     leveldb::Status status;
 
     if(m_subdirectory.empty())
     {
-        status = leveldb::DB::Open(options, path + "/" + this->m_dbName, &db);
+        m_open_db_path = path + "/" + this->m_dbName;
+        status = leveldb::DB::Open(m_options, m_open_db_path, &db);
         LOG_GENERAL(INFO, path + "/" + this->m_dbName);
     }
     else
@@ -58,8 +58,9 @@ LevelDB::LevelDB(const string& dbName, const string& path, const string& subdire
         {
             boost::filesystem::create_directories(path + "/" + this->m_subdirectory);
         }
-        status = leveldb::DB::Open(options, 
-            path + "/" + this->m_subdirectory + "/" + this->m_dbName,
+        m_open_db_path = path + "/" + this->m_subdirectory + "/" + this->m_dbName;
+        status = leveldb::DB::Open(m_options, 
+            m_open_db_path,
             &db);
         LOG_GENERAL(INFO, path + "/" + this->m_subdirectory + "/" + this->m_dbName);
     }
@@ -77,9 +78,8 @@ LevelDB::LevelDB(const std::string & dbName, const std::string& subdirectory, bo
     this->m_subdirectory = subdirectory;
     this->m_dbName = dbName;
 
-    leveldb::Options options;
-    options.max_open_files = 256;
-    options.create_if_missing = true;
+    m_options.max_open_files = 256;
+    m_options.create_if_missing = true;
 
     leveldb::DB* db;
     leveldb::Status status;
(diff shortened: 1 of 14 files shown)


src/libPersistence/ContractStorage2.cpp
```diff
@@ -277,6 +277,8 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
     }
   }
 
+  delete it;
```
ansnunez, line 280: There's some return statements above which will skip iterator deletion.

May I suggest to use something similar to what we do in [P2PComm](https://github.com/Zilliqa/Zilliqa/blob/ead49ba5d26b9526ab231de7de5afd42198fcbfc/src/libNetwork/P2PComm.cpp#L196)

For example:
```
void DeleteLevelDBIterator(leveldb::Iterator* it)
{
  delete it;
}

leveldb::Iterator* it = m_diagnosticDBNodes->GetDB()->NewIterator(leveldb::ReadOptions());
unique_ptr<leveldb::Iterator*, void (*)(leveldb::Iterator*)> ldbiter_deleter(it, DeleteLevelDBIterator);
```

This way we don't need to manually call `delete it` before every return statement.

src/depends/libDatabase/LevelDB.cpp
```diff
@@ -334,8 +344,22 @@ bool LevelDB::BatchInsert(const std::unordered_map<std::string, std::string>& kv
...
+
+bool LevelDB::BatchDelete(const std::vector<dev::h256>& toDelete) {
+    ldb::WriteBatch batch;
+    for (const auto& i : toDelete) {
+        batch.Delete(leveldb::Slice(i.hex()));
+    }
+
+    ldb::Status s = m_db->Write(leveldb::WriteOptions(), &batch);
+
+    if (!s.ok()) {
```
ansnunez, line 362: Can use `return s.ok();`
Please make the same improvement in the other functions that return bool too.

src/depends/libDatabase/MemoryDB.cpp
```diff
@@ -147,25 +147,51 @@ namespace dev
         m_aux[_h] = make_pair(_v.toBytes(), true);
     }
 
-    void MemoryDB::purge()
+    void MemoryDB::purge(std::vector<h256>& purged, bool haveMutex)
```
ansnunez, line 150: `haveMutex` name seems confusing because if it's true then you create mutex.
Maybe rename to `calledWithoutMutex` to be clearer.

Also, I suggest to rewrite like this to avoid duplicate code:

```
void MemoryDB::purgeMain(std::vector<h256>& purged)
{
  // code for purge m_main
  // code for purge m_aux
}

void MemoryDB::purge(std::vector<h256>& purged, bool calledWithoutMutex)
{
  if (calledWithoutMutex) {
    unique_lock<shared_timed_mutex> lock(x_this);
    purgeMain(purged);
  }
  else {
    purgeMain(purged);
  }
}
```

src/depends/libDatabase/MemoryDB.cpp
```diff
@@ -147,25 +147,51 @@ namespace dev
...
+        /// can be without the mutex
+        if (haveMutex) {
+            unique_lock<shared_timed_mutex> lock(x_this);
+            // purge m_main
+            for (auto it = m_main.begin(); it != m_main.end(); ) {
+                if (it->second.second) {
+                    ++it;
+                } else {
+                    // LOG_GENERAL(INFO, "purged: " << it->first.hex())
+                    purged.push_back(it->first);
```
ansnunez, line 162: Can use `emplace_back`

src/libPersistence/ContractStorage2.cpp
kaikawaliu: Simply replaced with unique_ptr for garbage collection

src/depends/libDatabase/LevelDB.cpp
kaikawaliu: Added some log under the false case

src/depends/libDatabase/MemoryDB.cpp
kaikawaliu: thx, fixed

src/depends/libDatabase/MemoryDB.cpp
kaikawaliu: noted

PR #2016, opened 2020-03-12
Optimize StartRetrieveHistory->RetrieveTxBlocks

kaikawaliu:
## Description

Related Issue:
https://github.com/Zilliqa/Issues/issues/614 

Actually the retrieval of all the txn blocks into memory is not essential, which were used to serve getting the lastest block number...
Besides, `CircularArray` was also modified so no need to populate the array with every blocks in history to get the latest view of the blockchain.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


sandipbhoir:
Looks good. Please run the small scale test with at least recovery of `ds node` and `level2lookup`

PR #2040, opened 2020-04-17
Add missing deletion

ansnunez:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libZilliqa/Zilliqa.cpp
```diff
@@ -452,5 +452,6 @@ void Zilliqa::Dispatch(pair<bytes, Peer>* message) {
   // Queue message
   if (!m_msgQueue.bounded_push(message)) {
     LOG_GENERAL(WARNING, "Input MsgQueue is full");
+    delete message;
```
kaikawaliu, line 455: The deletion of account, prototxn and job is fine.
But the message in Zilliqa::Dispatch should be handled with more care, since it takes a pointer externally, deleting it might be dangerous if the pointer is further reused after the function, it's better to make it wrapped with shared_ptr

PR #2059, opened 2020-04-24
[release-6.2] Add exception handling and logging to auto_backup.py 

sandipbhoir:
## Description
Reference: Zilliqa/Issues#618

- Added logging to file auto_backup-log.txt
- Added exception handling to avoid crash and log the exception and continue.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/scripts/auto_backup.py
+++ b/scripts/auto_backup.py
@@ -22,12 +22,38 @@ import json
 import socket
 import tarfile
 import xml.etree.cElementTree as xtree
+import math
+import shutil
+import logging
+from logging import handlers
 
 TAG_NUM_FINAL_BLOCK_PER_POW = "NUM_FINAL_BLOCK_PER_POW"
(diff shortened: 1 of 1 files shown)


scripts/auto_backup.py
```diff
@@ -133,20 +160,27 @@ def main():
...
+                logging.info("Starting to back-up persistence at blockNum : %s" % curr_blockNum)
+                backUp(curr_blockNum)
+                logging.info("Backing up persistence successfully.")
+                isBackup = True
+
+            time.sleep(frequency)
+        except Exception as e:
+            logging.warning(e)
+            time.sleep(frequency)
+            pass
```
bb111189, line 183: actually no need pass here

ansnunez: ok, for removal next time

PR #2072, opened 2020-05-03
Add txn log to ws

KaustubhShamshery:
## Description

Add TXN LOG to websocket query

The query takes input an address and if a txn is sent from or to that account, displays the following

ID,amount,success,fromAddr,toAddr

```json
{"query" : "TxnLog" , "addresses" : ["0xb50c2404e699fd985f71b2c3f032059f13d6543b"]}
< {"query" : "TxnLog" , "addresses" : ["0xb50c2404e699fd985f71b2c3f032059f13d6543b"]}
< {
	"type" : "Notification",
	"values" : 
	[
		{
			"query" : "TxnLog",
			"value" : 
			[
				{
					"address" : "b50c2404e699fd985f71b2c3f032059f13d6543b",
					"log" : 
					[
						{
							"ID" : "cb773dbd95d2ec7a4ec9539f92dc92c7f62d515fcc9b052e0b64bc06dde50e7b",
							"amount" : "2718000000000",
							"fromAddr" : "42fb82623b9ea0b9dbf41e74304a39908a378cfd",
							"success" : true,
							"toAddr" : "b50c2404e699fd985f71b2c3f032059f13d6543b"
						}
					]
				}
			]
		}
	]
}

```

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libServer/WebsocketServer.cpp
+++ b/src/libServer/WebsocketServer.cpp
@@ -49,6 +49,15 @@ std::map<websocketpp::connection_hdl, std::unordered_map<Address, Json::Value>,
          std::owner_less<connection_hdl>>
     WebsocketServer::m_eventLogDataBuffer;
 
+std::mutex WebsocketServer::m_mutexTxnLogDataBuffer;
+
+std::map<websocketpp::connection_hdl, std::unordered_map<Address, Json::Value>,
+         std::owner_less<websocketpp::connection_hdl>>
+    WebsocketServer::m_txnLogDataBuffer;
+
+EventLogAddrHdlTracker WebsocketServer::m_txnLogAddrHdlTracker;
+std::mutex WebsocketServer::m_mutexTxnLogAddrHdlTracker;
+
 bool WebsocketServer::start() {
   LOG_MARKER();
   clean();
@@ -146,6 +155,14 @@ void WebsocketServer::clean() {
     lock_guard<mutex> g(m_mutexTxnBlockNTxnHashes);
     m_jsonTxnBlockNTxnHashes.clear();
   }
+  {
+    lock_guard<mutex> g(m_mutexTxnLogDataBuffer);
+    m_txnLogDataBuffer.clear();
+  }
+  {
+    lock_guard<mutex> g(m_mutexTxnLogAddrHdlTracker);
+    m_txnLogAddrHdlTracker.clear();
+  }
 }
 
 bool WebsocketServer::sendData(const connection_hdl& hdl, const string& data) {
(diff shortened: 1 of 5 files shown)


src/libServer/WebsocketServer.cpp
```diff
@@ -290,8 +349,56 @@ void WebsocketServer::PrepareTxBlockAndTxHashes(
...
+  }
+  const auto log_json = CreateReturnAddressJson(twr);
+
+  lock_guard<mutex> h(m_mutexTxnLogDataBuffer);
+  for (const connection_hdl& hdl : find_addr->second) {
+    m_txnLogDataBuffer[hdl][addr_confirmed].append(log_json);
+  }
+}
+
+void WebsocketServer::ParseTxn(const TransactionWithReceipt& twr) {
```
ansnunez, line 393: I'm concerned about the performance here, because each transaction is processed twice and it looks like both ParseTxnxxx functions have common code blocks (including mutex acquisition and map searches). Can you merge the two functions instead to make this more efficient?

src/libServer/WebsocketServer.cpp
```diff
@@ -260,10 +282,47 @@ void WebsocketServer::on_message(const connection_hdl& hdl,
         }
         break;
       }
+      case TXNLOG: {
+        set<Address> track_addresses;
+        if (j_query.isMember("addresses") && j_query["addresses"].isArray() &&
+            !j_query["addresses"].empty()) {
+          for (const auto& addr_str : j_query["addresses"]) {
+            try {
+              if (!JSONConversion::checkStringAddress(addr_str.asString())) {
```
ansnunez, line 291: In normal case where the address is valid, `asString` is called twice. Better to store the result in a local variable to be more efficient.

src/libServer/WebsocketServer.cpp
KaustubhShamshery: The mutex acquistions and search variables are different, if you look closely.

src/libServer/WebsocketServer.cpp
kaikawaliu, line 291: Actually if the address is not valid, it will be caught in the try catch because `Address addr(addr_str.asString())` throw error, so maybe the `checkStringAddress` here is not that necessary?

KaustubhShamshery: It actually does not check the address size (for eg, 0x11 is a valid address for conversion)

kaikawaliu: Oh, in that case would you mind to help to apply that to EVENTLOG as well? Thank you!

KaustubhShamshery: hey the EVENTLOG code already has  
```
if (acc == nullptr || !acc->isContract()) {
```

PR #2077, opened 2020-05-08
Use smart pointer instead of raw pointer

ShengguangXiao:
## Description

Use smart pointer instead of raw pointer, don't repeat allocate same object inside a loop

Optimization

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libMessage/Messenger.cpp
+++ b/src/libMessage/Messenger.cpp
@@ -4810,12 +4810,12 @@ bool Messenger::SetNodeForwardTxnBlock(
 
   unsigned int txnsCurrentCount = 0, txnsGeneratedCount = 0, msg_size = 0;
 
-  const auto protoTxn = std::make_unique<ProtoTransaction>();
   for (const auto& txn : txnsCurrent) {
     if (msg_size >= PACKET_BYTESIZE_LIMIT) {
       break;
     }
 
+    auto protoTxn = std::make_unique<ProtoTransaction>();
     TransactionToProtobuf(txn, *protoTxn);
     unsigned txn_size = protoTxn->ByteSize();
     if ((msg_size + txn_size) > PACKET_BYTESIZE_LIMIT &&
@@ -4832,6 +4832,7 @@ bool Messenger::SetNodeForwardTxnBlock(
       break;
     }
 
+    auto protoTxn = std::make_unique<ProtoTransaction>();
     TransactionToProtobuf(txn, *protoTxn);
     unsigned txn_size = protoTxn->ByteSize();
     if ((msg_size + txn_size) > PACKET_BYTESIZE_LIMIT &&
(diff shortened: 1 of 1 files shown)


src/libMessage/Messenger.cpp
```diff
@@ -4808,16 +4808,14 @@ bool Messenger::SetNodeForwardTxnBlock(
...
+  unsigned int txnsCurrentCount = 0, txnsGeneratedCount = 0, msg_size = 0;
 
+  const auto protoTxn = std::make_unique<ProtoTransaction>();
   for (const auto& txn : txnsCurrent) {
     if (msg_size >= PACKET_BYTESIZE_LIMIT) {
       break;
     }
-    ProtoTransaction* protoTxn = new ProtoTransaction();
+
     TransactionToProtobuf(txn, *protoTxn);
```
sandipbhoir, line 4819: https://github.com/Zilliqa/Zilliqa/blob/b65aeecb8c14e2d7198bbfd3ad295501930d05de/src/libMessage/Messenger.cpp#L1143

https://github.com/Zilliqa/Zilliqa/blob/b65aeecb8c14e2d7198bbfd3ad295501930d05de/src/libMessage/Messenger.cpp#L1157

If we reuse the same pointer,  the data and/or code from previous txn can be set for next txn??
or else need some way to clear ProtoTransaction before reusing it.

bb111189: hmm. If it is the case, we can reset the pointer to nullptr first?

ansnunez: Probably better to keep the smart pointer initialization in the same line within the for loop.

src/libMessage/Messenger.cpp
```diff
@@ -4827,14 +4825,13 @@ bool Messenger::SetNodeForwardTxnBlock(
...
     msg_size += protoTxn->ByteSize();
-    delete protoTxn;
   }
 
   for (const auto& txn : txnsGenerated) {
     if (msg_size >= PACKET_BYTESIZE_LIMIT) {
       break;
     }
-    ProtoTransaction* protoTxn = new ProtoTransaction();
+
```
ansnunez, line 4834: Better to re-initialize the smart pointer here as well

src/libMessage/Messenger.cpp
```diff
@@ -4890,25 +4886,23 @@ bool Messenger::SetNodeForwardTxnBlock(bytes& dst, const unsigned int offset,
...
-  unsigned int msg_size = 0;
+  unsigned int txnsCount = 0, msg_size = 0;
+  const auto protoTxn = std::make_unique<ProtoTransaction>();
 
   for (const auto& txn : txns) {
     if (msg_size >= PACKET_BYTESIZE_LIMIT) {
       break;
     }
-    ProtoTransaction* protoTxn = new ProtoTransaction();
+
```
ansnunez, line 4896: And here

src/libMessage/Messenger.cpp
sandipbhoir: yes better to initialize in loop itself.

ShengguangXiao: TransactionToProtobuf should always reinitialize the whole buffer, so the pointer should can be reused. Can add a test to verify it.

ansnunez: Hmm nope not completely true, because there are optional fields like `code` and `data` which may have been set in the previous loop iteration but unmodified in the current iteration.

ShengguangXiao: Good point

ShengguangXiao: @ansnunez Updated. Please review again.

PR #2086, opened 2020-05-22
Add dropped txns to pending Txns Api

KaustubhShamshery:
## Description

Add dropped txns and the reason why to the pending txns api

https://docs.google.com/spreadsheets/d/1AosU85sg8IJnWMxcH00oDz32RQoj_xYKyy1oQRSndmE/edit#gid=1043100870

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 #add_subdirectory (Consensus)
 #add_subdirectory (Contracts)
 add_subdirectory (cmd)
-add_subdirectory (Data)
+#add_subdirectory (Data)
```
ansnunez, line 4: Was the removal intended?

src/libServer/StatusServer.cpp
```diff
@@ -203,26 +203,40 @@ Json::Value StatusServer::IsTxnInMemPool(const string& tranID) {
...
-        _json["code"] = PoolTxnStatus::PRESENT_GAS_EXCEEDED;
-        _json["info"] = "Could not fit in as microblock gas limit reached";
-        return _json;
-      case PoolTxnStatus::ERROR:
-        throw JsonRpcException(RPC_INTERNAL_ERROR, "Processing transactions");
-      default:
-        throw JsonRpcException(RPC_MISC_ERROR, "Unable to process");
+    const auto& code = m_mediator.m_node->IsTxnInMemPool(tranHash);
+
+    if (!IsPoolTxnDropped(code)) {
```
ansnunez, line 208: User might wonder why some results have an "info" field while the rest don't. I would prefer no "info" across the board, as long as the API is sufficiently documented for developers.

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -792,6 +809,10 @@ void Node::ReinstateMemPool(
     m_unconfirmedTxns.emplace(t.GetTranID(),
                               PoolTxnStatus::PRESENT_GAS_EXCEEDED);
   }
+
+  for (const auto& txnHashStatus : droppedTxns) {
```
ansnunez, line 813: This could be a good opportunity for move semantics ([vector example](https://stackoverflow.com/questions/15004517/moving-elements-from-stdvector-to-another-one))

src/common/MempoolEnum.cpp
```diff
@@ -0,0 +1,59 @@
...
+                        << "Inserted " << txhash << " at " << epochNum);
+  return m_txnCode.emplace(txhash, status).second;
+}
+
+void DroppedTxnContainer::clear(const uint64_t& epochNum, const uint& TTL) {
+  LOG_MARKER();
+  if (TTL > epochNum) {
+    return;
+  }
+  const uint64_t& oldEpoch = epochNum - TTL;
```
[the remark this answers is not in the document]
KaustubhShamshery: actually it is uint64_t, so actually it would save some space.

src/libData/AccountData/AccountStoreBase.tpp
```diff
@@ -163,7 +174,9 @@ bool AccountStoreBase<MAP>::AddAccount(const Address& address,
 
   if (!IsAccountExist(address)) {
     m_addressToAccount->insert(std::make_pair(address, account));
-    // UpdateStateTrie(address, account);
+    LOG_GENERAL(WARNING, "Address "
```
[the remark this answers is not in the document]
KaustubhShamshery: You are correct, it already has a check, but why does it fail the? anyways I will change the language

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1213,10 +1210,29 @@ bool Node::AddPendingTxn(const HashCodeMap& pendingTxns, const PubKey& pubkey,
...
+
+    if (BlockStorage::GetBlockStorage().CheckTxBody(entry.first)) {
+      LOG_GENERAL(INFO, "TranHash: " << entry.first << " sent by pubkey "
+                                     << pubkey << " of shard " << shardId
+                                     << " is already confirmed");
+      continue;
+    }
+
+    if (!IsPoolTxnDropped(entry.second)) {
+      m_pendingTxns.insert(entry.first, entry.second, currentEpochNum);
```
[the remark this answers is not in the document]
KaustubhShamshery: Ohhh nice catch, I acquired `m_unconfirmed txns` instead

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1213,10 +1210,29 @@ bool Node::AddPendingTxn(const HashCodeMap& pendingTxns, const PubKey& pubkey,
     }
   }
 
-  unique_lock<shared_timed_mutex> g(m_unconfirmedTxnsMutex);
+  const auto& currentEpochNum =
```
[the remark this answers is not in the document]
KaustubhShamshery: This seems appropriate for a separate PR/

tests/CMakeLists.txt
KaustubhShamshery: No, currently fixing some tests inside.

src/libNode/Node.cpp
```diff
@@ -552,8 +552,35 @@ bool Node::CheckIntegrity(bool fromIsolatedBinary) {
...
-  m_unconfirmedTxns.clear();
+  LOG_MARKER();
+  {
+    unique_lock<shared_timed_mutex> g(m_unconfirmedTxnsMutex);
+    m_unconfirmedTxns.clear();
+  }
+}
+
+void Node::ClearPendingAndDroppedTxn() {
+  const auto& currentEpochNum =
```
sandipbhoir, line 563: want to consider renaming it to`prevEpochNum` or `latestBlkNum`?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -913,7 +910,7 @@ bool Node::ProcessFinalBlockCore(const bytes& message, unsigned int offset,
     if (toSendPendingTxn) {
       SendPendingTxnToLookup();
     }
-
+    ClearUnconfirmedTxn();
```
sandipbhoir, line 913: Are we missing clearing `UnconfirmedTxn` and `ClearPendingAndDroppedTxn` for lookup nodes `LOOKUP_NODE_MODE = true`.

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -281,6 +281,7 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone() {
   if (toSendPendingTxn) {
     m_mediator.m_node->SendPendingTxnToLookup();
   }
+  m_mediator.m_node->ClearUnconfirmedTxn();
```
sandipbhoir, line 284: want to consider clearing `PendingAndDroppedTxn` as well

KaustubhShamshery: Actually is you see,`Unconirmed txn` are only added in shard nodes. So, it is already empty in lookup and vice versa for `PendingAndDropped`

src/libNode/MicroBlockPreProcessing.cpp
KaustubhShamshery: I tried doing this and turns out you cannot add vector into a map this way. Moreover `emplace` already uses move semantics

PR #2090, opened 2020-06-01
Moved consensus TOLERANCE_FRACTION into constants

ansnunez:
## Description

Moved `ConsensusCommon::TOLERANCE_FRACTION` into constants.xml to allow value adjustment in benchmarking experiments.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/common/Constants.h
```diff
@@ -154,6 +154,7 @@ extern const unsigned int SEED_TXN_COLLECTION_TIME_IN_SEC;
 extern const unsigned int TXN_STORAGE_LIMIT;
 
 // Consensus constants
+extern const double TOLERANCE_FRACTION;
```
KaustubhShamshery, line 157: I dont know if it is feasible but can we add some sanity check (cannot be greater than 1 )

PR #2095, opened 2020-06-05
Block incoming fb/vcds msg while normal node syncing

sandipbhoir:
## Description
This PR address the issue where epoch is incremented twice due to following situation:

1. Incoming message of FinalBlock `437115`  arrived while node was syncing. <-- BUG
2. Node fetched the latest FinalBlock `437115`  from lookup and epoch was increased to `437116` and FB stored to chain.
3. message received at `step 1` is also being processed at same time and while storing to blockchain it does complains

```
[WARN][  270][20-05-27T12:28:54.183][ata/BlockChain.h:113][AddBlock            ] Failed to add 437715 437715
```

However, it still incremented epoch by another one i.e. 437717

**Issue :**  FB Message is not being blocked while in sync. This could potentially happen also with DSBlock message.

**Fix** is to block FB/VCDS block message while normal node is syncing

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -2474,6 +2474,10 @@ bool Node::ToBlockMessage([[gnu::unused]] unsigned char ins_byte) {
                      SyncType::GUARD_DS_SYNC &&
                  GUARD_MODE) {
         return true;
+      } else if (m_mediator.m_lookup->GetSyncType() == SyncType::NORMAL_SYNC &&
+                 (ins_byte == NodeInstructionType::DSBLOCK ||
+                  ins_byte == NodeInstructionType::FINALBLOCK)) {
+        return true;
       }
       if (!m_fromNewProcess) {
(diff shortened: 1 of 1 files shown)


src/libNode/Node.cpp
```diff
@@ -2474,6 +2474,10 @@ bool Node::ToBlockMessage([[gnu::unused]] unsigned char ins_byte) {
                      SyncType::GUARD_DS_SYNC &&
                  GUARD_MODE) {
         return true;
+      } else if (m_mediator.m_lookup->GetSyncType() == SyncType::NORMAL_SYNC &&
```
KaustubhShamshery, line 2477: Can we have some log here?

sandipbhoir: we do log the message at caller of `ToBlockMessage` i.e. "Node is not yet in network, so ignore message"

PR #2097, opened 2020-06-08
fix: consider gas fee while doing transaction validation

renlulu:
## Description
This PR is to make LookupServer::ValidateTxn more clever by checking both gas fee and amount.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -420,6 +420,26 @@ bool ValidateTxn(const Transaction& tx, const Address& fromAddr,
                                to_string(sender->GetNonce()) + ")");
   }
 
+  // Check if transaction amount is valid
+  uint128_t gasDeposit = 0;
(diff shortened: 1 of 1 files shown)


KaustubhShamshery:
Hi @renlulu , a more appropriate place would be to add this check in the lookup,

https://github.com/Zilliqa/Zilliqa/blob/3d67ab7b0c4e67f1382e0db2ee23f945fda000a2/src/libServer/LookupServer.cpp#L355.

renlulu:
> Hi @renlulu , a more appropriate place would be to add this check in the lookup,
> 
> https://github.com/Zilliqa/Zilliqa/blob/3d67ab7b0c4e67f1382e0db2ee23f945fda000a2/src/libServer/LookupServer.cpp#L355
> 
> .

Ok, let me take a look at this.

PR #2104, opened 2020-06-15
Add option to skip mining for DS difficulty

ansnunez:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libNode/PoWProcessing.cpp
```diff
@@ -227,38 +227,44 @@ bool Node::StartPoW(const uint64_t& block_num, uint8_t ds_difficulty,
                                  winning_result.result, winning_result.mix_hash,
                                  lookupId, m_proposedGasPrice)) {
         return false;
+      }
+
+      if (SKIP_POW_REATTEMPT_FOR_DS_DIFF) {
+        LOG_GENERAL(INFO, "Skipping PoW reattempt for DS difficulty");
```
sandipbhoir, line 233: we still need to call `checkerThread`. Its responsible for checking if DS block was received and Rejoin if required.

PR #2106, opened 2020-06-17
Fix issue with  bootstrap joining of non-sharded node

sandipbhoir:
## Description
This PR fix issue where in when testnet is bootsrapped from existing previous testnet and when zilliqa node is not sharded it triggers rejoin. However, the incremental db is not yet uploaded by lookup to S3.

lookup starts uploading incr data to S3 once new txblock is mined.

So fix is to wait for next another txblock so that lookup can perform upload. After which `RejoinAsNormal` will proceed as usual.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -963,6 +963,29 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
     LOG_GENERAL(WARNING,
                 "Node " << m_mediator.m_selfKey.second
                         << " is not in network, apply re-join process instead");
+    // wait until next two txblocks are mined to give lookup enough time to
+    // upload incr data to S3.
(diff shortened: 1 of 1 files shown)


src/libNode/Node.cpp
```diff
@@ -818,6 +818,8 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
       unique_lock<mutex> lock(m_mediator.m_lookup->m_MutexCVSetTxBlockFromSeed);
       m_mediator.m_lookup->SetSyncType(SyncType::LOOKUP_SYNC);
 
+      // wait for two txblocks to give lookup after bootstrap enough time to
```
ansnunez, line 821: Is this comment correct?

PR #2111, opened 2020-06-23
Make get function const to prevent member data modified

ShengguangXiao:
## Description

Make get function const to prevent member data modified, and try to avoid use [] operator for map containter, because it impliciyly insert object to the container if object not exists.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/FinalBlockPreProcessing.cpp
+++ b/src/libDirectoryService/FinalBlockPreProcessing.cpp
@@ -58,7 +58,7 @@ void DirectoryService::ExtractDataFromMicroblocks(
 
   auto& microBlocks = m_microBlocks[m_mediator.m_currentEpochNum];
 
-  for (auto& microBlock : microBlocks) {
+  for (const auto& microBlock : microBlocks) {
     LOG_STATE("[STATS][" << std::setw(15) << std::left
                          << m_mediator.m_selfPeer.GetPrintableIPAddress()
                          << "][" << i << "    ]["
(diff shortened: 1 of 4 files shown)


src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -489,9 +489,13 @@ bool DirectoryService::CheckLegitimacyOfMicroBlocks() {
 
   {
     lock_guard<mutex> g(m_mutexMicroBlocks);
+    std::set<MicroBlock> microBlocks;
+    if (m_microBlocks.find(m_mediator.m_currentEpochNum) !=
```
ansnunez, line 493: Good call to add this check. However, the creation of `microBlocks` adds performance and memory penalty. Maybe you can instead encapsulate the for loop below within the check, so that no microblocks iteration is done if we can't find the epoch number in the map.

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -735,8 +739,12 @@ bool DirectoryService::CheckMicroBlockInfo() {
     //                 << "; hashes: " << hashesInMicroBlocks[i]
     //                 << "; IsMicroBlockEmpty: "
     //                 << m_finalBlock->GetIsMicroBlockEmpty()[i]);
-    auto& microBlocks = m_microBlocks[m_mediator.m_currentEpochNum];
-    for (auto& microBlock : microBlocks) {
+    std::set<MicroBlock> microBlocks;
+    if (m_microBlocks.find(m_mediator.m_currentEpochNum) !=
```
ansnunez, line 743: Same as previous comment

src/libDirectoryService/FinalBlockPreProcessing.cpp
ShengguangXiao: This is not inside the loop

ShengguangXiao: Updated

src/libDirectoryService/FinalBlockPreProcessing.cpp
ShengguangXiao: Now changed to if not found micro block info, then continue to check next next one.

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -735,8 +738,13 @@ bool DirectoryService::CheckMicroBlockInfo() {
     //                 << "; hashes: " << hashesInMicroBlocks[i]
     //                 << "; IsMicroBlockEmpty: "
     //                 << m_finalBlock->GetIsMicroBlockEmpty()[i]);
-    auto& microBlocks = m_microBlocks[m_mediator.m_currentEpochNum];
-    for (auto& microBlock : microBlocks) {
+
+    if (m_microBlocks.find(m_mediator.m_currentEpochNum) ==
+        m_microBlocks.end()) {
+      continue;
```
ansnunez, line 744:
```suggestion
      break;
```

src/libLookup/Lookup.cpp
```diff
@@ -3372,8 +3372,13 @@ void Lookup::FindMissingMBsForLastNTxBlks(const uint32_t& num) {
   }
 }
 
-const vector<Transaction>& Lookup::GetTxnFromShardMap(uint32_t index) {
-  return m_txnShardMap[index];
+const vector<Transaction>& Lookup::GetTxnFromShardMap(uint32_t shardId) const {
```
sandipbhoir, line 3375: This could be overkill. No need to find it always before fetching. Because empty value `vector` is assigned if key `index` does not exist & we any ways check for empty at caller. Note this is always called at https://github.com/Zilliqa/Zilliqa/blob/6157d930d4098b0c58f0fe02846f4e7e432957c4/src/libServer/LookupServer.cpp#L260

Also, if we stick with additional checks, we need to have it also at https://github.com/Zilliqa/Zilliqa/blob/6157d930d4098b0c58f0fe02846f4e7e432957c4/src/libLookup/Lookup.cpp#L5330 otherwise entry will be created before its cleared.

ShengguangXiao: It is ok to return empty vector, but the change of this function is to make get function to be const.

PR #2114, opened 2020-06-24
Check m_sendSCCallsToDS in CreateTransaction

ansnunez:
## Description

PR #1976 only enforces sending contract call transactions to DS inside `ProcessForwardTxn`, which is called when a lookup receives txn packets from seed nodes.

This means if a lookup node directly processes `CreateTransaction` itself, the enforcement is not triggered.

This PR adds the checking for `m_sendSCCallsToDS` inside `CreateTransaction` so that the enforcement is covered here as well. It will use the `priority` field only if `m_sendSCCallsToDS` is false, to preserve original behavior.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -493,7 +493,7 @@ Json::Value LookupServer::CreateTransaction(
       }
 
       toAccountExist = (toAccount != nullptr);
-      toAccountIsContract = toAccount->isContract();
+      toAccountIsContract = toAccountExist && toAccount->isContract();
     }
 
     const unsigned int shard = Transaction::GetShardIndex(fromAddr, num_shards);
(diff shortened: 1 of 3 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -540,9 +540,11 @@ Json::Value LookupServer::CreateTransaction(
 
         unsigned int to_shard =
             Transaction::GetShardIndex(tx.GetToAddr(), num_shards);
-        bool sendToDs = false;
+        // Use m_sendSCCallsToDS as main setting, then priority field if
+        // m_sendSCCallsToDS is false
+        bool sendToDs = m_mediator.m_lookup->m_sendSCCallsToDS;
         if (_json.isMember("priority")) {
-          sendToDs = _json["priority"].asBool();
+          sendToDs = sendToDs ? sendToDs : _json["priority"].asBool();
```
KaustubhShamshery, line 547: Can just use OR (||) for readability

PR #2117, opened 2020-06-25
Fix/python dependencies

joel-zilliqa:
## Description
Version locks python2 and 3 dependencies to requirement files.

## Backward Compatibility

## Review Suggestion
Create a testnet cluster from the image built by this commit (2aab02a) and test functionalities.

### Implementation
- [x] **ready for review**

- [x] local machine test (commit: 2aab02a)
- [x] small-scale cloud test (commit: 2aab02a)

### DIFF
--- a/docker/Dockerfile.k8s
+++ b/docker/Dockerfile.k8s
@@ -14,6 +14,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 # This Dockerfile is incomplete and should only be used with other Dockerfiles
+
+COPY requirements2.k8s.txt ./
+
 RUN apt-get update && apt-get install -y --no-install-recommends \
     dnsutils \
     gdb \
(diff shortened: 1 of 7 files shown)


docker/Dockerfile.k8s
```diff
@@ -14,6 +14,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 # This Dockerfile is incomplete and should only be used with other Dockerfiles
+
+COPY requirements2.k8s.txt ./
```
Revolution1, line 18: Don‘t think we need this file, we can just install these deps in a common image.  Image spacifically for k8s doen't make sense.
one for normal use, one for cuda is enough.

joel-zilliqa: Yea, make sense, gonna merge the k8s dependencies into the main image, I avoided that initially to mitigate image bloating for public images

PR #2119, opened 2020-06-26
Avoid duplicate txnpkt bufferring and processing

sandipbhoir:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.h
+++ b/src/libNode/Node.h
@@ -160,7 +160,7 @@ class Node : public Executable {
       m_pendingTxnBuffer;
 
   std::mutex m_mutexTxnPacketBuffer;
-  std::vector<bytes> m_txnPacketBuffer;
+  std::map<bytes, bytes> m_txnPacketBuffer;
 
   // txn proc timeout related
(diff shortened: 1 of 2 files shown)


src/libNode/Node.h
```diff
@@ -160,7 +160,7 @@ class Node : public Executable {
       m_pendingTxnBuffer;
 
   std::mutex m_mutexTxnPacketBuffer;
-  std::vector<bytes> m_txnPacketBuffer;
+  std::map<bytes, bytes> m_txnPacketBuffer;
```
KaustubhShamshery, line 163: Can we use unordered_map instead

sandipbhoir: Don't think its of any gain considering we are going to store not more than 1 or 2 pkts.
Its O(logN) vs O(1). Considering N = 2, benefits over each other are negligible.
Also, if we go with unordered_map or unordered_set, we would need to have hash function for `bytes` which i think is over kill to do in such situation.

PR #2127, opened 2020-06-30
Add websocket to isolated server

KaustubhShamshery:
Enable Websockets on Zilliqa Isolated server.

Also add block-related APIs to time-trigger setup

* GetTxBlock
* GetLatestTxBlock

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -20,6 +20,7 @@
 #include "depends/safeserver/safehttpserver.h"
 #include "libServer/IsolatedServer.h"
 #include "libServer/LookupServer.h"
+#include "libServer/WebsocketServer.h"
 
 #define SUCCESS 0
 #define ERROR_IN_COMMAND_LINE -1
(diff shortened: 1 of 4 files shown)


src/cmd/isolated_server.cpp
```diff
@@ -158,6 +159,14 @@ int main(int argc, const char* argv[]) {
       cout << "Server listening on " << port << endl;
     }
 
+    if (ENABLE_WEBSOCKET && timeDelta > 0) {
+      (void)WebsocketServer::GetInstance();
+    }
+    else if(ENABLE_WEBSOCKET)
+    {
+      LOG_GENERAL(INFO,"Websocket can only be enabled if run in time increment mode");
+    }
```
kaikawaliu, line 168: This paragraph looks too weird, can change to this?
```
if (ENABLE_WEBSOCKET) {
  if (timeDelta>0) {
    ...
  } else {
    ...
  }
}
```

src/libServer/WebsocketServer.cpp
```diff
@@ -63,6 +63,7 @@ bool WebsocketServer::start() {
   clean();
   // Initialising websocketserver
   m_server.init_asio();
+  m_server.set_reuse_addr(true);
```
KaustubhShamshery, line 66: @kaikawaliu can u check this if this makes any difference to the core protocol. In my opinion, it does not.

kaikawaliu: Yup I think it's good to have

PR #2130, opened 2020-07-01
ShardGuard Pod deletion and few optimization/bug fix

sandipbhoir:
## Description
This PR cover following:

- Shard guard pod deletion scenario where new IP of shard guard is broadcasted to its shard and it joins to shard eventually.

- Recovered shardguard and miner node which is already part of shard will now starts at state `WAITING_FINALBLOCK` instead of `MICROBLOCK_CONSENSUS`. This allows get rid of timing issue and starts with being ready to receive next FinalBlock.
 Note: If the recovery takes time which usually will be the case, it will get some missing finalblock which will trigger `RejoinAsNormal`.
   - Optimization is done now to check if number of missing txblocks are not more than `NUM_FINALBLOCK_NUM`. If so, it synchronizes those missing ones from seed nodes instead which will be quick. If not, it continues from S3 as usual. 

- Fix the deadlock issue due to `mutex m_mutexShardMember` in `Node::StartFirstTxEpoch`.

- Code refactoring `WaitForNextTwoBlocksBeforeRejoin`

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNetwork/RumorManager.cpp
+++ b/src/libNetwork/RumorManager.cpp
@@ -17,6 +17,8 @@
 
 #include "RumorManager.h"
 
+#include <boost/bimap/support/lambda.hpp>
+
 #include <chrono>
 #include <string>
 #include <thread>
@@ -176,6 +178,24 @@ bool RumorManager::Initialize(const VectorOfNode& peers, const Peer& myself,
                                      (ROUND_TIME_IN_MS);  // milliseconds
 
   return true;
+}
+
+void RumorManager::UpdatePeerInfo(const Peer& newPeerInfo,
+                                  const PubKey& pubKey) {
+  std::lock_guard<std::mutex> guard(m_mutex);  // critical section
+  auto it = m_pubKeyPeerBiMap.left.find(pubKey);
+  if (it != m_pubKeyPeerBiMap.left.end()) {
+    Peer oldPeerInfo = it->second;
+    m_pubKeyPeerBiMap.left.modify_data(it, boost::bimaps::_data = newPeerInfo);
+    auto it2 = m_peerIdPeerBimap.right.find(oldPeerInfo);
+    if (it2 != m_peerIdPeerBimap.right.end()) {
+      m_peerIdPeerBimap.right.modify_key(it2,
+                                         boost::bimaps::_key = newPeerInfo);
+      LOG_GENERAL(INFO, "Updated peer info successfully!");
+      return;
+    }
+  }
+  LOG_GENERAL(WARNING, "Failed to updated peer info!");
 }
 
 void RumorManager::SpreadBufferedRumors() {
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -629,6 +629,32 @@ void Node::Prepare(bool runInitializeGenesisBlocks) {
       FULL_DATASET_MINE);
 }
 
+void Node::WaitForNextTwoBlocksBeforeRejoin() {
+  // wait until next two txblocks are mined to give lookup enough time to
+  // upload incr data to S3.
+  unique_lock<mutex> lock(m_mediator.m_lookup->m_MutexCVSetTxBlockFromSeed);
+  m_mediator.m_lookup->SetSyncType(SyncType::RECOVERY_ALL_SYNC);
+
+  uint64_t oldBlkCount = m_mediator.m_txBlockChain.GetBlockCount();
+  LOG_GENERAL(INFO, "Wait until next two txblock are recvd from lookup..");
+  while (true) {
+    do {
+      m_mediator.m_lookup->GetTxBlockFromSeedNodes(
+          m_mediator.m_txBlockChain.GetBlockCount(), 0);
+    } while (m_mediator.m_lookup->cv_setTxBlockFromSeed.wait_for(
+                 lock, chrono::seconds(RECOVERY_SYNC_TIMEOUT)) ==
+             cv_status::timeout);
+
+    if (m_mediator.m_txBlockChain.GetBlockCount() > oldBlkCount + 1) {
+      LOG_GENERAL(INFO, "Received next two txblocks. Ok to rejoin now!")
+      break;
+    }
+    this_thread::sleep_for(chrono::seconds(RECOVERY_SYNC_TIMEOUT));
+  }
+
+  m_mediator.m_lookup->SetSyncType(SyncType::NO_SYNC);
+}
+
 bool Node::StartRetrieveHistory(const SyncType syncType,
                                 bool rejoiningAfterRecover) {
   LOG_MARKER();
@@ -931,6 +957,7 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
   }
 
   bool bInShardStructure = false;
+  bool bIpChanged = false;
 
   if (bDS) {
     m_myshardId = m_mediator.m_ds->m_shards.size();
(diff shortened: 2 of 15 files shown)


src/libNode/Node.cpp
```diff
@@ -2408,6 +2427,199 @@ bool Node::ProcessDoRejoin(const bytes& message, unsigned int offset,
...
+      MessageType::NODE, NodeInstructionType::NEWSHARDGUARDIDENTITY};
+
+  uint64_t curDSEpochNo =
+      m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() + 1;
+
+  if (!Messenger::SetNodeNewShardGuardNetworkInfo(
+          updateshardguardidentitymessage, MessageOffset::BODY, curDSEpochNo,
+          m_mediator.m_selfPeer, get_time_as_int(), m_mediator.m_selfKey)) {
+    LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
+              "Messenger::SetNodeSetNewDSGuardNetworkInfo failed.");
```
ansnunez, line 2462:
```suggestion
              "Messenger::SetNodeNewShardGuardNetworkInfo failed.");
```

src/libNode/Node.cpp
```diff
@@ -2408,6 +2427,199 @@ bool Node::ProcessDoRejoin(const bytes& message, unsigned int offset,
...
+  if (!GUARD_MODE) {
+    LOG_GENERAL(
+        WARNING,
+        "Not in guard mode. Unable to update shard guard network info.");
+    return false;
+  }
+
+  if (!Guard::GetInstance().IsNodeInShardGuardList(
+          m_mediator.m_selfKey.second)) {
+    LOG_GENERAL(WARNING,
```
ansnunez, line 2445: To avoid non-guard miner from unnecessarily receiving this message (which might cause confusion), I suggest to just return false here and instead add a "Current node is a shard node. Updating network info." log in line 2450.

src/libNode/Node.cpp
```diff
@@ -2408,6 +2427,199 @@ bool Node::ProcessDoRejoin(const bytes& message, unsigned int offset,
...
+    for (auto& it : *m_myShardMembers) {
+      peerInfo.push_back(it.second);
+    }
+  }
+
+  {
+    // Multicast to all DS committee
+    lock_guard<mutex> lock(m_mediator.m_mutexDSCommittee);
+    for (auto const& i : *m_mediator.m_DSCommittee) {
+      if (i.second.m_listenPortHost != 0) {
```
ansnunez, line 2486: This check may not be necessary?

src/libNode/Node.cpp
```diff
@@ -2408,6 +2427,199 @@ bool Node::ProcessDoRejoin(const bytes& message, unsigned int offset,
...
+                               << from
+                               << " does not match any of my shard members");
+      return false;
+    }
+    // update requestor's ( ShardGuard ) new IP
+    lock_guard<mutex> g(m_mutexShardMember);
+
+    unsigned int indexOfShardNode;
+    for (indexOfShardNode = 0;
+         indexOfShardNode < Guard::GetInstance().GetNumOfShardGuard();
```
ansnunez, line 2576: Just to check: do shard guards all occupy the first slots in shard members list?

ansnunez, line 2576: Since `GetNumOfShardGuard()` calls a non-inline function that must also acquire a mutex, it might be more efficient to put `GetNumOfShardGuard()` result into a const value instead.

src/libNode/Node.cpp
```diff
@@ -2408,6 +2427,199 @@ bool Node::ProcessDoRejoin(const bytes& message, unsigned int offset,
...
+    m_mediator.m_ds->UpdateShardNodeNetworkInfo(shardGuardNewNetworkInfo,
+                                                shardGuardPubkey);
+    if (!BlockStorage::GetBlockStorage().PutShardStructure(
+            m_mediator.m_ds->m_shards, 0)) {
+      LOG_GENERAL(WARNING, "BlockStorage::PutShardStructure failed");
+    }
+  }
+  // I am sharded node and requestor is also from my shard
+  else if (m_mediator.m_ds->m_mode == DirectoryService::IDLE) {
+    if (!IsShardNode(shardGuardPubkey)) {
```
ansnunez, line 2565: Here we make an initial search through `m_myShardMembers` and then in line 2575 we also go through `m_myShardMembers` to make the actual replacement. Maybe we can skip this initial search and just try to find and replace immediately? If it's not found during replacement, you can log the "does not match" message as well.

src/libNode/Node.cpp
```diff
@@ -2408,6 +2427,199 @@ bool Node::ProcessDoRejoin(const bytes& message, unsigned int offset,
...
+                m_mediator.m_ds->m_shards, m_mediator.m_node->m_myshardId)) {
+          LOG_GENERAL(WARNING, "BlockStorage::PutShardStructure failed");
+        }
+        break;
+      }
+    }
+  }
+  // I am ds node and requestor is from one of shards
+  else if (m_mediator.m_ds->m_mode != DirectoryService::IDLE) {
+    if (!m_mediator.m_ds->CheckIfShardNode(shardGuardPubkey)) {
```
ansnunez, line 2603: Here we make an initial search through `m_shards` and then in line 2610 we also go through `m_shards` to make the actual replacement. Maybe we can skip this initial search and just try to find and replace immediately? If it's not found during replacement, you can log the "does not match" message as well.

src/libNode/Node.cpp
```diff
@@ -958,35 +989,23 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
...
-        break;
+  if (REJOIN_NODE_NOT_IN_NETWORK && !LOOKUP_NODE_MODE && !bDS) {
+    if (!bInShardStructure) {
+      LOG_GENERAL(
+          WARNING,
+          "Node " << m_mediator.m_selfKey.second
+                  << " is not in network, apply re-join process instead");
+      WaitForNextTwoBlocksBeforeRejoin();
+      return false;
+    } else if (bIpChanged) {
```
ansnunez, line 1000: What happens if `bIpChanged` is false?

src/libNode/Node.cpp
ansnunez: I reviewed a recent chat with @bb111189 and it seems shard guards are randomly positioned inside each shard, so this for loop might not be accurate

sandipbhoir: Will be removing the usage of this now.

src/libNode/Node.cpp
sandipbhoir: It will follow usual path i.e. it will join a shard because `bInShardStructure` is true ( IP is same in this case )

src/libNetwork/RumorManager.cpp
```diff
@@ -178,6 +178,24 @@ bool RumorManager::Initialize(const VectorOfNode& peers, const Peer& myself,
...
+void RumorManager::UpdatePeerInfo(const Peer& newPeerInfo,
+                                  const PubKey& pubKey) {
+  std::lock_guard<std::mutex> guard(m_mutex);  // critical section
+  auto it = m_pubKeyPeerBiMap.left.find(pubKey);
+  if (it != m_pubKeyPeerBiMap.left.end()) {
+    Peer oldPeerInfo = it->second;
+    m_pubKeyPeerBiMap.left.modify_data(it, boost::bimaps::_data = newPeerInfo);
+    auto it2 = m_peerIdPeerBimap.right.find(oldPeerInfo);
+    if (it2 != m_peerIdPeerBimap.right.end()) {
+      m_peerIdPeerBimap.right.modify_key(it2,
```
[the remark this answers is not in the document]
sandipbhoir: We are modfiying key i.e. PeerInfo (IP/Port) here for corresponding PeerId.

PR #2154, opened 2020-07-21
Soft confirmation by earlier forwarding of MB and TxnBodies

kaikawaliu:
## Description

- Added new API `GetSoftConfirmedTransaction`, the result have a new field `softconfirm` to indicate whether it's soft confirmed transaction, otherwise will be permanently stored already.
- Bring `MBnForwardTransaction` from shard node to MicroBlockConsensusDone, thus lookup node can be earlier notified about the status of transaction by soft confirmation. 
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1384,8 +1465,9 @@ bool Node::ProcessPendingTxn(const bytes& message, unsigned int cur_offset,
   const auto& currentEpochNum =
       m_mediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum();
 
-  if (currentEpochNum > epochNum) {
-    LOG_GENERAL(WARNING, "PENDINGTXN sent of an older epoch " << epochNum);
+  if (currentEpochNum > epochNum + 1) {
```
KaustubhShamshery, line 1468: Can I ask why was this change needed? allowing one epoch later pending txn

kaikawaliu: Because now the pending txn will be forwarded upon the completion of MicroBlockConsensus by the MBnForwardTxn message, the lookup may receive it later than the FinalBlock message due to lagging network. Besides, there is a checker for existing transactions, it can avoid duplication that if the transaction has already been confirmed but the pending transaction tries to mislead it.

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1277,6 +1350,14 @@ bool Node::ProcessMBnForwardTransaction(const bytes& message,
                           << " shard "
                           << entry.m_microBlock.GetHeader().GetShardId());
 
+    if (entry.m_microBlock.GetHeader().GetShardId() !=
+        m_mediator.m_ds->m_shards.size()) {
+      // pre-process of early MBnForwardTxn submission
+      // soft confirmation
+      SoftConfirmForwardedTransactions(entry);
+      // invoke txn distribution
```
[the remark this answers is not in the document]
kaikawaliu: ok

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1211,6 +1269,21 @@ bool Node::ProcessMBnForwardTransaction(const bytes& message,
...
 
+  if (entry.m_microBlock.GetHeader().GetVersion() != MICROBLOCK_VERSION) {
+    LOG_CHECK_FAIL("MicroBlock version",
+                   entry.m_microBlock.GetHeader().GetVersion(),
+                   MICROBLOCK_VERSION);
+    return false;
+  }
+
+  // Verify the co-signature
+  if (!m_mediator.m_ds->VerifyMicroBlockCoSignature(
```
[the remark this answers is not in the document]
kaikawaliu: noted, good catch

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -427,6 +427,10 @@ void Node::CallActOnFinalblock() {
 
   LOG_MARKER();
 
+  if (m_mediator.m_ds->m_mode == DirectoryService::IDLE) {
```
ansnunez, line 430: Or can just remove the call to `CallActOnFinalblock` inside `DirectoryService::ProcessFinalBlockConsensusWhenDone`?

src/libServer/LookupServer.cpp
```diff
@@ -627,6 +632,47 @@ Json::Value LookupServer::GetTransaction(const string& transactionHash) {
...
+  }
+
+  try {
+    TxBodySharedPtr tptr;
+    TxnHash tranHash(txnHash);
+    if (txnHash.size() != TRAN_HASH_SIZE * 2) {
+      throw JsonRpcException(RPC_INVALID_PARAMS, "Size not appropriate");
+    }
+    bool isPresent = BlockStorage::GetBlockStorage().GetTxBody(tranHash, tptr);
+    bool isPresentHistorical = false;
```
ansnunez, line 649: Maybe cross-checking against historical DB is not necessary? Also, this db is for diagnostic purposes, so using it as a canonical source might not be a good idea.

src/libNode/MicroBlockPostProcessing.cpp
```diff
@@ -280,11 +291,24 @@ bool Node::ProcessMicroBlockConsensusCore(const bytes& message,
...
+      // To Lookup -> ProcessMBnForwardTxn
+      DataSender::GetInstance().SendDataToOthers(
+          *m_microblock, *m_myShardMembers, {}, {},
+          m_mediator.m_lookup->GetLookupNodes(),
+          m_mediator.m_txBlockChain.GetLastBlock().GetBlockHash(),
+          m_consensusMyID, composeMBnForwardTxnMessageForSender, false,
+          SendDataToLookupFuncDefault, sendMbnFowardTxnToShardNodes);
+      // pending Txns
+      if (!IsUnconfirmedTxnEmpty()) {
+        SendPendingTxnToLookup();
```
ansnunez, line 309: Just to check: We still have code for `SendPendingTxnToLookup` inside `Node::ProcessFinalBlockCore`. I'm guessing that code will now be relevant only for DS committee?

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1277,6 +1350,14 @@ bool Node::ProcessMBnForwardTransaction(const bytes& message,
                           << " shard "
                           << entry.m_microBlock.GetHeader().GetShardId());
 
+    if (entry.m_microBlock.GetHeader().GetShardId() !=
```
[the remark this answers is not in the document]
kaikawaliu: yes agree

src/libNode/FinalBlockProcessing.cpp
kaikawaliu: fixed

src/libServer/LookupServer.cpp
kaikawaliu: noted

src/libNode/MicroBlockPostProcessing.cpp
kaikawaliu: Yes we still need to check this on shard, in case the microblock from this shard is not included in the FinalBlock.

PR #2159, opened 2020-07-27
Modify docker file to upgrade Ubuntu os running inside docker to 18.04

chetan-zilliqa:
## Description

The goal of the PR is to upgrade OS running inside docker to `Ubuntu 18.04`.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,15 +14,15 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-ARG BASE=zilliqa/scilla:v0.7.0
+ARG BASE=zilliqa/scilla:v0.8.0-alpha.0
 FROM ${BASE}
 
 COPY requirements3.txt ./
 
+
 # Format guideline: one package per line and keep them alphabetically sorted
 RUN apt-get update \
     && apt-get install -y software-properties-common \
-    && add-apt-repository ppa:tah83/secp256k1 -y \
     && apt-get update && apt-get install -y --no-install-recommends \
     autoconf \
     build-essential \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.cuda
```diff
@@ -47,65 +49,70 @@ RUN apt-get update \
...
-    python \
-    python-pip \
     python3-setuptools \
     python3-pip \
     libsecp256k1-dev \
     && rm -rf /var/lib/apt/lists/* \
-    && pip install 'setuptools<45' \
-    && pip install request requests clint futures \
-    && pip3 install request requests clint future
+    && pip3 install requests==2.24.0 clint==0.5.1 future==0.18.2
```
Revolution1, line 56: better put this inside a requirements.txt

docker/Dockerfile
```diff
@@ -14,15 +14,15 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-ARG BASE=zilliqa/scilla:v0.7.0
+ARG BASE=zilliqa/scilla:scilla_ubuntu_upgrade_b25539d
```
Revolution1, line 17: why not tag this image more readable, like give it a version number

joel-zilliqa: that's cause we're testing a pre-release build for scilla instead of a officially released build

joel-zilliqa: so we tagged the scilla image with the hash so that we can easily reference which commit it's from, but in a proper release, it'll revert to the proper tagging

.gitmodules
```diff
@@ -14,3 +14,4 @@
 [submodule "src/depends/Schnorr"]
 	path = src/depends/Schnorr
 	url = https://github.com/Zilliqa/schnorr.git
+	branch = fix/openssl-compatibility-ubuntu-20
```
ansnunez, line 17: The PR for this branch has just been merged. Maybe we can remove this change now?

PR #2161, opened 2020-07-27
Improve multicast in non-ideal cases with connectivity with some of peers

sandipbhoir:
## Description
https://github.com/Zilliqa/Issues/issues/647

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libNetwork/P2PComm.cpp
```diff
@@ -373,9 +376,25 @@ void SendJobPeers<T>::DoSend() {
...
     }
 
-    SendMessageCore(peer, m_message, m_startbyte, m_hash);
+    // Launch the thread.
+    auto hThread = make_shared<std::future<void>>(
+        std::async(std::launch::async, SendMessageCore, peer, m_message,
+                   m_startbyte, m_hash));
+
+    // Park the thread if message sending taking more than expected
+    auto status = hThread->wait_for(std::chrono::seconds(5));
```
ansnunez, line 385: If I recall correctly, messages that go beyond a few MB can potentially take (maybe) more than 5 seconds. Maybe we can add this duration to constants.xml instead so we can adjust easily if necessary.

src/libNetwork/P2PComm.cpp
```diff
@@ -373,9 +376,25 @@ void SendJobPeers<T>::DoSend() {
...
+
+    if (status == std::future_status::timeout) {
+      LOG_GENERAL(WARNING, "Sending message to "
```
ansnunez, line 388: In the interest of brevity, I suggest to shorten to: "Sending delayed for <IP>"
Anyway the presence of the `LOG_MARKER` will already help us find the scope for this entire execution in the logs.

sandipbhoir: sure. but i still prefer the message logging for peer, ofcourse short one like you suggested.

PR #2187, opened 2020-08-26
Divert naming of rewards field in different epoch

kaikawaliu:
## Description

Since the `rewards` field in the actual TxBlock can mean for Transaction fees incurred during non-vacuous epoch and also coinbase reward during vacuous epoch. For avoid misunderstanding we decided to explain it into two fields in API level, `TxnFees` and `Rewards` respectively as below:
```
                  | TxnFees | Rewards
Non-vacuous       | fees    | 0
Vacuous           | 0       | coinbase
```
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/JSONConversion.cpp
+++ b/src/libServer/JSONConversion.cpp
@@ -27,6 +27,7 @@
 #include "libData/AccountData/Transaction.h"
 #include "libData/AccountData/TransactionReceipt.h"
 #include "libData/BlockData/Block.h"
+#include "libMediator/Mediator.h"
 #include "libUtils/DataConversion.h"
 #include "libUtils/Logger.h"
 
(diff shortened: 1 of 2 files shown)


evesnow91 (approved):

KaustubhShamshery:
I don't think you should use the current block num instead we should use the block num of the tx block to figure out if it is a vacuous epoch.

kaikawaliu:
> I don't think you should use the current block num instead we should use the block num of the tx block to figure out if it is a vacuous epoch.

Yes I'm doing this exactly. which line doesn't look correct to you?

ansnunez (approved):

src/libServer/JSONConversion.cpp
```diff
@@ -58,7 +59,8 @@ const Json::Value JSONConversion::convertTxBlocktoJson(const TxBlock& txblock) {
   ret_head["Version"] = txheader.GetVersion();
   ret_head["GasLimit"] = to_string(txheader.GetGasLimit());
   ret_head["GasUsed"] = to_string(txheader.GetGasUsed());
-  ret_head["Rewards"] = txheader.GetRewards().str();
+  ret_head["Rewards"] = isVacuous ? txheader.GetRewards().str() : 0;
```
KaustubhShamshery, line 62: here `isVacous` is governed by some outside parameter, why dont just call m_mediator.GetIsVacuousEpoch(Latest.GetHeader().GetBlockNum()) inside the function. This extra parameter is not needed

ansnunez: Good point. Can use `txheader.GetBlockNum()`

KaustubhShamshery (approved):

PR #2216, opened 2020-09-11
[master] Revised pBFT consensus with txn processing

sandipbhoir:
## Description
This PR enforces newly revised pBFT consensus which will be applicable for shard microblock consensus and final block consensus.

Attached is the revised new flow for microblock consensus. Similar applies for final block consensus.

This PR also includes following:
- Early packet dispatching for shards.
- Fixes to stress test suite. `scan_consolidate.py` due to early packet dispatching.
- Multiple fixes for handling cases arising due to early packet dispatching and get stress test suite working. (https://zilliqa-jira.atlassian.net/browse/ZIL-1745)
- Txn packet is now processed in every epoch except in vacuous epoch.
- Fix for Leader Consensus State change to INITIAL even on receiving expected commits (issue : https://github.com/Zilliqa/Issues/issues/658

**Revised pBFT consensus flow:**

revised_new_flow(1)
)

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libConsensus/ConsensusBackup.h
```diff
@@ -98,8 +104,19 @@ class ConsensusBackup : public ConsensusCommon {
...
+          nullptr,  // function handler for validating the preprep content of
+                    // message for consensus (e.g., Tx block)
+      PostPrePrepValidationFunc post_preprep_validation =
+          nullptr,  // function handler to execute
+                    // any post activity after validation of preprep
+                    // message
+      // function to give a chance to check if the commit failure is resolved
+      // before changing to ERROR state
+      CollectiveSigReadinessFunc collsig_readiness_func =
+          nullptr  // function handler for waits until some cond is met
```
ansnunez, line 119: Seems like two comment descriptions for this func handler

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -1033,28 +1294,64 @@ bool Node::RunConsensusOnMicroBlockWhenShardBackup() {
...
+    return PrePrepMicroBlockValidator(input, offset, errorMsg, consensusID,
+                                      blockNumber, blockHash, leaderID,
+                                      leaderKey, messageToCosign);
+  };
+
+  auto postPreprepValidationFunc = [this]() -> void {
+    StartTxnProcessingThread();
+  };
+
+  /* auto postFailedPreprepValidationFunc = [this]() -> bool {
```
ansnunez, line 1337: Can remove?

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -1226,11 +1528,24 @@ unsigned char Node::CheckLegitimacyOfTxnHashes(bytes& errorMsg) {
...
-      LOG_GENERAL(WARNING, "The leader may have composed wrong order");
+      if (!m_prePrepRunning) {
+        LOG_GENERAL(WARNING, "The leader may have composed wrong order");
+      }
       return LEGITIMACYRESULT::WRONGORDER;
     }
 
     if (missingTxnHashes.size() > 0) {
+      if (!m_prePrepRunning) {
+        LOG_GENERAL(WARNING,
```
ansnunez, line 1539: Maybe better to say "Unexpected missing txns during prepare phase - rejecting microblock"

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -1226,11 +1528,24 @@ unsigned char Node::CheckLegitimacyOfTxnHashes(bytes& errorMsg) {
...
+                    "Handling missing txns in prepare phase should not be "
+                    "needed anymore!!");
+        return false;
+      }
+      {
+        lock_guard<mutex> g(m_mutexPrePrepMissingTxnhashes);
+        m_prePrepMissingTxnhashes.clear();
+        m_prePrepMissingTxnhashes = missingTxnHashes;
```
ansnunez, line 1547: Is `move` semantics possible here?

src/libNode/Node.cpp
```diff
@@ -1656,6 +1664,29 @@ bool Node::ProcessSubmitMissingTxn(const bytes& message, unsigned int offset,
     return false;
   }
 
+  if (m_prePrepRunning) {
```
ansnunez, line 1667: If not pre-prepare, can function return false already here?

src/libNode/Node.cpp
```diff
@@ -1753,14 +1784,15 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
...
   // excess data beyond the TxnPacket
-  bytes message2 = {MessageType::NODE, NodeInstructionType::FORWARDTXNPACKET};
-  if (!Messenger::SetNodeForwardTxnBlock(
-          message2, MessageOffset::BODY, epochNumber, dsBlockNum, shardId,
-          lookupPubKey, transactions, signature)) {
-    LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
-              "Messenger::GetNodeForwardTxnBlock failed.");
-    return false;
-  }
+  // bytes message2 = {MessageType::NODE,
```
ansnunez, line 1787: Good optimization (similar to issue 629). Anyway can also delete this code block now to clean up.

src/libNode/Node.cpp
```diff
@@ -1785,55 +1817,72 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
...
       return true;
     }
   }
 
+  // for shard:
+  // 1. DS epoch: lookup dispatch in new DS epoch, buffer and dispatch when mb
+  // finish, so normally first DS epoch won't have txns
+  // 2. FB epoch: saying the 2nd epoch after ds epoch, lookup dispatch txn upon
+  // mb, they distribute right away until mb consensus started. During the mb
+  // submitting mb, all the packet received should be buffered
```
ansnunez, line 1833: This and next line sound confusing. Anyway I think we can re-phrase these comments more clearly (can discuss offline). Also, these will be documented in detail in the ZIP.

ansnunez (changes requested):

src/libNode/Node.cpp
sandipbhoir: @haichuan

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -1481,6 +1807,23 @@ bool Node::CheckMicroBlockValidity(bytes& errorMsg,
   // persistent storage lib)
 }
 
+bool Node::PrePrepMicroBlockValidator(
+    const bytes& message, unsigned int offset, bytes& errorMsg,
+    const uint32_t consensusID, const uint64_t blockNumber,
+    const bytes& blockHash, const uint16_t leaderID, const PubKey& leaderKey,
+    bytes& messageToCosign) {
+  LOG_MARKER();
+  if (!MicroBlockValidator(message, offset, errorMsg, consensusID, blockNumber,
```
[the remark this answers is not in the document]
sandipbhoir: I think we don't need to. Just the way we didn't need to have it previously

ansnunez (changes requested):

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -912,6 +1043,78 @@ void Node::UpdateBalanceForPreGeneratedAccounts() {
...
+  DetachedFunction(1, t);
+}
+
+bool Node::WaitUntilTxnProcessingDone() {
+  LOG_MARKER();
+  // wait for txn processing being ready by me (backup)
+  unique_lock<mutex> lock(m_mutexCVTxnProcFinished);
+  int timeout_time = std::max(
+      0,
+      ((int)MICROBLOCK_TIMEOUT -
```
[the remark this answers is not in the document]
sandipbhoir: yes. formula should be still same. `TX_DISTRIBUTE_TIME_IN_MS` still exist. However, now we can lower it.
We better have this formula so that we can fine tune ( possibly reduce ) constants like `MICROBLOCK_TIMEOUT` and `TX_DISTRIBUTE_TIME_IN_MS` to target new block time.

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -186,24 +266,44 @@ bool Node::OnNodeMissingTxns(const bytes& errorMsg, const unsigned int offset,
...
     } else {
-      LOG_GENERAL(INFO,
-                  "Leader unable to find txn proposed in microblock " << hash);
+      LOG_GENERAL(
+          INFO, "Leader unable to find txn in own created txns list " << hash);
       continue;
     }
   }
 
+  /*
```
ansnunez, line 284: Can do some cleanup here

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -186,24 +266,44 @@ bool Node::OnNodeMissingTxns(const bytes& errorMsg, const unsigned int offset,
 
   std::vector<Transaction> txns;
 
-  const std::unordered_map<TxnHash, TransactionWithReceipt>&
-      processedTransactions = (epochNum == m_mediator.m_currentEpochNum)
-                                  ? t_processedTransactions
-                                  : m_processedTransactions[epochNum];
+  lock_guard<mutex> g(m_mutexCreatedTransactions);
```
ansnunez, line 269: Maybe can scope this mutex to exclude `SendMessage` below, in case multiple backups are fetching the missing txns

src/libMessage/ZilliqaMessage.proto
```diff
@@ -1048,6 +1048,11 @@ message ConsensusCollectiveSig
     ConsensusInfo consensusinfo = 1;
     ByteArray pubkey            = 2;
     ByteArray signature         = 3;
+    oneof oneof1
```
ansnunez, line 1051: I think `oneof` isn't necessary. There are 2 purposes for `oneof` in ZilliqaMessage: (1) to preserve proto2 data format by force-adding a value even if it's default (e.g., for primitive numeric types), and (2) in case there are multiple possible data types (like `union` in C).

ansnunez (changes requested):

src/libMessage/ZilliqaMessage.proto
sandipbhoir: I thought its also used for specifying optional field. The `FinalCollectiveSig` is not going to have `newannouncement` field.

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -290,17 +328,63 @@ bool DirectoryService::RunConsensusOnFinalBlockWhenDSPrimary() {
...
+    m_mediator.m_node->m_microblock = nullptr;
+  } else {
+    lock_guard<mutex> g(m_mutexMicroBlocks);
+    m_microBlocks[m_mediator.m_currentEpochNum].emplace(
+        *(m_mediator.m_node->m_microblock));
+  }
+
+  // stores it in m_finalBlock
+  if (!ComposeFinalBlock()) {
+    LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
```
ansnunez, line 384: Might wanna change the message to `DS ComposeFinalBlock (prep) Failed` for consistency with `ComposePrePrepMicroBlock` above and `ComposeMicroBlock` below.

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -184,8 +184,46 @@ bool DirectoryService::ComposeFinalBlock() {
...
+  m_cvCompleteFinalBlockReady.notify_all();
+
+  return true;
+}
+
+bool DirectoryService::WaitUntilCompleteFinalBlockIsReady() {
+  LOG_MARKER();
+  unique_lock<mutex> lock(m_mediator.m_node->m_mutexMicroBlock);
+  int timeout_time = std::max(
+      0, ((int)MICROBLOCK_TIMEOUT -
```
ansnunez, line 205: This timeout definition seems a bit strange now because `MICROBLOCK_TIMEOUT` and `TX_DISTRIBUTE_TIME_IN_MS` have no more meaning for leader when it processes txns while the backups are confirming the pre-prepare txn list in the announcement message and then also processing the same txns on their end.

src/libDirectoryService/FinalBlockPreProcessing.cpp
```diff
@@ -184,8 +184,46 @@ bool DirectoryService::ComposeFinalBlock() {
...
+
+  m_completeFinalBlockReady = true;
+  m_cvCompleteFinalBlockReady.notify_all();
+
+  return true;
+}
+
+bool DirectoryService::WaitUntilCompleteFinalBlockIsReady() {
+  LOG_MARKER();
+  unique_lock<mutex> lock(m_mediator.m_node->m_mutexMicroBlock);
```
ansnunez, line 203: What's the reason for holding the MB mutex during the wait for FB generation?

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -908,6 +1021,84 @@ void Node::UpdateBalanceForPreGeneratedAccounts() {
...
+                "need to adjust timeouts.");
+      return false;
+    }
+  }
+  return true;
+}
+
+bool Node::WaitUntilCompleteMicroBlockIsReady() {
+  LOG_MARKER();
+  unique_lock<mutex> lock(m_mutexMicroBlock);
```
ansnunez, line 1075: What's the reason for holding the MB mutex during the wait for MB generation?

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -908,6 +1021,84 @@ void Node::UpdateBalanceForPreGeneratedAccounts() {
...
+  int timeout_time = std::max(
+      0, ((int)MICROBLOCK_TIMEOUT -
+          ((int)TX_DISTRIBUTE_TIME_IN_MS +
+           (int)((m_mediator.m_ds->m_mode == DirectoryService::Mode::IDLE)
+                     ? SHARD_ANNOUNCEMENT_DELAY_IN_MS
+                     : DS_ANNOUNCEMENT_DELAY_IN_MS)) /
```
ansnunez, line 1081: I believe `WaitUntilCompleteFinalBlockIsReady` is the one used by DS leader, so can probably just use `SHARD_ANNOUNCEMENT_DELAY_IN_MS`

src/libDirectoryService/FinalBlockPreProcessing.cpp
ansnunez: My understanding is:
* `Node::NotifyTimeout` - actual timeout that cuts off txn processing in both leader and backup, in both DS- and shard-level consensus
* `Node::WaitUntilTxnProcessingDone` - timeout used by backup for MB readiness check, in both DS- and shard-level consensus
* `Node::WaitUntilCompleteMicroBlockIsReady` - timeout used by leader for MB readiness check, in shard-level consensus
* `DirectoryService::WaitUntilCompleteFinalBlockIsReady` - timeout used by leader for FB readiness check, in DS-level consensus

I wonder if we should change the timeout at shard- and DS-level consensus. Shard-level can probably use the current definition, but for DS-level maybe we have something that does not involve `MICROBLOCK_TIMEOUT` and `TX_DISTRIBUTE_TIME_IN_MS`?

ansnunez (changes requested):

src/libNode/Node.cpp
```diff
@@ -1582,6 +1590,34 @@ bool Node::ProcessSubmitMissingTxn(const bytes& message, unsigned int offset,
...
+    if (txns.size() != m_prePrepMissingTxnhashes.size()) {
+      LOG_GENERAL(WARNING,
+                  "Expected and received number of missing txns mismatched!");
+      return false;
+    }
+    unsigned int numfound = 0;
+    for (const auto& missingtxn : m_prePrepMissingTxnhashes) {
+      for (const auto& receivedtxn : txns) {
+        if (missingtxn == receivedtxn.GetTranID()) {
+          ++numfound;
```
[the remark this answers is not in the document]
sandipbhoir: cannot break here. we need to get number of missing txn received.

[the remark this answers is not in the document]
sandipbhoir: i do agree on slowness optimization

src/libDirectoryService/FinalBlockPreProcessing.cpp
sandipbhoir: I think it not holding but making sure to wait until the complete DS MB(with txns) is generated before we proceed.

src/libNode/MicroBlockPreProcessing.cpp
sandipbhoir: same reason as for final block wait.

src/libNode/Node.cpp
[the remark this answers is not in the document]
ansnunez: The break would just be for the second for loop; since `missingtxn` is already equal to `receivedtxn`, no need to keep reading through `txns` right?

src/libNode/MicroBlockPreProcessing.cpp
```diff
@@ -950,6 +1063,81 @@ void Node::UpdateBalanceForPreGeneratedAccounts() {
...
+    }
+  };
+  DetachedFunction(1, t);
+}
+
+bool Node::WaitUntilTxnProcessingDone() {
+  LOG_MARKER();
+  // wait for txn processing being ready by me (backup)
+  unique_lock<mutex> lock(m_mutexCVTxnProcFinished);
+  int timeout_time = std::max(
```
ansnunez, line 1090: Does this also need to be updated to use `DS_TX_PROCESSING_TIMEOUT`?

ansnunez (changes requested):

src/libNode/MicroBlockPreProcessing.cpp
sandipbhoir: yes. good catch

ansnunez (approved):

chetan-zilliqa (approved):

PR #2230, opened 2020-09-18
[release-7.0] Change server selection timeout to 1s

KaustubhShamshery:
## Description

* Make server selection a constant
* Do not try to write if no txns being inserted or updated

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libServer/LookupServer.cpp
```diff
@@ -2019,6 +2019,11 @@ Json::Value LookupServer::GetTransactionStatus(const string& txnhash) {
                              "Txn Hash size not appropriate");
     }
 
+    if (!REMOTESTORAGE_DB_ENABLE) {
```
ansnunez, line 2022: I suggest to make this the very first check, and also change the message to "API not supported" (remote storage is an internal detail)

ansnunez (changes requested):

constants.xml
```diff
@@ -40,6 +40,7 @@
         <REMOTESTORAGE_DB_PORT>27017</REMOTESTORAGE_DB_PORT>
         <REMOTESTORAGE_DB_ENABLE>false</REMOTESTORAGE_DB_ENABLE>
         <REMOTESTORAGE_DB_TLS_FILE/>
+        <REMOTESTORAGE_DB_SERVER_SELECTION_TIMEOUT_MS>1000</REMOTESTORAGE_DB_SERVER_SELECTION_TIMEOUT_MS>
```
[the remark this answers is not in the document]
KaustubhShamshery: I think the write operation is called atleast 3 times by lookup in an epoch. 5 seconds * 3 would be too much if mongo server is not available

ansnunez (approved):

tests/RemoteStorageDB/CMakeLists.txt
```diff
@@ -3,5 +3,5 @@ link_directories(${CMAKE_BINARY_DIR}/lib)
 
 add_executable(Test_Mongo Test_mongo.cpp)
 target_include_directories(Test_Mongo PUBLIC ${CMAKE_SOURCE_DIR}/src ${CMAKE_SOURCE_DIR}/tests)
-target_link_libraries(Test_Mongo PUBLIC RemoteStorageDB TestUtils)
+target_link_libraries(Test_Mongo PUBLIC RemoteStorageDB TestUtils AccountData)
```
kaikawaliu, line 6: seems not needed

KaustubhShamshery: If you try and build test_mongo it is needed as account data is not linked.

kaikawaliu: okay

kaikawaliu (approved):

PR #2231, opened 2020-09-18
[release-7.0] Fix rsync call inside auto_backup.py

ansnunez:
## Description

Change introduced in #2179 seems to not work, so executing rsync twice instead as separate bash commands

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/auto_backup.py
+++ b/scripts/auto_backup.py
@@ -110,9 +110,10 @@ def GetCurrentTxBlockNum():
     return blockNum + 1
 
 def CreateTempPersistence():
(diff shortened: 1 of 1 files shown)


sandipbhoir (approved):

Revolution1 (approved):

Revolution1:
another way to do this is like this
```python
subprocess.Popen("ls -l | grep xxx && echo ok", stdout=subprocess.PIPE, shell=True)
```
when `shell=True`, python will use a shell to execute the line of commands

PR #2238, opened 2020-09-22
Bring transaction distribution earlier for shard

kaikawaliu:
Two major changes:
- Lookup dispatch transaction upon receiving of soft confirmation with microblock.
- Shard node distribute gossip for transaction upon receiving of packet from lookup, hold and commit into txn pool upon new epoch. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -5455,11 +5455,14 @@ void Lookup::SendTxnPacketToShard(const uint32_t shardId, bool toDS) {
   bool result = false;
 
   {
-    lock_guard<mutex> g1(m_txnShardMapMutex);
-    lock_guard<mutex> g2(m_txnShardMapGeneratedMutex);
-    if (shardId > m_txnShardMapGenerated.size()) {
+    unique_lock<mutex> g1(m_txnShardMapMutex, defer_lock);
+    unique_lock<mutex> g2(m_txnShardMapGeneratedMutex, defer_lock);
+    lock(g1, g2);
+
+    if (m_txnShardMapGenerated.find(shardId) == m_txnShardMapGenerated.end()) {
       return;
     }
+
     auto transactionNumber = m_txnShardMapGenerated[shardId].size();
 
     LOG_GENERAL(INFO, "Txn number generated: " << transactionNumber);
@@ -5550,7 +5553,7 @@ void Lookup::SendTxnPacketToShard(const uint32_t shardId, bool toDS) {
         }
 
         if (toSend.size() >= NUM_NODES_TO_SEND_LOOKUP) {
-          return;
+          break;
         }
       }
     }
@@ -5623,7 +5626,7 @@ void Lookup::SendTxnPacketToNodes(const uint32_t oldNumShards,
     return;
   }
 
-  SendTxnPacketToShard(oldNumShards, newNumShards);
+  SendTxnPacketPrepare(oldNumShards, newNumShards);
 
   for (unsigned int i = 0; i < newNumShards + 1; i++) {
     SendTxnPacketToShard(i, i == newNumShards);
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -5357,7 +5357,8 @@ bool Lookup::DeleteTxnShardMap(uint32_t shardId) {
   return true;
 }
 
-void Lookup::SenderTxnBatchThread(const uint32_t oldNumShards) {
+void Lookup::SenderTxnBatchThread(const uint32_t oldNumShards,
+                                  bool newDSEpoch) {
   if (!LOOKUP_NODE_MODE) {
     LOG_GENERAL(WARNING,
                 "Lookup::SenderTxnBatchThread not expected to be called from "
@@ -5372,13 +5373,17 @@ void Lookup::SenderTxnBatchThread(const uint32_t oldNumShards) {
     return;
   }
 
-  auto main_func = [this, oldNumShards]() mutable -> void {
+  auto main_func = [this, oldNumShards, newDSEpoch]() mutable -> void {
     m_startedTxnBatchThread = true;
     uint32_t numShards = 0;
     while (true) {
       if (!m_mediator.GetIsVacuousEpoch()) {
         numShards = m_mediator.m_ds->GetNumShards();
-        SendTxnPacketToNodes(oldNumShards, numShards);
+        if (newDSEpoch) {
+          SendTxnPacketToNodes(oldNumShards, numShards);
+        } else {
+          SendTxnPacketToDS(oldNumShards, numShards);
+        }
       }
       break;
     }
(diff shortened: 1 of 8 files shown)


sandipbhoir (changes requested):

src/libNode/Node.cpp
```diff
@@ -1923,6 +1937,13 @@ bool Node::ProcessTxnPacketFromLookupCore(const bytes& message,
     P2PComm::GetInstance().SendBroadcastMessage(toSend, message);
   }
 
+  if (m_mediator.m_ds->m_mode == DirectoryService::Mode::IDLE &&
+      m_state != MICROBLOCK_CONSENSUS_PREP) {
+    unique_lock<mutex> lk(m_mutexCVWaitDSBlock);
+    cv_txnPacket.wait(lk,
+                      [this] { return m_state == MICROBLOCK_CONSENSUS_PREP; });
+  }
+
```
sandipbhoir, line 1946: by the time cv_txnPacket is notified ( after moving to MICROBLOCK_CONSENSUS_PREP), its seems possible that leader/backup already start processing transaction.
Here i think we assume all pkts are checked, validated and added to m_createdTxns by then which may not hold true.

lock_guard<mutex> g(m_mutexCreatedTransactions) at Line 2026 will give sequential processing. so with multiple packets, it may fail to check and add to mem pool before , leader actually start processing at Line 1100

sandipbhoir (changes requested):

src/libNode/Node.cpp
sandipbhoir: i see there is sleep for `TXN_DISTRIBUTION_TIME_MS` but still there is chance for race condition provided it can be as small as 1 sec.
May be need some cv to indicate all pkts were checked and added to pool before starting processing txns. ofcorse need to some timeout for that as well.
Basically give enough time to do validation and adding to mempool for all pkts that were received.

kaikawaliu: yup very good point, will address it

src/libNode/Node.cpp
```diff
@@ -2059,6 +2082,12 @@ bool Node::ProcessTxnPacketFromLookupCore(const bytes& message,
                               << processed_count << "]");
   }
 
+  if (m_txnPacketThreadOnHold > 0) {
+    m_txnPacketThreadOnHold--;
+    if (m_txnPacketThreadOnHold == 0) {
```
sandipbhoir, line 2087: Can remove this?

sandipbhoir (changes requested):

sandipbhoir (approved):

src/libLookup/Lookup.cpp
```diff
@@ -5436,145 +5441,193 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
...
-  if (m_mediator.m_disableTxns) {
-    LOG_GENERAL(INFO, "Txns disabled - skipping dispatch to shards");
-    return;
-  }
+  bytes msg = {MessageType::NODE, NodeInstructionType::FORWARDTXNPACKET};
+  bool result = false;
 
-  const uint32_t numShards = newNumShards;
+  {
+    lock_guard<mutex> g1(m_txnShardMapMutex);
```
KaustubhShamshery, line 5458: use `lock` to avoid deadlock?

src/libLookup/Lookup.cpp
```diff
@@ -5436,145 +5441,193 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
...
+      for (auto const& i : *m_mediator.m_DSCommittee) {
+        if (toSend.size() < NUM_NODES_TO_SEND_LOOKUP &&
+            i.second != dsLeader.second) {
+          toSend.push_back(i.second);
         }
-        if (m_mediator.m_ds->m_shards.at(i).empty()) {
-          continue;
+
+        if (toSend.size() >= NUM_NODES_TO_SEND_LOOKUP) {
+          return;
```
sandipbhoir, line 5553: While testing i encountered that this should be `break` instead of `return`.

kaikawaliu: will fix

src/libLookup/Lookup.cpp
```diff
@@ -5436,145 +5441,193 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
...
+  LOG_MARKER();
+
+  if (!LOOKUP_NODE_MODE) {
+    LOG_GENERAL(WARNING,
+                "Lookup::SendTxnPacketToNodes not expected to be called from "
+                "other than the LookUp node.");
+    return;
+  }
+
+  SendTxnPacketToShard(oldNumShards, newNumShards);
```
sandipbhoir, line 5626: we are sending to DS shard twice??

kaikawaliu: Typo here, should be SendTxnPacketPrepare

src/libLookup/Lookup.cpp
```diff
@@ -5436,145 +5441,195 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
...
+  bytes msg = {MessageType::NODE, NodeInstructionType::FORWARDTXNPACKET};
+  bool result = false;
 
-  const uint32_t numShards = newNumShards;
+  {
+    unique_lock<mutex> g1(m_txnShardMapMutex, defer_lock);
+    unique_lock<mutex> g2(m_txnShardMapGeneratedMutex, defer_lock);
+    lock(g1, g2);
+
+    if (shardId > m_txnShardMapGenerated.size()) {
```
sandipbhoir, line 5462: when this is being called from `SendTxnPacketToDS`, map `m_txnShardMapGenerated` is cleared in `SendTxnPacketPrepare` and populated only for given shard id ( say 3 for ds ).
So above check will always fail and no txns are sent to DS. For ex: shard id 3 and `m_txnShardMapGenerated.size() = 1`
May be should better check for existing of key `shardid` in map `m_txnShardMapGenerated` instead?

sandipbhoir (changes requested):

src/libLookup/Lookup.cpp
```diff
@@ -5436,145 +5441,196 @@ void Lookup::RectifyTxnShardMap(const uint32_t oldNumShards,
...
+void Lookup::SendTxnPacketToDS(const uint32_t oldNumShards,
+                               const uint32_t newNumShards) {
+  LOG_MARKER();
 
-        for (auto const& i : *m_mediator.m_DSCommittee) {
-          if (toSend.size() < NUM_NODES_TO_SEND_LOOKUP &&
-              i.second != dsLeader.second) {
-            toSend.push_back(i.second);
-          }
+  SendTxnPacketPrepare(oldNumShards, newNumShards);
```
sandipbhoir, line 5574: This will generates txns for all shards ( not just ds shard ). This one most likely will cause high nonce issue on ds shard because txn pkt to ds will be sent immediately.

sandipbhoir (changes requested):

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1364,20 +1371,20 @@ bool Node::ProcessMBnForwardTransaction(const bytes& message,
...
     }
 
     // shard microblock only:
     // pre-process of early MBnForwardTxn submission
     // soft confirmation
     SoftConfirmForwardedTransactions(entry);
-    // [TODO] invoke txn distribution
+    // invoke txn distribution
+    if (!m_mediator.GetIsVacuousEpoch()) {
+      m_mediator.m_lookup->SendTxnPacketToShard(
```
sandipbhoir, line 1385: This sends the same txns which were sent when new dsepoch takes place.
so as soon as early confirmation is received for microblock of epoch 5, it will send the same txns fror epoch 6 which were already send after ds block were processed.
This causes no new txns being processed for epoch 6 by shards.

sandipbhoir (changes requested):

src/libNode/Node.cpp
```diff
@@ -1782,55 +1783,68 @@ bool Node::ProcessTxnPacketFromLookup([[gnu::unused]] const bytes& message,
...
     }
     SHA2<HashType::HASH_VARIANT_256> sha256;
-    sha256.Update(message2);  // message hash
+    sha256.Update(message);  // message hash
     bytes msg_hash = sha256.Finalize();
-    m_txnPacketBuffer.emplace(msg_hash, message2);
+    m_txnPacketBuffer.emplace(msg_hash, message);
   } else {
+    // not from lookup and in proper state
     LOG_GENERAL(INFO,
```
sandipbhoir, line 1844: This message may not be always true. since we do process messages received directly from lookup now due to change at line Line 1822.

sandipbhoir (changes requested):

sandipbhoir:
Already merged to master as part of https://github.com/Zilliqa/Zilliqa/pull/2216

PR #2271, opened 2020-10-05
[release-6.4] Redefine recovery process of normal and dsguard nodes

sandipbhoir:
## Description
Addresses issue https://github.com/Zilliqa/Issues/issues/670 and also remediate issue https://github.com/Zilliqa/Issues/issues/669

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] small-scale cloud test
 - [x] Recover entire testnet from old testnet
 - [x] Recover one normal node part of any existing shard
 - [x] Recover one normal node not part of any existing shard 
 - [x] Recover one dsguard node part of current dscommittee
 - [x] scale new node (miner node)

### DIFF
--- a/src/libZilliqa/Zilliqa.cpp
+++ b/src/libZilliqa/Zilliqa.cpp
@@ -177,6 +177,18 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
     LOG_GENERAL(FATAL, "Archvial lookup is true but not lookup ");
   }
 
+  // when individual node is being recovered and persistence is not available
+  // locally, then Rejoin as if new miner node which will download persistence
+  // from S3 incremental db and identify if already part of any
+  // shard/dscommittee and proceed accordingly.
+
+  if (!LOOKUP_NODE_MODE && (SyncType::RECOVERY_ALL_SYNC == syncType)) {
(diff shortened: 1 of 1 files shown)


chetan-zilliqa (approved):

src/libZilliqa/Zilliqa.cpp
```diff
@@ -201,6 +189,30 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
...
+      m_lookup.SetSyncType(SyncType::NEW_SYNC);
+    }
+  }
+
+  if (SyncType::NEW_SYNC == syncType) {
+    // Setting it earliest before even p2pcomm is instantiated
+    m_n.m_runFromLate = true;
+  }
+
+  P2PComm::GetInstance().SetSelfPeer(peer);
```
ansnunez, line 209: Will `IsNodeInDSGuardList` and `IsNodeInShardGuardList` above work if `SetSelfPeer` and `SetSelfKey` are now set later?

ansnunez (changes requested):

src/libZilliqa/Zilliqa.cpp
sandipbhoir: ohh ya..good catch. when tested against guard nodes it works. But it won't log correctly.

ansnunez (approved):

PR #2329, opened 2020-10-30
[release-7.0] Keep one mutex for both reading and deletion of txn shard map

ansnunez:
## Description

Newly created transactions can be accidentally deleted before being dispatched to upper_seed, because the mutex on the txn-shard map is acquired twice: first for creating FORWARDTXN message, and then for deletion of the map. If a transaction is added to the map in between the two steps, it is deleted before dispatch.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -5209,8 +5209,6 @@ bool Lookup::DeleteTxnShardMap(uint32_t shardId) {
     return true;
   }
 
-  lock_guard<mutex> g(m_txnShardMapMutex);
-
   m_txnShardMap[shardId].clear();
 
   return true;
@@ -5396,6 +5394,7 @@ void Lookup::SendTxnPacketToNodes(const uint32_t oldNumShards,
 
       P2PComm::GetInstance().SendBroadcastMessage(toSend, msg);
 
+      lock_guard<mutex> g(m_txnShardMapMutex);
       DeleteTxnShardMap(i);
     } else if (i == numShards) {
       // To send DS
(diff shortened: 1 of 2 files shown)


kaikawaliu (approved):

src/libLookup/Lookup.cpp
```diff
@@ -5575,6 +5573,7 @@ void Lookup::SendTxnPacketToNodes(const uint32_t oldNumShards,
 
       P2PComm::GetInstance().SendBroadcastMessage(toSend, msg);
 
+      lock_guard<mutex> g(m_mediator.m_lookup->m_txnShardMapMutex);
```
KaustubhShamshery, line 5576: is this different from just using `m_txnShardMapMutex`?

ansnunez: Should be the same. I can make the change.

src/libServer/LookupServer.cpp
```diff
@@ -349,14 +349,13 @@ bool LookupServer::StartCollectorThread() {
                     SEND_TYPE::ARCHIVAL_SEND_DS))) {
           continue;
         }
+        for (auto const& i :
```
KaustubhShamshery, line 352: Need to acquire mutex somewhere here?

ansnunez: Mutex is already acquired at line 343

KaustubhShamshery: Oh did not see, thanks

KaustubhShamshery (approved):

PR #2361, opened 2020-11-17
[master] Fix dm3 for commit failed node catch up with consensus

kaikawaliu:
## Description

Fixing deprecated dm3 test due to timing issue. After this fix, a DS node missing MBs can replay the consensus with the patch-up message from leader and proceed to the next epoch.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] small-scale cloud test


src/libDirectoryService/Coinbase.cpp
```diff
@@ -160,12 +160,14 @@ void DirectoryService::InitCoinbase() {
       ++it;
   }
 
-  const auto& vecLookup = m_mediator.m_lookup->GetLookupNodesStatic();
   const auto& epochNum = m_mediator.m_currentEpochNum;
 
-  for (const auto& lookupNode : vecLookup) {
-    m_coinbaseRewardees[epochNum][CoinbaseReward::LOOKUP_REWARD].push_back(
-        lookupNode.first);
+  if (coinbaseAppendLookup) {
```
kaikawaliu, line 165: The reason for this is `InitCoinbase` will be invoked every time a ds trying to `PrepareRunConsensusOnFinalBlockNormal
`, which will also when the failed node received the missing microblocks from leader and restart the consensus. 
If we don't have the flag here, the number of lookup node will duplicate and affect the result of coinbase rewards.

ansnunez: How about if we just do `m_coinbaseRewardees[epochNum][CoinbaseReward::LOOKUP_REWARD].clear()` before going through the lookups?

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -558,7 +561,16 @@ bool DirectoryService::ProcessFinalBlockConsensusCore(
...
     m_viewChangeCounter = 0;
-    ProcessFinalBlockConsensusWhenDone();
+    // Update the final block with the co-signatures from the consensus
+    m_finalBlock->SetCoSignatures(*m_consensusObject);
+
+    // Update the DS microblock with the same co-signatures from the consensus
+    // If we don't do this, DataSender won't be able to process it
+    m_mediator.m_node->m_microblock->SetCoSignatures(*m_consensusObject);
+
+    auto func = [this]() -> void { ProcessFinalBlockConsensusWhenDone(); };
```
ansnunez, line 571: Why do we need to spawn another thread?

ansnunez (changes requested):

src/libDirectoryService/FinalBlockPostProcessing.cpp
kaikawaliu: Because the mutex `m_mutexConsensus` will be occupied until the consensus end if we didn't throw out a thread. While missing microblock fetching still requires the consensus to be in activated state.

src/libDirectoryService/Coinbase.cpp
kaikawaliu: good idea

chetan-zilliqa (approved):

src/libDirectoryService/FinalBlockPostProcessing.cpp
sandipbhoir: But `ProcessFinalBlockConsensusWhenDone` won't be called if `state == ConsensusCommon::State::ERROR` and while MB is being fetched.
So `m_mutexConsensus` is not needed right? or am i missing something

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -279,6 +280,8 @@ void DirectoryService::ProcessFinalBlockConsensusWhenDone() {
   }
   m_mediator.m_node->ClearUnconfirmedTxn();
 
+  this_thread::sleep_for(chrono::seconds(FB_DONE_SET_STATE_DELAY_IN_SECONDS));
```
sandipbhoir, line 283: why we need a delay?

src/libDirectoryService/FinalBlockPostProcessing.cpp
sandipbhoir: still didn't get clear reason for spawning new thread
@ansnunez did you?

PR #2364, opened 2020-11-19
[release-7.0] Minor fix to RefreshDB(TX_BODY)

ansnunez:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


chetan-zilliqa (approved):
Just to confirm the motivation here.
We want a behavior where if any of the `txBodyDB->RefreshDB()` refresh fails, then return the error instead of just checking on `ret = m_txEpochDB->RefreshDB();`
correct ?

sandipbhoir (approved):

ansnunez:
> Just to confirm the motivation here.
> We want a behavior where if any of the `txBodyDB->RefreshDB()` refresh fails, then return the error instead of just checking on `ret = m_txEpochDB->RefreshDB();`
> correct ?

Yes correct

PR #2377, opened 2020-12-03
[master] Add importing and exporting of persistence

KaustubhShamshery:
## Description

Add an API called `ExportPersistence` for exporting persistence

Add bootstrap option --path for choosing a tar file from where to load persistence

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -28,6 +28,7 @@
 
 using namespace std;
 namespace po = boost::program_options;
+namespace fs = boost::filesystem;
 
 int readAccountJsonFromFile(const string& path) {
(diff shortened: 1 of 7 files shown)


bb111189 (changes requested):
1. sanitize input
2. disable such API by default. togglable by constants
3. avoid using system()

bb111189 (changes requested):
Can we make the reinit api togglable? 

For instance, for public isolation, it should not be available. For private isolated like Ceres, it should be enabled.

bb111189 (approved):

ansnunez (approved):

PR #2419, opened 2021-01-19
[master] Allow txn dispatching to ds-shard when no other shard exists

sandipbhoir:
## Description
This PR addresses: https://github.com/Zilliqa/Issues/issues/675

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -5581,11 +5581,6 @@ bool Lookup::ProcessForwardTxn(const bytes& message, unsigned int offset,
   if (!ARCHIVAL_LOOKUP) {
     uint32_t shard_size = m_mediator.m_ds->GetNumShards();
 
-    if (shard_size == 0) {
-      LOG_GENERAL(WARNING, "Shard size 0");
-      return false;
-    }
-
     if (!m_sendSCCallsToDS) {
       for (const auto& txn : txnsShard) {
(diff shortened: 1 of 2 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -5581,11 +5581,6 @@ bool Lookup::ProcessForwardTxn(const bytes& message, unsigned int offset,
   if (!ARCHIVAL_LOOKUP) {
     uint32_t shard_size = m_mediator.m_ds->GetNumShards();
 
-    if (shard_size == 0) {
-      LOG_GENERAL(WARNING, "Shard size 0");
-      return false;
```
KaustubhShamshery, line 5586: should u return?

KaustubhShamshery (changes requested):

src/libLookup/Lookup.cpp
sandipbhoir: i think thats the purpose, the transaction needs to be forwarded to shard 0 ( ds-shard in above case ).
should not return right? if i got you correctly..

KaustubhShamshery (approved):

KaustubhShamshery (approved):

ansnunez (approved):

PR #2427, opened 2021-01-25
[master] Changed version check logic in ProcessPoWSubmission

ansnunez:
## Description

This is a temporary measure to allow mining nodes on newer versions (that are optional) to continue submitting PoW tagged with older version.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


sandipbhoir (approved):

sandipbhoir (approved):

constants.xml
```diff
@@ -240,6 +240,7 @@
         <TXN_DS_TARGET_NUM>360</TXN_DS_TARGET_NUM>
         <PRIORITY_TOLERANCE_IN_PERCENT>5</PRIORITY_TOLERANCE_IN_PERCENT>
         <SKIP_POW_REATTEMPT_FOR_DS_DIFF>false</SKIP_POW_REATTEMPT_FOR_DS_DIFF>
+        <POW_SUBMISSION_VERSION_TAG/>
```
chetan-zilliqa, line 243: is this ok ? 
are we not providing any default values here.

ansnunez: Will put the version once it's used in a particular configuration (e.g., "v7.2.0").
Should be blank by default so that small-scale testnets use the VERSION_TAG value.
In next disruptive upgrade, we either leave it blank or remove the constant entirely.

chetan-zilliqa: got it.Thanks

chetan-zilliqa (approved):

PR #2434, opened 2021-02-05
Improve Daemon to start new process when failed to monitor previous process

sandipbhoir:
## Description
Daemon is improved to start the process if it fails to monitor the previous running process.
I also stop the `scilla-server` process if running whenever new `zilliqa` process is started.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/daemon/ZilliqaDaemon.cpp
+++ b/daemon/ZilliqaDaemon.cpp
@@ -23,6 +23,8 @@ namespace po = boost::program_options;
 #define SUCCESS 0
 #define ERROR_IN_COMMAND_LINE -1
 
+#define MONITORING_FAIL_COUNT 10
+
 const vector<string> programName = {"zilliqa"};
 const string SYNCTYPE_OPT = "--synctype";
 const char* synctype_descr =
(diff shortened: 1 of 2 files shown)


daemon/ZilliqaDaemon.cpp
```diff
@@ -72,18 +74,33 @@ ZilliqaDaemon::ZilliqaDaemon(int argc, const char* argv[], std::ofstream& log)
...
+          (m_nodeType == "dsguard" || m_nodeType == "normal")) {
+        auto it = m_failedMonitorProcessCount.find(name);
+        if (it == m_failedMonitorProcessCount.end()) {
+          m_failedMonitorProcessCount[name] = 1;
+        } else {
+          m_failedMonitorProcessCount[name]++;
+        }
+        if (m_failedMonitorProcessCount[name] >= MONITORING_FAIL_COUNT) {
+          StartNewProcess(true);
+          m_failedMonitorProcessCount[name] = 0;
```
ansnunez, line 93: Do you also need to populate `tmp` at this point with the new process ID? So that when it goes to line 98 below, it will report the process ID.

daemon/ZilliqaDaemon.cpp
```diff
@@ -230,8 +247,10 @@ vector<pid_t> ZilliqaDaemon::GetProcIdByName(const string& procName) {
   return result;
 }
 
-void ZilliqaDaemon::StartNewProcess() {
-  KillProcess();
+void ZilliqaDaemon::StartNewProcess(bool cleanPersistence) {
```
ansnunez, line 250: For consistency with KillProcess signature change, should we pass `procName` here also, instead of using `programName`?

daemon/ZilliqaDaemon.h
```diff
@@ -62,9 +64,9 @@ class ZilliqaDaemon {
   bool DownloadPersistenceFromS3();
```
ansnunez, line 64: I think we can delete this function, as it has not really been used for a long time

sandipbhoir: We can keep it for now. May be future work for recovery of exchange seed nodes can use it.
It was added for the community seed node but is currently broken.

daemon/ZilliqaDaemon.cpp
sandipbhoir: No need i think. It will monitor the process in next run of `MonitorProcess` i.e. after 5 sec.
Note that with `StartNewProcess`, the actual process may take some time to get spawned.  So reading for pid can fail if checked immediately.

ansnunez (approved):

chetan-zilliqa (approved):

PR #2437, opened 2021-02-08
[master] Fix for libevent deadlock in p2pseed

chetan-zilliqa:
## Description

In `p2pseed` branch, we observed deadlock situation during testing for testnet termination scenarios or this could probably also happen if all seed nodes are down for long time(say 2-4 mins).Below is explanation.

When the  testnet is terminated, the seedpubs will always send `BEV_EVENT_ERROR`.
On the client side `GetVCFinalBlockFromL2lDataProvider` is run in the infinite loop.The sender application thread calling `SendMsgToSeedNodeOnWire()` keeps waiting always after taking `m_mutexBufferEvent` in function `bufferevent_write`.This happens randomly if one of the libevent callbacks is received out of order.
The `CloseAndFreeBevP2PSeedConn` function is also waiting to acquire lock.This cause deadlock.

Solution:
We should perform `bufferevent_write()` only if the socket is connected.This way, it will not go to perform `bufferevent_write` in the `BEV_EVENT_ERROR`(i.e when the server is down) scenarios.

Note : We won't have face this problem if `bufferevent_socket_connect()` had returned immediately on failure.But it does not happen. Libevent provides socket connection error only in event callback for non blocking operation.

 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNetwork/P2PComm.cpp
+++ b/src/libNetwork/P2PComm.cpp
@@ -531,7 +531,7 @@ void P2PComm::CloseAndFreeBufferEvent(struct bufferevent* bufev) {
   bufferevent_free(bufev);
 }
 
-void P2PComm::CloseAndFreeBevP2PSeedConn(struct bufferevent* bufev) {
+void P2PComm::CloseAndFreeBevP2PSeedConnServer(struct bufferevent* bufev) {
   lock(m_mutexPeerConnectionCount, m_mutexBufferEvent);
   unique_lock<mutex> lock(m_mutexPeerConnectionCount, adopt_lock);
   lock_guard<mutex> g(m_mutexBufferEvent, adopt_lock);
@@ -542,7 +542,7 @@ void P2PComm::CloseAndFreeBevP2PSeedConn(struct bufferevent* bufev) {
   char* strAdd = inet_ntoa(cli_addr.sin_addr);
   int port = cli_addr.sin_port;
   // TODO Remove log
-  LOG_GENERAL(DEBUG, "P2PSeed CloseAndFreeBevP2PSeedConn ip="
+  LOG_GENERAL(DEBUG, "P2PSeed CloseAndFreeBevP2PSeedConnServer ip="
                          << strAdd << " port=" << port << " bev=" << bufev);
   uint128_t ipAddr = cli_addr.sin_addr.s_addr;
   {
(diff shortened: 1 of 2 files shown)


src/libNetwork/P2PComm.cpp
```diff
@@ -1020,19 +1052,30 @@ void P2PComm ::EventCbClientSeed([[gnu::unused]] struct bufferevent* bev,
...
+    if (events & BEV_EVENT_TIMEOUT) {
+      LOG_GENERAL(DEBUG, "P2PSeed BEV_EVENT_TIMEOUT");
+    }
+    if (events & BEV_EVENT_CONNECTED) {
+      LOG_GENERAL(DEBUG, "P2PSeed BEV_EVENT_CONNECTED req msg len="
+                             << destBytes->size());
+    }
+  }
+  if (events & BEV_EVENT_CONNECTED) {
+    if (bufferevent_write(bev, &(destBytes->at(0)), destBytes->size()) < 0) {
```
sandipbhoir, line 1064: may be want to check `ctx/destBytes` for `null` before accessing?

sandipbhoir (changes requested):

sandipbhoir (approved):

src/libNetwork/P2PComm.cpp
```diff
@@ -1020,19 +1053,32 @@ void P2PComm ::EventCbClientSeed([[gnu::unused]] struct bufferevent* bev,
...
+    if (events & BEV_EVENT_CONNECTED) {
+      LOG_GENERAL(DEBUG, "P2PSeed BEV_EVENT_CONNECTED req msg len="
+                             << destBytes->size());
+    }
+  }
+  if (events & BEV_EVENT_CONNECTED) {
+    if (destBytes != NULL) {
+      if (bufferevent_write(bev, &(destBytes->at(0)), destBytes->size()) < 0) {
+        LOG_GENERAL(WARNING, "Error: P2PSeed bufferevent_write failed !!!");
+        return;
```
ansnunez, line 1068: For this case, which function deallocates `destBytes`?

chetan-zilliqa: When we receive the response message from the server seed, `ReadCbClientSeed` would be called.After dispatching of message, we are calling `CloseAndFreeBevP2PSeedConnClient` and it will free the memory.
If there's no response then `BEV_EVENT_TIMEOUT` would be called.
Basically `CloseAndFreeBevP2PSeedConnClient` is the function where memory will be freed.

ansnunez (approved):

PR #2440, opened 2021-02-11
[master] De-blacklist ejected dscommittee members

sandipbhoir:
This PR makes sure that `ejected dscommittee members` which are now `shard-members` are removed from blacklist of existing 
ds-committee members.

This change should take effect only at `UPGRADE_TARGET_DS_NUM`

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPreProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPreProcessing.cpp
@@ -30,6 +30,7 @@
 #include "libCrypto/Sha2.h"
 #include "libMediator/Mediator.h"
 #include "libMessage/Messenger.h"
+#include "libNetwork/Blacklist.h"
 #include "libNetwork/Guard.h"
 #include "libNetwork/P2PComm.h"
 #include "libPOW/pow.h"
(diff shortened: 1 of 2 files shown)


src/libDirectoryService/DSBlockPreProcessing.cpp
```diff
@@ -304,6 +305,12 @@ void DirectoryService::InjectPoWForDSNode(
       LOG_GENERAL(INFO, "Injecting into PoW connections " << rit->second);
     }
 
+    if (m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() >=
+        UPGRADE_TARGET_DS_NUM) {
+      // Remove this node from blacklist if it exists
```
ansnunez, line 310: Do you think everyone (DS, former DS, and new shard nodes) should remove them from blacklist? If so, doing it here might be easier: https://github.com/Zilliqa/Zilliqa/blob/d5cc251e6b3703dc9c70b8f871a63d6c76ba5593/src/libDirectoryService/DSComposition.cpp#L137

sandipbhoir: Yes. Can do

ansnunez (approved):

chetan-zilliqa (approved):

PR #2441, opened 2021-02-16
[master] P2pseed: close connection from server if the key is not whitelisted.

chetan-zilliqa:
## Description

In p2pseed code, client seed is always responsible for closing socket connection on successful response message or no response message.server seed will close upon receipt of `BEV_EVENT_EOF` from client.

However if the key is not whitelisted in our seedpubs, it is better to close connection from server side immediately.In this scenario, client will receive `BEV_EVENT_EOF` from server and terminate socket connection on it's end.

In the same PR, added some optimization in client seed code.
1) To set timeout on bufferevent only if socket connect is successful.
2) Remove `RemoveBevFromMap` from `Zillqa.cpp` for clearing entry from map.Anyway this would be cleared in another call to `RemoveBevFromMap` in `EventCbServerSeed`

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNetwork/P2PComm.cpp
+++ b/src/libNetwork/P2PComm.cpp
@@ -1029,6 +1029,53 @@ void P2PComm::RemoveBevFromMap(const Peer& peer) {
   }
 }
 
+void P2PComm::RemoveBevAndCloseP2PConnServer(const Peer& peer,
+                                             const unsigned& startByteType) {
+  LOG_MARKER();
+  if (startByteType == START_BYTE_SEED_TO_SEED_REQUEST) {
+    lock(m_mutexPeerConnectionCount, m_mutexBufferEvent);
+    unique_lock<mutex> lock(m_mutexPeerConnectionCount, adopt_lock);
+    lock_guard<mutex> g(m_mutexBufferEvent, adopt_lock);
+    string bufKey = peer.GetPrintableIPAddress() + ":" +
+                    boost::lexical_cast<string>(peer.GetListenPortHost());
+    LOG_GENERAL(DEBUG,
+                "P2PSeed RemoveBufferEvent=" << peer << " bufKey =" << bufKey);
+    auto it = m_bufferEventMap.find(bufKey);
+    if (it != m_bufferEventMap.end()) {
+      // TODO Remove this log
+      if (DEBUG_LEVEL == 4) {
+        LOG_GENERAL(DEBUG, "P2PSeed clearing bufferevent for bufKey="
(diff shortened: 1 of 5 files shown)


src/libNetwork/P2PComm.cpp
```diff
@@ -1029,6 +1029,55 @@ void P2PComm::RemoveBevFromMap(const Peer& peer) {
...
+      struct sockaddr_in cli_addr {};
+      socklen_t addr_size = sizeof(struct sockaddr_in);
+      getpeername(fd, (struct sockaddr*)&cli_addr, &addr_size);
+      char* strAdd = inet_ntoa(cli_addr.sin_addr);
+      int port = cli_addr.sin_port;
+      // TODO Remove log
+      LOG_GENERAL(DEBUG, "P2PSeed RemoveBevAndCloseP2PConnServer ip="
+                             << strAdd << " port=" << port << " bev=" << bufev);
+      uint128_t ipAddr = cli_addr.sin_addr.s_addr;
+      {
```
ansnunez, line 1065: Unnecessary bracket?

ansnunez (approved):

src/libNetwork/P2PComm.cpp
chetan-zilliqa: Done

src/libNetwork/P2PComm.cpp
```diff
@@ -1065,15 +1112,11 @@ void P2PComm ::EventCbClientSeed([[gnu::unused]] struct bufferevent* bev,
...
+      struct timeval tv = {SEED_SYNC_LARGE_PULL_INTERVAL, 0};
+      bufferevent_set_timeouts(bev, &tv, NULL);
     }
-  }
-  if (events & (BEV_EVENT_EOF | BEV_EVENT_ERROR)) {
-    bufferevent_set_timeouts(bev, NULL, NULL);
-    CloseAndFreeBevP2PSeedConnClient(bev, ctx);
-  }
-  if (events & BEV_EVENT_TIMEOUT) {
+  } else if (events & (BEV_EVENT_EOF | BEV_EVENT_ERROR | BEV_EVENT_TIMEOUT)) {
```
sandipbhoir, line 1119: No need to clear the timeout anymore?

chetan-zilliqa: No.Actually, we are freeing bufferevent , so timeout would be automatically cleared.

sandipbhoir (approved):

PR #2167, opened 2020-08-11
Contract MPT storage and Remote State Fetching Support

kaikawaliu:
## Description

This PR contains these following points:
1. Support Remote State Fetching
2. Contract Merkle Partricia Tree storage
3. Data migrate
Since these two features are dependent to each others so submitting in one PR as a whole.
### Support Remote State Fetching
We mainly introduced:
- Constant `TYPE_INDICATOR` for all scilla data entry, this is introduced by parsing the result from Scilla Checker
- Constant `CONTRACT_ADDR_INDICATOR` for quick locating a remote contract.
- Constant `SCILLA_VERSION_INDICATOR` for version consistency when reading states from remote contract.
- Function `ContractStorage2::FetchExternalStateValue` algorithm for fetching external contract state value.
- Some refactoring to related functions
### Contract MPT storage
We mainly introduced:
- A series of new classes in `ContractStorage2Data.h` (e.g. `AddDeleteMap`, `RecordableAddDeleteMap` ...) for interacting with Trie library.
- Adopted new composing mechanism for contract storage root based on MPT.
- Revised most of contract storage related basic functions for supporting MPT storage.
### Data migrate
- Basically we will let every contracts existing in the previous mainnet to run through the Scilla checker again to make up the new parameters needed and deprecate the old ones.
 
## Backward Compatibility

- [ ] This is not a breaking change
- [x] This is a breaking change

### Implementation

- [x] **ready for review**

### Integration Test (Core Team)

- [ ] local machine test

- [ ] small-scale cloud test

PR #2452, opened 2021-02-19
[master] Remote State Reads

vaivaswatha:
## Description
This is a redo of #2167 without the MPT / Trie parts.

## Backward Compatibility

## Status
It is ready for review, but requires more testing. I might need @kaikawaliu's and @renlulu's  help for that.

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libPersistence/ContractStorage2.cpp
```diff
@@ -342,6 +364,58 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
...
 
+bool ContractStorage2::FetchExternalStateValue(
+    const dev::h160& caller, const dev::h160& target, const bytes& src,
+    unsigned int s_offset, bytes& dst, unsigned int d_offset, bool& foundVal,
+    string& type, uint32_t caller_version) {
+  // get caller version if not available
+  if (caller_version == std::numeric_limits<uint32_t>::max()) {
+    std::map<std::string, bytes> t_caller_version;
+    string version_key =
+        GenerateStorageKey(caller, SCILLA_VERSION_INDICATOR, {});
```
jjcnn, line 375: I'm surprised we need scilla_version here. It looks like it's used to store several version of the contract state - is that to make it easier to deal with future migrations?

ansnunez: I believe we based it on the remote state specs where, yes, the version is also included: https://github.com/Zilliqa/scilla-compiler/wiki/Interaction-with-Blockchain#fetching-states-of-other-contracts

jjcnn: "...a fail happens when
...
- The version of the external contract is not the same as the contract being executed."

That's not going to work. If someone deploys a contract that other contracts are supposed to interact with, then we can't expect the owner of that contract to update the contract (i.e., deploy a new one and migrate the contract state) when we update to a new Scilla version, nor can we expect the interacting contracts to stay at the earlier Scilla version (which we may choose to no longer support).

vaivaswatha: @jjcnn I recall writing this point because we wanted to disallow interaction b/w contracts of different versions. Shouldn't we have this anymore?

jjcnn: We have no restriction whatsoever on the versions of the sender and the recipient of messages, so I doubt we made that decision - we either made the opposite decision, or didn't decide on anything yet. 

In any case I really don't think it will work to restrict contract interaction like that. The result would be that contracts will exist in isolated silos based on their version number. If a token contract is written in version 0, then users of that token would forever be restricted to using Scilla version 0 to trade in those tokens, and that isn't a viable solution in the long-term.

jjcnn:
I've looked through the changes, and apart from the version issue it passes the smell test. I don't know enough about the blockchain layer to do a proper review, though.

src/libPersistence/ContractStorage2.cpp
vaivaswatha: In any case, I suggest we ignore this point as far as this PR is concerned and if we finalize on allowing cross-version remote state reads, I'll implement that as a separate PR. 

The reason I prefer to do that is: this PR is a re-do of earlier PR with some parts removed. It's confusing as it is to keep track. So I'd prefer it be as near to the original one as we practically can, and do further fixes as a separate PR.

jjcnn: That's fine - it doesn't make any difference as long as there's only one Scilla version in play.

vaivaswatha:
Ping !. I have some more changes pending that I need to do on top of this change after its merged. So it would be nice to have this merged at the earliest.

sandipbhoir (changes requested):

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -967,22 +970,36 @@ bool AccountStoreSC<MAP>::ParseContractCheckerOutput(
...
+            if (!hasMap && field["depth"].asInt() > 0) {
+              hasMap = true;
+            }
+            metadata.emplace(
+                Contract::ContractStorage2::GetContractStorage()
+                    .GenerateStorageKey(addr, TYPE_INDICATOR,
+                                        {field["vname"].asString()}),
+                DataConversion::StringToCharArray(field["type"].asString()));
+          } else {
+            LOG_GENERAL(WARNING, "Unexpected field detected" << checkerPrint);
```
ansnunez, line 993: Might be useful to also log the field that caused the error, and not just the entire checkerPrint

src/libPersistence/ContractStorage2.cpp
```diff
@@ -140,11 +142,31 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
...
+    FetchStateDataForKey(t_type, type_key, true);
+    if (t_type.empty()) {
+      LOG_GENERAL(WARNING, "Failed to fetch type for addr: "
+                               << addr.hex() << " vname: " << query.name());
+      return false;
+    }
+    try {
+      type = DataConversion::CharArrayToString(t_type[type_key]);
+    } catch (const std::exception& e) {
+      LOG_GENERAL(WARNING, "invalid type: " << type_key << endl << e.what());
```
ansnunez, line 166:
```suggestion
      LOG_GENERAL(WARNING, "Invalid type fetched for key=" << type_key << " for addr=" << addr.hex() << ": " << e.what());
```

src/libServer/ScillaIPCServer.cpp
```diff
@@ -81,6 +83,23 @@ bool ScillaIPCServer::fetchStateValue(const string &query, string &value,
   return true;
 }
 
+bool ScillaIPCServer::fetchExternalStateValue(const std::string &addr,
```
ansnunez, line 86: Do we need to bind this method also in `ScillaIPCServer::ScillaIPCServer`?

ansnunez (changes requested):

src/libData/AccountData/AccountStore.cpp
```diff
@@ -495,3 +493,126 @@ void AccountStore::RevertCommitTemp() {
...
+      return false;
+    }
+
+    if (!ExportCreateContractFiles(account, is_library, scilla_version,
+                                   extlibs_exports)) {
+      LOG_GENERAL(WARNING, "ExportCreateContractFiles failed");
+      return false;
+    }
+
+    // invoke scilla checker
```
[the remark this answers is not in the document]
kaikawaliu: Agreed

ansnunez (changes requested):

src/libPersistence/ContractStorage2.cpp
```diff
@@ -140,11 +142,31 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
...
 
+  if (getType) {
+    std::map<std::string, bytes> t_type;
+    std::string type_key =
+        GenerateStorageKey(addr, TYPE_INDICATOR, {query.name()});
+    FetchStateDataForKey(t_type, type_key, true);
+    if (t_type.empty()) {
+      LOG_GENERAL(WARNING, "Failed to fetch type for addr: "
+                               << addr.hex() << " vname: " << query.name());
+      return false;
```
ansnunez, line 161: Do we need to set `foundVal` to false?

src/libPersistence/ContractStorage2.cpp
```diff
@@ -140,11 +142,31 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
...
+    }
+    try {
+      type = DataConversion::CharArrayToString(t_type[type_key]);
+    } catch (const std::exception& e) {
+      LOG_GENERAL(WARNING, "invalid type: " << type_key << endl << e.what());
+    }
```
ansnunez, line 167: Should we set `foundVal` to false and also return false here?

src/libPersistence/ContractStorage2Data.h
```diff
@@ -0,0 +1,263 @@
+/*
+ * Copyright (C) 2020 Zilliqa
```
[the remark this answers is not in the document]
ansnunez: I think this whole file might be for MPT-related work? In that case, we can remove it for now.

tests/depends/Trie/Test_StateTrie.cpp
```diff
@@ -0,0 +1,264 @@
+/*
```
ansnunez, line 1: I think this whole file is only relevant to MPT-related work and can be removed

tests/depends/Trie/CMakeLists.txt
```diff
@@ -14,3 +14,8 @@ add_executable(Test_TriePerformance Test_TriePerformance.cpp)
 target_include_directories(Test_TriePerformance PUBLIC ${CMAKE_SOURCE_DIR}/src)
 target_link_libraries(Test_TriePerformance PUBLIC Trie Utils Constants)
 add_test(NAME Test_TriePerformance COMMAND Test_TriePerformance)
+
```
ansnunez, line 17: We can remove instead of comment, since this is for MPT-related work

tests/Data/Test_Contract.cpp
```diff
@@ -2051,4 +2051,171 @@ BOOST_AUTO_TEST_CASE(testCreateContractJsonOutput) {
   }
 }
 
+BOOST_AUTO_TEST_CASE(simplemap) {
```
ansnunez, line 2054: This test is for outdated MPT-related work and can be removed

tests/Data/AccountData/Test_AccountStore.cpp
```diff
@@ -681,74 +681,77 @@ void CheckRFContract(
                       "ContrBalance doesn't match");
 }
 
-BOOST_AUTO_TEST_CASE(serializeAndDeserialize) {
-  INIT_STDOUT_LOGGER();
+// BOOST_AUTO_TEST_CASE(serializeAndDeserialize) {
```
ansnunez, line 684: Let's try reverting this change also

ansnunez (changes requested):

src/libData/AccountData/AccountStoreBase.tpp
```diff
@@ -181,10 +181,10 @@ bool AccountStoreBase<MAP>::IsAccountExist(const Address& address) {
 
 template <class MAP>
 bool AccountStoreBase<MAP>::AddAccount(const Address& address,
-                                       const Account& account) {
+                                       const Account& account, bool toReplace) {
   // LOG_MARKER();
 
-  if (!IsAccountExist(address)) {
+  if (toReplace || !IsAccountExist(address)) {
```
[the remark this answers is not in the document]
vaivaswatha: `insert_or_assign` is C++17 onwards. Shall I just do `m_addressToAccount[address] = account;`?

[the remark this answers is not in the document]
ansnunez: Yep sure. `toReplace` only applies to migration, so there'll be no inefficiency considerations afterwards.

src/libPersistence/ContractStorage2.cpp
vaivaswatha: If the function returns `false`, that's an error, `foundVal` will just be ignored after that. So no.

src/libPersistence/ContractStorage2.cpp
vaivaswatha: Same as above.

src/libData/AccountData/AccountStoreSC.h
```diff
@@ -46,10 +46,10 @@ class AccountStoreAtomic
   GetAddressToAccount();
 };
 
+enum INVOKE_TYPE { CHECKER, RUNNER_CREATE, RUNNER_CALL };
```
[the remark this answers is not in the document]
vaivaswatha: I tried this, but turns out, uses such as this: https://github.com/Zilliqa/Zilliqa/home/vaivaswatha/d/Zilliqa/src/libData/AccountData/AccountStore.cpp#L567 won't work because it becomes private. I can either redo this change (i.e., move it out of the class) or make it public I guess.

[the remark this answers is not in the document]
ansnunez: Oh okay, I did not notice the migration needed it. You can ignore this comment then.

sandipbhoir (approved):

ansnunez (approved):

PR #2453, opened 2021-02-21
[release-7.2] Ignore pow submission which are received late

sandipbhoir:
## Description
This PR addresses issue : https://github.com/Zilliqa/Issues/issues/706

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/PoWProcessing.cpp
+++ b/src/libDirectoryService/PoWProcessing.cpp
@@ -151,6 +151,12 @@ bool DirectoryService::ProcessPoWSubmission(
     return true;
   }
 
+  if (m_powSubmissionWindowExpired) {
+    LOG_GENERAL(INFO, "Submission recvd too late from "
+                          << from.GetPrintableIPAddress());
+    return true;
+  }
+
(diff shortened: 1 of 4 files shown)


src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -151,6 +151,12 @@ bool DirectoryService::ProcessPoWSubmission(
     return true;
   }
 
+  if (m_powSubmissionWindowExpired) {
```
ansnunez, line 154: Should we put this within UPDATE_TARGET_DS_NUM?

sandipbhoir: We don’t need right. We can deploy one node ( non-leader) at a time.

ansnunez: ok, agree

ansnunez (approved):

bb111189 (approved):

PR #2459, opened 2021-02-24
[master] Add bech32 utilities and initial migration to bech32 address

sunny-teo9000:
## Description
To add Bech32 utilities and allow the following API to accept both base16 and bech32 for now.

GetBalance
GetSmartContractCode
GetSmartContractInit
GetSmartContracts
GetSmartContractState
GetSmartContractSubState

## Backward Compatibility

### Implementation

- [x] **ready for review**

Was tested with a local isolated server, testing with the bech32 and base16 address

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -30,6 +30,7 @@
 #include "libNetwork/Peer.h"
 #include "libPersistence/BlockStorage.h"
 #include "libRemoteStorageDB/RemoteStorageDB.h"
+#include "libUtils/AddressConversion.h"
 #include "libUtils/DetachedFunction.h"
 #include "libUtils/Logger.h"
 #include "libUtils/TimeUtils.h"
@@ -40,9 +41,32 @@ using namespace std;
 CircularArray<std::string> LookupServer::m_RecentTransactions;
 std::mutex LookupServer::m_mutexRecentTxns;
 
+namespace {
 const unsigned int PAGE_SIZE = 10;
 const unsigned int NUM_PAGES_CACHE = 2;
 const unsigned int TXN_PAGE_SIZE = 100;
+
+Address ToBase16AddrHelper(const std::string& addr) {
+  using RpcEC = ServerBase::RPCErrorCode;
+
+  Address convertedAddr;
+  auto retCode = ToBase16Addr(addr, convertedAddr);
+
+  if (retCode == AddressConversionCode::INVALID_ADDR) {
+    throw JsonRpcException(RpcEC::RPC_INVALID_ADDRESS_OR_KEY,
+                           "invalid address");
+  } else if (retCode == AddressConversionCode::INVALID_BECH32_ADDR) {
+    throw JsonRpcException(RpcEC::RPC_INVALID_ADDRESS_OR_KEY,
+                           "Bech32 address is invalid");
+  } else if (retCode == AddressConversionCode::WRONG_ADDR_SIZE) {
+    throw JsonRpcException(RpcEC::RPC_INVALID_PARAMETER,
+                           "Address size not appropriate");
+  }
+
+  return convertedAddr;
+}
+
+}  // namespace
 
 //[warning] do not make this constant too big as it loops over blockchain
 const unsigned int REF_BLOCK_DIFF = 1;
@@ -800,17 +824,7 @@ Json::Value LookupServer::GetBalance(const string& address) {
   }
 
   try {
-    if (address.size() != ACC_ADDR_SIZE * 2) {
-      throw JsonRpcException(RPC_INVALID_PARAMETER,
-                             "Address size not appropriate");
-    }
-
-    bytes tmpaddr;
-    if (!DataConversion::HexStrToUint8Vec(address, tmpaddr)) {
-      throw JsonRpcException(RPC_INVALID_ADDRESS_OR_KEY, "invalid address");
-    }
-    Address addr(tmpaddr);
-
+    Address addr{ToBase16AddrHelper(address)};
     shared_lock<shared_timed_mutex> lock(
         AccountStore::GetInstance().GetPrimaryMutex());
 
@@ -853,18 +867,7 @@ Json::Value LookupServer::GetSmartContractState(const string& address,
   }
 
   try {
-    Json::Value _json;
-    if (address.size() != ACC_ADDR_SIZE * 2) {
-      throw JsonRpcException(RPC_INVALID_PARAMETER,
-                             "Address size not appropriate");
-    }
-    bytes tmpaddr;
-    if (!DataConversion::HexStrToUint8Vec(address, tmpaddr)) {
-      throw JsonRpcException(RPC_INVALID_ADDRESS_OR_KEY, "invalid address");
-    }
-
-    Address addr(tmpaddr);
-
+    Address addr{ToBase16AddrHelper(address)};
     shared_lock<shared_timed_mutex> lock(
         AccountStore::GetInstance().GetPrimaryMutex());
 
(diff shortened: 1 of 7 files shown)


src/libUtils/Bech32.cpp
```diff
@@ -0,0 +1,342 @@
...
+}
+
+bool Bech32::isZilBase16(const std::string& input) {
+  return std::regex_match(input, BASE16_ZIL_REGEX_FORMAT);
+}
+
+bool Bech32::HasZilHrp(const std::string& input) {
+  if (input.length() < 4) return false;
+
+  auto hrp = boost::string_view(input);
```
KaustubhShamshery, line 195: const auto& hrp

src/libUtils/Bech32.cpp
```diff
@@ -0,0 +1,342 @@
...
+  bool lower = false;
+  bool upper = false;
+
+  size_t inputSize = input.length();
+  if (inputSize < 8 || inputSize > 90) {
+    LOG_GENERAL(WARNING, "Cannot decode bech32 with invalid size: " << input);
+    return "";
+  }
+
+  for (size_t i = 0; i < inputSize; ++i) {
```
KaustubhShamshery, line 210: modernize loop

src/libUtils/Bech32.cpp
```diff
@@ -0,0 +1,342 @@
...
+    char c = input[i];
```
KaustubhShamshery, line 211: can use const auto& here too

src/libUtils/Bech32.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+#include <algorithm>
+#include <boost/algorithm/string.hpp>
+#include <boost/utility/string_view.hpp>
+#include <iostream>
+#include <regex>
+
+#include "Bech32.h"
+#include "libServer/AddressChecksum.h"
+
```
KaustubhShamshery, line 26: can use using namespace std in .cpp file

src/libUtils/Bech32.cpp
```diff
@@ -0,0 +1,342 @@
...
+const std::string Bech32::ZIL_HRP = "zil";
+
+const char Bech32::HEX_ARRAY[16] = {'0', '1', '2', '3', '4', '5', '6', '7',
+                                    '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'};
+
+const std::regex Bech32::BECH32_ZIL_REGEX_FORMAT(
+    "^zil1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{38}");
+const std::regex Bech32::BASE16_ZIL_REGEX_FORMAT("^(0x)?[0-9a-fA-F]{40}");
+
+int Bech32::hexCharToInt(char hexChar) {
```
ansnunez, line 47: Using streams to convert individual characters might be inefficient. Something like this should suffice:
```
int i = -1;
if (hexChar >= '0' && hexChar <= '9') {
  i = hexChar - '0';
} else if (hexChar >= 'a' && hexChar <= 'f') {
  i = 10 + hexChar - 'a';
}
return i;
```

ansnunez: Additionally, if you also check for `(hexChar >= 'A' && hexChar <= 'F')` then `hexStringToCharVec` won't need to call `boost::to_lower_copy` anymore

tests/Utils/Test_Bech32.cpp
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright (C) 2019 Zilliqa
```
ansnunez, line 2:
```suggestion
 * Copyright (C) 2021 Zilliqa
```

tests/Utils/Test_Bech32.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  for (auto& pair : testPairs) {
+    bool result = boost::iequals(
+        Bech32::EncodeZilFromBase16ToBase32(pair.first), pair.second);
+    BOOST_CHECK_MESSAGE(result, "Wrong encode result for " + pair.first +
+                                    ", expecting: " + pair.second);
+  }
+}
+
+BOOST_AUTO_TEST_SUITE_END()
```
ansnunez, line 143: Can you also add some tests for negative/failure cases?

ansnunez (changes requested):

ansnunez (approved):

bb111189:
https://github.com/Zilliqa/cryptoutils/tree/main/lib

Can we use bech32 from cryptoutils? This is to avoid 2 copy of bech32. @vaivaswatha  has recently created it and will be using for his VM. I think on the core side, ethash is already using cryptoutils

vaivaswatha:
> https://github.com/Zilliqa/cryptoutils/tree/main/lib
> 
> Can we use bech32 from cryptoutils? This is to avoid 2 copy of bech32. @vaivaswatha has recently created it and will be using for his VM. I think on the core side, ethash is already using cryptoutils

Example usage: https://github.com/Zilliqa/scilla-vm/blob/4c58dae89aa3bc1567f21ad1ad9aa952ee089142/libsrtl/ScillaBuiltins.cpp#L608. This function and the next one do decoding and encoding respectively.

src/libServer/LookupServer.cpp
```diff
@@ -40,9 +43,59 @@ using namespace std;
 CircularArray<std::string> LookupServer::m_RecentTransactions;
 std::mutex LookupServer::m_mutexRecentTxns;
 
+namespace {
 const unsigned int PAGE_SIZE = 10;
 const unsigned int NUM_PAGES_CACHE = 2;
 const unsigned int TXN_PAGE_SIZE = 100;
+const unsigned UINT8_ADDR_SIZE = ACC_ADDR_SIZE * 2;
+
+bool HasZilHrp(const string& input) {
```
ansnunez, line 52: I think this could actually be reduced to just
```
inline bool HasZilHrp(const string& input) {
   return input.substr(0, 4) == string("zil1");
}
```
If above is used, then the boost hpps can also be un-included

src/libServer/LookupServer.cpp
```diff
@@ -40,9 +43,59 @@ using namespace std;
 CircularArray<std::string> LookupServer::m_RecentTransactions;
 std::mutex LookupServer::m_mutexRecentTxns;
 
+namespace {
 const unsigned int PAGE_SIZE = 10;
 const unsigned int NUM_PAGES_CACHE = 2;
 const unsigned int TXN_PAGE_SIZE = 100;
+const unsigned UINT8_ADDR_SIZE = ACC_ADDR_SIZE * 2;
```
ansnunez, line 50:
```suggestion
const unsigned int UINT8_ADDR_SIZE = ACC_ADDR_SIZE * 2;
```

ansnunez (changes requested):

src/libServer/LookupServer.cpp
sunny-teo9000: the boost string_view is actually a read-only wrapper to allow substring without constructing a new string.
But I guess since this is a short string, it should be ok. Will reduce this. :)

src/libServer/LookupServer.cpp
```diff
@@ -40,9 +41,55 @@ using namespace std;
 CircularArray<std::string> LookupServer::m_RecentTransactions;
 std::mutex LookupServer::m_mutexRecentTxns;
 
+namespace {
 const unsigned int PAGE_SIZE = 10;
 const unsigned int NUM_PAGES_CACHE = 2;
 const unsigned int TXN_PAGE_SIZE = 100;
+const unsigned int UINT8_ADDR_SIZE = ACC_ADDR_SIZE * 2;
```
KaustubhShamshery, line 48: should be HEX_ADDR_SIZE

KaustubhShamshery:
Can we have the unit test again?

ansnunez (approved):

KaustubhShamshery (approved):

PR #2461, opened 2021-02-26
[master] Updates to external state fetch in ContractStorage2 

vaivaswatha:
## Description
1. `query.mapdepth` may be set to -1 by Scilla (in case of external states, Scilla cannot know this value). The blockchain needs to fetch the value from its database and use it for the lookup.
2. `query.ignoreval` may be set for nonmap fields too. In such a case, the blockchain only needs to return the type of the field and not the value.
3. Field not found should no more be an error. `foundVal` needs to be set to `false` and that's it. This will enable Scilla to check for presence of a field in a contract without it becoming an error.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -449,6 +449,8 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       // reset the storageroot update buffer atomic per transaction
       m_storageRootUpdateBufferAtomic.clear();
 
+      m_originAddr = fromAddr;
+
       Account* fromAccount = this->GetAccount(fromAddr);
       if (fromAccount == nullptr) {
         LOG_GENERAL(WARNING, "Sender has no balance, reject");
@@ -879,6 +881,7 @@ bool AccountStoreSC<MAP>::ExportCallContractFiles(
     msgObj["_sender"] =
         prepend +
         Account::GetAddressFromPublicKey(transaction.GetSenderPubKey()).hex();
+    msgObj["_origin"] = prepend + m_originAddr.hex();
     msgObj["_amount"] = transaction.GetAmount().convert_to<std::string>();
 
     JSONUtils::GetInstance().writeJsontoFile(INPUT_MESSAGE_JSON, msgObj);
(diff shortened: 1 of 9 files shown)


src/libPersistence/ContractStorage2.cpp
```diff
@@ -415,9 +423,28 @@ bool ContractStorage2::FetchExternalStateValue(
...
+
+  ProtoScillaQuery query;
+  query.ParseFromArray(src.data() + s_offset, src.size() - s_offset);
+
+  map<string, bytes> map_depth;
+  string map_depth_key =
+      GenerateStorageKey(target, MAP_DEPTH_INDICATOR, {query.name()});
+  FetchStateDataForKey(map_depth, map_depth_key, true);
+
+  auto map_depth_val = !map_depth.empty()
```
ansnunez, line 440: Might want to put inside try-catch

ansnunez (changes requested):

src/libPersistence/ContractStorage2.cpp
```diff
@@ -160,15 +168,19 @@ bool ContractStorage2::FetchStateValue(const dev::h160& addr, const bytes& src,
...
     try {
       type = DataConversion::CharArrayToString(t_type[type_key]);
     } catch (const std::exception& e) {
       LOG_GENERAL(WARNING, "Invalid type fetched for key="
                                << type_key << " for addr=" << addr.hex() << ": "
                                << e.what());
+      return false;
     }
+    // If not interested in the value, exit early.
+    if (query.indices().empty() && query.ignoreval()) return true;
```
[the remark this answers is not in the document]
vaivaswatha: `query.indices().empty()` implies that we are only looking at a state variable (field) as a whole, not indexing into a map (if the state variable is a map).

In case the indices list is not empty, that means we're checking to see if a (nested) map key exists (and depending on whether `ignoreval` is set then, the value for that key may be needed too).

In this case, if the indices list is empty, we only care whether the field itself exists (and its type) or not.

ansnunez (approved):

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -971,8 +974,14 @@ bool AccountStoreSC<MAP>::ParseContractCheckerOutput(
 
       bool hasMap = false;
 
-      if (root["contract_info"].isMember("fields")) {
+      if (root["contract_info"].isMember("fields") ||
+          root["contract_info"].isMember("params")) {
+        // We include contract parameters along with the mutable state.
+        auto stateVars = root["contract_info"]["params"];
```
ansnunez, line 980: Because of the `or` check in previous line, wouldn't we need to make sure `params` is also a member here before we access it?

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -971,8 +974,14 @@ bool AccountStoreSC<MAP>::ParseContractCheckerOutput(
...
         for (const auto& field : root["contract_info"]["fields"]) {
```
ansnunez, line 981: Because of the `or` check in previous line, wouldn't we need to make sure `fields` is also a member here before we access it?

ansnunez (changes requested):

src/libData/AccountData/AccountStoreSC.tpp
vaivaswatha: Fixed now.

ansnunez (approved):

sandipbhoir (approved):

PR #2475, opened 2021-03-18
[master] Fix timing issue with GetVCFinalBlock

sandipbhoir:
## Description

This fixes the timing issue with syncing of `extseed` which uses pubkey whitelisting.

 - Basically flag `m_vcFinalBlockProcessed` can remains unset in some corner case with timing issue in `GetVCFinalBlockFromL2lDataProvider`. 
- Same also applies for `GetDSBlockFromL2lDataProvider` 

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -849,11 +849,11 @@ bool Lookup::GetDSBlockFromL2lDataProvider(uint64_t blockNum) {
                   "GetDSBlockFromL2lDataProvider Timeout... may be ds block "
                   "yet to be mined");
     } else {
-      m_mediator.m_lookup->m_vcDsBlockProcessed = false;
-      return true;
+      break;
     }
   }
-  return false;
+  m_mediator.m_lookup->m_vcDsBlockProcessed = false;
+  return true;
 }
 
 bool Lookup::GetVCFinalBlockFromL2lDataProvider(uint64_t blockNum) {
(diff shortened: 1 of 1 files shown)


ansnunez (approved):

chetan-zilliqa:
Hi @sandipbhoir , One question about the fix, Maybe I'm missing somewhere.
Imagine the flow when the block is received and timeout happens.
1) Initially received is false and we go into a while loop.
2) Both timeouts happened and the final block is received. so `m_vcFinalBlockProcessed= true` now.
3) next time it will not go into a loop and we set `m_vcFinalBlockProcessed=false` and return received which is set to `false`.
4) But since we returned `false` from `GetVCFinalBlockFromL2lDataProvider` function, it will not go to fetch `FetchMbTxPendingTxMessageFromL2l` ?

sandipbhoir:
> Hi @sandipbhoir , One question about the fix, Maybe I'm missing somewhere.
> Imagine the flow when the block is received and timeout happens.
> 
>     1. Initially received flag is set to `false` and we go into a while loop.
> 
>     2. Both timeouts happened and the final block is received. so `m_vcFinalBlockProcessed= true` now.
> 
>     3. next time it will not go into a loop and we set `m_vcFinalBlockProcessed=false` and return received which is set to `false`.
> 
>     4. But since we returned `false` from `GetVCFinalBlockFromL2lDataProvider` function, it will not go to fetch `FetchMbTxPendingTxMessageFromL2l` ?

Good valid point.

chetan-zilliqa (approved):

PR #2478, opened 2021-03-19
[master] Remove pending transaction pool

ansnunez:
## Description

This PR removes the pending transaction pool, which has already been replaced by the off-chain transaction status feature.

Changes are mostly on lookup side, involving removal of `m_pendingTxns` and `m_droppedTxns`, which lookup and seed nodes usually populate when receiving `PENDINGTXN` message from shards. `m_pendingTxnStore` - used by seed nodes for seed-to-seed syncing - has also been removed.

Shard side will continue to send `PENDINGTXNS` message in order for lookup nodes to populate the off-chain database. For this reason, shard side will also continue to use `m_unconfirmedTxns`.

`GetPendingTxn` and `GetPendingTxns` API are also removed.

For StatusServer, `IsTxnInMempool` continues to be available, but only for shard side.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libServer/LookupServer.h
```diff
@@ -289,8 +280,6 @@ class LookupServer : public Server,
 
   // gets the number of transaction starting from block blockNum to most recent
   // block
-  Json::Value GetPendingTxn(const std::string& tranID);
```
ansnunez, line 292: Todo: Reroute to `GetTransactionStatus`

KaustubhShamshery (approved):

src/libServer/LookupServer.h
```diff
@@ -289,8 +280,6 @@ class LookupServer : public Server,
...
-  Json::Value GetPendingTxns();
```
ansnunez, line 293: Todo: Reroute to remote DB query in separate PR

src/libLookup/Lookup.cpp
```diff
@@ -5986,12 +5883,6 @@ void Lookup::FetchMbTxPendingTxMessageFromL2l(uint64_t blockNum) {
           }
         }
 
-        // for each shard, send request for pending txn message to l2l data
```
sandipbhoir, line 5989: Can rename `FetchMbTxPendingTxMessageFromL2l` to `FetchMBnForwardTxMessageFromL2l`?

sandipbhoir (changes requested):

sandipbhoir (approved):

PR #2480, opened 2021-03-22
[master] P2pseed server side socket cleanup

chetan-zilliqa:
## Description

Add a timeout of 20 seconds on the server-side for clearing socket connections for the two outstanding scenarios - 2 & 3 mentioned in the slack discussion [https://zilliqa-team.slack.com/archives/GRFEL9S1H/p1613455520095000](url)

while coding for a solution as discussed in slack chat [https://zilliqa-team.slack.com/archives/GRFEL9S1H/p1615803397080500
](URL) , I have realized that `bufferevent_set_timeouts()` gets called even after the first successful read operation.

As per the latest documentation, For the below scenarios, the timeout events would be called.
1) After the first successful read operation, timeout seconds elapse and timeout event is called.
2) If there's no read callback is called, then timeout seconds elapses and timeout event is called.

http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/bufferevent_8h.html#a3686d744e67e9a530ebf8ab888ade2cf

So instead of going with the above solution, we can set a timeout on the buffer event when `AcceptCbServerSeed` is called on the server-side(For eg: 20 secs). so both 2 & 3 scenarios are covered.
Scenario -2) If there's no activity on the socket on the server-side, `ReadCbServerSeed` won't be called and timeout would be fired after 20 secs and the connection and resources(buffer event) would be released.

Scenarios-3) If a successful read callback `ReadCbServerSeed` is called on the server-side and the client doesn't close the connection, the timeout will be called after 20 secs on the server-side after the read and the connection would be closed. 
Ideally, the client will close the connection immediately after receiving the response or if the client timeout of 10 secs expires without a response. The client will send an event callback to the server to close the connection and release the resources like bufferevent. If a malicious client doesn't do so then the server timeout of 20 secs will expire and the connection would be close on both sides.

## Test Scenarios.

1) Do telnet from the local machine. The connection would be closed after 20 secs
`telnet 54.203.153.154 33135`

2) Comment out `bufferevent_set_timeouts()`- line 1295 and `CloseAndFreeBevP2PSeedConnClient()` - line1235 from the client code and run a seed node. We will see sockets would be cleared on both sides after 20 secs.

3) Send garbage data on the socket - Connection would be closed immediately.
`echo 010003440000008B04240A1E084710021A180A120A100000000000000000|netcat 54.203.153.154 33135`

4) Test with the following settings:
Client side = SEED_SYNC_LARGE_PULL_INTERVAL = 10 , Server side = P2P_SEED_SERVER_CONNECTION_TIMEOUT =20
Client side = SEED_SYNC_LARGE_PULL_INTERVAL = 20 , Server side = P2P_SEED_SERVER_CONNECTION_TIMEOUT =20

5) Delete seedpubs pod.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


src/libNetwork/P2PComm.cpp
```diff
@@ -1288,6 +1289,10 @@ void P2PComm::AcceptCbServerSeed([[gnu::unused]] evconnlistener* listener,
   bufferevent_setwatermark(bev, EV_READ, MIN_READ_WATERMARK_IN_BYTES,
                            MAX_READ_WATERMARK_IN_BYTES);
   bufferevent_setcb(bev, ReadCbServerSeed, NULL, EventCbServerSeed, NULL);
+  struct timeval tv = {P2P_SEED_SERVER_CONNECTION_TIMEOUT, 0};
+  LOG_GENERAL(INFO, "Timeout Added =" << P2P_SEED_SERVER_CONNECTION_TIMEOUT
```
chetan-zilliqa, line 1293: Will remove this log before the final merge.

ansnunez (approved):

sandipbhoir (approved):

PR #2483, opened 2021-03-24
[master] Adding DNS query function for seed nodes

sunny-teo9000:
## Description
To move upper_seeds, l2l_dataprovider and multiplier IPs from constants.xml to getting fetched from DNS

The constants.xml IP still stays, in event where dns is unable to be accessed, we will fall back to the static IPs in there.

**DNS cache**
- This is maintained on a separate thread, to be updated on every DS epoch.
- The `AttemptPopulateLookupsDnsCache` is called at every end of a DS epoch.
- It is separate so that it would not be blocking our nodes function while querying.
- A retry constant `QUERY_DNS_MAX_TRIES` can be use to query again if DNS query failed.

**Local list**
- These are all in the lookup's class. `m_lookupNodes`, `m_seedNodes`, `m_l2lDataProviders`
- These will be updated via looking into the **DNS cache** at every end of DS epoch.
- Because of the DNS querying delay, when the local list look at **DNS cache**, it is looking at the data from last DS epoch.
- This is separate from the **DNS cache** because we want to use it without the delay of blocking calls from DNS querying.
- **DNS cache** will be updated after **Local list** populate from it.

**constants.xml list**
- In the event when **Local list** is empty, this will be used to populate the **Local list**
- This is usually used at the startup, as the DNS needs time to start up, querying to DNS will fail at this point.

## Backward Compatibility

## Review Suggestion

Scaling can be done by by
`kubectl scale sts <testnetname>-<nodetype> --replicas=2`

Tests done
- Scaling up and down seedpubs and multipliers to see if other nodes can obtain the ip list
- Check if community seeds and miner nodes can obtain the ip list from dns
- Simulate scenario where DNS is down by setting /etc/resolv.conf namesevers ip to random IP to block dns query.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -95,7 +95,7 @@
         <TX_DISTRIBUTE_TIME_IN_MS>30000</TX_DISTRIBUTE_TIME_IN_MS>
         <EXTRA_TX_DISTRIBUTE_TIME_IN_MS>15000</EXTRA_TX_DISTRIBUTE_TIME_IN_MS>
         <DS_TX_PROCESSING_TIMEOUT>55</DS_TX_PROCESSING_TIMEOUT>
-        <NEW_LOOKUP_SYNC_DELAY_IN_SECONDS>300</NEW_LOOKUP_SYNC_DELAY_IN_SECONDS>
+        <NEW_LOOKUP_SYNC_DELAY_IN_SECONDS>10</NEW_LOOKUP_SYNC_DELAY_IN_SECONDS>
         <GETSHARD_TIMEOUT_IN_SECONDS>3</GETSHARD_TIMEOUT_IN_SECONDS>
         <GETSTATEDELTAS_TIMEOUT_IN_SECONDS>5</GETSTATEDELTAS_TIMEOUT_IN_SECONDS>
         <GETCOSIGREWARDS_TIMEOUT_IN_SECONDS>5</GETCOSIGREWARDS_TIMEOUT_IN_SECONDS>
@@ -398,6 +398,14 @@
             <wallet_address>ee06b3c906612cc5bdb087a30e6093c9f0aa04fd</wallet_address>
         </account>
     </ds_accounts>
+    <node_discovery>
+        <QUERY_DNS_FOR_SEED>true</QUERY_DNS_FOR_SEED>
+        <QUERY_DNS_MAX_TRIES>5</QUERY_DNS_MAX_TRIES>
+        <l2l_data_providers_dns/>
+        <upper_seed_dns/>
+        <multiplier_dns/>
+        <default_seed_port>33133</default_seed_port>
+    </node_discovery>
     <!-- The remaining constants below are auto-populated -->
     <lookups>
         <!--IP to be provided after public testnet launch.
--- a/src/libDirectoryService/Coinbase.cpp
+++ b/src/libDirectoryService/Coinbase.cpp
@@ -161,8 +161,21 @@ void DirectoryService::InitCoinbase() {
       ++it;
   }
 
-  const auto& vecLookup = m_mediator.m_lookup->GetLookupNodesStatic();
+  auto vecLookup = m_mediator.m_lookup->GetLookupNodesStatic();
+
+  // Sorting is done here, as different sequences may occur from obtaining the
+  // IP through the DNS Different sequences will affect the coinbase
+  // calculation, therefore we need to sort here
+  std::sort(vecLookup.begin(), vecLookup.end(),
+            [](const auto& node1, const auto& node2) {
+              return node1.first < node2.first;
+            });
+
   const auto& epochNum = m_mediator.m_currentEpochNum;
+
+  // In VC case, if this wasn't cleared, it will accumulate the same epochNum
+  // previous entries
+  m_coinbaseRewardees[epochNum][CoinbaseReward::LOOKUP_REWARD].clear();
 
   for (const auto& lookupNode : vecLookup) {
     m_coinbaseRewardees[epochNum][CoinbaseReward::LOOKUP_REWARD].push_back(
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -74,11 +74,13 @@ Lookup::Lookup(Mediator& mediator, SyncType syncType, bool multiplierSyncMode,
       ignorable_syncTypes.end()) {
     m_syncType = syncType;
   }
-  SetLookupNodes();
-  SetAboveLayer(m_seedNodes, "node.upper_seed");
-  if (!MULTIPLIER_SYNC_MODE) {
-    SetAboveLayer(m_l2lDataProviders, "node.l2l_data_providers");
-  }
+
+  DNSUtils::InitDNSCacheList();
+
+  UpdateAllSeedsAndMultipliers();
+
+  SetGenesisWallets();
+
   if (LOOKUP_NODE_MODE) {
     SetDSCommitteInfo();
   }
@@ -90,7 +92,7 @@ Lookup::Lookup(Mediator& mediator, SyncType syncType, bool multiplierSyncMode,
   }
 }
 
-Lookup::~Lookup() {}
+Lookup::~Lookup() { DNSUtils::ShutDownDNSCacheList(); }
 
 void Lookup::GetInitialBlocksAndShardingStructure() {
   LOG_MARKER();
@@ -132,6 +134,20 @@ void Lookup::GetInitialBlocksAndShardingStructure() {
   }
 }
 
+void Lookup::UpdateAllSeedsAndMultipliers() {
+  LOG_MARKER();
+
+  SetLookupNodes();
+  SetUpperSeedsInner();
+  if (!MULTIPLIER_SYNC_MODE) {
+    SetL2lDataProvidersInner();
+  }
+
+  if (QUERY_DNS_FOR_SEED) {
+    DNSUtils::AttemptPopulateLookupsDNSCache();
+  }
+}
+
 void Lookup::InitSync() {
   LOG_MARKER();
   auto func = [this]() -> void {
@@ -139,14 +155,12 @@ void Lookup::InitSync() {
     // and register me with multiplier.
     this_thread::sleep_for(chrono::seconds(NEW_LOOKUP_SYNC_DELAY_IN_SECONDS));
 
-    if (m_seedNodes.empty()) {
-      SetAboveLayer(
-          m_seedNodes,
-          "node.upper_seed");  // since may have called CleanVariable earlier
+    if (IsUpperSeedEmpty()) {
+      SetUpperSeedsInner();  // since may have called CleanVariable earlier
     }
 
-    if (!MULTIPLIER_SYNC_MODE && m_l2lDataProviders.empty()) {
-      SetAboveLayer(m_l2lDataProviders, "node.l2l_data_providers");
+    if (!MULTIPLIER_SYNC_MODE && IsL2lDataProvidersEmpty()) {
+      SetL2lDataProvidersInner();
     }
 
     // Send whitelist request to seeds, in case it was blacklisted if was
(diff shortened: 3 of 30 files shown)


src/libNetwork/DataSender.cpp
```diff
@@ -21,40 +21,34 @@
...
-    if (!url.empty()) {
-      uint128_t tmpIp;
-      if (IPConverter::ResolveDNS(url, node.second.GetListenPortHost(),
-                                  tmpIp)) {
-        resolved_ip = tmpIp;  // resolved one
-      } else {
-        LOG_GENERAL(WARNING, "Unable to resolve DNS for " << url);
-      }
-    }
+  ObtainIpListFromDns(ipList, LOOKUPS_URL);
```
ansnunez, line 40: I think we want to avoid obtaining IP list every time we have to send a message. Better to still retain local list and only obtain from DNS server at pre-determined intervals (e.g., new DS epoch)

Krishnakumar0033:
ladida

src/libLookup/Lookup.cpp
```diff
@@ -1252,88 +1323,201 @@ bool Lookup::ProcessGetDSInfoFromSeed(const bytes& message, unsigned int offset,
...
+      int index = RandomGenerator::GetRandomInt(ipFromDns.size());
+      auto ip = ipFromDns[index];
+      auto ipStr = IPConverter::ToStrFromNumericalIP(ip);
+
+      randomPeer = Peer{ip, port};
+
+      // Need to insert into our local list for fetching verification
+      bytes pubkeyBytes;
+      if (!ObtainPubKeyFromUrl(pubkeyBytes, ipStr, L2L_DATA_PROVIDERS_DNS)) {
+        LOG_GENERAL(WARNING, "Pubkey cannot be found!");
```
chetan-zilliqa, line 1377: If the public key is not found for one node because of an error, should we should still try for other nodes(if the key is found) so that `localL2lDataProvider` is not empty

chetan-zilliqa (changes requested):

src/libLookup/Lookup.cpp
```diff
@@ -132,21 +134,29 @@ void Lookup::GetInitialBlocksAndShardingStructure() {
   }
 }
 
+void Lookup::UpdateAllSeedsAndMultipliers() {
+  LOG_MARKER();
+
+  SetLookupNodes();
```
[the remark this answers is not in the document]
sandipbhoir: lookup does not support scalable feature for reason line coinbase reward.
And we don't forsee any need to scale lookup since they are not involved in syncing or api service

src/libLookup/Lookup.cpp
```diff
@@ -167,65 +177,148 @@ void Lookup::SetLookupNodes(const VectorOfNode& lookupNodes) {
...
+  UpdateNodes(m_seedNodes, DEFAULT_SEED_PORT, UPPER_SEED_DNS,
+              "node.upper_seed");
+}
+
+void Lookup::SetL2lDataProvidersInner() {
+  LOG_MARKER();
+  auto port = ENABLE_SEED_TO_SEED_COMMUNICATION ? P2P_SEED_CONNECT_PORT
+                                                : DEFAULT_SEED_PORT;
+  lock_guard<shared_timed_mutex> lock(m_mutexL2lDataProviders);
+  UpdateNodes(m_l2lDataProviders, port, L2L_DATA_PROVIDERS_DNS,
```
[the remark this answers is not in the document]
sunny-teo9000: `TryGettingResolvedIP` will check for this

If hostname is filled, it was from config, we will use it to get the resolve ip.

Else, empty hostname will be ignored, we will use the latest ip that was obtain from the dns previously

Both of these cases are when current DNS are not accessible.

src/libLookup/Lookup.cpp
sunny-teo9000: Got it, will add this.

Revolution1 (approved):
Tested OK

chetan-zilliqa (changes requested):

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,348 @@
...
+  }
+}
+
+string GetPubKeyUrl(const std::string &ip, const std::string &url) {
+  auto tmpIp = ip;
+  std::replace(tmpIp.begin(), tmpIp.end(), '.', '-');
+  return "pub-" + tmpIp + url.substr(url.find_first_of('.'));
+}
+
+uint128_t ConvertIpStringToUint128(const string &ipStr) {
```
chetan-zilliqa, line 303: We already have a utility function for this.
`https://github.com/Zilliqa/Zilliqa/blob/1571634cc97baf1d0b81310d1c72395be75a9ac4/src/libUtils/IPConverter.cpp#L88`

chetan-zilliqa (changes requested):

constants.xml
```diff
@@ -392,6 +392,15 @@
             <wallet_address>ee06b3c906612cc5bdb087a30e6093c9f0aa04fd</wallet_address>
         </account>
     </ds_accounts>
+    <node_discovery>
+        <QUERY_DNS_FOR_SEED>true</QUERY_DNS_FOR_SEED>
+        <QUERY_DNS_MAX_TRIES>1</QUERY_DNS_MAX_TRIES>
```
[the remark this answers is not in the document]
sunny-teo9000: As the dns is not always accessible during my test, I added this in to retry in hope to access the dns.

But so far in the devop testnet, I don't see significant success rate with more retries.

Though I'm not quite sure would this be the same in mainnet, it could be good if we have this configurable to try out during the time when this is deployed to mainnet.

src/libLookup/Lookup.cpp
```diff
@@ -74,11 +74,14 @@ Lookup::Lookup(Mediator& mediator, SyncType syncType, bool multiplierSyncMode,
...
-  SetLookupNodes();
-  SetAboveLayer(m_seedNodes, "node.upper_seed");
-  if (!MULTIPLIER_SYNC_MODE) {
-    SetAboveLayer(m_l2lDataProviders, "node.l2l_data_providers");
-  }
+
+  InitDnsCacheList();
+
+  UpdateAllSeedsAndMultipliers();
+  m_isFirstTimeSetNodes = false;
```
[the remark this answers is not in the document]
sunny-teo9000: hmm, thanks for mentioning this, i realised this is not required.

We will always query from dns, but if failed
- resultNodes are empty = We need to read from constants.xml
- resultNodes are not empty = No changes, we rely on this current result nodes as our latest.

Will remove this.

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,348 @@
...
+  if (isShuttingDown) return;
+
+  if (dnsAddresses[listType].empty()) {
+    LOG_GENERAL(INFO, "DNS address is empty for type " << (int)listType);
+    return;
+  }
+
+  auto &dataListCache = dnsCacheListDataMap[listType];
+
+  if (dataListCache.isQuerying) {
```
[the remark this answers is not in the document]
sunny-teo9000: dataListCache will exist, as it is initialized via `InitDnsCacheList` in the Lookup's constructor.

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,348 @@
...
+        break;
+      case AF_INET6:
+        ptr = &((struct sockaddr_in6 *)res->ai_addr)->sin6_addr;
+        break;
+    }
+
+    if (ptr) {
+      inet_ntop(res->ai_family, ptr, addrstr, 100);
+      LOG_GENERAL(DEBUG, "Address: " << addrstr);
+      ipStrList.emplace_back(addrstr);
```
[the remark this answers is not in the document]
sunny-teo9000: got it, will change this

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,348 @@
...
+  auto &dataListCache = dnsCacheListDataMap[listType];
+
+  if (dataListCache.isQuerying) {
+    LOG_GENERAL(INFO, "Another thread is querying " << dnsAddresses[listType]);
+    return;
+  }
+
+  auto &ipList = dataListCache.ipList;
+
+  if (!QueryIpStrListFromDns(ipList, listType)) {
```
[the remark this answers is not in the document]
sunny-teo9000: got it, will change this

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,348 @@
...
+    currentIpKeys.emplace_back(ipKey);
+
+    if (pubKeyList.find(ipKey) != pubKeyList.end()) {
+      // Already exists, don't need to query again, unlikely to change pubkey
+      // for an ip
+      continue;
+    }
+
+    bytes pubKeyResult;
+    if (QueryPubKeyFromUrl(pubKeyResult, ipStr, listType)) {
```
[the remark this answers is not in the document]
sunny-teo9000: got it, will change this

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,348 @@
...
+  LOG_MARKER();
+  QueryDnsList(DnsListType::UPPER_SEED);
+  QueryDnsList(DnsListType::L2L_DATA_PROVIDERS);
+  QueryDnsList(DnsListType::MULTIPLIERS);
+}
+
+void AttemptPopulateLookupsDnsCacheImmediately(DnsListType listType) {
+  LOG_MARKER();
+
+  auto &dataListCache = dnsCacheListDataMap[listType];
```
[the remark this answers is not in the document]
sunny-teo9000: yea, but this will be required at the bottom.

src/libUtils/DnsUtils.cpp
sunny-teo9000: got it, will use this

sandipbhoir (approved):

src/libUtils/DnsUtils.h
```diff
@@ -0,0 +1,45 @@
...
+#include <vector>
+#include "common/BaseType.h"
+
+enum class DnsListType : int {
+  UPPER_SEED = 0,
+  L2L_DATA_PROVIDERS,
+  MULTIPLIERS
+};
+
+void InitDnsCacheList();
```
ansnunez, line 31: Can you add some level of scoping here? Like namespace or class? So we can call the methods like `DnsUtils::GetPubKeyUrl` or (if needed) `DnsUtils::GetInstance().GetPubKeyUrl`

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  auto url = dnsAddresses[listType];
+  if (url.empty()) {
+    LOG_GENERAL(INFO, "DNS address is empty for type " << (int)listType);
+    return;
+  }
+
+  auto &dataListCache = dnsCacheListDataMap[listType];
+
+  if (!dataListCache.dataAccessMutex.try_lock()) {
```
ansnunez, line 198: Just curious, are there situations where multiple threads may actually be doing query? Is it if two DS blocks are received (or mined) at somehow the same time?

In any case, good to have this in place.

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,329 @@
...
+    LOG_GENERAL(INFO, "Another thread is querying " << dnsAddresses[listType]);
+    return;
+  }
+
+  if (!QueryIpStrListFromDns(listType)) {
+    LOG_GENERAL(WARNING, "Failed to obtain ip list from "
```
ansnunez, line 204: Minor request, please make consistent the logs around "IP" and "DNS" (e.g., IP vs ip, Dns vs DNS), because some editors we use during debugging logs are case-sensitive.

(My personal preference is "IP" and "DNS").

(Also, if you agree with that, then maybe the files and function and variable names can also be modified for consistency, e.g., DNSUtils.cpp, currentIPKeys, etc.)

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,329 @@
...
+struct DnsCacheList {
+  mutex dataAccessMutex;
+
+  DnsIpListData ipList;
+  DnsPubKeyListData pubKeyList;
+};
+
+unordered_map<DnsListType, string> dnsAddresses;
+
+// End of dsepoch, a separate thread will be spawned to query
```
ansnunez, line 56: I think to be more exact, it's end of DS block consensus or processing, right? Not end of DS epoch

ansnunez (changes requested):
@sunny-teo9000 here's a first batch of comments which you can address perhaps before I continue

src/libUtils/DnsUtils.cpp
```diff
@@ -0,0 +1,329 @@
...
+  dnsAddresses[DnsListType::UPPER_SEED] = UPPER_SEED_DNS;
+  dnsAddresses[DnsListType::L2L_DATA_PROVIDERS] = L2L_DATA_PROVIDERS_DNS;
+  dnsAddresses[DnsListType::MULTIPLIERS] = MULTIPLIER_DNS;
+
+  isShuttingDown = false;
+}
+
+void ShutDownDnsCacheList() { isShuttingDown = true; }
+
+void AttemptPopulateLookupsDnsCache() {
```
chetan-zilliqa, line 267: Is DNS timeout resolution removed for some reason? 
When the `QueryDnsList`  is called for the first DNS resolution, it locks the mutex and call `QueryIpStrListFromDns` and unlocks the mutex at the last. If there's no timeout, it may be possible that the calling thread blocks at the call to `QueryIpStrListFromDns`, and this may result in other `QueryDnsList` not getting called.

chetan-zilliqa (changes requested):

src/libUtils/DnsUtils.cpp
sunny-teo9000: This mutex is trying to protect when querying is in progress.

As the query for `getaddrinfo` is a blocking call, sometimes it might take from 2ms to 15 seconds (So far when I observed querying fails).

In the chance that it blocks [indefinitely](https://man7.org/linux/man-pages/man3/getaddrinfo_a.3.html#:~:text=then%20the%20call-,blocks%20indefinitely,-(until%20one%0A%20%20%20%20%20%20%20%20%20%20of), at least this try_lock() will return false for those who are attempting to access the cache, so that we can continue to progress.

src/libDirectoryService/Coinbase.cpp
```diff
@@ -161,9 +161,18 @@ void DirectoryService::InitCoinbase() {
       ++it;
   }
 
-  const auto& vecLookup = m_mediator.m_lookup->GetLookupNodesStatic();
+  auto vecLookup = m_mediator.m_lookup->GetLookupNodesStatic();
+  std::sort(vecLookup.begin(), vecLookup.end(),
```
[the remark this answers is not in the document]
sunny-teo9000: Done

src/libDirectoryService/Coinbase.cpp
```diff
@@ -161,9 +161,18 @@ void DirectoryService::InitCoinbase() {
...
+            [](const auto& node1, const auto& node2) {
+              return node1.first < node2.first;
+            });
+
   const auto& epochNum = m_mediator.m_currentEpochNum;
 
+  // In VC case, if this wasn't cleared, it will accumulate the same epochNum
+  // previous entries
+  m_coinbaseRewardees[epochNum][CoinbaseReward::LOOKUP_REWARD].clear();
```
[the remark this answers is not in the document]
sunny-teo9000: In the case of VC found here, `InitCoinbase` was called repeatedly by `PrepareRunConsensusOnFinalBlockNormal`

This caused CoinbaseReward::LOOKUP_REWARD to have duplicate entries of lookups.

shard_id wasn't changed here, nor it was repeatedly adding on, so it didn't needed to be cleared.

Not sure if such VC had happened in the past, but I think all nodes will still have the same duplicates entries and still proceed.

src/libUtils/DnsUtils.cpp
sunny-teo9000: The `try_lock()` won't be blocked, it will check if the current thread cannot acquire the lock, it will return false.

In that case, we will fall back to the local list and continue to progress.

Previous timeout isn't really a timeout for getaddrinfo, but just a timeout to move on to local list, while getaddrinfo is still being blocked.

With this implementation, there's 1 lesser thread required.

getaddrinfo can block indefinitely, the failing cases i saw can block up to 15 seconds so far.

Instead of spawning more threads to query when DNS is down, I think it can wait for getaddrinfo to end, and retry again.

src/libLookup/Lookup.cpp
```diff
@@ -372,10 +441,10 @@ bool Lookup::RemoveFromWhitelistExtSeed(const PubKey& pubKey) {
 
 VectorOfNode Lookup::GetSeedNodes() const {
   if (!MULTIPLIER_SYNC_MODE) {
-    lock_guard<mutex> g(m_mutexL2lDataProviders);
+    shared_lock<shared_timed_mutex> g(m_mutexL2lDataProviders);
```
ansnunez, line 444: Can use `IsL2lDataProvidersEmpty` here?

src/libLookup/Lookup.cpp
```diff
@@ -372,10 +441,10 @@ bool Lookup::RemoveFromWhitelistExtSeed(const PubKey& pubKey) {
...
     return m_l2lDataProviders;
   } else {
-    lock_guard<mutex> g(m_mutexSeedNodes);
+    shared_lock<shared_timed_mutex> g(m_mutexSeedNodes);
```
ansnunez, line 447: Can use `IsUpperSeedEmpty` here?

ansnunez (changes requested):

src/libLookup/Lookup.cpp
```diff
@@ -244,7 +347,11 @@ void Lookup::SetLookupNodes() {
       <account></account> -> for lookup 2 shard 1 set 2
       <account></account> -> for lookup 2 shard 2 set 2
   */
+  m_startedTxnBatchThread = false;
```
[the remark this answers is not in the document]
sunny-teo9000: I think this variable is used for gentxn in `SenderTxnBatchThread` part.

It used to be mixed together with SetLookupNodes in the past.

But now with this feature, SetLookupNodes will be called at more places that do not require to call this.

It used to cause some problem in VC during my test on this feature, then I realised this boolean should not be mixed in SetLookupNodes, and moved it to `SetGenesisWallets`.

src/libLookup/Lookup.cpp
```diff
@@ -167,65 +181,154 @@ void Lookup::SetLookupNodes(const VectorOfNode& lookupNodes) {
...
 
-  m_startedTxnBatchThread = false;
-  m_lookupNodes.clear();
-  m_lookupNodesOffline.clear();
-  // Populate tree structure pt
+bool Lookup::GetNodesFromDnsCache(VectorOfNode& currentLocalNodes,
+                                  DNSUtils::DNSListType listType,
+                                  unsigned int port) {
+  vector<string> listOfIPFromDNS;
+  if (!DNSUtils::GetIPStrListFromDNSCache(listOfIPFromDNS, listType)) {
```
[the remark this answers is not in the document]
sunny-teo9000: ahh, you're right, the way you mentioned is better, I'll change this.

ansnunez (approved):

chetan-zilliqa (approved):

PR #2485, opened 2021-03-25
[master] Fix shard guard gaps filling

sunny-teo9000:
## Description
To fix a bug in filling up the gaps when there is not enough nodes after trimming

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


ansnunez (approved):

chetan-zilliqa (approved):
Updating the sequence of nodes going into sharding structure.
1) Fill all shard guards equal to SHARD_GUARD_TOL of the total.
2) Fill all nonshard guards till the capacity is reached.
3)  Fill all remaining nodes(shard+nonshard) if the capacity is still there.

PR #2486, opened 2021-03-25
[master] Add generated test txns to remote storage DB

ansnunez:
## Description

`Lookup::GenTxnToSend` is used by developer testnets to generate test transactions.
Currently these test transactions are not inserted into the remote db.

Notes:
1. In `SendTxnPacketToNodes`, `SendTxnPacketPrepare` is called twice (`SendTxnPacketPrepare` in turn calls `GenTxnToSend`), so we need to avoid duplicate insertion into the remote db during the second call.
1. I also moved the check for `USE_REMOTE_TXN_CREATOR` from `GenTxnToSend` to `SendTxnPacketPrepare` so that we no longer see pointless "GenTxnToSend failed" messages in mainnet.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -454,7 +454,8 @@ bool Lookup::GenTxnToSend(size_t num_txn, vector<Transaction>& shardTxn,
 
 bool Lookup::GenTxnToSend(size_t num_txn,
                           map<uint32_t, deque<pair<Transaction, uint32_t>>>& mp,
-                          uint32_t numShards) {
+                          uint32_t numShards,
+                          const bool updateRemoteStorageDBForGenTxns) {
   LOG_MARKER();
   vector<Transaction> txns;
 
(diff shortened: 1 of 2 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -529,6 +545,10 @@ bool Lookup::GenTxnToSend(size_t num_txn,
     }
   }
 
+  if (hasTransactions) {
```
chetan-zilliqa, line 548: `GenTxnToSend()`  call will happen twice, But the scenario where we pass `updateRemoteStorageDBForGenTxns = false` , `hasTransactions` will be true and `ExecuteWrite` would be called.This should not happen. Correct ?
The condition should be like this.
` if (updateRemoteStorageDBForGenTxns && hasTransactions) {
    RemoteStorageDB::GetInstance().ExecuteWrite();
  }`

chetan-zilliqa (changes requested):

chetan-zilliqa (approved):

sandipbhoir (approved):

PR #2488, opened 2021-03-26
Improve DS-Rejoin

sandipbhoir:
## Description
This PR covers following:

- DS node now rejoin from upper_seed instead of S3 as far as its not lagging by more than a ds epoch.

- Older txn packets are ignored after rejoined ( i.e. txn packets from older epochs )

- Clean `m_createdtxns`, `t_createdtxns` and `t_processedTransactions` before rejoining because they are no more relevant after rejoining.

- Commit MB submission buffer after rejoined so that now we have good chance of moving from `MICROBLOCK_SUBMISSION` to `FINALBLOCK_CONSENSUS_PREP` immediately.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DSBlockPostProcessing.cpp
+++ b/src/libDirectoryService/DSBlockPostProcessing.cpp
@@ -335,6 +335,7 @@ void DirectoryService::UpdateDSCommitteeComposition() {
 }
 
 void DirectoryService::StartNextTxEpoch() {
+  LOG_MARKER();
   if (LOOKUP_NODE_MODE) {
     LOG_GENERAL(WARNING,
                 "DirectoryService::StartNextTxEpoch not expected to be "
@@ -478,10 +479,13 @@ void DirectoryService::StartNextTxEpoch() {
       }
     };
     DetachedFunction(1, func);
+
+    CommitMBSubmissionMsgBuffer();
   }
 }
 
 void DirectoryService::StartFirstTxEpoch() {
+  LOG_MARKER();
   if (LOOKUP_NODE_MODE) {
     LOG_GENERAL(WARNING,
                 "DirectoryService::StartFirstTxEpoch not expected to be "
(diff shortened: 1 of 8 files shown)


src/libDirectoryService/DSBlockPostProcessing.cpp
```diff
@@ -365,13 +365,6 @@ void DirectoryService::StartNextTxEpoch() {
   ResetPoWSubmissionCounter();
   m_viewChangeCounter = 0;
 
-  {
```
ansnunez, line 368: Reason for removal?

src/libNode/Node.cpp
```diff
@@ -2158,9 +2158,11 @@ void Node::CommitTxnPacketBuffer() {
                 "Messenger::GetNodeForwardTxnBlock failed.");
       return;
```
ansnunez, line 2159: Hmm I just noticed this. Should we change it to `continue` instead of `return`?

ansnunez (changes requested):
Minor comments only. Also, how did you simulate lagging DS node in testing? Just curious.

src/libDirectoryService/DSBlockPostProcessing.cpp
sandipbhoir: these buffer as such are supposed to be cleared at start of first tx epoch only.
while this func `StartNextTxEpoch` is being invoked after rejoin.

sandipbhoir:
> Minor comments only. Also, how did you simulate lagging DS node in testing? Just curious.

vc8 test.
Its joins back succesfully, but again goes back in to vc8 ( failure simulation ) on start of next FB consensus.
**Note:** I tried avoiding loop but finally turns out we can avoid that for the same reason why we had a loop of VC8 simulation at first place.

bb111189 (approved):

ansnunez (approved):

PR #2516, opened 2021-04-19
[master] Fix memory increase issue of lookups/seed nodes

chetan-zilliqa:
## Description

To fix the continuous increase of memory, clean up the memory cache for STL containers for `seeds` , `lookup`, `ds` and `normal` node.

Lookup, seed = Clear every Tx epoch
DS, Miner node = Clear every vacuous epoch.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -45,6 +45,7 @@
 #include "libServer/LookupServer.h"
 #include "libServer/WebsocketServer.h"
 #include "libUtils/BitVector.h"
+#include "libUtils/CommonUtils.h"
 #include "libUtils/DataConversion.h"
 #include "libUtils/DetachedFunction.h"
 #include "libUtils/HashUtils.h"
@@ -968,6 +969,11 @@ bool Node::ProcessFinalBlockCore(uint64_t& dsBlockNumber,
       return false;
     }
 
+    // Clear STL memory cache
+    if (LOOKUP_NODE_MODE) {
+      DetachedFunction(1, CommonUtils::ReleaseSTLMemoryCache);
+    }
+
     // if lookup and loaded microblocks, then skip
     lock_guard<mutex> g(m_mutexUnavailableMicroBlocks);
     if (!(LOOKUP_NODE_MODE &&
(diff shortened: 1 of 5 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -995,6 +1004,15 @@ bool Node::ProcessFinalBlockCore(uint64_t& dsBlockNumber,
       LOG_GENERAL(WARNING, "StoreFinalBlock failed!");
       return false;
     }
+    // Clear STL memory cache
+    if (!LOOKUP_NODE_MODE) {
+      auto clearStlMemoryCache = []() -> void {
+        LOG_GENERAL(INFO, "Clearing STL container cache for mining nodes");
```
sandipbhoir, line 1010: May be we have generic log like - releasing unused memory back to system

sandipbhoir (changes requested):

src/libNode/FinalBlockProcessing.cpp
chetan-zilliqa: I will remove this log

chetan-zilliqa: Also, I will remove the time diff code logic.

src/libUtils/CommonUtils.cpp
```diff
@@ -0,0 +1,15 @@
+#include <malloc.h>
+#include <chrono>
+
+#include "libUtils/Logger.h"
+#include "libUtils/CommonUtils.h"
+
+using namespace std;
+
+void CommonUtils ::ReleaseSTLMemoryCache() {
+  auto startTime = chrono::high_resolution_clock::now();
```
sandipbhoir, line 10: Add `LOG_MARKER()` to this function

sandipbhoir (changes requested):

ansnunez (approved):

src/libUtils/CommonUtils.cpp
chetan-zilliqa: done.

chetan-zilliqa:
@sandipbhoir , I have bit refactored the code. Please review the final code.

sandipbhoir (approved):

PR #2527, opened 2021-04-27
[release-8.0] fix processing of duplicate updates

KaustubhShamshery:
## Description
Also detach `BulkWrite`

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -687,6 +687,10 @@ bool Node::ProcessFinalBlock(const bytes& message, unsigned int offset,
       }
       // Clear the vc blocks store
       m_vcBlockStore.clear();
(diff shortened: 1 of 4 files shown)


src/libNode/FinalBlockProcessing.cpp
```diff
@@ -687,6 +687,9 @@ bool Node::ProcessFinalBlock(const bytes& message, unsigned int offset,
       }
       // Clear the vc blocks store
       m_vcBlockStore.clear();
+    } else if (REMOTESTORAGE_DB_ENABLE && !ARCHIVAL_LOOKUP) {
```
sandipbhoir, line 690: Would be better to add LOOKUP_NODE_MODE in check as well ?

sandipbhoir (changes requested):

ansnunez (approved):

sandipbhoir (approved):

PR #2542, opened 2021-05-04
[release-8.0] Avoid deadlock issue

sandipbhoir:
## Description
This PR fixes deadlock issue.
- Basically, set `m_txn_distribute_window_open = true` in `Node::StartFirstTxEpoch()`
- Avoid tx packet cleanup in `CleanCreatedTransaction()` when already have packets waiting for state `MICROBLOCK_CONSENSUS_PREP` in order to be distributed to others. Otherwise, it would cause deadlock.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -2351,23 +2351,36 @@ void Node::CleanCreatedTransaction() {
     std::lock_guard<mutex> g(m_mutexCreatedTransactions);
     m_createdTxns.clear();
     t_createdTxns.clear();
+    LOG_GENERAL(INFO, "Cleaned created txns!");
   }
-  {
-    std::lock_guard<mutex> g(m_mutexTxnPacketBuffer);
-    m_txnPacketBuffer.clear();
-  }
-  {
-    std::lock_guard<mutex> g(m_mutexTxnPktInProcess);
-    m_txnPktInProcess.clear();
+  // Avoid cleaning when buffer already have packets waiting to be picked up for
+  // distribution. Thus avoid deadlock.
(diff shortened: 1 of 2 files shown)


src/libNode/Node.cpp
```diff
@@ -2352,13 +2352,21 @@ void Node::CleanCreatedTransaction() {
...
-    std::lock_guard<mutex> g(m_mutexTxnPktInProcess);
-    m_txnPktInProcess.clear();
+  // Avoid cleaning when buffer already have packets waiting to be picked up for
+  // distribution. Thus avoid deadlock.
+  if (m_txnPacketThreadOnHold == 0) {
+    {
+      // extra safety
+      std::unique_lock<std::mutex> lock(m_mutexTxnPacketBuffer,
+                                        std::try_to_lock);
+      if (!lock.owns_lock()) {
```
ansnunez, line 2362: Should this be `if (lock.owns_lock())`?

src/libNode/Node.cpp
```diff
@@ -2352,13 +2352,21 @@ void Node::CleanCreatedTransaction() {
...
+        m_txnPacketBuffer.clear();
```
ansnunez, line 2363: Maybe before every `clear` in this function we can add a short log message so we know if and when the different clearings are done.

ansnunez (changes requested):

ansnunez (approved):

chetan-zilliqa (approved):
I think the first fix has to uncomment out before the merge.

PR #2564, opened 2021-05-18
Add missing property to constants config to make isolatedServer start successfully

hivtushok:
CC: @KaustubhShamshery 

## Description

This PR is to make `/usr/local/bin/isolatedServer` start successfully in https://github.com/Zilliqa/Zilliqa/blob/master/docker/Dockerfile.

Currently, running it causes the following error: 

```# /usr/local/bin/isolatedServer -f isolated-server-accounts.json
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::property_tree::ptree_bad_path> >'
  what():  No such node (node.smart_contract.disambiguate_exclusion_list)
Aborted
```

It points out that `disambiguate_exclusion_list` property inside the `constants.xml` is missing. This PR fills it in with the default empty value. After this change, I was able to run it successfully inside the Docker image.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test (commit: 1b340c04) 

- [ ] small-scale cloud test

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -284,6 +284,7 @@
         <LOG_SC>false</LOG_SC>
         <DISABLE_SCILLA_LIB>false</DISABLE_SCILLA_LIB>
         <SCILLA_SERVER_PENDING_IN_MS>1500</SCILLA_SERVER_PENDING_IN_MS>
+        <disambiguate_exclusion_list></disambiguate_exclusion_list>
     </smart_contract>
     <tests>
         <ENABLE_CHECK_PERFORMANCE_LOG>false</ENABLE_CHECK_PERFORMANCE_LOG>
--- a/constants_local.xml
+++ b/constants_local.xml
@@ -283,6 +283,7 @@
         <LOG_SC>true</LOG_SC>
         <DISABLE_SCILLA_LIB>false</DISABLE_SCILLA_LIB>
         <SCILLA_SERVER_PENDING_IN_MS>1500</SCILLA_SERVER_PENDING_IN_MS>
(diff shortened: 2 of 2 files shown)


bb111189 (approved):

KaustubhShamshery:
@hivtushok it seems that you are using one of the alpha releases. The following constant has not been added in the current master (ref : https://github.com/Zilliqa/Zilliqa/blob/master/src/common/Constants.h). Kindly use one of the releases to run the iso server.

hivtushok:
> @hivtushok it seems that you are using one of the alpha releases. The following constant has not been added in the current master (ref : https://github.com/Zilliqa/Zilliqa/blob/master/src/common/Constants.h). Kindly use one of the releases to run the iso server.

I'm using `zilliqa/zilliqa:v8.0.3` image and it seems, that it uses the alpha version of scilla image. Can it updated to the release version?

bb111189:
hey, can you check out our release branch over at https://github.com/Zilliqa/Zilliqa/blob/release-8.0/docker/Dockerfile
It should be updated.

Thanks a lot. We will also update the master accordingly.

hivtushok:
> hey, can you check out our release branch over at https://github.com/Zilliqa/Zilliqa/blob/release-8.0/docker/Dockerfile
> It should be updated.
> 
> Thanks a lot. We will also update the master accordingly.

@bb111189 thanks!

KaustubhShamshery:
Closing because the issue will be in a separate fix.

PR #2565, opened 2021-05-19
[master] Avoid cleaning txns if no shard mb was included in fb itself

sandipbhoir:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/FinalBlockProcessing.cpp
+++ b/src/libNode/FinalBlockProcessing.cpp
@@ -132,6 +132,7 @@ bool Node::LoadUnavailableMicroBlockHashes(const TxBlock& finalBlock,
   const auto& microBlockInfos = finalBlock.GetMicroBlockInfos();
 
   bool foundMB = false;
+  bool foundMismatchedMB = false;
   // bool doRejoin = false;
 
   for (const auto& info : microBlockInfos) {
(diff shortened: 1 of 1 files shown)


KaustubhShamshery (approved):

src/libNode/FinalBlockProcessing.cpp
```diff
@@ -1443,13 +1454,15 @@ bool Node::ProcessMBnForwardTransaction(
        m_mediator.m_lookup->GetSyncType() == SyncType::NEW_LOOKUP_SYNC) ||
       (LOOKUP_NODE_MODE && !ARCHIVAL_LOOKUP &&
```
sunny-teo9000, line 1455: Maybe the if condition can be combined here since there are 2 using `LOOKUP_NODE_MODE && ARCHIVAL_LOOKUP`

sunny-teo9000 (approved):

src/libNode/FinalBlockProcessing.cpp
sandipbhoir: well, one is `ARCHIVAL_LOOKUP` and other is `!ARCHIVAL_LOOKUP`

bb111189:
Will need to perform a View changes test. Esp those on the DM series

sandipbhoir:
> Will need to perform a View changes test. Esp those on the DM series

yup. good to run vc test but vc timer always starts in `RunConsensusOnFinalBlock`. With above change we are just delaying running final block consensus by extra time.

PR #2567, opened 2021-05-20
[master] Check blocknum of pow submission

sandipbhoir:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/PoWProcessing.cpp
+++ b/src/libDirectoryService/PoWProcessing.cpp
@@ -184,6 +184,13 @@ bool DirectoryService::ProcessPoWSubmission(
     return false;
   }
 
+  uint64_t expectedBlockNum =
+      m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetBlockNum() + 1;
+  if (blockNumber != expectedBlockNum) {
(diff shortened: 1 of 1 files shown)


src/libDirectoryService/PoWProcessing.cpp
```diff
@@ -117,8 +117,8 @@ bool DirectoryService::ProcessPoWPacketSubmission(
 
   LOG_GENERAL(INFO, "PoW solutions received in this packet: " << tmp.size());
   for (auto& sol : tmp) {
-    // No point processing the other solutions if DS Block consensus is starting
-    if ((m_state == DSBLOCK_CONSENSUS_PREP) || (m_state == DSBLOCK_CONSENSUS)) {
+    // No point processing other solutions if no more pow submission state.
+    if (m_state != POW_SUBMISSION) {
```
bb111189, line 121: There can be a timing issue. Let's say a DS node is at waiting to receive the final block at vacuous epoch before transiting to POW submission, it may drop POW submission from nodes that are faster than it

bb111189 (changes requested):

bb111189 (approved):

bb111189 (approved):

chetan-zilliqa (approved):

PR #2569, opened 2021-05-22
[release-8.0] Make checking for an empty key more optimized

KaustubhShamshery:
## Description
to check if the key was empty, we used to fetch the whole key's value.
Optimize it to just return if it is non-empty

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libPersistence/ContractStorage2.cpp
+++ b/src/libPersistence/ContractStorage2.cpp
@@ -698,6 +698,66 @@ void ContractStorage2::FetchStateDataForKey(map<string, bytes>& states,
   }
 }
 
+bool ContractStorage2::CheckIfKeyIsEmpty(const string& key, bool temp) {
+  std::map<std::string, bytes>::iterator p;
+  unordered_set<string> keys_to_be_deleted;
+
+  auto checkIfKeyIsToBeDeleted = [&](const string& key) mutable {
+    if (keys_to_be_deleted.find(key) != keys_to_be_deleted.end()) {
+      return true;
+    }
+    if (temp) {
+      if (t_indexToBeDeleted.find(key) != t_indexToBeDeleted.cend()) {
+        keys_to_be_deleted.insert(key);
+        return true;
+      }
+    }
+    if (m_indexToBeDeleted.find(key) != m_indexToBeDeleted.cend()) {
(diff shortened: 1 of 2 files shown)


vaivaswatha (approved):

sandipbhoir (approved):

src/libPersistence/ContractStorage2.cpp
```diff
@@ -698,6 +698,66 @@ void ContractStorage2::FetchStateDataForKey(map<string, bytes>& states,
...
+      return true;
+    }
+    if (temp) {
+      if (t_indexToBeDeleted.find(key) != t_indexToBeDeleted.cend()) {
+        keys_to_be_deleted.insert(key);
+        return true;
+      }
+    }
+    if (m_indexToBeDeleted.find(key) != m_indexToBeDeleted.cend()) {
+      keys_to_be_deleted.insert(key);
```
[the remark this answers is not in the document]
chetan-zilliqa: Ohh. Ignore the above comment. I understand now that it is intended.

src/libPersistence/ContractStorage2.cpp
```diff
@@ -698,6 +698,66 @@ void ContractStorage2::FetchStateDataForKey(map<string, bytes>& states,
...
+      return false;
+    }
+    ++p;
+  }
+
+  std::unique_ptr<leveldb::Iterator> it(
+      m_stateDataDB.GetDB()->NewIterator(leveldb::ReadOptions()));
+
+  it->Seek({key});
+  if (!it->Valid() || it->key().ToString().compare(0, key.size(), key) != 0) {
```
chetan-zilliqa, line 747: @KaustubhShamshery , Can we shift this if statement on the top and make everything else as "else" statement.
If `  if (!it->Valid() || it->key().ToString().compare(0, key.size(), key) != 0)` is not `true` then we anyways returning true.It will save time for filling map `keys_to_be_deleted` which won't be required for this particular condition.

chetan-zilliqa: Thanks @KaustubhShamshery for clearing doubts here. I understand now. We are checking in the map first before traversing through level DB.

PR #2572, opened 2021-05-24
[release-8.0] Make unix domain server timeout configurable

KaustubhShamshery:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


cmake/InstallJsonrpccpp.cmake
```diff
@@ -46,8 +46,8 @@ set(CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
 
 ExternalProject_Add(jsonrpc-project
     PREFIX src/depends/jsonrpc
-    URL https://github.com/cinemast/libjson-rpc-cpp/archive/v1.2.0.tar.gz
-    URL_HASH SHA256=485556bd27bd546c025d9f9a2f53e89b4460bf820fd5de847ede2539f7440091
+    URL https://github.com/KaustubhShamshery/libjson-rpc-cpp/archive/v1.2.0-time-patch.tar.gz
```
bb111189, line 49: change to this repo https://github.com/Zilliqa/libjson-rpc-cpp

bb111189 (changes requested):

tests/Server/Test_ScillaIPCServer.cpp
```diff
@@ -43,6 +43,7 @@ BOOST_AUTO_TEST_SUITE(scillaipc)
 BOOST_AUTO_TEST_CASE(test_connection) {
   INIT_STDOUT_LOGGER();
   UnixDomainSocketServer s(SCILLA_IPC_SOCKET_PATH);
+  s.SetWaitTime(10);
```
sandipbhoir, line 46: May be we add wait time to constants.xml

sandipbhoir (changes requested):

vaivaswatha:
I see that you're calling `SetWaitTime()` at each place that `UnixDomainSocketServer` is instantiated. An alternate to this is to derive from `UnixDomainSocketServer`, a class that sets `SetWaitTime` in its constructor. That way you only create instances of this class and don't manually set the time everytime.

sandipbhoir (approved):

bb111189 (approved):

KaustubhShamshery:
> I see that you're calling `SetWaitTime()` at each place that `UnixDomainSocketServer` is instantiated. An alternate to this is to derive from `UnixDomainSocketServer`, a class that sets `SetWaitTime` in its constructor. That way you only create instances of this class and don't manually set the time everytime.

For now since only scilla server needs setting of this value and it is only intastiated once, the current design does not have any major disadvantage. However, the concern is valid, can be made a lower priority issue

PR #2573, opened 2021-05-25
[release-8.0] Ask for cosig when only one shard

sandipbhoir:
## Description
PR addressess issue https://github.com/Zilliqa/Issues/issues/724

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -1065,10 +1065,9 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
          blockNum++) {
       const auto& it = coinbaseRewardeesTmp.find(blockNum);
       if (it == coinbaseRewardeesTmp.end() ||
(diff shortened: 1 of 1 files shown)


bb111189 (approved):

chetan-zilliqa (approved):

chetan-zilliqa:
Just for my understanding, In the case of 1 shard, MB size would be 1 so any check like.
1 or more <= 1 will pass.

sandipbhoir:
> Just for my understanding, In the case of 1 shard, MB size would be 1 so any check like.
> 1 or more <= 1 will pass.

1 <= 1 will pass (one shard only)
(>1) <=1 will not pass.

chetan-zilliqa:
ohh.yeah.Correct

PR #2586, opened 2021-05-31
[master] Add health check in jsonrpc

sunny-teo9000:
## Description
Adding a empty healthcheck so that we can check if zilliqa process is running

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/miner_info.py
+++ b/scripts/miner_info.py
@@ -103,6 +103,7 @@ def make_options_dictionary(options_dict):
 	options_dict["dsepoch"] = "GetCurrentDSEpoch"
 	options_dict["type"] = "GetNodeType"
 	options_dict["ds"] = "GetDSCommittee"
+	options_dict["health"] = "healthcheck"
 	options_dict["state"] = "GetNodeState"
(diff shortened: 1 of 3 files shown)


sandipbhoir (approved):

joel-zilliqa (approved):

Revolution1:
uh, do adding a status server rpc method really meet our need?

The code shows it always returns 'ok'

If it's OK, of course we can get a 'ok'

But, if it's not OK
1. first, we need to `POST` a JSONRPC request to status server's listening port, then only `miner_info.py` can do it, cause most healthchecker don't support `HTTP POST` method
2. second
     1. if status server stucked (cause it's single threaded now), the request (and miner_info.py) will stuck too, which leads to timeout
     2. if status server stopped, the port will not be listened too, `miner_info.py` will just raise a error, which is identical to healthcheck command like `nc -zv localhost 4301`

So, if we need a healthcheck, it better be accessible through HTTP GET method
and returns some internal metrics, like `transaction process time summary`, `last time of it successfully connected to a node` etc.

PR #2593, opened 2021-06-08
[master] Set limit on querying DS Block

sunny-teo9000:
## Description
To limit the amount of DS blocks being retrieved from ProcessGetDSBlockFromSeed
https://github.com/Zilliqa/Issues/issues/686

Any call for `RetrieveDSBlocks` that is not partial will be limit to `FETCH_DS_BLOCK_LIMIT` of blocks starting from `lowBlockNum`

## Backward Compatibility

- [] This is a breaking change

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


constants.xml
```diff
@@ -198,6 +198,7 @@
         <BROADCAST_INTERVAL>60</BROADCAST_INTERVAL>
         <BROADCAST_EXPIRY>600</BROADCAST_EXPIRY>
         <FETCH_LOOKUP_MSG_MAX_RETRY>3</FETCH_LOOKUP_MSG_MAX_RETRY>
+        <FETCH_DS_BLOCK_LIMIT>2</FETCH_DS_BLOCK_LIMIT>
```
chetan-zilliqa, line 201: Should this constant be in `seed` section ? P2PComm deals with the network layer.

src/libLookup/Lookup.cpp
```diff
@@ -1855,6 +1855,18 @@ void Lookup::RetrieveDSBlocks(vector<DSBlock>& dsBlocks, uint64_t& lowBlockNum,
     highBlockNum = curBlockNum;
   }
 
+  uint64_t diff = 0;
+  if (!partialRetrieve &&
+      SafeMath<uint64_t>::sub(highBlockNum, lowBlockNum, diff)) {
+    if (diff > FETCH_DS_BLOCK_LIMIT) {
```
chetan-zilliqa, line 1861: Do we think that we should increase the `FETCH_DS_BLOCK_LIMIT`?  As on the client-side, the getter function `FetchLatestDSBlocksSeed` is called in a loop, So there would be many calls for a large block range and every time lookup has to execute function for sliced block range.

sandipbhoir (changes requested):

chetan-zilliqa (approved):

sandipbhoir (approved):

PR #2600, opened 2021-06-13
[master] Purge MPT roots for normal node

KaustubhShamshery:
## Description
Add purging of historical data for normal node (non-lookup)

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libData/DataStructures/TraceableDB.h
```diff
@@ -110,9 +116,22 @@ class TraceableDB : public dev::OverlayDB {
  public:
   void DetachedExecutePurge() {
     LOG_MARKER();
-    auto detached_func = [this]() { return ExecutePurge(0, true); };
+    auto detached_func = [this]() -> void {
+      if (!m_purgeRunning) {
+        m_purgeRunning = true;
```
sandipbhoir, line 121: set m_stopSignal to false otherwise ExecutePurge might stop immediately

sandipbhoir (changes requested):

src/libData/DataStructures/TraceableDB.h
KaustubhShamshery: It is false by default and also set false after every function, it won't be true if the thread isn't running.

sandipbhoir (approved):

src/libZilliqa/Zilliqa.cpp
```diff
@@ -236,6 +236,7 @@ Zilliqa::Zilliqa(const PairOfKey& key, const Peer& peer, SyncType syncType,
     if (!AccountStore::GetInstance().RefreshDB()) {
       LOG_GENERAL(WARNING, "AccountStore::RefreshDB failed");
     }
+    AccountStore::GetInstance().PurgeUnnecessary();
```
chetan-zilliqa, line 239: Is this function meant to be called for `seedpubs or seedprvs or seednodes`?

chetan-zilliqa (approved):

src/libZilliqa/Zilliqa.cpp
KaustubhShamshery: No

sandipbhoir (approved):

chetan-zilliqa (approved):

PR #2612, opened 2021-06-21
[master] Add VC9 - Suspend before response

sunny-teo9000:
## Description
Add VC9 test
DS Leader halt consensus at final challenge 1
Then VC
Then the next leader halt consensus again once

In the same period got to spam smart contracts

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/FinalBlockPostProcessing.cpp
+++ b/src/libDirectoryService/FinalBlockPostProcessing.cpp
@@ -567,6 +567,20 @@ bool DirectoryService::ProcessFinalBlockConsensusCore(
     return false;
   }
 
+#ifdef VC_TEST_FB_SUSPEND_RESPONSE
+  ConsensusCommon::State state = m_consensusObject->GetState();
+
+  if (state == FINALCHALLENGE_DONE && m_mode == PRIMARY_DS &&
+      m_viewChangeCounter == 0 &&
+      m_mediator.m_txBlockChain.GetBlockCount() % NUM_FINAL_BLOCK_PER_POW !=
+          0) {
+    LOG_EPOCH(WARNING, m_mediator.m_currentEpochNum,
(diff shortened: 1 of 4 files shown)


src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -567,6 +567,23 @@ bool DirectoryService::ProcessFinalBlockConsensusCore(
     return false;
   }
 
+#ifdef VC_TEST_FB_SUSPEND_RESPONSE
+  ConsensusCommon::State state = m_consensusObject->GetState();
+
+  static const int afterVcCount = 0;
+  if (state == CHALLENGE_DONE && m_mode == PRIMARY_DS)
```
KaustubhShamshery, line 574: I think we want to re-trigger if state is `FINALCHALLENGE_DONE`

src/libDirectoryService/FinalBlockPostProcessing.cpp
```diff
@@ -567,6 +567,23 @@ bool DirectoryService::ProcessFinalBlockConsensusCore(
...
+    if (m_viewChangeCounter == 0  // Initial ds leader
```
bb111189, line 575: I think `m_viewChangeCounter == 0` will be enough. It will stop consensus, view change, new leader restart consensus and move on. Then next epoch repeat

bb111189: To make the test more efficient, maybe don't halt it at vaccous epoch

bb111189 (changes requested):

sandipbhoir (approved):

bb111189 (approved):

PR #2632, opened 2021-06-28
[master] Upgrade jsonrpc to v1.3.0 and add epoll flags to MHD methods

sunny-teo9000:
## Description
Upgrade JSONRPC and add new MHD flags
Add logs to safehttpserver

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


joel-zilliqa (approved):
LGTM, deployed the pod and tested, the 502 issue is no longer able to be reproduced with the changes

sandipbhoir (approved):

PR #2642, opened 2021-07-08
[master] fix purging

KaustubhShamshery:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/DataStructures/TraceableDB.cpp
+++ b/src/libData/DataStructures/TraceableDB.cpp
@@ -22,7 +22,9 @@ using namespace std;
 
 bool TraceableDB::commit(const uint64_t& dsBlockNum) {
   std::vector<dev::h256> toPurge;
-  if (!OverlayDB::commit(KEEP_HISTORICAL_STATE && LOOKUP_NODE_MODE, toPurge)) {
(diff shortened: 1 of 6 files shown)


src/libData/DataStructures/TraceableDB.cpp
```diff
@@ -115,11 +128,13 @@ bool TraceableDB::RefreshDB() {
...
 void TraceableDB::DetachedExecutePurge() {
   LOG_MARKER();
+
   auto detached_func = [this]() -> void {
     if (!m_purgeRunning) {
+      unordered_set<dev::h256> inserted;
       m_stopSignal = false;
       m_purgeRunning = true;
-      ExecutePurge(0, true);
+      ExecutePurge(0, inserted, true);
```
chetan-zilliqa, line 137: Hi @KaustubhShamshery , Just to confirm here,  `inserted` set is not of any use in `DetachedExecutePurge`. It is there to keep for the sake of the signature of function `ExecutePurge` and not to duplicate the code around. Right?

KaustubhShamshery:
Added in #2656

PR #2667, opened 2021-07-20
[release-8.1] Change JsonRpc port start stop mechanism

chetan-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -3351,22 +3351,27 @@ bool Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
           m_rejoinInProgress = false;
           m_rejoinNetworkAttempts = 0;  // reset rejoining attempts
           cv_setRejoinRecovery.notify_all();
-          if (m_lookupServer) {
-            if (m_lookupServer->StartListening()) {
-              LOG_GENERAL(INFO, "API Server started to listen again");
-            } else {
-              LOG_GENERAL(WARNING, "API Server couldn't start");
-            }
-          }
           m_isFirstLoop = true;
-
-          if (m_stakingServer) {
-            if (m_stakingServer->StartListening()) {
-              LOG_GENERAL(INFO, "Staking Server started to listen again");
-            } else {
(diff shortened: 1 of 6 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -3351,22 +3351,32 @@ bool Lookup::CommitTxBlocks(const vector<TxBlock>& txBlocks) {
...
-              LOG_GENERAL(INFO, "API Server started to listen again");
-            } else {
-              LOG_GENERAL(WARNING, "API Server couldn't start");
-            }
-          }
           m_isFirstLoop = true;
+          auto startJsonRpc = [this]() mutable -> void {
+            LOG_GENERAL(INFO, "In startJsonRpc func");
+            std::unique_lock<std::mutex> cv_lk(m_mutexJsonRpcPortOpen);
+            if (m_mediator.m_lookup->cv_jsonRpcPortOpen.wait_for(
```
sandipbhoir, line 3358: should consider checking some flag before waiting? Otherwise, here it will always wait for 180 sec before starting listening, if `listening stops `finished before `listening start` is called.
Flag should be set/reset before notify.

src/libLookup/Lookup.cpp
```diff
@@ -4490,6 +4495,7 @@ void Lookup::RejoinAsNewLookup(bool fromLookup) {
       this_thread::sleep_for(chrono::seconds(SEED_SYNC_SMALL_PULL_INTERVAL));
     }
     auto func1 = [this]() mutable -> void {
+      std::lock_guard<mutex> lock(m_mutexJsonRpc);
```
sandipbhoir, line 4498: Can add log before acquiring mutex - indicating stop-listening func started

sandipbhoir (changes requested):

src/libLookup/Lookup.cpp
chetan-zilliqa: done

sandipbhoir (approved):

bb111189 (approved):

sunny-teo9000 (approved):

PR #2690, opened 2021-08-19
[release-8.1] Add extra txn distribution time before FB consensus when no shard

sandipbhoir:
## Description
Fix issue:
https://github.com/Zilliqa/Issues/issues/743
https://github.com/Zilliqa/Issues/issues/744

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


bb111189 (approved):

chetan-zilliqa (approved):
@sandipbhoir, Just to understand, we have received `DSMBMISSINGTXN` for the `ANNOUNCE`response in yesterday's mainnet issue, so are we addressing the issue of missing txn's by increasing the transaction distribution window from current 6 secs to 15 secs, so that all DSgurads have a same number of txns.

KaustubhShamshery (approved):

chetan-zilliqa (approved):

PR #2696, opened 2021-08-23
[master] Fix for CLOSE_WAIT on EPOLL

sunny-teo9000:
## Description
Fixing the CLOSE_WAIT issue by adding 2 timeout
1st timeout is at `notify_connection_callback` for all connections passing by (Usually set to 1 second)
2nd timeout is at `SafeHttpServer::callback` where it look for our callback methods (Here we set it back to unlimited)

This filter out the CLOSE_WAIT connections, which will get eliminated in 1 second.

If we were to turn off the timeout feature, we can just set all timeout in constant to zero. (Zero = No timeout limit)

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/depends/safeserver/safehttpserver.cpp
+++ b/src/depends/safeserver/safehttpserver.cpp
@@ -29,12 +29,11 @@ struct mhd_coninfo {
     int code;
 };
 
-SafeHttpServer::SafeHttpServer(int port, bool useEpoll, const std::string &sslcert, const std::string &sslkey, int threads) :
+SafeHttpServer::SafeHttpServer(int port, const std::string &sslcert, const std::string &sslkey, int threads) :
     AbstractServerConnector(),
     port(port),
     threads(threads),
     running(false),
-    useEpoll(useEpoll),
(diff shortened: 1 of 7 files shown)


KaustubhShamshery (approved):

src/depends/safeserver/safehttpserver.cpp
```diff
@@ -222,6 +229,26 @@ int SafeHttpServer::callback(void *cls, MHD_Connection *connection, const char *
   return MHD_YES;
 }
 
+void SafeHttpServer::notify_connection_callback(void* cls,
+                      struct MHD_Connection* connection,
+                      void** socket_context,
+                      enum MHD_ConnectionNotificationCode toe)
```
sandipbhoir, line 235: just curious to know why arg name `toe`? :)

sandipbhoir (approved):

Kooriii (approved):

PR #2723, opened 2021-09-15
[master] Avoid txn loss during upgrade of seedpubs and lookups

sandipbhoir:
Issue: https://github.com/Zilliqa/Issues/issues/704

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


src/libLookup/Lookup.cpp
```diff
@@ -4759,7 +4822,9 @@ bool Lookup::ToBlockMessage(unsigned char ins_byte) {
           ins_byte != LookupInstructionType::SETSTATEDELTAFROMSEED &&
           ins_byte != LookupInstructionType::SETSTATEDELTASFROMSEED &&
           ins_byte != LookupInstructionType::SETDIRBLOCKSFROMSEED &&
-          ins_byte != LookupInstructionType::SETMINERINFOFROMSEED);
+          ins_byte != LookupInstructionType::SETMINERINFOFROMSEED &&
+          ins_byte != LookupInstructionType::FORWARDTXN);  // Allow fwded txn to
```
chetan-zilliqa, line 4826: Here ,Adding forward txn here, do we mean to ignore such requests if the node is not synced?
I didn't get what is meant by comment `Allow fwded txn to to buffer`

chetan-zilliqa (approved):

src/libLookup/Lookup.cpp
sandipbhoir: To allow/**Not-Block** `FORWARDTXN` msgs even if not synced.

PR #2727, opened 2021-09-20
[master] Remove contract states from GetSmartContractState

KaustubhShamshery:
## Description
Only print the contract addresses associated with the account.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


chetan-zilliqa (approved):
I think , we will have to remove the state from API-docs as well.
https://dev.zilliqa.com/docs/apis/api-contract-get-smartcontracts

bb111189 (approved):

bb111189:
Please inform viewblock about this breaking change

PR #2735, opened 2021-10-04
[release-8.1] Improve API GetSmartContracts to avoid holding primary mutex for long time

sandipbhoir:
This PR addresses issue : https://github.com/Zilliqa/Issues/issues/752

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -1012,31 +1012,39 @@ Json::Value LookupServer::GetSmartContracts(const string& address) {
 
   try {
     Address addr{ToBase16AddrHelper(address)};
+    uint64_t nonce = 0;
+    {
+      shared_lock<shared_timed_mutex> lock(
+          AccountStore::GetInstance().GetPrimaryMutex());
 
-    shared_lock<shared_timed_mutex> lock(
-        AccountStore::GetInstance().GetPrimaryMutex());
+      const Account* account =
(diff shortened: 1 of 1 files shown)


KaustubhShamshery (approved):

src/libServer/LookupServer.cpp
```diff
@@ -1012,31 +1012,39 @@ Json::Value LookupServer::GetSmartContracts(const string& address) {
...
+
     //[TODO] find out a more efficient way (using storage)
     Json::Value _json;
 
     for (uint64_t i = 0; i < nonce; i++) {
       Address contractAddr = Account::GetAddressForContract(addr, i);
-      const Account* contractAccount =
-          AccountStore::GetInstance().GetAccount(contractAddr, true);
+      {
+        shared_lock<shared_timed_mutex> lock(
```
chetan-zilliqa, line 1040: With moving mutex inside the for loop, we are expecting that if it's released even once in for loop continuation, then we would get a chance to proceed with other execution flow of `DeserializeDelta`, and then once the `DeserializeDelta` is finished, the API execution can carry on. Right ?

chetan-zilliqa (approved):

src/libServer/LookupServer.cpp
sandipbhoir: Right

PR #2747, opened 2021-10-28
[master] Persistence upload/download optimization feature(primary/secondary bucket support)

chetan-zilliqa:
## Description

Changes in `upload_incr_DB.py`:
1  Added support for primary/secondary bucket.
2 Upload will happen to both the buckets using two threads.

Changes in `download_incr_DB.py`
1 Download script will always choose PRIMARY_BUCKET for downloading.

2 If there is an exception that happens during downloading persistence(for eg fetching lock files, key files), the script will try downloading `MAX_RETRY_DOWNLOAD_PERSISTENCE_FROM_PRIMARY` times from primary and then fallback to download from the secondary bucket. The script will give up download after `MAX_RETRY_DOWNLOAD_OVERALL_PERSISTENCE_ATTEMPT` attempts.

3 If there is an issue in file download then the download script will `MAX_RETRY_DOWNLOAD_INDIVIDUAL_FILE` for individual files.

4 There could be the possibility of connection timeout while downloading individual files. In such a scenario, the download script will try downloading from an alternate bucket after `MAX_RETRY_DOWNLOAD_INDIVIDUAL_FILE`

5 The PR also requires changes in the testnet. Below is the PR for that.
https://github.com/Zilliqa/testnet/pull/939

**Note**: Testnet PR and this PR have to go **TOGETHER** because we have changed the `--bucket` parameter name to `--primary-bucket` in bootstrap.py script.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


scripts/download_incr_DB.py
```diff
@@ -180,27 +213,49 @@ def GetAllObjectsFromS3(url, folderName=""):
...
 	mutex.release()
 	while True:
 		try:
-			response = requests.get(key_url, stream=True)
-		except Exception as e:
-			print("Exception occurred while downloading " + key_url + ": " + str(e))
+			# Test code to reproduce connection timeout issue
+			if key_url == 	"http://temp-chetan-test-bucket-devops.s3.amazonaws.com/incremental/chetan-cd17b90-2/persistence/txEpochs/000024.ldb":
+				key_url = 	"http://temp-chetan-test-bucket-devops.s3.amazonas.com/incremental/chetan-cd17b90-2/persistence/txEpochs/000024.ldb"
+			elif key_url == "http://temp-chetan-test-bucket-devops.s3.amazonaws.com/incremental/chetan-cd17b90-2/persistence/txEpochs/000024.ldb":
```
sandipbhoir, line 239: seems same condition for both if & elif check

scripts/download_incr_DB.py
```diff
@@ -125,10 +134,34 @@ def LaunchParallelUrlFetch(list_of_keyurls):
...
+			modified_url = modified_url.replace(PRIMARY_BUCKET_NAME,SECONDARY_BUCKET_NAME)
+			# Remove testcode
+			#modified_url = TestCodeCorrectUrl(modified_url)
+			replaced_failed_url.append(modified_url)
+			# Remove the log
+		elif modified_url.find(SECONDARY_BUCKET_NAME) > 0:
+			modified_url = modified_url.replace(SECONDARY_BUCKET_NAME,PRIMARY_BUCKET_NAME)
+			#modified_url = TestCodeCorrectUrl(modified_url)
+			replaced_failed_url.append(modified_url)
+			# Remove the log
```
sandipbhoir, line 157: To be removed in addition to `Remove testcode` comments.

sandipbhoir (commented):
Looks good. Will review again after code cleanup

sandipbhoir (approved):

joel-zilliqa (approved):

PR #2750, opened 2021-11-10
[master] Individual POW submission to DS Committee

Kooriii:
## Description

Change sending of POW submission to send individually as to prevent shards pow submission being delayed or gone missing leading to mainnet having little shard nodes

Issue: https://github.com/Zilliqa/Issues/issues/751

Current testnet is running at `epoch 150` without any VC.

## Backward Compatibility

## Review Suggestion

If there is any other suggestions for the above issue, please do let me know on how to improve this PR

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/PoWProcessing.cpp
+++ b/src/libNode/PoWProcessing.cpp
@@ -373,7 +373,14 @@ bool Node::SendPoWResultToDSComm(const uint64_t& block_num,
     }
   }
 
-  P2PComm::GetInstance().SendMessage(peerList, powmessage);
+  // Instead of sending whole list at once, send 1 by 1 to prevent incidents
+  // where that particular node is down and hence the whole list is being
+  // delayed or gone
+  auto& p2pComm = P2PComm::GetInstance();
+  for (const auto& p : peerList) {
+    p2pComm.SendMessage(p, powmessage);
+  }
+
   return true;
 }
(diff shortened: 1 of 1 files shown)


src/libNode/PoWProcessing.cpp
```diff
@@ -373,7 +373,14 @@ bool Node::SendPoWResultToDSComm(const uint64_t& block_num,
     }
   }
 
-  P2PComm::GetInstance().SendMessage(peerList, powmessage);
+  // Instead of sending whole list at once, send 1 by 1 to prevent incidents
+  // where that particular node is down and hence the whole list is being
+  // delayed or gone
+  auto& p2pComm = P2PComm::GetInstance();
+  for (const auto& p : peerList) {
+    p2pComm.SendMessage(VectorOfPeer{p}, powmessage);
```
sandipbhoir, line 381: should use this version instead
```
void P2PComm::SendMessage(const Peer& peer, const bytes& message,
                          const unsigned char& startByteType) 
```

sandipbhoir (changes requested):

sandipbhoir (approved):

chetan-zilliqa (approved):

PR #2754, opened 2021-11-29
[master] Fix multiple external libraries import in contract

chetan-zilliqa:
## Description

Below two issues are fixed.
1) Currently if more than one external library is imported then the second library is not found by the blockchain.
2) If library code has some syntax error then also it's successfully deployed. In such a case we should throw an exception when the scilla-checker reports an error.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


chetan-zilliqa:
Closing this pull request as these changes are added  in another pull request https://github.com/Zilliqa/Zilliqa/pull/2755

PR #2755, opened 2021-12-07
[master] Invoke scilla for external library call

chetan-zilliqa:
## Description

Following issues are observed and fixed during scilla external library testing.
1) Return an error if the library has any syntax error as reported by the scilla checker.
2) Allow library calls to silently pass to scilla-checker if someone calls library address. Below is the slack discussion around it.
https://zilliqa-team.slack.com/archives/GCBRD8Z1B/p1638291802234700
3) Allow multiple external libraries to import in the contract. Earlier if the user imports more than one library in the contract then blockchain says "Library not found" for the second library.
4) Distinguish between a contract and a library. Below is the related discussion for the same.
https://zilliqa-team.slack.com/archives/GF28P7PRQ/p1638247188267500?thread_ts=1638181641.263200&cid=GF28P7PRQ

Below are the test cases executed on this branch.
https://docs.google.com/spreadsheets/d/1bL1LwMXVlctcyI7RgfzISkvv050K9Wss0x85m8dTs_I/edit#gid=0

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-ARG BASE=zilliqa/scilla:v0.10.0-alpha.1
+ARG BASE=zilliqa/scilla:7652ad30
 FROM ${BASE} AS scilla
 # run a copy -L to unfold the symlinkes, and strip all exes
 RUN mkdir -p /scilla/0/bin2/ && cp -L /scilla/0/bin/* /scilla/0/bin2/ && strip /scilla/0/bin2/*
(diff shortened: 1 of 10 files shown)


docker/Dockerfile
```diff
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-ARG BASE=zilliqa/scilla:v0.10.0-alpha.1
+ARG BASE=zilliqa/scilla:7652ad30
```
chetan-zilliqa, line 17: TODO : Rollback the image version which points from the current master scilla branch to the old one.

KaustubhShamshery (approved):

sandipbhoir (approved):

PR #2768, opened 2022-01-20
[master] Remove templates from AccountStore classes

ansnunez:
## Description

(Reference: https://github.com/Zilliqa/Issues/issues/759)

This PR removes the MAP template around the AccountStore classes.

In the actual instantiations:
- Our `AccountStore` singleton uses `unordered_map<Address, Account>` along its hierarchy
- Its `m_accountStoreTemp` member uses both `unordered_map<Address, Account>` and `map<Address, Account>` along its own hierarchy, depending on whether some ordering is needed or not.

Since some ordering is needed in some parts of `AccountStoreTemp`, we can potentially just use `map` everywhere and remove the template structure altogether.

PR tested on small-scale with both payment and contract transactions.

Note that the TPP files are now renamed to CPP and removed from inclusion in the H files. As a consequence, some `include` statements need to be filled into other files.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


sandipbhoir:
@ansnunez  will this have any performance issue? Accessing underlying data will be slower when moved from unordered to ordered map?

ansnunez:
> @ansnunez will this have any performance issue? Accessing underlying data will be slower when moved from unordered to ordered map?

It comes down to BST (ordered map) vs hash table (unordered map) performance. Seeing as the ordered map one is already the one used during transaction processing during consensus (AccountStoreTemp), I foresee no impact there. The impact will be more on the AccountStore side, if ever. I can consider adding some performance tests perhaps.

src/libData/AccountData/AccountStore.h
```diff
@@ -67,16 +67,14 @@ class AccountStoreTemp : public AccountStoreSC<std::map<Address, Account>> {
...
-      Singleton<AccountStore> {
+class AccountStore : public AccountStoreTrie, Singleton<AccountStore> {
   /// instantiate of AccountStoreTemp, which is serving for the StateDelta
   /// generation
   std::unique_ptr<AccountStoreTemp> m_accountStoreTemp;
 
   /// used for states reverting
-  std::unordered_map<Address, Account> m_addressToAccountRevChanged;
-  std::unordered_map<Address, Account> m_addressToAccountRevCreated;
+  std::map<Address, Account> m_addressToAccountRevChanged;
```
KaustubhShamshery, line 76: Any reason to change from unordered to ordered? Because unordered map actually gives faster access

ansnunez: Upon testing, nodes could not reach consensus if I used unordered, so I presumed it was due to account ordering in AccountStoreTemp. Not sure if we iterate across AccountStoreTemp during consensus (is state delta ordering part of consensus at any stage?). I will look into it further after the next release to see where exactly the ordering matters and if we can somehow just switch to unordered map everywhere to preserve performance.

PR #2772, opened 2022-01-25
[master] Added flag nonisoload to allow generate state from statedeltas

sandipbhoir:
Allow isolated server to run testnet or mainnet persistence.
New paramter added : `nonisoload`

This flag should be set so that isolated server can load mainnet/testnet persistence which can also have statedeltas.

```
desc.add_options()("help,h", "Print help message")(
        "file,f", po::value<string>(&accountJsonFilePath),
        "Json file containing bootstrap accounts")(
        "port,p", po::value<uint>(&port),
        "Port to run server on {default: 5555")(
        "blocknum,b", po::value<string>(&blocknum_str),
        "Initial blocknumber {default : 1 }")(
        "time,t", po::value<uint>(&timeDelta),
        "the automatic blocktime for incrementing block number (in ms)  "
        "(Disabled by default)")(
        "load,l", po::bool_switch()->default_value(false),
        "Load from persistence folder (False by default)")(
        "nonisoload,n", po::bool_switch()->default_value(false),    <================== New Flag
        "Load is either of testnet or mainnet having state-deltas (False by default)")(
        "uuid,u", po::value<string>(&uuid),
        "unique id to be provided upon startup (can be any string)");
```

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -77,6 +77,7 @@ int main(int argc, const char* argv[]) {
   string blocknum_str{"1"};
   uint timeDelta{0};
   bool loadPersistence{false};
+  bool nonisoload{false};
   string uuid;
   try {
     po::options_description desc("Options");
@@ -93,6 +94,9 @@ int main(int argc, const char* argv[]) {
         "(Disabled by default)")(
         "load,l", po::bool_switch()->default_value(false),
         "Load from persistence folder (False by default)")(
+        "nonisoload,n", po::bool_switch()->default_value(false),
+        "Load is either of testnet or mainnet having state-deltas (False by "
+        "default)")(
         "uuid,u", po::value<string>(&uuid),
         "unique id to be provided upon startup (can be any string)");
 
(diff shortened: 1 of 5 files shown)


src/cmd/isolated_server.cpp
```diff
@@ -195,6 +202,10 @@ int main(int argc, const char* argv[]) {
       cout << "Server listening on " << port << endl;
     }
 
+    if (timeDelta > 0) {
```
KaustubhShamshery, line 205: Don't need it here as called here https://github.com/Zilliqa/Zilliqa/blob/a5890ef13b5b0115912f4865d12a9f011732565e/src/libServer/IsolatedServer.cpp#L111-L140

sandipbhoir: But then block will be incremented and generated even if timedelta = 0, Is this expected behavior?
I thought we wanted to have control over block generation for some special case.
If not, then can remove it.

KaustubhShamshery (approved):

ansnunez (approved):

PR #2798, opened 2022-04-29
[master] Refactoring: split longer methods into short ones, remove code duplication

valeryz:
## Description

Remove code duplication between `IsolatedServer` and `LookupServer`. This is needed to prepare the codebase for changes related to EVM integration. EVM integration will have to changes codepaths related to transaction creation, and it would be better to change in fewer places.

Other changes: added m_priority field to Transaction class to avoid carrying it around. It is not part of `TransactionCoreInfo`, does not affect the signature, but because it is mentioned in the CreateTransaction API and affects how transaction is processed, it is better for the flag to live inside the class.

## Backward Compatibility

## Review Suggestion
Turned out a rather large change. I managed to extract many methods from long functions, and move them up the hierarchy from `IsolatedServer` to `LookupServer`

### Implementation
- [x] **ready for review**

Existing tests should cover the change, as it only a refactoring.


KaustubhShamshery:
Two prima facie observarions:

1. Try not use to raw pointers, I think value can be passed by reference or if it is really necessary use smart pointers
2. Might be useful to do a https://zilliqa-docs.atlassian.net/wiki/spaces/ZILLIQA/pages/588546391/Network+Txn+Test+Giant.sclla coz it does affect transaction handling.

valeryz:
Ok, running the network test did indeed help isolate a bug, fixed in 4fb5fec.  @KaustubhShamshery PTAL

CSideSteve (commented):
Looks fine to me

CSideSteve (approved):
Looks fine to me

PR #2803, opened 2022-05-31
[evm_support_main] Steve evm branch work

valeryz:


src/libUtils/EvmUtils.cpp
```diff
@@ -98,3 +102,59 @@ Json::Value EvmUtils::GetCallContractJson(const RunnerDetails& details) {
 
   return arr_ret;
 }
+
+bool EvmUtils::EvmUpdateContractStateAndAccount(
```
valeryz, line 106: This is clean and nice, thank you!

Looking forward to seeing account deletion and handling reset_storage.  Cannot really test without reset_storage - I expect it could be frequent.

src/libData/AccountData/AccountStoreSC.h
```diff
@@ -196,10 +198,10 @@ class AccountStoreSC : public AccountStoreBase<MAP> {
                          const boost::multiprecision::uint128_t& balance,
                          bool& ret, TransactionReceipt& receipt);
 
-  void InvokeEvmInterpreter(INVOKE_TYPE invoke_type, RunnerDetails& details,
-                            const uint32_t& version, bool& ret,
-                            TransactionReceipt& receipt,
-                            struct EvmReturn& result);
+  uint64_t InvokeEvmInterpreter(Account* account, INVOKE_TYPE invoke_type,
```
valeryz, line 201: is this the contract account or the caller account?   Might reflect in the name.

src/libUtils/EvmCallParameters.h
```diff
@@ -15,18 +15,20 @@
...
+// input parameters to Json call
+
+struct EvmCallParameters {
   std::string m_from;
   std::string m_to;
   std::string m_code;
   std::string m_data;
-  const uint64_t m_available_gas;
+  uint64_t m_available_gas;
   const boost::multiprecision::uint128_t m_balance;
```
valeryz, line 31: nit: either remove this const, or make other members const too.  I don't think part of this struct could be immutable, and part not.

src/libUtils/EvmUtils.cpp
CSideSteve: Account deletion is being looked at by @KaustubhShamshery , handle reset we are working on right now, doesn't exist in current usage, need to make sure it travels full depth through the objects temps,cached etc.

src/libUtils/EvmCallParameters.h
CSideSteve: The old code used to use a parameter as both an in/out this has been changed so the are input parameters to the function, also balance is not currently passed through it is hard coded to send "00" for this field, currently the evm interface doesn't handle anything other than the "00" needs to be addressed,

src/libData/AccountData/AccountStoreSC.h
CSideSteve: For Evm it is the contact account, it has been renamed to reflect this, will see that on next commit

src/libUtils/EvmCallParameters.h
CSideSteve: This is actually Steve who wishes to merge the changes onto evm_support_main, valeriy raised the PR on my behalf.

CSideSteve:
Closing this pull request as rasing a new pull request on my own behalf for a similar review.

PR #2806, opened 2022-06-06
Add Eth-style transactions (v1)

n-hutton:
Initial implementation of tx signing using eth-style.

TL;DR: Transactions with version number 2 are verified with ECDSA signatures. Additionally, their public key is converted to an Eth style address when it is being used. Currently, the signature is of an empty string as I don't yet have a good comparison (typed data v4).

Technical details/findings:

-Addresses of ethereum and zil differ in that zil is taking the hash of the public key using sha256 while ethereum is using keccak.

- Confusingly, ethereum uses keccak that is NOT sha3 for legacy reasons.

- Elliptic curve public keys are essentially an x and a y coordinate. This means that you can 'compress' the key by providing the x coordinate and whether the y-coordinate is even/odd. Zil uses compressed keys while ethereum and its toolchains do not. This means there is some time spent converting between compressed and uncompressed.

- The TX holds a pubkey object which must have a compressed version

- There is a minor issue that when you don't have a TX/version number, you are forced to assume that pubkeys correspond to Zil style addresses. This could be for example in the genesis file.

- There is a possible security issue/increased collision probability in mixing addresses generated using two different hash functions. Needs some review/thought.

- Further refactoring is needed to add a field to the TX that will signify the signing scheme when using version 2

- Ethereum crypto header/library needs to be put in the correct place (Zilliqa/cryptoutils ? src/libCrypto ?)

- Before going to production it should be audited

- There is a mutex in the schnorr signature verification scheme. Something about openssl not handling multithreading? Need to look into this. Schnorr.cpp:398

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -4,7 +4,7 @@
     <general>
         <DEBUG_LEVEL>3</DEBUG_LEVEL>
         <ENABLE_DO_REJOIN>false</ENABLE_DO_REJOIN>
-        <LOOKUP_NODE_MODE>true</LOOKUP_NODE_MODE>
+        <LOOKUP_NODE_MODE>false</LOOKUP_NODE_MODE>
         <MAX_ENTRIES_FOR_DIAGNOSTIC_DATA>25</MAX_ENTRIES_FOR_DIAGNOSTIC_DATA>
         <CHAIN_ID>1</CHAIN_ID>
         <NETWORK_ID>3</NETWORK_ID>
@@ -284,7 +284,7 @@
     </recovery>
     <smart_contract>
         <ENABLE_SC>true</ENABLE_SC>
-        <SCILLA_ROOT>/home/nathan/repos/Zilliqa/scilla</SCILLA_ROOT>
+        <SCILLA_ROOT>/scilla</SCILLA_ROOT>
         <SCILLA_CHECKER>bin/scilla-checker</SCILLA_CHECKER>
         <SCILLA_BINARY>bin/scilla-runner</SCILLA_BINARY>
         <SCILLA_FILES>scilla_files</SCILLA_FILES>
@@ -304,12 +304,6 @@
         <DISABLE_SCILLA_LIB>false</DISABLE_SCILLA_LIB>
         <SCILLA_SERVER_PENDING_IN_MS>1500</SCILLA_SERVER_PENDING_IN_MS>
         <SCILLA_SERVER_LOOP_WAIT_MICROSECONDS>10</SCILLA_SERVER_LOOP_WAIT_MICROSECONDS>
-        ## Steps to Enable EVM for a run this temporarily replaces the Scilla Interpreter
-        <ENABLE_EVM>true</ENABLE_EVM>
-        <EVM_ROOT>evm-ds</EVM_ROOT>
-        <EVM_SERVER_SOCKET_PATH>/tmp/evm-server.sock</EVM_SERVER_SOCKET_PATH>
-        <EVM_SERVER_BINARY>evm-ds</EVM_SERVER_BINARY>
-        <ENABLE_EVM_MULTI_VERSION>false</ENABLE_EVM_MULTI_VERSION>
     </smart_contract>
     <tests>
         <ENABLE_CHECK_PERFORMANCE_LOG>false</ENABLE_CHECK_PERFORMANCE_LOG>


src/libCrypto/EthCrypto.h
```diff
@@ -0,0 +1,187 @@
...
+// Prefix signed txs in Ethereum with Keccak256("\x19Ethereum Signed
+// Message:\n32" + Keccak256(message))
+constexpr uint8_t prelude[] = {25,  69,  116, 104, 101, 114, 101, 117, 109,
+                               32,  83,  105, 103, 110, 101, 100, 32,  77,
+                               101, 115, 115, 97,  103, 101, 58,  10,  48};
+
+constexpr unsigned int UNCOMPRESSED_SIGNATURE_SIZE = 65;
+
+// https://stackoverflow.com/questions/57385412/
+inline void SetOpensslSignature(const std::string& sSignatureInHex,
```
valeryz, line 43: Inlining such long methods, and placing them in .h files?   Please move most of this file to .cpp.  If we don't to this, out build times will keep growing and become completely unmanageable.

valeryz (changes requested):
Hey @n-hutton could you address the .h / .cpp problem before we review further?
It will be very hard to review otherwise once we have some comments, and you then do this later.

src/libCrypto/EthCrypto.h
```diff
@@ -0,0 +1,187 @@
+/*
+ * Copyright (C) 2019 Zilliqa
```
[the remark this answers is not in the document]
n-hutton: :+1:

cmake/FindLevelDB.cmake
```diff
@@ -23,5 +23,5 @@ if (NOT BUILD_SHARED_LIBS AND APPLE)
 endif()
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(leveldb DEFAULT_MSG
+find_package_handle_standard_args(LevelDB DEFAULT_MSG
```
n-hutton, line 26: fixes warnings during build

src/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ add_subdirectory (common)
 add_subdirectory (depends)
 add_subdirectory (diagnostic)
 add_subdirectory (libConsensus)
+add_subdirectory (libCrypto)
```
n-hutton, line 6: crypto now a library so as to have symbols in cpp file

src/libServer/JSONConversion.cpp
```diff
@@ -421,14 +435,9 @@ Address JSONConversion::checkJsonGetEthCall(const Json::Value& _json) {
...
-  string lower_case_addr;
-  if (!AddressChecksum::VerifyChecksumAddress(_json["toAddr"].asString(),
-                                              lower_case_addr)) {
-    LOG_GENERAL(INFO,
-                "To Address checksum wrong " << _json["toAddr"].asString());
-    throw jsonrpc::JsonRpcException(Server::RPC_INVALID_PARAMETER,
-                                    "To Addr checksum wrong");
-  }
+
+  string lower_case_addr = boost::to_lower_copy(_json["toAddr"].asString());
```
n-hutton, line 439: @KaustubhShamshery is it ok to forgo the address checksum check for this call? It seems that it's not technically neccessary

KaustubhShamshery: Well might wanna check if the address contains "0x" and remove it if does not

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+  } else if (sPubKeyString[1] == '3') {
+    y_chooser_bit = 1;
+  } else {
+    std::cout << "Received badly set signature bit! Should be 2 or 3 and got: "
+              << sPubKeyString[1] << std::endl;
+  }
+
+  // Don't want the first byte
+  if (!BN_hex2bn(&gx_ptr, sPubKeyString + 2)) {
+    std::cout << "***** Error getting to x binary format" << std::endl;
```
valeryz, line 91: I'm not sure that logging to std::cout is a good error handling strategy.

src/libServer/JSONConversion.cpp
n-hutton: Do you mean if it contains 0x you should strip it off, right?

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+// Message:\n32" + Keccak256(message))
+constexpr uint8_t prelude[] = {25,  69,  116, 104, 101, 114, 101, 117, 109,
+                               32,  83,  105, 103, 110, 101, 100, 32,  77,
+                               101, 115, 115, 97,  103, 101, 58,  10,  48};
+
+// https://stackoverflow.com/questions/57385412/
+void SetOpensslSignature(const std::string& sSignatureInHex, ECDSA_SIG* pSign) {
+  // Openssl uses raw pointers and macros for freeing but can force it into
+  // smart pointers methodology
+  std::unique_ptr<BIGNUM, std::function<void(BIGNUM*)>> rr(
```
valeryz, line 47: As much as I like unique_ptr, if the usage of this pointer is only within what otherwise would be a 5 line function, I see no harm from just using `BN_free()` at the end.  Because it will not throw. Much less cognitive load in pure C style.  I don't insist though. (To me it just looks more modern, but solves a non-existing problem).  Wouldn't something like this be easier on the reader?

```
std::string sSignatureR =
      sSignatureInHex.substr(0, sSignatureInHex.size() / 2);
  std::string sSignatureS = sSignatureInHex.substr(sSignatureInHex.size() / 2);
BIGNUM *r = nullptr, *s = nullptr;   // Well known r, s numbers
BN_hex2n(&r, sSignatureR.c_str());
BN_hex2n(&s, sSignatureS.c_str());
ECDSA_SIG_set0(pSign, r, s);
BN_free(r);
BN_free(s);
```

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+  BIGNUM* gx_ptr = gx.get();
+
+  EC_KEY_set_asn1_flag(pKey, OPENSSL_EC_NAMED_CURVE);
+
+  // From
+  // https://www.oreilly.com/library/view/mastering-ethereum/9781491971932/ch04.html
+  // The first byte indicates whether the y coordinate is odd or even
+  int y_chooser_bit = 0;
+
+  if (sPubKeyString[1] == '2') {
```
valeryz, line 81: where did `sPubKeyString[0]` go?  Why are we starting with `[1]`?

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  // Don't want the first byte
+  if (!BN_hex2bn(&gx_ptr, sPubKeyString + 2)) {
+    std::cout << "***** Error getting to x binary format" << std::endl;
+  }
+
+  // Create a new curve group
+  // Refactor: probably can have static/constexpr curve, need to check
+  std::unique_ptr<EC_GROUP, std::function<void(EC_GROUP*)>> curve_group(
```
valeryz, line 98: Ok maybe worth having it here, but maybe a `using` directrive or a typedef for this somewhere, so that you don't have to repeat the lambda. Or a few unique_ptr specializations etc.

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <openssl/ec.h>  // for EC_GROUP_new_by_curve_name, EC_GROUP_free, EC_KEY_new, EC_KEY_set_group, EC_KEY_generate_key, EC_KEY_free
+#include <openssl/obj_mac.h>  // for NID_secp192k1
+#include <openssl/sha.h>      //for SHA512_DIGEST_LENGTH
+#include <ethash/keccak.hpp>
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <string>
+
+using namespace std;
```
valeryz, line 30: Most  style guides I saw forbid `using namespace` even in .cpp files.  You could probably use `using std::cout` etc.  Or just remove it as you are using `std::` qualifiers everywhere explicitly anyways.

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+  if (EC_KEY_check_key(pKey) == 1) {
+    return true;
+  } else {
+    std::cout << "ec key invalid " << std::endl;
+    return false;
+  }
+}
+
+bool VerifyEcdsaSecp256k1(const std::string&, const std::string& sSignature,
```
valeryz, line 121: please document the 1st argument (either /* randomNumber */, or use one of the ways to tame the args unused  warnings.

Why is it unused btw, and why do you need it then?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -86,8 +86,6 @@ void AccountStore::Init() {
 }
 
 void AccountStore::InitSoft() {
-  LOG_MARKER();
```
valeryz, line 89: Are you sure removing those LOG_MARKERs is safe?  Surely don't have to do this to bring Eth-style transactions in.

valeryz (commented):
Thanks for this PR!  Many many good things in it, please don't let my critical points make an impression I'm not happy with it - I am super happy about this PR overall (still a few things to address though).

src/libCrypto/EthCrypto.cpp
```diff
@@ -0,0 +1,183 @@
...
+using namespace std;
+
+// Inspiration from:
+// https://stackoverflow.com/questions/10906524
+// https://stackoverflow.com/questions/57385412/
+// https://medium.com/mycrypto/the-magic-of-digital-signatures-on-ethereum-98fe184dc9c7
+
+// Prefix signed txs in Ethereum with Keccak256("\x19Ethereum Signed
+// Message:\n32" + Keccak256(message))
+constexpr uint8_t prelude[] = {25,  69,  116, 104, 101, 114, 101, 117, 109,
```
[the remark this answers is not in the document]
n-hutton: I would suggest that we leave it as uint8_t because it contains non-ascii values which can cause issues around when there are x0 in it etc. Happy to move it into the one function

src/libCrypto/EthCrypto.cpp
n-hutton: There is always a way your code could throw, even just when allocating memory :) I would opt to leave it in although I agree that it looks horrible. Perhaps we can get a second opinion from @CSideSteve @KaustubhShamshery or @moboware

src/libCrypto/EthCrypto.cpp
n-hutton: That's right - I have added a check for this

src/libCrypto/EthCrypto.cpp
n-hutton: sounds great

src/libCrypto/EthCrypto.cpp
n-hutton: That's fair. I will remove it - although there are 100 other files in the codebase that already do this!

src/libCrypto/EthCrypto.cpp
n-hutton: Sure! It is unused because it will be the data going into keccak that follows the data (typed data / EIP-712)

src/libCrypto/EthCrypto.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#ifndef ZILLIQA_SRC_LIBCRYPTO_ETHCRYPTO_H_
+#define ZILLIQA_SRC_LIBCRYPTO_ETHCRYPTO_H_
+
+#include <openssl/ecdsa.h>  // for ECDSA_do_sign, ECDSA_do_verify
+
+#include <string>
+
+constexpr unsigned int UNCOMPRESSED_SIGNATURE_SIZE = 65;
```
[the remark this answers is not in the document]
n-hutton: Nope, fair point. Internal only functions now hidden, functions correctly cased

src/libCrypto/EthCrypto.h
```diff
@@ -0,0 +1,45 @@
...
+
+bool VerifyEcdsaSecp256k1(const std::string& sRandomNumber,
+                          const std::string& sSignature,
+                          const std::string& sDevicePubKeyInHex);
+
+// Given a hex string representing the pubkey (secp256k1), return the hex
+// representation of the pubkey in uncompressed format.
+// The input will have the '02' prefix, and the output will have the '04' prefix
+// per the 'Standards for Efficient Cryptography' specification
+std::string toUncompressedPubKey(const std::string& pubKey);
```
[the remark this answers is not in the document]
n-hutton: :+1:

src/libData/AccountData/Account.h
```diff
@@ -195,6 +195,8 @@ class Account : public AccountBase {
   /// Computes an account address from a specified PubKey.
   static Address GetAddressFromPublicKey(const PubKey& pubKey);
 
+  static Address GetAddressFromPublicKeyEth(const PubKey& pubKey);
```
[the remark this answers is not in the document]
n-hutton: Agree, I think add it to the TODO refactor list :)

src/libData/AccountData/AccountStore.cpp
n-hutton: I did not consider that this might not be safe - I thought that they just print 'BEG' and 'END' according to its lifetime (and so only useful if you are debugging that function, it is just generic cleanup). In what way might it not be safe?

src/libData/AccountData/Transaction.cpp
```diff
@@ -171,6 +178,43 @@ const bytes& Transaction::GetData() const { return m_coreInfo.data; }
...
+  sigString = sigString.substr(2);
+  pubKeyStr = pubKeyStr.substr(2);
+
+  return VerifyEcdsaSecp256k1("", sigString, pubKeyStr);
+}
+
+// Function to return whether the TX is signed
+bool Transaction::IsSigned() const {
+  // Use the version number to tell which signature scheme it is using
+  if (GetVersion() == 65538) {
```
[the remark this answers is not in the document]
n-hutton: :+1:

src/libData/AccountData/AccountStore.cpp
KaustubhShamshery: Agree useful to just keep the logs

src/libCrypto/EthCrypto.cpp
valeryz: ok as I said I don't insist here.

valeryz (approved):

constants.xml
```diff
@@ -304,12 +304,6 @@
         <DISABLE_SCILLA_LIB>false</DISABLE_SCILLA_LIB>
         <SCILLA_SERVER_PENDING_IN_MS>1500</SCILLA_SERVER_PENDING_IN_MS>
         <SCILLA_SERVER_LOOP_WAIT_MICROSECONDS>10</SCILLA_SERVER_LOOP_WAIT_MICROSECONDS>
-        ## Steps to Enable EVM for a run this temporarily replaces the Scilla Interpreter
-        <ENABLE_EVM>true</ENABLE_EVM>
```
valeryz: don't you have to ENABLE_EVM section intact, just set it to false by default?

src/libServer/JSONConversion.cpp
valeryz: Well, @n-hutton did you change this as advised by Kaustubh? It broke the tests (which don't run on CI yet :))

n-hutton: no - it should be fixed in this PR though https://github.com/Zilliqa/Zilliqa/pull/2828

PR #2810, opened 2022-06-09
[evm_support_main] Clean up constants related to evm-ds.

valeryz:
Set EVM_SERVER_BINARY to the full path of the binary. The default is /usr/local/bin/evm-ds.


CSideSteve:
Merged

PR #2807, opened 2022-06-07
[master] Evm support main

CSideSteve:
This Update provides the functionality required to host an evm facility within the Node.
## Description
The update provides the ability to create evm transactions within the node and make the evm smart contracts available to core development for testing and further development of evm facilities.
## Backward Compatibility

- [X] This is not a breaking change
- [ ] This is a breaking change

## Review Suggestion
Make sure all tests run fine

## Review Suggestion

pull it , build it and run local tests to verify that release is complete
### Implementation

- [x] **ready for review**
see new section in IsolatedServer setup to see new EVM configuration parameters
### Integration Test (Core Team)

- [x] local machine test (commit: 3187f63d266) 

- [x] small-scale cloud test (commit: 3187f63d266)

PR #2811, opened 2022-06-14
[evm_support_main] add resetStorage

KaustubhShamshery:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -298,6 +298,7 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       // generate address for new contract account
       Address contractAddress =
           Account::GetAddressForContract(fromAddr, fromAccount->GetNonce());
+      toAddr = contractAddress;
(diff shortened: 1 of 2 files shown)


valeryz (commented):
Seems okay, but needs to apply on top latest Steve's changes.

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -298,6 +298,7 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       // generate address for new contract account
       Address contractAddress =
           Account::GetAddressForContract(fromAddr, fromAccount->GetNonce());
+      toAddr = contractAddress;
```
valeryz, line 301: Thanks, well spotted.

Could we resolve this by renaming the other variable down below to `contractAddress`.  toAddr should mean the address to which the transaction was sent, which would be zero in this case.

valeryz (approved):

KaustubhShamshery:
Already in #2807

PR #2812, opened 2022-06-16
[master] Add support for missing queries from EVM

valeryz:
## Description

Add infrastructure and methods for blockchain data to be available for EVM contracts.

Support for queries like "BLOCKNUMBER", "TIMESTAMP", "CHAINID", as well as moving this blockchain metadata into a separate class cherry picked from the previous master.

This is backported from the old master (master-05-2022 branch)

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libData/AccountData/AccountStoreSC.h
```diff
@@ -64,6 +64,9 @@ class AccountStoreSC : public AccountStoreBase<MAP> {
   /// the blocknum while executing each txn
   uint64_t m_curBlockNum{0};
 
+  /// the DS blocknum
+  uint64_t m_curDSBlockNum{0};
```
KaustubhShamshery, line 68: variable is not really set anywhere. Is it just a placeholder?

valeryz: indeed, the reason is that Zilliqa has more than one blocknumber - and EVM has just one NUMBER opcode ( https://www.ethervm.io/#43), because one Ethereum block type.

So added both block numbers, but only one is really used. Until we decide how to support both in the EVM. 

So really a placeholder for now.

valeryz: It is zero by default, so should be safe regardless.

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -1694,9 +1701,12 @@ bool AccountStoreSC<MAP>::ParseCallContractJsonOutput(
...
 
-      // prepare IPC with the recipient contract address
-      m_scillaIPCServer->setContractAddressVerRoot(recipient, scilla_version,
-                                                   account->GetStorageRoot());
+      // prepare IPC with current blockchain info provider.
+      auto sbcip1 = std::make_unique<ScillaBCInfo>(
+          m_curBlockNum, m_curDSBlockNum, m_originAddr, recipient,
+          account->GetStorageRoot(), scilla_version);
+      m_scillaIPCServer->setBCInfoProvider(std::move(sbcip1));
+
```
n-hutton, line 1709: So, is my understanding correct: this is the first PR towards enabling the EVM to access the block info (so it does not currently), but scilla can do so already?

I understand why this is added in UpdateAccounts, but not ParseCallContractJsonOutput function (and why it is set twice - does the storage root change between calls?). My reading of the code is that ParseCallContractJsonOutput is called once the VM is finished executing, so why does it need the block info etc. to be attached?

n-hutton (commented):
I'd find it valuable to have a chat about this code so I can keep updated - perhaps tomorrow standup. I would also suggest slightly more comments like in AccountStoreSC where the functions are very long

src/libData/AccountData/AccountStoreSC.tpp
valeryz: EVM needs more blockchain info than Scilla as of 8.2.  Scillla in 8.3 .. old master needs more hence they added this BC Info provider, which I just extended.

Spot on re ParseContractJsonOutput - it just really rebased the change first, and changes related to Scilla also got in.
I actually overlooked this. But it indeed must get in, because if we change the way this gets handled for fetchBlockchainInfo method, it also affects Scilla, so the object must be in place.

Why does `ParseCallContractJsonOutput` call Scilla again? Well, because there is recursion there - Scilla must be able to call other smart contracts.  In fact, I think this recursion is not needed.  Scilla only allows tail calls to other smart contracts - meaning this should be a loop rather than recursion.  File https://zilliqa-jira.atlassian.net/browse/ZIL-4650

KaustubhShamshery (approved):

PR #2813, opened 2022-06-16
[master] Fix rust building on devnet

KaustubhShamshery:
## Description
Rust needs to be build on an 18.04 enviornment otherwise it crashes when being tried to be used.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,11 +20,20 @@ FROM ${BASE} AS scilla
 RUN mkdir -p /scilla/0/bin2/ && cp -L /scilla/0/bin/* /scilla/0/bin2/ && strip /scilla/0/bin2/*
 
 FROM ubuntu:18.04 as rust_builder
-RUN apt-get update && apt-get install -y protobuf-compiler
+RUN apt-get update && apt-get install -y protobuf-compiler && \
+apt-get install -y git && \
+apt-get install -y curl && \
+apt-get install -y build-essential && \
+apt-get install -y pkg-config && \
+apt-get install -y libssl-dev && \
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o install_script.sh && \
+sh install_script.sh -y 
+ENV PATH="/root/.cargo/bin:${PATH}"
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -19,14 +19,22 @@ FROM ${BASE} AS scilla
 # run a copy -L to unfold the symlinkes, and strip all exes
 RUN mkdir -p /scilla/0/bin2/ && cp -L /scilla/0/bin/* /scilla/0/bin2/ && strip /scilla/0/bin2/*
 
-FROM rust:latest as rust_builder
-RUN rustup component add clippy rustfmt
-RUN apt-get update && apt-get install -y protobuf-compiler
+FROM ubuntu:18.04 as rust_builder
```
valeryz, line 22: What's the original problem and why we should build on ubuntu 18.04 base image instead of one coming from Rust?

KaustubhShamshery:
```
root@ip-10-3-37-186:/run/zilliqa# ldd /usr/local/bin/evm-ds                     
/usr/local/bin/evm-ds: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /usr/local/bin/evm-ds)
        linux-vdso.so.1 (0x00007ffd4815b000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f653cf55000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f653cd36000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f653c998000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f653c794000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f653c3a3000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f653d562000)

valeryz: Got it, thanks!

valeryz (approved):

PR #2814, opened 2022-06-20
[release-8.2] Fix CPU starvation during node start

chetan-zilliqa:
## Description

Introduce sleep or extend sleep duration for long-running loops in code to prevent CPU starvation.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/ProcessStats.cpp
+++ b/src/libUtils/ProcessStats.cpp
@@ -14,15 +14,32 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-#include "MemoryStats.h"
+#include "ProcessStats.h"
 #include <errno.h>
+#include <sys/times.h>
 #include <cstring>
 #include "libUtils/Logger.h"
 #include "sys/sysinfo.h"
 #include "sys/types.h"
 
 using namespace std;
-int parseLine(char* line) {
+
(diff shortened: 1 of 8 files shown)


KaustubhShamshery (approved):

src/libUtils/ProcessStats.cpp
```diff
@@ -105,4 +119,44 @@ int64_t DisplayPhysicalMemoryStats(const string& str, int64_t startMem) {
     LOG_GENERAL(INFO, "PM diff = " << diff << " " << str);
   }
   return processPhysMemUsedMB;
-}
\ No newline at end of file
+}
```
valeryz, line 122: I'm a little suspicious about in-process collection of memory and cpu stats about the process itself. 

Why isn't it done with some form of probes that are then exported to a centralized metrics collection. We had discussed OpenTelemetry etc, - those solutions already have probes set up at hosts, that can collect this information from all nodes, aggregated any way w/o having to write and maintain this code.

valeryz (approved):
Happy to additionally LGTM this, since it brings the much needed stability and peace of mind!

However, we should address the underlying issues. Taming the process with sleeps seems suboptimal to me. Collecting metrics in the process itself, instead of using standard node exporters (why do we have a zilliqa one?) is also adding to the maintenance burden in my view.

PR #2816, opened 2022-06-20
[master] Process logs a unformatted data, just supply char array of contents

CSideSteve:
## Description
Treats the contents of the logs field as raw data and simply passes data back throught the standard receipt mechanism.
review and grant permission to merge

## Backward Compatibility

- [ X] This is not a breaking change
- [ ] This is a breaking change

### Implementation

- [ ] **ready for review**

### Integration Test (Core Team)
merge or pull from evm_log_fix branch

- [ ] local machine test

- [ ] small-scale cloud test

PR #2817, opened 2022-06-20
[master] Get eth_call analogue working

KaustubhShamshery:
## Description
1. GetEthCall for viewing calls
2. Handle Logs being added in the receipt

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStore.cpp
+++ b/src/libData/AccountData/AccountStore.cpp
@@ -62,7 +62,9 @@ AccountStore::AccountStore() {
     }
   }
   // EVM required to run on Lookup nodes too for view calls
-  if (ENABLE_EVM) EvmClient::GetInstance().Init();
+  if (ENABLE_EVM) {
+    EvmClient::GetInstance().Init();
+  }
 }
 
 AccountStore::~AccountStore() {
--- a/src/libData/AccountData/AccountStore.cpp
+++ b/src/libData/AccountData/AccountStore.cpp
@@ -49,7 +49,6 @@ AccountStore::AccountStore() {
         make_shared<ScillaIPCServer>(*m_scillaIPCServerConnector);
 
     ScillaClient::GetInstance().Init();
-    if (ENABLE_EVM) EvmClient::GetInstance().Init();
 
     if (m_scillaIPCServer == nullptr) {
       LOG_GENERAL(WARNING, "m_scillaIPCServer NULL");
@@ -62,6 +61,8 @@ AccountStore::AccountStore() {
       }
     }
   }
+  // EVM required to run on Lookup nodes too for view calls
+  if (ENABLE_EVM) EvmClient::GetInstance().Init();
 }
 
 AccountStore::~AccountStore() {
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -177,17 +177,21 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
   ret = evmReturnValues.isSuccess() ? ret : false;
 
   if (!evmReturnValues.Logs().empty()) {
-    LOG_GENERAL(WARNING, evmReturnValues.Logs());
-    Json::Value _json;
-    Json::Reader _reader;
+    Json::Value _json = Json::arrayValue;
     bool success = true;
-    try {
-      success = _reader.parse(evmReturnValues.Logs(), _json);
-    } catch (std::exception& e) {
-      LOG_GENERAL(WARNING, "Caught Exception: " << e.what());
-    }
-    if (!success) {
-      LOG_GENERAL(WARNING, "Parsing json unsuccessuful");
+    for (const auto& lval : evmReturnValues.Logs()) {
+      LOG_GENERAL(INFO, "Logs: " << lval);
+      Json::Reader _reader;
+      Json::Value tmp;
+      try {
+        success = _reader.parse(lval, tmp);
+      } catch (std::exception& e) {
+        LOG_GENERAL(WARNING, "Exception: " << e.what());
+      }
+      if (!success) {
+        LOG_GENERAL(WARNING, "Parsing json unsuccessful " << lval)
+      }
+      _json.append(tmp);
     }
     receipt.AddJsonEntry(_json);
   }
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -128,8 +128,7 @@ template <class MAP>
 uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
     Account* contractAccount, INVOKE_TYPE invoke_type,
     EvmCallParameters& params, const uint32_t& version, bool& ret,
-    TransactionReceipt& receipt) {
-  evmproj::CallResponse evmReturnValues;
+    TransactionReceipt& receipt, evmproj::CallResponse& evmReturnValues) {
   uint64_t gas = params.m_available_gas;
 
   bool call_already_finished = false;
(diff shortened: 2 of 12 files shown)


CSideSteve (approved):
Looking good.

KaustubhShamshery:
@CSideSteve I have found an another way of dealing with the log exception issue, it makes more sense to me that the whole log is a string so that it is easily convertible to a JSON again. If you agree then no need for #2816

src/libServer/LookupServer.cpp
```diff
@@ -876,7 +881,58 @@ Json::Value LookupServer::GetBalance(const string& address) {
...
+    if (contractAccount == nullptr) {
+      throw JsonRpcException(RPC_INVALID_PARAMS, "Account does not exist");
+    }
+    code = contractAccount->GetCode();
+  }
+  Address fromAddr;
+  string result;
+  uint64_t amount{0};
+  try {
+    if (_json.isMember("fromAddr")) {
```
valeryz, line 907: fromAddr makes sense to me, but for some reason CreateTransaction uses pubkey, and derives the address from it.  I wonder whether we should do the same here or not.

src/libServer/LookupServer.cpp
```diff
@@ -876,7 +881,58 @@ Json::Value LookupServer::GetBalance(const string& address) {
...
+      fromAddr = Address(_json["fromAddr"].asString());
+    }
+    if (_json.isMember("amount")) {
+      const auto& amount_str = _json["amount"].asString();
+      amount = strtoull(amount_str.c_str(), NULL, 0);
+    }
+    if (_json.isMember("gasLimit")) {
+      const auto& gasLimit_str = _json["gasLimit"].asString();
+      gasRemained = strtoull(gasLimit_str.c_str(), NULL, 0);
```
valeryz, line 916: Please see the comment above re gasLimit.  Don't let the user override the maximum by setting this to a huge number.

src/libUtils/EvmJsonResponse.h
```diff
@@ -77,13 +77,13 @@ struct CallResponse {
...
   }
-  const std::vector<std::string>& Logs() const { return m_logs; }
+  const std::string Logs() const { return m_logs; }
   const std::string& ExitReason() const { return m_exitReason; }
   uint64_t Gas() const { return m_gasRemaining; }
   const std::string& ReturnedBytes() const { return m_return; }
   bool isSuccess();
   std::vector<std::shared_ptr<ApplyInstructions>> m_apply;
-  std::vector<std::string> m_logs;
+  std::string m_logs;
```
valeryz, line 86: I don't understand the reason for this change.  

This is meant to be a separate entry for every LOGx opcode. Why do we merge them into one string?

valeryz (approved):
Overall approved, but please explain and/or address the logs change from vector to just a single string.

(Fine to merge and send the followup changes, like gasLimit etc  in next PRs)

valeryz (approved):

src/libUtils/EvmJsonResponse.h
KaustubhShamshery: actually the logs recieved are a JSON array, and the output required to be put in the receipt is also a JSON. So, it is just easier to convert a whole string to JSON. Otherwise would have to go element by element and convert to JSON.

src/libServer/LookupServer.cpp
KaustubhShamshery: Noted

src/libServer/LookupServer.cpp
KaustubhShamshery: I used the eth_call config for this purpose however happy to change it

valeryz: I am not sure which way is better, just thought that mimicking `CreateTransaction` would make more sense for consistency of the Zilliqa API

src/libUtils/EvmJsonResponse.h
valeryz: I'm surprised the logs received is a JSON array, I believe this is a reminiscent of JSONRPC mechanism being used for communication.  If we decide to change it (and we did plan to move to gRPC, unless we also decide to embed it), and it won't be JSONRPC anymore, it will break.

KaustubhShamshery: ya but it will break either way i.e. vector<string> or string as both expect a JSON now.

valeryz: Well, that's because `EvmResult` in Rust is

```rust
#[derive(serde::Serialize)]
pub struct EvmResult {
    exit_reason: evm::ExitReason,
    return_value: String,
    apply: Vec<DirtyState>,
    logs: Vec<ethereum::Log>,
    remaining_gas: u64,
}
```

And Vec<ethereum::Log> in JSONRPC ends up encoded, as, well, a JSON array :)  it is a coincidence of Vec being encoded that way.

I'd recommend reverting. I guess for the code to convey the meaning is better than just a small convenience of parsing in one place.  Or at least leave a big fat comment about this.

KaustubhShamshery: Got it, will revert.

src/libServer/LookupServer.cpp
```diff
@@ -913,7 +912,8 @@ string LookupServer::GetEthCall(const Json::Value& _json) {
     }
     if (_json.isMember("gasLimit")) {
       const auto& gasLimit_str = _json["gasLimit"].asString();
-      gasRemained = strtoull(gasLimit_str.c_str(), NULL, 0);
+      gasRemained =
+          min(gasRemained, (uint64_t)strtoull(gasLimit_str.c_str(), NULL, 0));
```
valeryz, line 916: Why not prefer `std::stol` ?   This one will return `ULONG_MAX` on error, and mess with `errno`.  Because of the `min` it will end up fine, but it is soo non-obvious...

valeryz (approved):

src/libData/AccountData/AccountStore.cpp
```diff
@@ -62,6 +61,8 @@ AccountStore::AccountStore() {
       }
     }
   }
+  // EVM required to run on Lookup nodes too for view calls
+  if (ENABLE_EVM) EvmClient::GetInstance().Init();
```
moboware, line 65: please add {} after the if

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -728,7 +758,7 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       isScilla = !EvmUtils::isEvm(contractAccount->GetCode());
       bool is_library;
       uint32_t scilla_version;
```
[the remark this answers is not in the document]
KaustubhShamshery: Legacy code

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -165,18 +160,40 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
...
   if (not evmReturnValues.isSuccess()) {
     LOG_GENERAL(WARNING, evmReturnValues.ExitReason());
   }
   // switch ret to reflect our overall success
   ret = evmReturnValues.isSuccess() ? ret : false;
 
-  if (evmReturnValues.Logs().size() > 0) {
+  if (!evmReturnValues.Logs().empty()) {
+    Json::Value _json = Json::arrayValue;
+    bool success = true;
```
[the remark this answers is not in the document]
valeryz: Not keen to block submission of this because of this comment: we need to move fast. Don't mind if the new style is being iterated over in parallel.

I feel this should also be discussed with other C++ style guideline questions. Sure it can be changed, if there is any merit.

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -165,18 +160,40 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
...
+      Json::Reader _reader;
+      Json::Value tmp;
+      try {
+        success = _reader.parse(lval, tmp);
+      } catch (std::exception& e) {
+        LOG_GENERAL(WARNING, "Exception: " << e.what());
+      }
+      if (!success) {
+        LOG_GENERAL(WARNING, "Parsing json unsuccessful " << lval)
+      }
```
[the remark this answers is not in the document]
valeryz: Actually seems indeed a bug, don't you need to move the `bool success = true` inside the for loop? Or even inside the `try`?

Or why not just 

```c++
try {
   if (!_reader.parse(lval, tmp)) {
           LOG_GENERAL(WARNING, "Parsing json unsuccessful " << lval);
   }
 }
 catch (std::exception& e) { ...
```

I also don't like catching the most generic exception and pretending it didn't happen, with only a warning, but I realized it was there before this PR.

PR #2820, opened 2022-06-24
[master] v1 metamask connect to isolated server + testnet

n-hutton:
MVP getting metamask to connect to the testnet and move funds about.

TL;DR: by using the old zil api you can transfer to a metamask account. You can then transfer between these accounts. The mechanism for tracking mined TX is not yet in place so metamask thinks they are still pending:
Screenshot from 2022-07-14 10-43-22

This required adding a library for ECDSA public key recovery (that is, given a message and a signature, derive the public key), which was forked from bitcoin-core. I do not know how esteemed/well tested this library is. It was used by Aleth.

Details:

The main difficulty is getting the details of the raw transaction submission correct - it is detailed here: https://eips.ethereum.org/EIPS/eip-155

The current solution I think only works if metamask thinks your chain height is over 2,675,000 - otherwise, it would use a different signing scheme. The solution I go with might be the code just checks for both.

In order to add a raw eth transaction, you do the following:
Parse all the fields from the transaction using ethereum serialization format 'RLP'
Replace the signature with 0 and hash the transaction to determine the hash that was created client side
Put it back into RLP format and hash it
Use this hash to recreate the public key

There are some dummy values returned for the queries that metamask likes to see that need to be filled in - getting blocks by hash etc.

Note: get balance for eth returns a hex value, whereas zil returns a decimal string. So there is potential for massively incorrect balances to be shown if it is not used correctly. 

Adds the following API calls:

eth_blockNumber
net_version
eth_getBalance
eth_getBlockByNumber
eth_gasPrice
eth_getCode
eth_estimateGas
eth_getTransactionCount
eth_sendRawTransaction
eth_getTransactionReceipt - needs to be added to lookup server (!)


CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ find_package(g3logger CONFIG REQUIRED)
 
 include(InstallCryptoUtils)
 
+include(ProjectSecp256k1)
```
n-hutton, line 48: 3rd party library for public key recovery

src/libEth/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_library(Eth Eth.cpp)
```
n-hutton, line 1: Add a new library for functions relating to Eth stuff. I think we should put most stuff here basically

.gitmodules
```diff
@@ -14,6 +14,7 @@
 [submodule "src/depends/Schnorr"]
 	path = src/depends/Schnorr
 	url = https://github.com/Zilliqa/schnorr.git
+	branch = fix_bn2
```
n-hutton, line 17: this points to this branch https://github.com/Zilliqa/schnorr/pull/14 - which is needed for parsing the pubkey in uncompressed format (could work around to compressed format if need be but it is untidy)

CMakeLists.txt
n-hutton, line 48: external library from a bitcoin-core fork

src/libUtils/DataConversion.cpp
```diff
@@ -45,6 +45,20 @@ bool DataConversion::HexStrToUint8Vec(const string& hex_input, bytes& out) {
   return true;
 }
 
+bytes DataConversion::HexStrToUint8VecRet(const string& hex_input) {
```
n-hutton, line 48: I have added these convenience functions which mirror HexStrToUint8Vec etc. except they return the thing. Usually used for debugging.

constants.xml
```diff
@@ -4,7 +4,7 @@
     <general>
         <DEBUG_LEVEL>3</DEBUG_LEVEL>
         <ENABLE_DO_REJOIN>false</ENABLE_DO_REJOIN>
-        <LOOKUP_NODE_MODE>false</LOOKUP_NODE_MODE>
+        <LOOKUP_NODE_MODE>true</LOOKUP_NODE_MODE>
```
CSideSteve: This is an error is assume, you didnt mean to change this in the master

constants.xml
```diff
@@ -287,7 +287,7 @@
     </recovery>
     <smart_contract>
         <ENABLE_SC>true</ENABLE_SC>
-        <SCILLA_ROOT>/scilla</SCILLA_ROOT>
+        <SCILLA_ROOT>/home/nathan/repos/Zilliqa/scilla</SCILLA_ROOT>
```
CSideSteve: This is an error is assume, you didnt mean to change this in the master

constants.xml
```diff
@@ -307,6 +307,12 @@
         <DISABLE_SCILLA_LIB>false</DISABLE_SCILLA_LIB>
         <SCILLA_SERVER_PENDING_IN_MS>1500</SCILLA_SERVER_PENDING_IN_MS>
         <SCILLA_SERVER_LOOP_WAIT_MICROSECONDS>10</SCILLA_SERVER_LOOP_WAIT_MICROSECONDS>
+        ## Steps to Enable EVM for a run this temporarily replaces the Scilla Interpreter
```
CSideSteve: Need clarification this should be committed.

src/libServer/LookupServer.cpp
```diff
@@ -792,6 +847,187 @@ Json::Value LookupServer::CreateTransaction(
   }
 }
 
+Json::Value LookupServer::CreateTransactionEth(EthFields const& fields, bytes const& pubKey, const unsigned int num_shards,
```
CSideSteve: Is it neccassary to have this code in both isolated server and lookup server, could it not be common to both? just asking :-) or is this just temporary ?

CSideSteve (commented):
Fairly understandble, just a couple of comments will seek clarification tommorw.

src/libServer/LookupServer.cpp
n-hutton: the code that can be common is common - so the isolated server calls the lookup server's implementation when it is simple (like getting the chain id)

constants.xml
n-hutton: Not intentional

constants.xml
n-hutton: nope

constants.xml
n-hutton: nope

src/libData/AccountData/Transaction.cpp
```diff
@@ -147,6 +147,15 @@ const TransactionCoreInfo& Transaction::GetCoreInfo() const {
 
 const uint32_t& Transaction::GetVersion() const { return m_coreInfo.version; }
 
+// Check if the version is 1 or 2 - the only valid ones for now
+// this will look like 65538 or 65537
+bool Transaction::VersionCorrect() const {
+  auto version = DataConversion::UnpackB(this->GetVersion());
+
+  return !(version != TRANSACTION_VERSION &&
+           version != TRANSACTION_VERSION_ETH);
```
bzawisto, line 156: Nitpick: it's perhaps a matter of style, but what about:
`return (version == TRANSACTION_VERSION || version == TRANSACTION_VERSION_ETH)`
Also version could be marked as const.

src/libData/AccountData/Transaction.cpp
```diff
@@ -202,7 +211,9 @@ bool Transaction::IsSignedECDSA() const {
   sigString = sigString.substr(2);
   pubKeyStr = pubKeyStr.substr(2);
 
-  return VerifyEcdsaSecp256k1("", sigString, pubKeyStr);
+  auto hash = GetOriginalHash(GetCoreInfo(), ETH_CHAINID_INT);
```
bzawisto, line 214: If we don't modify `hash` in `VerifyEcdsaSecp256k1` then it should be marked as const.

src/libServer/IsolatedServer.cpp
```diff
@@ -491,6 +549,166 @@ Json::Value IsolatedServer::CreateTransaction(const Json::Value& _json) {
...
 }
 
+Json::Value IsolatedServer::CreateTransactionEth(EthFields const& fields,
+                                                 bytes const& pubKey) {
+  try {
+    if (m_pause) {
+      throw JsonRpcException(RPC_INTERNAL_ERROR, "IsoServer is paused");
+    }
+
+    lock_guard<mutex> g(m_blockMutex);
```
bzawisto, line 559: Nitpick: Do we need this lock here? Can we move it to the relevant scope it really protects?

src/libServer/IsolatedServer.cpp
```diff
@@ -491,6 +549,166 @@ Json::Value IsolatedServer::CreateTransaction(const Json::Value& _json) {
   }
 }
 
+Json::Value IsolatedServer::CreateTransactionEth(EthFields const& fields,
```
bzawisto, line 552: Presumably challenging, but do you think it's feasible to split this function into smaller entities (now it's > 150 loc).

src/libServer/IsolatedServer.h
```diff
@@ -49,6 +50,24 @@ class IsolatedServer : public LookupServer,
     response = this->CreateTransaction(request[0u]);
   }
 
+  inline virtual void GetEthSendRawTransactionI(const Json::Value& request,
+                                                Json::Value& response) {
+    (void)request;
```
bzawisto, line 55: If we don't use request, should we enclose it with / * */ in the arg list?

src/libServer/IsolatedServer.h
```diff
@@ -49,6 +50,24 @@ class IsolatedServer : public LookupServer,
...
+    auto rawTx = request[0u].asString();
+
+    // Erase '0x' at the beginning if it exists
+    if (rawTx[1] == 'x') {
+      rawTx.erase(0, 2);
+    }
+
+    auto pubKey = RecoverECDSAPubSig(rawTx, ETH_CHAINID_INT);
```
bzawisto, line 63: Nitpick: immutable vars could be marked as const.

src/libEth/Eth.cpp
```diff
@@ -0,0 +1,150 @@
...
+  ret["transactions"] = Json::arrayValue;
+  ret["transactionsRoot"] =
+      "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421";
+  ret["uncles"] = Json::arrayValue;
+
+  return ret;
+}
+
+// Given a RLP message, parse out the fields and return a EthFields object
+EthFields parseRawTxFields(std::string const& message) {
```
bzawisto, line 98: It's a bit of a subtle piece of code, can we write some unit test just to have a mean of checking if it still works in case someone wants to change it one day?

src/libCrypto/EthCrypto.cpp
```diff
@@ -188,4 +210,121 @@ std::string ToUncompressedPubKey(std::string const& pubKey) {
...
+      s_ctx{secp256k1_context_create(SECP256K1_CONTEXT_SIGN |
+                                     SECP256K1_CONTEXT_VERIFY),
+            &secp256k1_context_destroy};
+  return s_ctx.get();
+}
+
+// EIP-155 : assume the chain height is high enough that the signing scheme
+// is in line with EIP-155.
+// message shall not contain '0x'
+bytes RecoverECDSAPubSig(std::string const& message, int chain_id) {
```
bzawisto, line 227: It'd be nice to have a unit test for some predefined input and known output (just to ensure it works in case someone needs to change it).

bzawisto (approved):
Well done Nathan! There are a couple of nitpick comments so if you don't want to address them that's still fine!

src/libCrypto/EthCrypto.cpp
```diff
@@ -69,15 +74,18 @@ void SetOpensslSignature(const std::string& sSignatureInHex, ECDSA_SIG* pSign) {
...
   BIGNUM* gx_ptr = gx.get();
+  BIGNUM* gy_ptr = gy.get();
 
   EC_KEY_set_asn1_flag(pKey, OPENSSL_EC_NAMED_CURVE);
 
   // From
   // https://www.oreilly.com/library/view/mastering-ethereum/9781491971932/ch04.html
   // The first byte indicates whether the y coordinate is odd or even
   int y_chooser_bit = 0;
+  bool notCompressed = false;
```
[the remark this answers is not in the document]
n-hutton: sounds good 👍

src/libCrypto/EthCrypto.cpp
```diff
@@ -188,4 +210,121 @@ std::string ToUncompressedPubKey(std::string const& pubKey) {
...
+
+// EIP-155 : assume the chain height is high enough that the signing scheme
+// is in line with EIP-155.
+// message shall not contain '0x'
+bytes RecoverECDSAPubSig(std::string const& message, int chain_id) {
+  // First we need to parse the RSV message, then set the last three fields
+  // to chain_id, 0, 0 in order to recreate what was signed
+  bytes asBytes;
+  int v = 0;
+  bytes rs;
```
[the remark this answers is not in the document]
n-hutton: 👍

src/libCrypto/EthCrypto.cpp
```diff
@@ -188,4 +210,121 @@ std::string ToUncompressedPubKey(std::string const& pubKey) {
...
+  rlpStreamRecreated << info.gasPrice;
+  rlpStreamRecreated << info.gasLimit;
+  rlpStreamRecreated << info.toAddr;
+  rlpStreamRecreated << info.amount;
+  rlpStreamRecreated << info.data;
+  rlpStreamRecreated << chainId;
+  rlpStreamRecreated << bytes{};
+  rlpStreamRecreated << bytes{};
+
+  auto signingHash = ethash::keccak256(rlpStreamRecreated.out().data(),
```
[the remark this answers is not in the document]
n-hutton: 👍

src/libData/AccountData/Transaction.cpp
n-hutton: 👍

src/libData/AccountData/Transaction.cpp
n-hutton: 👍

src/libServer/IsolatedServer.cpp
n-hutton: yes, I can do this - I was basically copying the code from the scilla way of adding TX so can refactor both in a seperate PR if you like?

src/libServer/IsolatedServer.cpp
n-hutton: Again, I was following the code from add TX from what was there before. It looks like the mutex should be protecting changes to the account store (UpdateAccountsTemp) and the m_mediator.m_txBlockChain. 

I have moved the locks to the last line which looks safe (just before validateTX). The way the code is written, I think that if you lock any later there is the potential of multiple TXs passing the nonce check and proceeding to execute, although the code is hard to reason about so I am not totally sure.

src/libServer/IsolatedServer.h
n-hutton: Oh. Request is used. The (void) was a copy/paste error in this instance. I don't mind refactoring to use /* */ instead of a void cast in a future PR if you like?

src/libServer/IsolatedServer.h
n-hutton: 👍

src/libEth/Eth.cpp
n-hutton: sounds great - done

src/libCrypto/EthCrypto.cpp
n-hutton: sounds great - done

PR #2821, opened 2022-06-25
[master] Gas fixes, needs test routine to check handling of gas, not currently…

CSideSteve:
Use the evm calculated gas as oppossed to the charging strategy deployed for Scilla

## Description

Return an EVM caclulated gas charge and return accumulated gas used to client

Address part of the issues raised in 4646, no tests exist in the suite for checing gas, need to develop tests

## Backward Compatibility

### Implementation

- [ ] **ready for review**

Observe gas utilisation in reciepts of client and balance in account

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -198,8 +198,7 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
     receipt.AddJsonEntry(_json);
   }
 
-  auto gas = EvmUtils::UpdateGasRemaining(
-      receipt, invoke_type, params.m_available_gas, evmReturnValues.Gas());
+  auto gas = evmReturnValues.Gas();
 
   std::map<std::string, bytes> states;
   std::vector<std::string> toDeletes;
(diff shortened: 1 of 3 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -520,7 +519,14 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       auto sbcip = std::make_unique<ScillaBCInfo>(
           m_curBlockNum, m_curDSBlockNum, m_originAddr, contractAddress,
           contractAccount->GetStorageRoot(), scilla_version);
-      m_scillaIPCServer->setBCInfoProvider(std::move(sbcip));
+
+      if (m_scillaIPCServer) {
```
valeryz, line 523: At this moment I realized the relationship is too complicated between AccountStoreSC and `m_scillaIPCServer`, which can indeed be nullptr, and even if not, it could probably not be started/listening.  Probably need to simplify this later on.  Not blocking submission of course.

valeryz (approved):

PR #2825, opened 2022-06-29
Ut for evm lookup server

moboware:
## Description
- Make unit test that is able to mock the evm-ds server response and see how we can implement a test that manages that.
- This review is to share these changes and make a conclusion on the next steps.
- It only implements a GetEthCall test in good weather scenario, more tests and checks should be added to check error situations.
- Response data from the evm-ds server should generated, but is now copied from the live server to make this test run. Need a clear picture of evm-ds protocol for that.

## Backward Compatibility

- [-] This is a breaking change

## Review Suggestion
Check the unit test structural implementation and the logic that is checked, if this covers enough cases for this test now.

### Implementation
- [ ] **ready for review**

### Conclusion
- To able to mock classes and override of singleton is not the best choice. Dependency injection is a better way to solve this but needs more code changes.
- Boost test framework is not sufficient to as a test/mock framework. Google Mock is a better choice
- Reverse engineering code and build a test for that will check code but is no guarantee that all test cases are tested. (Use TDD)


src/common/Singleton.h
```diff
@@ -26,20 +28,24 @@ class Singleton {
...
 
  public:
-  static T& GetInstance() noexcept(std::is_nothrow_constructible<T>::value) {
-    // Guaranteed to be destroyed.
-    // Instantiated on first use.
-    // Thread safe in C++11
-    static T instance;
-
-    return instance;
+  static T& GetInstance(const std::function<std::shared_ptr<T>()>& _allocator =
```
KaustubhShamshery, line 40: Any need to remove this? https://stackoverflow.com/questions/11711920/how-to-implement-multithread-safe-singleton-in-c11-without-using-mutex

This is how it is in C++11

valeryz:
Is this PR adding a unit test or does the cleanup?

Would be good to have separate PRs (not for this one, but please keep in mind for the future).

src/common/Singleton.h
```diff
@@ -26,20 +28,24 @@ class Singleton {
...
+                            []() { return std::make_shared<T>(); }) {
```
valeryz, line 41: Do you plan to use the `allocator`?  I quite strongly feel about the YAGNI principle and think things should be added only when they are needed, not proactively.

valeryz (commented):
Thanks, lots of very good cleanups, only a few questions in the comments.

src/common/Singleton.h
moboware: there was no reason to remove this, have added it back

src/common/Singleton.h
moboware: Overriding the default allocater in the singleton is kind of a hack, but gave me to possibility to inject a different class to override the CallRunner method. 
Would be better to remove all singletons and use a different design that will give use the override possibility to test and implement different versions.

valeryz (approved):

src/libData/AccountData/EvmClient.h
```diff
@@ -23,36 +23,33 @@
...
+  virtual bool CallRunner(uint32_t version, const Json::Value& _json,
+                          evmproj::CallResponse& result,
+                          uint32_t counter = MAXRETRYCONN);
 
- private:
-  EvmClient() {}
-  virtual ~EvmClient() {}
-  bool OpenServer(bool force = false);
+ protected:
+  virtual bool OpenServer(bool force = false);
```
CSideSteve, line 51: In EvmClient, are we supporting multiple versions now, as this was deliberatly forced to a single version with a note that we should get rid of the map later as we were minimising changes at the time. if we are not supporting multiple versions using the ScillaClient technique then it would be better to just ghet rid of the maps entirely and replace with single pointers.

src/libData/AccountData/EvmClient.cpp
```diff
@@ -75,21 +75,20 @@ bool EvmClient::OpenServer(bool force) {
...
 
-  if (m_initialised) return true;
+  if (m_initialised) {
+    return true;
+  }
 
   if (!OpenServer(enforce)) {
     LOG_GENERAL(WARNING, "OpenServer for version " << version << "failed");
     return false;
   }
```
CSideSteve, line 86: See previous comment on map or not

src/libServer/LookupServer.cpp
```diff
@@ -896,42 +897,47 @@ string LookupServer::GetEthCall(const Json::Value& _json) {
...
     if (_json.isMember("fromAddr")) {
       fromAddr = Address(_json["fromAddr"].asString());
     }
+
+    uint64_t amount{0};
     if (_json.isMember("amount")) {
       const auto amount_str = _json["amount"].asString();
       amount = strtoull(amount_str.c_str(), NULL, 0);
     }
+
```
CSideSteve, line 913: Why are we playing with gas here?

src/libUtils/EvmJsonResponse.cpp
```diff
@@ -34,23 +34,24 @@ namespace evmproj {
  * throws std::exception or the original exception is passed up to the caller.
  * */
 
-evmproj::CallResponse& GetReturn(const Json::Value& oldJason,
+evmproj::CallResponse& GetReturn(const Json::Value& oldJson,
                                  evmproj::CallResponse& fo) {
-  nlohmann::json newJason;
+  nlohmann::json newJson;
```
CSideSteve, line 40: Thanks for catching that one, pity we have to kill of Jason, never noticed him before :-)

src/libServer/LookupServer.cpp
moboware: the gas implementation was there already, I only moved the gasRemained closer to where it is used

src/libData/AccountData/EvmClient.h
moboware: fair point, but would change this when we refactor

src/libUtils/EvmJsonResponse.cpp
CSideSteve: It was probably a spellcheck, but a funny one that I didn't notice, its like one of those examples where you can read a sentence clearly with letters mixed up and numbers inserted, the brain corrects.

PR #2827, opened 2022-07-04
Zil 4609 make sure evm contracts can be created through the create transaction api

CSideSteve:
## Description

Creates a new contract if EVM submits apply instructions with a non existant contract account.

## Backward Compatibility

## Review Suggestion

Check your existing tests that use EVM-DS as all apply instructions receive this new behavior, an account is created as oppossed to an error message and a halt in processing.

### Implementation

- [x ] **ready for review**

internal core dev only

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -154,7 +154,7 @@ do
 done
 
 cmake -H. -B${dir} ${CMAKE_EXTRA_OPTIONS} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DTESTS=ON -DCMAKE_INSTALL_PREFIX=..
-cmake --build ${dir} -- -j4
+cmake --build ${dir} -- -j20
 ./scripts/license_checker.sh
 ./scripts/depends/check_guard.sh
 [ ${run_clang_tidy_fix} -ne 0 ] && cmake --build ${dir} --target clang-tidy-fix
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -210,13 +210,12 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
     } else {
       // Get the account that this apply instruction applies to
       Account* targetAccount = this->GetAccount(Address(it->Address()));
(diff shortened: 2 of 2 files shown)


build.sh
```diff
@@ -154,7 +154,7 @@ do
 done
 
 cmake -H. -B${dir} ${CMAKE_EXTRA_OPTIONS} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DTESTS=ON -DCMAKE_INSTALL_PREFIX=..
-cmake --build ${dir} -- -j4
+cmake --build ${dir} -- -j20
```
n-hutton, line 157: is this intentional?

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -211,12 +211,18 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
       // Get the account that this apply instruction applies to
       Account* targetAccount = this->GetAccount(Address(it->Address()));
       if (targetAccount == nullptr) {
-        LOG_GENERAL(
-            WARNING,
-            "Cannot find account for address given in  Apply operation from "
-            "EVM-DS"
-            " ");
-        return gas;
+        if (!this->AddAccount(Address(it->Address()), {0, 0})) {
```
n-hutton, line 214: is this creating an empty account if there is an apply instruction? In my mind, the apply instruction would usually contain code/data

build.sh
CSideSteve: No, but it helps a lot if you have the cpus, I will revert that one.

src/libData/AccountData/AccountStoreSC.tpp
CSideSteve: The code gets added later which could be a small optimisation, to avoid creating with zero and then setting later, it allows for the code to be modified on an existing account. It also allows an account to be created with balance and no code if the code is omitted by EVM-ds

valeryz:
Ok I got confused because of the title of the PR. Seemed something we've done long ago.

valeryz (approved):

PR #2832, opened 2022-07-08
[master] Zil 4673 implement basic ethereum api support into lookup server

moboware:
## Description
Implementation of the Ethereum basics API messages in the LookupServer, see ticket ZIL-4373 for details

## Backward Compatibility
- [-] This is a breaking change

### Implementation
- [x] **ready for review**

There are no integration test implemented in this branch

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -1007,6 +1007,12 @@ string LookupServer::GetWeb3Sha3(const Json::Value& _json) {
       reinterpret_cast<const uint8_t*>(str.data()), str.size()));
 }
 
+std::string LookupServer::GetEthMining() {
+  LOG_MARKER();
+
+  return "false";
+}
+
 Json::Value LookupServer::GetSmartContractState(const string& address,
                                                 const string& vname,
                                                 const Json::Value& indices) {
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -1009,7 +1009,8 @@ string LookupServer::GetWeb3Sha3(const Json::Value& _json) {
 
 std::string LookupServer::GetEthMining() {
   LOG_MARKER();
-
+  // @todo : the mining state a could be retrieved from the WorkServer if it can
+  // provide the exact state of mining.
   return "false";
 }
 
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -382,6 +382,31 @@ LookupServer::LookupServer(Mediator& mediator,
                          jsonrpc::JSON_STRING, NULL),
       &LookupServer::GetEthCoinbaseI);
 
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("net_listening", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetNetListeningI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("protocol_version", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetProtocolVersionI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("net_peerCount", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetNetPeerCountI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_chainId", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetEthChainIdI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_syncing", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetEthSyncingI);
+
   m_StartTimeTx = 0;
   m_StartTimeDs = 0;
   m_DSBlockCache.first = 0;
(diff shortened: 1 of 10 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -96,216 +99,324 @@ LookupServer::LookupServer(Mediator& mediator,
...
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("web3_sha3", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, "param01", jsonrpc::JSON_STRING,
+                         NULL),
+      &LookupServer::GetWeb3Sha3I);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_mining", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetWeb3Sha3I);
```
valeryz, line 378: copy paste error here.

src/libServer/LookupServer.cpp
```diff
@@ -96,216 +99,324 @@ LookupServer::LookupServer(Mediator& mediator,
...
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_coinbase", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetEthCoinbaseI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("net_version", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetNetListeningI);
```
valeryz, line 388: another copy/paste error

src/libServer/LookupServer.cpp
```diff
@@ -939,6 +1051,70 @@ string LookupServer::GetEthCall(const Json::Value& _json) {
...
+}
+
+std::string LookupServer::GetProtocolVersion() {
+  LOG_MARKER();
+  return "";
+}
+
+std::string LookupServer::GetEthChainId() {
+  LOG_MARKER();
+  return "0x0";
```
valeryz, line 1103: My proposal is to use 0x8001 for Mainnet, or 0x814d for Tesnet.  See https://zilliqa-jira.atlassian.net/browse/ZIL-4668 for more details. Also https://dev.zilliqa.com/docs/apis/api-blockchain-get-network-id/

So we need to read constants.xml here, or in the constructor of the lookupServer, or, wherever Zilliqa chain_id can be found.

src/libServer/LookupServer.cpp
```diff
@@ -939,6 +1051,70 @@ string LookupServer::GetEthCall(const Json::Value& _json) {
...
+}
+
+Json::Value LookupServer::GetEthSyncing() {
+  LOG_MARKER();
+  const Json::Value returnJson = false;
```
valeryz, line 1108: `return Json::Value(false);`  like elsewhere?

src/libServer/LookupServer.cpp
```diff
@@ -87,6 +89,7 @@ LookupServer::LookupServer(Mediator& mediator,
       jsonrpc::Procedure("GetCurrentDSEpoch", jsonrpc::PARAMS_BY_POSITION,
                          jsonrpc::JSON_STRING, NULL),
       &Server::GetCurrentDSEpochI);
+
```
valeryz, line 92: Much whitespace makes fewer lines fit on the screen.  I understand the preference, but we first should agree about this style choice, then make the change.

Lack of this space is not a bug, so need not be fixed.  Also tagging @CSideSteve, @n-hutton, @art-gor, @KaustubhShamshery for an opinion about blank lines.

Another question is whether to keep style cleanup PRs separate from the functionality implementation.

tests/EvmLookupServer/Test_EvmLookupServer.cpp
```diff
@@ -29,21 +29,37 @@ class AbstractServerConnectorMock : public jsonrpc::AbstractServerConnector {
...
 
-BOOST_AUTO_TEST_SUITE(lookup_server)
+/**
+ * @brief Default Mock implementation for the evm client
+ */
+class EvmClientMock : public EvmClient {
+ public:
+  EvmClientMock() = default;
+
+  auto OpenServer(bool /*force = false*/) -> bool { return true; };
```
valeryz, line 39: The fact that newer C++ allows arrow function declaration/definition, doesn't make it automatically preferred syntax, does it?

Especially given that the current codebase uses the more traditional way, which is *not* declared obsolete in any way, AFAIK.  

Don't want to be judgemental about style, really would want to leave it to the team, but I strongly prefer consistency whatever the decisions is.

moboware: I understand your point, but this is a c++11 style and that is the current standard that we use. I am afraid that as we don't adopt new language features and styles we will have exceptions on exceptions and will stay in c++99. It also shows that our coding standard is not up-to-date.

src/libServer/LookupServer.cpp
moboware: IMHU, an extra line after a code-block makes the code more readable

tests/EvmLookupServer/Test_EvmLookupServer.cpp
valeryz: Which style guide recommends always using newer forms?  The newer forms are available to use when it makes sense, but whether to encourage its use is probably left to the team.

E.g. I've been following this style, and find it useful: https://google.github.io/styleguide/cppguide.html#trailing_return

Same about `auto`, which is great when the type is obscure, and hurts readability otherwise. Just because something is available, doesn't necessarily mean it is better:


> The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type. When judging whether the code is clearer, keep in mind that your readers are not necessarily on your team, or familiar with your project, so types that you and your reviewer experience as unnecessary clutter will very often provide useful information to others. For example, you can assume that the return type of make_unique<Foo>() is obvious, but the return type of MyWidgetFactory() probably isn't.

CSideSteve (approved):
GTG style looks fine to me, the blank lines do improve readability for the new reader. We can style things later in the project when all the functionality is complete and we are comfortable to go public, hopefully just automate the formatting process on checkin, you can then unformat it anyway you like when you are editing if you prefer to read it some other way.

PR #2837, opened 2022-07-15
[master] Zil 4672 resolve the 128 256 bit discrepancy between zilliqa and evm

moboware:
## Description
- When an evm balance (256 uint) or nonce (128 uint) is received the values can exceed the zilliqa balance and nonce maxes and will overflow the internal balance and nonce values. To prevent this situation there is now a check in the account data that checks the overflow condition and will return false or throws and exception.

- This branch also contains a code cleanup on the boost::multiprecision and the using declaration on uint128_t etc. because it did not use them in all cases.

- Added 2 unit tests to check the behavior on the overflow condition and when it is returned from the evm

## Backward Compatibility

### Implementation
- [X] **ready for review**

- [X] local machine test


valeryz:
This PR is impossible to review.

@moboware would you please make a new one just with the changes relevant to the 128-256 bit problem and nothing else?

PR #2849, opened 2022-07-27
[master] Evm rpc plus integration test extensions

n-hutton:
Suggestion: use a python integration script to run an isolated server and test the API works as part of the CI


scripts/integration_test.sh
```diff
@@ -0,0 +1,45 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+# This script will start an isolated server and run the python API against it
+#
+
+find ./ -name isolatedServer
```
valeryz, line 20: This `find` just prints the path to the isolated server and that's it? Was the intent to use the result later on?

valeryz (approved):
Please address the comments below, otherwise LGTM

scripts/integration_test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Modify constants.xml for use by isolated server
+cp constants.xml constants_backup.xml
+sed -i 's/.ENABLE_SC.true/<ENABLE_SC>false/g' constants.xml
```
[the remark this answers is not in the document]
n-hutton: For this test, it would fail when it tries to use the VMs. In a subsequent PR that does use the VM then I will arrange the test environment so that it has it

scripts/integration_test.sh
n-hutton: ah, this is a mistake and I will remove it. Thanks!

PR #2853, opened 2022-08-01
[master] Block by hash support

bzawisto:
## Description
This PR adds possibility to retrieve transaction block by providing a hash. GetBlockByHash is not fully implemented (although some parts of LookupServer has already been touched) as the PR would've become even bigger. Sorry for the unreviewable changes in `TrieDB.h`, but this file became a victim of clang format. The only relevant change made the is `const` specifier for two functions + `mutable` for a class field. That was a consequence of fixing const-ness in Serialize method (from AccountStore classes). 
This is not a breaking change since I'm only adding new things to new leveldb tables. 

## Backward Compatibility

- [x ] This is not a breaking change

## Review Suggestion

- [x] **ready for review**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -337,7 +337,10 @@ endif()
 
 # installation
 
-set_target_properties(zilliqa sendcmd genaccounts genkeypair getpub getaddr gentxn signmultisig verifymultisig zilliqad gensigninitialds grepperf getnetworkhistory getrewardhistory validateDB restore genTxnBodiesFromS3 isolatedServer data_migrate
+set_target_properties(buildTxBlockHashesToNums data_migrate genaccounts genkeypair genTxnBodiesFromS3 
+                      getpub getaddr gentxn signmultisig verifymultisig gensigninitialds getnetworkhistory grepperf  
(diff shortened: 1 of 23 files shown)


valeryz (approved):
I like the PR, but we need to make sure new persistence files are supported by all the downstream tools and the size is not a problem.

An LGTM from @chetan-zilliqa is required before submission.

chetan-zilliqa:
LGTM !!!
Space-wise, I believe we should be ok, as the two newly persistence folder looks to be storing minimal information.
However,  In the `BuildHashToNumberMappingForTxBlocks`  function, we are iterating over all the blocks, so we will need to track the execution time of the function when it runs on large persistence(In the future). The execution goes through all the txBlocks for finding the `MAX_TX_BLOCK_NUM_KEY`.
This call will execute during the start of blockchain or any rejoin scenarios, so would not be a problem. But anyway, just a note.

chetan-zilliqa (approved):

bzawisto:
Thank you both for approval. `BuildHashToNumberMappingForTxBlocks` is supposed to be run on very limited range (this is why we have external program that should offline build as many hash->num pairs as possible before we deploy new code to dev/test/mainnet).

PR #2854, opened 2022-08-02
[master] Local performance fixes

art-gor:
## Description

Minor performance improvements in internal messaging.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libDirectoryService/DirectoryService.cpp
+++ b/src/libDirectoryService/DirectoryService.cpp
@@ -1190,7 +1190,8 @@ uint8_t DirectoryService::CalculateNewDifficultyCore(uint8_t currentDifficulty,
                                                      int64_t expectedNodes,
                                                      uint32_t powChangeoAdj) {
   if (powChangeoAdj == 0) {
-    // Special mode for developers: just to make thee difficulty flat and not growing
+    // Special mode for developers: just to make thee difficulty flat and not
+    // growing
     return minDifficulty;
   }
 


src/depends/safeserver/safehttpserver.cpp
```diff
@@ -9,35 +9,44 @@
...
+      threads(threads),
+      running(false),
+      path_sslcert(sslcert),
+      path_sslkey(sslkey),
+      daemon(NULL),
+      bindlocalhost(false) {
+  if (threads < 1) {
+    this->threads = 1;
+  } else if (threads > 50) {
+    this->threads = 50;
```
valeryz, line 47: Why are we capping threads at 50 here? How do we know this won't cause problems in prod if all threads get busy?

src/depends/safeserver/safehttpserver.cpp
```diff
@@ -107,20 +114,20 @@ bool SafeHttpServer::StartListening() {
             SafeHttpServer::callback, this, MHD_OPTION_HTTPS_MEM_KEY,
             this->sslkey.c_str(), MHD_OPTION_HTTPS_MEM_CERT,
             this->sslcert.c_str(), MHD_OPTION_THREAD_POOL_SIZE, this->threads,
+            MHD_OPTION_LISTEN_BACKLOG_SIZE, LISTEN_BACKLOG_SIZE,
```
valeryz, line 117: Could you add a comment about these new options. Why are they needed here and what they do?

src/libData/AccountData/AccountStore.cpp
```diff
@@ -67,7 +67,8 @@ AccountStore::AccountStore() {
   }
   // EVM required to run on Lookup nodes too for view calls
   if (ENABLE_EVM) {
-    if (not ipcScillaInit && !LOOKUP_NODE_MODE) {
+    // TODO lookup nodes may also need it
+    if (not ipcScillaInit /*&& !LOOKUP_NODE_MODE*/) {
```
valeryz, line 71: This change seems wrong, as we should not change anything related to Scilla.

src/libDirectoryService/DirectoryService.cpp
```diff
@@ -1189,6 +1189,11 @@ uint8_t DirectoryService::CalculateNewDifficultyCore(uint8_t currentDifficulty,
                                                      int64_t powSubmissions,
                                                      int64_t expectedNodes,
                                                      uint32_t powChangeoAdj) {
+  if (powChangeoAdj == 0) {
```
valeryz, line 1192: Can you explain in the comments what this special mode for developers mean?

src/libNetwork/P2PComm.h
```diff
@@ -19,7 +19,7 @@
 #define ZILLIQA_SRC_LIBNETWORK_P2PCOMM_H_
 
 #include <event2/util.h>
-#include <boost/lockfree/queue.hpp>
+// XXX #include <boost/lockfree/queue.hpp>
```
valeryz, line 22: Please remove this.

src/libNetwork/P2PComm.h
```diff
@@ -114,8 +115,11 @@ class P2PComm {
 
   ThreadPool m_SendPool{MAXSENDMESSAGE, "SendPool"};
 
-  boost::lockfree::queue<SendJob*> m_sendQueue;
-  void ProcessSendJob(SendJob* job);
+  // TODO shared instead of unique due to lambda move capture limitations
```
valeryz, line 118: TODO comment seems not needed.

valeryz (commented):
A nice change!

But please address all the comments, and get @CSideSteve feedback.

src/libData/AccountData/AccountStore.cpp
CSideSteve: Changes in this area were introduced for the work Kaustubh was doing on the API interface.

We have some states that the logic here doesn't quit match, these are :

Running as a Zilliqa process and not the Lookup Node or LOOKUP_NODE_MODE setting

Running as the isolatedServer

Running the testroutines as part of the build causes an invocation of this method and has requirements too.

valeryz: Kaustub's change were to enable EVM on seeds/lookups.  But Scilla should not be affected.

CSideSteve (approved):
Looks Good

PR #2862, opened 2022-08-04
[master] More rpc tests

n-hutton:
Add more RPC tests, comments on the PR are inline

### DIFF
--- a/scripts/integration_test.sh
+++ b/scripts/integration_test.sh
@@ -17,6 +17,11 @@
 # This script will start an isolated server and run the python API against it
 #
 
+# Need to build evm...
+git clone git@github.com:Zilliqa/evm-ds.git
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+cargo build --release
+
 # Modify constants.xml for use by isolated server
 cp constants.xml constants_backup.xml
 sed -i 's/.ENABLE_SC.true/<ENABLE_SC>false/g' constants.xml
--- a/scripts/integration_test.sh
+++ b/scripts/integration_test.sh
@@ -28,7 +28,7 @@ cp constants.xml constants_backup.xml
 echo "Starting isolated server"
 ./build/bin/isolatedServer -f isolated-server-accounts.json -u 999 &
 
-sleep 15
+sleep 30
 
 echo "Starting python test"
 sudo apt-get install python3-pip python3-setuptools python3-pip python3-dev python-setuptools-doc python3-wheel || exit 1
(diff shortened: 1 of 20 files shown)


src/libCrypto/EthCrypto.cpp
```diff
@@ -324,7 +324,7 @@ bytes GetOriginalHash(TransactionCoreInfo const& info, uint64_t chainId) {
   rlpStreamRecreated << info.gasLimit;
   rlpStreamRecreated << info.toAddr;
   rlpStreamRecreated << info.amount;
-  rlpStreamRecreated << info.data;
+  rlpStreamRecreated << FromEVM(info.code);
```
n-hutton, line 327: when building up the rlp stream as it was created on the client side, we need to strip off the 'EVM' prefix otherwise it thinks the signature is incorrect

src/libEth/Eth.cpp
```diff
@@ -128,7 +128,7 @@ EthFields parseRawTxFields(std::string const& message) {
         ret.amount = uint128_t(*it);
         break;
       case 5:
-        ret.data = byteIt;
+        ret.code = byteIt;
```
n-hutton, line 131: I will mention on slack - Eth TXs only have a code field (not code and data)

valeryz: how come? They have code and calldata?

valeryz (approved):

scripts/integration_test.sh
```diff
@@ -17,20 +17,38 @@
 # This script will start an isolated server and run the python API against it
 #
 
+# Need to build evm...
+git clone git@github.com:Zilliqa/evm-ds.git
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+source "$HOME/.cargo/env"
+cargo build --release
+
+find ./ -name evm-ds
```
bzawisto, line 26: Nitpick: leftover?

src/libCrypto/EthCrypto.cpp
```diff
@@ -334,3 +334,40 @@ bytes GetOriginalHash(TransactionCoreInfo const& info, uint64_t chainId) {
...
+    return in;
+  }
+
+  // std::string ret =
+  // std::string(reinterpret_cast<const char*>(in.data()), in.size());
+  // ret.erase(0, 3);
+
+  // Copy all but the first three characters('EVM') into a string for conversion
+  // Advance iterator at beginning
+  std::string ret;
```
bzawisto, line 363: Nitpick: `ret{in.begin() + 3, in.end()};`

bzawisto (approved):
One nitpick, otherwise LGTM. Thanks!

scripts/integration_test.sh
n-hutton: yup, fixed

src/libCrypto/EthCrypto.cpp
n-hutton: 👍

PR #2864, opened 2022-08-08
Adding missing  eth_getBlockByHash api call

bzawisto:
## Description
This PR adds missing eth_getBlockByHash call

## Backward Compatibility

### Implementation

 [X] **ready for review**

[X ] local machine test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -387,6 +387,12 @@ LookupServer::LookupServer(Mediator& mediator,
       &LookupServer::GetEthBlockByNumberI);
 
   this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_getBlockByHash", jsonrpc::PARAMS_BY_POSITION,
+                         jsonrpc::JSON_STRING, "param01", jsonrpc::JSON_STRING,
+                         "param02", jsonrpc::JSON_BOOLEAN, NULL),
+      &LookupServer::GetEthBlockByHashI);
+
+  this->bindAndAddMethod(
       jsonrpc::Procedure("eth_gasPrice", jsonrpc::PARAMS_BY_POSITION,
                          jsonrpc::JSON_STRING, NULL),
       &LookupServer::GetEthGasPriceI);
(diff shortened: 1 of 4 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -1238,33 +1269,6 @@ Json::Value LookupServer::GetTxBlockByNum(const string& blockNum,
   }
 }
 
-Json::Value LookupServer::GetTxBlockByHash(const string& inputHash,
```
bzawisto, line 1241: This method was previously added but has never been used.

src/libServer/LookupServer.cpp
```diff
@@ -1034,52 +1040,77 @@ Json::Value LookupServer::GetEthBlockByNumber(const std::string& blockNumberStr,
...
-          continue;
-        }
-        transactions.push_back(std::move(transactioBodyPtr));
-      }
+Json::Value LookupServer::GetEthBlockByHash(const std::string& inputHash,
+                                            bool includeFullTransactions) {
+  try {
+    const BlockHash blockHash{inputHash};
+    const auto txBlock = m_mediator.m_txBlockChain.GetBlockByHash(blockHash);
+    static const TxBlock NON_EXISTING_TX_BLOCK{};
```
n-hutton, line 1058: Is it neccessary to use a static in this way? I consider it to be a little unsafe to have static classes as it means whatever is in its constructor will execute at t = 0. In this case it is fine as there is no constructor but this has caused issues for me in the past where you can get race conditions at start up.

Could be something like:

```
    if (!m_mediator.m_txBlockChain.GetBlockByHash(blockHash, setMe)) {
      return Json::nullValue;
    }
```

src/libServer/LookupServer.cpp
```diff
@@ -1034,52 +1040,77 @@ Json::Value LookupServer::GetEthBlockByNumber(const std::string& blockNumberStr,
...
+    LOG_GENERAL(INFO, "[Error]" << e.what() << " Input: " << inputHash
                                 << ", includeFullTransactions: "
                                 << includeFullTransactions);
     throw JsonRpcException(RPC_MISC_ERROR, "Unable To Process");
   }
 }
 
+Json::Value LookupServer::GetEthBlockCommon(const TxBlock& txBlock,
+                                            bool includeFullTransactions) {
+  const auto dsBlock =
```
n-hutton, line 1074: You use this for the difficulty field? That's very thorough of you :)

src/libServer/LookupServer.cpp
n-hutton: 👍

n-hutton (changes requested):
Minor change/discussion. Looks great otherwise!

src/libServer/LookupServer.cpp
bzawisto: Are we going to rely on automatically generated `operator bool()`? I'm afraid this may not work (some of the fields may not return false on an empty object). I think we can just use my approach without `static` though.

n-hutton (approved):

PR #2865, opened 2022-08-09
[master] add ds to ci build env

n-hutton:


scripts/ci_build.sh
```diff
@@ -19,6 +19,8 @@
 
 set -e
 
+time ./scripts/integration_test.sh || exit 1
```
bzawisto, line 22: Should we keep that wrapped up with `time` command?

scripts/integration_test.sh
```diff
@@ -17,20 +17,45 @@
...
+#find ./ -name evm-ds
+#find ./ -name cargo.toml 2>&1 > /dev/null
+#
+## Modify constants.xml for use by isolated server
+#cp constants.xml constants_backup.xml
+#sed -i 's/.LOOKUP_NODE_MODE.false/<LOOKUP_NODE_MODE>true/g' constants.xml
+#sed -i 's/.EVM_SERVER_BINARY.*/<EVM_SERVER_BINARY>\/tmp\/evm-ds<\/EVM_SERVER_BINARY>/g' constants.xml
+#sudo mkdir -p /usr/local/etc/
+
+exit 1
```
bzawisto, line 40: Is it supposed to exit at this point?

src/libCrypto/EthCrypto.cpp
```diff
@@ -334,3 +334,31 @@ bytes GetOriginalHash(TransactionCoreInfo const& info, uint64_t chainId) {
...
+bytes ToEVM(bytes const& in) {
+  if (in.empty()) {
+    return in;
+  }
+
+  std::string prefixedEvm{"EVM"};
+  prefixedEvm += DataConversion::Uint8VecToHexStrRet(in);
+  bytes ret;
+
+  for (const auto& a : prefixedEvm) {
```
bzawisto, line 347: Nitpick: it's a good candidate for `std::copy()` with `std::back_inserter`.

src/libCrypto/EthCrypto.cpp
```diff
@@ -334,3 +334,31 @@ bytes GetOriginalHash(TransactionCoreInfo const& info, uint64_t chainId) {
...
+
+  return ret;
+}
+
+bytes FromEVM(bytes const& in) {
+  if (in.empty()) {
+    return in;
+  }
+
+  std::string conv =
```
bzawisto, line 359: Nitpick: could we just use `in.data() + 3` instead with some comment why is so?

src/libServer/LookupServer.cpp
```diff
@@ -1423,12 +1430,115 @@ Json::Value LookupServer::GetEthSyncing() {
...
   const Json::Value expectedResponse = Json::arrayValue;
   return expectedResponse;
 }
 
+Json::Value LookupServer::GetEthStorageAt(std::string const& address,
+                                          std::string const& position,
+                                          std::string const& /*blockNum*/) {
+  LOG_MARKER();
+
+  string vname{};
```
bzawisto, line 1444: Nitpick: could these vars be moved closer to the relevant scope?

src/libServer/LookupServer.cpp
```diff
@@ -1423,12 +1430,115 @@ Json::Value LookupServer::GetEthSyncing() {
...
+    // Left-pad position with 0s up to 64
+    std::string zeroes =
+        "0000000000000000000000000000000000000000000000000000000000000000";
+
+    if (position.size() > zeroes.size()) {
+      throw JsonRpcException(RPC_INTERNAL_ERROR,
+                             "Position string is too long! " + position);
+    }
+
+    if (!position.empty()) {
```
bzawisto, line 1491: I don't think I fully understand the algorithm here, but that's probably not something that does leftPad. Also please note you dereference `end()` iterator in line 1497 on the first which is basically UB.
It could be a nice candidate for STL algorithms. Please consider the following:
`const auto firstPosOfX = position.find_first_of('x');
if(firstPosOfX == std::string::npos) {
   // Handle errors accordingly
}
// Calculate valid range
const auto maxEnd = std::min(zeros.size() - firstPosOfX, positions.size() - firstPosOfX);
// Copy positions to zeroed pad
std::copy(positions.begin() + firstPosOfX, positions.begin() + firstPosOfX + maxEnd, zeros.begin() + firstPosOfX);
`
Obviously this requires some attention and testing.

src/libServer/LookupServer.cpp
```diff
@@ -1423,12 +1430,115 @@ Json::Value LookupServer::GetEthSyncing() {
...
+    }
+
+    auto res = root["_evm_storage"][zeroes];
+    bytes resAsStringBytes;
+
+    for (const auto& item : res.asString()) {
+      resAsStringBytes.push_back(item);
+    }
+
+    auto resAsStringHex = std::string("0x") +
```
bzawisto, line 1513: nitpick: this could be `const`

bzawisto:
Hey Hathan! Well done! However, please review my comments (in particular one for leftPad). Cheers!

scripts/ci_build.sh
n-hutton: Oh, now I think about it, it probably interfers with the exit 1 code. Will remove on the other PR

scripts/integration_test.sh
n-hutton: Not relevant/on the main PR

src/libCrypto/EthCrypto.cpp
n-hutton: 👍

src/libServer/LookupServer.cpp
n-hutton: 👍

src/libServer/LookupServer.cpp
n-hutton: 👍

src/libServer/LookupServer.cpp
n-hutton: 👍

n-hutton: I think I made it slightly cleaner/safer using replace (on other branch - check it out?)

bzawisto: Yeah I checked there, looks ok, thanks.

PR #2868, opened 2022-08-11
[master] Fix payments in EVM

valeryz:
## Description

This is because we modeled it after Scilla, but in fact Scilla contracts by default don't accept payments, they need to specifically accept them. Where as in EVM, the incoming value is accepted unless the EVM execution reverts. 

Additionally, we've erroneously been modifying accounts in place, rather than modifying the "atomics".  Then atomics, when committed, would overwrite our modifications.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,7 +339,10 @@ endif()
 
 # installation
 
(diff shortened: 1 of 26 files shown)


CSideSteve (approved):
LGTM

PR #2873, opened 2022-08-11
[master] Enable metamask on devnet

n-hutton:
Add neccessary functions for metamask to operate against isolated server + devnet.

In order to provide a transaction receipt, it has to know the block which the TX can be found in. It does this currently by looking downward through the chain to see if it can find it. This is an interim solution as it will not scale.

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -75,7 +75,7 @@ int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
   string blocknum_str{"1"};
-  uint timeDelta{30000};
+  uint timeDelta{0};
   bool loadPersistence{false};
   bool nonisoload{false};
   string uuid;
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -75,7 +75,7 @@ int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
   string blocknum_str{"1"};
-  uint timeDelta{0};
+  uint timeDelta{50000};
   bool loadPersistence{false};
   bool nonisoload{false};
   string uuid;
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -75,7 +75,7 @@ int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
   string blocknum_str{"1"};
-  uint timeDelta{50000};
+  uint timeDelta{30000};
   bool loadPersistence{false};
   bool nonisoload{false};
   string uuid;
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -75,7 +75,7 @@ int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
   string blocknum_str{"1"};
-  uint timeDelta{1000};
+  uint timeDelta{0};
   bool loadPersistence{false};
   bool nonisoload{false};
   string uuid;
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -75,7 +75,7 @@ int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
   string blocknum_str{"1"};
-  uint timeDelta{0};
+  uint timeDelta{1000};
   bool loadPersistence{false};
   bool nonisoload{false};
   string uuid;
--- a/src/libServer/IsolatedServer.cpp
+++ b/src/libServer/IsolatedServer.cpp
@@ -252,7 +252,7 @@ IsolatedServer::IsolatedServer(Mediator& mediator,
   AbstractServer<IsolatedServer>::bindAndAddMethod(
       jsonrpc::Procedure("eth_blockNumber", jsonrpc::PARAMS_BY_POSITION,
                          jsonrpc::JSON_STRING, NULL),
-      &LookupServer::GetEthBlockNumberI);
+      &IsolatedServer::GetEthBlockNumberI);
 
   AbstractServer<IsolatedServer>::bindAndAddMethod(
       jsonrpc::Procedure("net_version", jsonrpc::PARAMS_BY_POSITION,
@@ -304,7 +304,7 @@ IsolatedServer::IsolatedServer(Mediator& mediator,
       jsonrpc::Procedure("eth_getTransactionReceipt",
                          jsonrpc::PARAMS_BY_POSITION, jsonrpc::JSON_STRING,
                          "param01", jsonrpc::JSON_STRING, NULL),
-      &LookupServer::GetTransactionReceiptI);
+      &LookupServer::GetEthTransactionReceiptI);
 
   AbstractServer<IsolatedServer>::bindAndAddMethod(
       jsonrpc::Procedure("eth_feeHistory", jsonrpc::PARAMS_BY_POSITION,
@@ -341,6 +341,12 @@ IsolatedServer::IsolatedServer(Mediator& mediator,
                          jsonrpc::JSON_STRING, "param01", jsonrpc::JSON_STRING,
                          "param02", jsonrpc::JSON_STRING, NULL),
       &LookupServer::GetEthCodeI);
+
+//  AbstractServer<IsolatedServer>::bindAndAddMethod(
+//      jsonrpc::Procedure("eth_blockNumber", jsonrpc::PARAMS_BY_POSITION,
+//                         jsonrpc::JSON_STRING,
+//                         NULL),
+//      &IsolatedServer::GetEthBlockNumberI);
 }
 
 bool IsolatedServer::ValidateTxn(const Transaction& tx, const Address& fromAddr,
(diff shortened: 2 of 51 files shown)


src/libServer/IsolatedServer.cpp
```diff
@@ -619,6 +619,8 @@ Json::Value IsolatedServer::CreateTransaction(const Json::Value& _json) {
 
 Json::Value IsolatedServer::CreateTransactionEth(EthFields const& fields,
                                                  bytes const& pubKey) {
+  Json::Value ret;
```
bzawisto, line 622: nitpick: any reason we moved this variable to the beginning of the method?

src/libServer/IsolatedServer.cpp
```diff
@@ -769,16 +769,26 @@ Json::Value IsolatedServer::CreateTransactionEth(EthFields const& fields,
...
     throw je;
   } catch (exception& e) {
     LOG_GENERAL(INFO, "[Error]" << e.what() << " Input code: NA");
     throw JsonRpcException(RPC_MISC_ERROR, "Unable to Process");
   }
+
+  // Double create a block to make sure TXs are 'flushed'
+  // This will make sure the block height advances, the
+  // TX can be found in a block etc.
+  if (m_timeDelta == 0) {
```
bzawisto, line 786: Sorry, I don't fully understand why we call it twice. Does it create genesis block on the first run? Is my understanding correct?

src/libServer/LookupServer.cpp
```diff
@@ -1106,37 +1135,79 @@ Json::Value LookupServer::GetEthBlockCommon(const TxBlock& txBlock,
                                                  includeFullTransactions);
 }
 
-Json::Value LookupServer::GetTransactionReceipt(const std::string& txnhash) {
-  Json::Value ret;
+Json::Value LookupServer::GetEthTransactionReceipt(const std::string& txnhash) {
+  auto hash = txnhash;
+
+  if (hash[0] == '0' && hash[1] == 'x') {
```
bzawisto, line 1141: I don't think you have to strip out the prefix as it's done automatically by `TransactionHash`.

bzawisto: Update: I see it's used at the later stage so it may be fine. However I'd recommend using Transaction/Block Hash types and do comparison on them instead of writing hand-written loops.

valeryz:
I like the title of this PR.  What does it to do the mainnet? :)

src/libServer/LookupServer.cpp
```diff
@@ -1106,37 +1135,79 @@ Json::Value LookupServer::GetEthBlockCommon(const TxBlock& txBlock,
...
+    auto height = txBlock.GetHeader().GetBlockNum() ==
+                          std::numeric_limits<uint64_t>::max()
+                      ? 1
+                      : txBlock.GetHeader().GetBlockNum();
+
+    std::string blockHash = "";
+
+    // Scan downwards through the chain until the TX can be found
+    do {
+      std::stringstream ss;
```
bzawisto, line 1160: nitpick: you can use std::to_string() which is more performant.

src/libUtils/DataConversion.cpp
```diff
@@ -81,6 +81,36 @@ bool DataConversion::Uint8VecToHexStr(const bytes& hex_vec, string& str) {
   return true;
 }
 
+bool DataConversion::HexStringsSame(const std::string& a,
```
bzawisto, line 84: This code may not be needed if we rely on `FixedHash<>` `operator==`.
So instead of comparing strings, you can do sth like this:
```TransactionHash hash_1{a};
TransactionHash hash_2{b};
if(a == b) {}```

File: src/libServer/IsolatedServer.cpp (hunk 1 above)
Comment (n-hutton) [reply src/libServer/IsolatedServer.cpp:634]: It now returns outside of the try... catch and so also has to be declared in that scope. It needs to return outside the scope it was originally defined in because there is a mutex that is locked in that scope which is also used by PostTxBlock (so would deadlock).

File: src/libServer/IsolatedServer.cpp (hunk 2 above)
Comment (n-hutton) [reply src/libServer/IsolatedServer.cpp:798]: I believe there is an issue around height 1 which causes this to be neccessary to satisfy metamask. The block height starts at 1 and if you submit a tx and mint a block it goes into block 1 which is not how the chain works usually as that is like injecting into something that is finalised.

For convenience I just mint 2 blocks but if if bothers you I can write a workaround so it only needs to mint 1.

File: src/libServer/LookupServer.cpp (hunk 3 above)
Comment (n-hutton) [reply src/libServer/LookupServer.cpp:1194]: I am just calling some other zil function `GetTransaction` and so satisfying its requirements (IIRC it breaks with the 0x prefix). Now I think of it, what we could/should do is to create a convenience function in DataProcessing that will strip off '0x' iff it exists, then we can do it cleanly when we need to. What do you think?

File: src/libUtils/DataConversion.cpp (hunk 5 above)
Comment (n-hutton) [reply src/libUtils/DataConversion.cpp:84]: that will work only if they are the right length, the casing is the same, and there are no '0x' to be seen - so I thought my function more general. I can switch it to your way if you like(?)

**Comment by n-hutton** [comment]
> I like the title of this PR. What does it to do the mainnet? :)

:) - that's funny

File: src/libServer/IsolatedServer.cpp (hunk 2 above)
Comment (bzawisto) [reply src/libServer/IsolatedServer.cpp:798]: It's still IsolatedServer so I think these type of workarounds are ok. Thanks for clarification.

File: src/libServer/LookupServer.cpp (hunk 3 above)
Comment (bzawisto) [reply src/libServer/LookupServer.cpp:1194]: Thanks for explanation. However, that's very interesting it breaks with `0x` prefix since `GetTransaction` internally creates immediately `TxnHash, which is of type dev::h256 == FixedHash<32>`. `FixedHash` itself, when instantiated with hex string uses:
`explicit FixedHash(std::string const& _s, ConstructFromStringType _t = FromHex, ConstructFromHashType _ht = FailIfDifferent): FixedHash(_t == FromHex ? fromHex(_s, WhenError::Throw) : dev::asBytes(_s), _ht) {}`
This calls underneath:
`bytes dev::fromHex(std::string const& _s, WhenError _throw)` which moves the index by 2 when hex strings has `0x` prefix. 
In general I think we should reduce any string comparisons and manual stripping of 0x and use built-in aliases for Block/Transaction/Other hashes. The only exception could be variable-length hex for contract code/calldata. What do you think?

File: src/libUtils/DataConversion.cpp (hunk 5 above)
Comment (bzawisto) [reply src/libUtils/DataConversion.cpp:84]: Thanks for explaining. But in the case here you use this method for comparing two tx hashes - which are supposed to be of the same length (if not, then one hash may be padded with zeros and would return false for equality anyway), different casing is ok (since hex representation doesn't matter as long underlying bytes are the same). What do you think?

Comment (n-hutton) [reply src/libUtils/DataConversion.cpp]: The code breaks not at this point but lower down when it checks the size of the string coming in (it throws). It is ok when you remove this checking code, so I have taken off the '0x' stripping and removed the checking code. Hashes that are too long or short fail in the same way as hashes that are not found, it seems, so all is well.

File: src/libServer/LookupServer.cpp (hunk 3 above)
Comment (n-hutton) [reply src/libServer/LookupServer.cpp]: The code breaks not at this point but lower down when it checks the size of the string coming in (it throws). It is ok when you remove this checking code, so I have taken off the '0x' stripping and removed the checking code. Hashes that are too long or short fail in the same way as hashes that are not found, it seems, so all is well.

I would say that I am not a massive fan (nor very confident it works) of the TxnHash style, but it is good to be consistent :+1:

File: src/libUtils/DataConversion.cpp (hunk 5 above)
Comment (n-hutton) [reply src/libUtils/DataConversion.cpp]: Yes, it is better this way - you can find the change in this most recent commit

File: src/cmd/isolated_server.cpp [hunk 6]
```diff
@@ -75,7 +75,7 @@ int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
   string blocknum_str{"1"};
-  uint timeDelta{1000};
```
n-hutton: I didn't intend to have this change committed

src/libServer/IsolatedServer.cpp
```diff
@@ -620,15 +620,22 @@ Json::Value IsolatedServer::CreateTransaction(const Json::Value& _json) {
...
     LOG_GENERAL(INFO,
                 "[Error]" << e.what() << " Input: " << _json.toStyledString());
     throw JsonRpcException(RPC_MISC_ERROR, "Unable to Process");
   }
+
+  // Double create a block to make sure TXs are 'flushed'
+  // This will make sure the block height advances, the
+  // TX can be found in a block etc.
+  if (m_timeDelta == 0) {
+    PostTxBlock();
```
n-hutton, line 635: Convenience - advance chain for both eth and non eth create TX methods

src/libServer/LookupServer.cpp
```diff
@@ -1198,14 +1198,8 @@ Json::Value LookupServer::GetEthBlockTransactionCountByNumber(
...
 Json::Value LookupServer::GetEthTransactionReceipt(const std::string& txnhash) {
-  auto hash = txnhash;
-
-  if (hash.size() >= 2 && hash[0] == '0' && hash[1] == 'x') {
-    hash.erase(0, 2);
-  }
-
   try {
-    auto const result = GetTransaction(hash);
+    auto const result = GetTransaction(txnhash);
```
n-hutton, line 1202: no need to strip 0x

bzawisto (approved):

PR #2878, opened 2022-08-15
[master] Fix constructor value sending and contracts creating contracts

valeryz:
## Description

This PR fixes multiple problems related to contracts creating other contracts.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStoreSC.tpp
+++ b/src/libData/AccountData/AccountStoreSC.tpp
@@ -468,11 +468,11 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
           Account::GetAddressForContract(fromAddr, fromAccount->GetNonce());
       // instantiate the object for contract account
       // ** Remeber to call RemoveAccount if deployment failed halfway
-      Account *contractAccount;
+      Account* contractAccount;
       if (isScilla) {
         if (!this->AddAccount(contractAddress, {0, 0})) {
           LOG_GENERAL(WARNING, "AddAccount failed for contract address "
-                      << contractAddress.hex());
+                                   << contractAddress.hex());
           error_code = TxnStatus::FAIL_CONTRACT_ACCOUNT_CREATION;
           return false;
         }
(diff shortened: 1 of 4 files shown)


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -442,45 +445,66 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
...
-                                 << contractAddress.hex());
-        error_code = TxnStatus::FAIL_CONTRACT_ACCOUNT_CREATION;
-        return false;
-      }
-      Account* contractAccount = this->GetAccount(contractAddress);
-      if (contractAccount == nullptr) {
-        LOG_GENERAL(WARNING, "contractAccount is null ptr");
-        error_code = TxnStatus::FAIL_CONTRACT_ACCOUNT_CREATION;
-        return false;
+      Account* contractAccount;
```
CSideSteve, line 471: This whole Atomics Area of the Accountstore is another obscure area of code that needs some clarity, should be addressed in a future refactoring.

src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -277,8 +277,7 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
         try {
           if (it->hasCode() && it->Code().size() > 0) {
             targetAccount->SetImmutable(
-                DataConversion::StringToCharArray("EVM" + it->Code()),
-                contractAccount->GetInitData());
+                DataConversion::StringToCharArray("EVM" + it->Code()), {});
```
CSideSteve, line 280: Is it ever possible that a smart contract would want to modify the contents of code retaining the previous value of data, for instance an account could have both scilla contracts and evm contracts and potentially an evm contract could change the contents of the scilla code field which is just a string and preserver whatever value of data was currently there.

CSideSteve (approved):
Apart from the questions I raised, LGTM

src/libData/AccountData/AccountStoreSC.tpp
valeryz: There's just no "data" in EVM accounts.
There's only balance, nonce, and storage (root).  If/when we start mixing EVM/Scilla in the same contract, we'll have to deal with this, but then we can't just all it SetImmutable anymore.

PR #2885, opened 2022-08-17
[master] Tweak incorrect chain ID string

n-hutton:
If you set the chainID to 0 or 1 you can get a confusing message when attempting pub key recover - this is less confusing.


valeryz (approved):
Seems safe and likely necessary.

valeryz (approved):

PR #2891, opened 2022-08-18
[master] Saeed fix when contract deploy

n-hutton:
When testing, it was discovered that the RLP will throw if you feed it incorrect data (in this case, it was the bytecode rather than a raw TX). I had thought that throws were caught by the http framework, but perhaps it only catches the rpc exceptions. In any case, this switches the behaviour to a fail, and it is handled further down.


bzawisto:
Nice! Can we have a unit test for this?

bzawisto (approved):

PR #2892, opened 2022-08-18
[master] Another two eth calls for getting tx from block

bzawisto:
## Description
This PR adds new eth calls for getting transactions from block by specifying index. Tested locally with unit tests.

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [X] local machine test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -393,6 +393,32 @@ LookupServer::LookupServer(Mediator& mediator,
       &LookupServer::GetEthBlockByHashI);
 
   this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_getBlockTransactionCountByHash",
+                         jsonrpc::PARAMS_BY_POSITION, jsonrpc::JSON_STRING,
+                         "param01", jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetEthBlockTransactionCountByHashI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_getBlockTransactionCountByNumber",
+                         jsonrpc::PARAMS_BY_POSITION, jsonrpc::JSON_STRING,
+                         "param01", jsonrpc::JSON_STRING, NULL),
+      &LookupServer::GetEthBlockTransactionCountByNumberI);
+
+  this->bindAndAddMethod(
+      jsonrpc::Procedure("eth_getTransactionByBlockHashAndIndex",
+                         jsonrpc::PARAMS_BY_POSITION, jsonrpc::JSON_STRING,
+                         "param01", jsonrpc::JSON_STRING, "param02",
+                         jsonrpc::JSON_STRING, NULL),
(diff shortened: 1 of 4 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -1159,6 +1185,93 @@ Json::Value LookupServer::GetEthBlockTransactionCountByNumber(
   }
 }
 
+Json::Value LookupServer::GetEthTransactionByBlockHashAndIndex(
+    const std::string& inputHash, const std::string& indexStr) const {
+  try {
+    const BlockHash blockHash{inputHash};
+    const auto txBlock = m_mediator.m_txBlockChain.GetBlockByHash(blockHash);
+    const uint64_t index = std::strtoull(indexStr.c_str(), nullptr, 0);
```
n-hutton, line 1193: This looks like it will not work as the index will likely be prepended with `0x` https://www.quicknode.com/docs/ethereum/eth_getTransactionByBlockHashAndIndex

src/libServer/LookupServer.cpp
```diff
@@ -1159,6 +1185,93 @@ Json::Value LookupServer::GetEthBlockTransactionCountByNumber(
...
+}
+
+Json::Value LookupServer::GetEthTransactionByBlockNumberAndIndex(
+    const std::string& blockNumberStr, const std::string& indexStr) const {
+  try {
+    TxBlock txBlock;
+    if (blockNumberStr == "latest") {
+      txBlock = m_mediator.m_txBlockChain.GetLastBlock();
+    } else if (blockNumberStr == "earliest") {
+      txBlock = m_mediator.m_txBlockChain.GetBlock(0);
```
n-hutton, line 1210: I think we might need to test on the isolated server + devnet that this shouldn't be `1` - IIRC the isolated server does start at 1, devnet might also depending on how genesis works

n-hutton (changes requested):

src/libServer/LookupServer.cpp
bzawisto: Thanks for raising that, however it seems it's going to work as expected. According to docs (https://en.cppreference.com/w/cpp/string/byte/strtoul):
Signature is:
`unsigned long long strtoull( const char *str, char **str_end, int base );`
And a comment on base (last parameter):
`If the value of base is ​0​, the numeric base is auto-detected: if the prefix is 0, the base is octal, if the prefix is 0x or 0X, the base is hexadecimal, otherwise the base is decimal.`

src/libServer/LookupServer.cpp
bzawisto: It's just a matter of interpretation I'd say. 0 is the first valid block mined ever (it's a genesis containing no transactions). So I'd say this is fine to access it here. In addition it's unlikely `earliest` has any useful use-case in our network - I guess no one really uses in on ethereum mainnet either.

n-hutton (approved):

PR #2894, opened 2022-08-19
[master] Move templates from .tpp to .cpp with explicit instantiation

valeryz:
## Description
Most of the change is actually recovering #includes lost because of lack of transitive includes from the removed .tpp

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/cmd/genTxnBodiesFromS3.cpp
+++ b/src/cmd/genTxnBodiesFromS3.cpp
@@ -25,6 +25,7 @@
 #include "libPersistence/BlockStorage.h"
 #include "libServer/JSONConversion.h"
 #include "libUtils/FileSystem.h"
+#include "libUtils/SysCommand.h"
 
 using namespace std;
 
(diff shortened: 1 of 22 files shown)


valeryz:
This is something like what @art-gor suggested.

And I just discussed with @CSideSteve .   

Measured the impact on build times, and the win is marginal.  On my 8-core Intel i7-6700 4 GHz (5 y.o. CPU) it is  7 minutes before with this PR vs. 6 minutes after.

Incremental build time impact on AccountStore is probably way more significant, but for the EVM work, we should just separate our code from the AccountStoreSC.tpp anyways. 

Bottom line: probably not worth it.

valeryz:
Update: this is now rebased to `release/8.2`

chetan-zilliqa (approved):

yaron-zilliqa (approved):

PR #2895, opened 2022-08-19
[master] Separate Scilla and EVM code to make it easier to verify changes from 8.1.2

CSideSteve:
First PR to tidy up code in AccountData

It is not writtent to win prizes or coding contests, it is a structure change only so if you criticise the tpp technique, find the original author and not me :-)

A simple change to seperate the code and reveet the scilla processing path back to the original code as used in production.

This EVM path is unchanged from previous versions so should have impact on current testing or development.
The EVM specific functions are now contained in their seperate file and not in AccountStoreSC.tpp

conversion of tpp to cpp has been halted as does not add value required today.

This is for internal testing Only. Has passed all tests with evm_ds_test.

## Description

To increase confidence that Scilla path is unchanged.
To make support of scilla and evm easier because you are dealing with one context and not two.
To improve general clarity.

## Backward Compatibility
should be 100% backwards compaitible for scilla and 100% compatible with EVM code today.

### Implementation

- [X] **ready for review**

Continue developing with this version.

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStore.cpp
+++ b/src/libData/AccountData/AccountStore.cpp
@@ -501,7 +501,24 @@ bool AccountStore::UpdateAccountsTemp(const uint64_t& blockNum,
   unique_lock<mutex> g2(m_mutexDelta, defer_lock);
   lock(g, g2);
 
-  if (ENABLE_EVM && EvmUtils::isEvm(transaction.GetCode())) {
+  bool isEvm{false};
+
+  if (Transaction:: GetTransactionType(transaction) ==
+      Transaction:: CONTRACT_CREATION) {
+    isEvm = EvmUtils:: isEvm(transaction.GetCode());
+  } else if (Transaction:: GetTransactionType(transaction) ==
+             Transaction:: CONTRACT_CALL) {
+    Account* contractAccount = this->GetAccount(transaction.GetToAddr());
+    isEvm = EvmUtils:: isEvm(contractAccount->GetCode());
+  } else {
+    isEvm = false;
+  }
+  if (ENABLE_EVM == false && isEvm) {
+    LOG_GENERAL(WARNING,
+                "EVM is disabled so not processing this EVM transaction ");
+    return false;
+  }
+  if (isEvm) {
     return m_accountStoreTemp->UpdateAccountsEvm(
         blockNum, numShards, isDS, transaction, receipt, error_code);
   } else {
--- a/src/libData/AccountData/AccountStore.cpp
+++ b/src/libData/AccountData/AccountStore.cpp
@@ -26,10 +26,10 @@
 #pragma GCC diagnostic ignored "-Wunused-parameter"
 #include "libPersistence/ScillaMessage.pb.h"
 #pragma GCC diagnostic pop
-#include "libServer/ScillaIPCServer.h"
-#include "libUtils/SysCommand.h"
 #include "EvmClient.h"
+#include "libServer/ScillaIPCServer.h"
 #include "libUtils/EvmUtils.h"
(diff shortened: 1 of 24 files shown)


valeryz (approved):

CSideSteve:
Ready to test - now a cpp file for evm.

src/libData/AccountData/AccountStore.cpp
```diff
@@ -497,8 +499,30 @@ bool AccountStore::UpdateAccountsTemp(const uint64_t& blockNum,
...
-  return m_accountStoreTemp->UpdateAccounts(blockNum, numShards, isDS,
-                                            transaction, receipt, error_code);
+  bool isEvm{false};
+
+  if (Transaction::GetTransactionType(transaction) ==
+      Transaction::CONTRACT_CREATION) {
+    isEvm = EvmUtils::isEvm(transaction.GetCode());
+  } else if (Transaction::GetTransactionType(transaction) ==
+             Transaction::CONTRACT_CALL) {
+    Account* contractAccount = this->GetAccount(transaction.GetToAddr());
```
valeryz, line 509: I think you should use `GetAccountTemp` here. probably there's no difference at the beginning. But it is more reliable. This is what's happening when the txn is processed in UpdateAccounts as it was before the change.

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -0,0 +1,661 @@
...
+template <class MAP>
+void AccountStoreSC<MAP>::EvmCallRunner(
+    INVOKE_TYPE invoke_type, EvmCallParameters& params, const uint32_t& version,
+    bool& ret, TransactionReceipt& receipt,
+    evmproj::CallResponse& evmReturnValues) {
+  auto call_already_finished{false};
+
+  auto worker = [this, &params, &invoke_type, &ret, &version,
+                 &call_already_finished, &evmReturnValues]() mutable -> void {
+    if (invoke_type == RUNNER_CREATE || invoke_type == RUNNER_CALL) {
```
valeryz, line 34: I think we don't need that invoke_type. Can probably remove the function argument.

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -0,0 +1,661 @@
...
+  if (invoke_type == RUNNER_CREATE) {
+    contractAccount->SetImmutable(DataConversion::StringToCharArray(
+                                      "EVM" + evmReturnValues.ReturnedBytes()),
+                                  contractAccount->GetInitData());
+  }
+  return gas;
+}
+
+template <class MAP>
+bool AccountStoreSC<MAP>::ViewAccounts(EvmCallParameters& params, bool& ret,
```
valeryz, line 251: ViewAccounts is not a very good name. I understand the analogy with UpdateAccounts, but that's not what it does.


How about "EvaluateEvmCall"?

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -0,0 +1,661 @@
...
+                                       std::string& result) {
+  TransactionReceipt rcpt;
```
valeryz, line 253: No transaction is involved, therefore there shouldn't be this `rcpt` variable.  If it is just needed by EvmCallRunner below, it would help to make this parameter optional (just a nullable pointer out output) and not create any recepipts when not necessary..

Having to create a suprios "receipt" and then discard it seems wrong.

valeryz (approved):
A few comments, please address. But otherwise a great change, thanks!

PR #2896, opened 2022-08-22
[master] Fix padding on R and S elements of the signature

valeryz:
## Description
Must handle R and S as numbers, and cast them to 256 bit arrays. Otherwise, bignums that start with at least 0 zeros would lose a byte in the encoding, making the signature invalid for the verifying code later.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


n-hutton (approved):
Looks great!

PR #2899, opened 2022-08-23
[master] Fix for incorrect chain ID when it is not 1

n-hutton:


src/libData/AccountData/AccountStoreSC.tpp
```diff
@@ -447,7 +447,9 @@ bool AccountStoreSC<MAP>::UpdateAccounts(const uint64_t& blockNum,
       // Check if the sender has enough balance to pay gasDeposit
       if (fromAccount->GetBalance() < gasDeposit + transaction.GetAmount()) {
         LOG_GENERAL(WARNING,
-                    "The account doesn't have enough gas to create a contract");
+                    "The account doesn't have enough gas to create a contract. Bal: "
```
valeryz, line 450: This is useful, but various stability reasons we like to keep the changes to AccountStoreSC.tpp to the bare minimum.    Makes sense to apply this change to Steve's PR: https://github.com/Zilliqa/Zilliqa/pull/2895

src/libServer/LookupServer.cpp
```diff
@@ -1324,6 +1324,9 @@ Json::Value LookupServer::GetEthTransactionReceipt(const std::string& txnhash) {
 
     std::string blockHash = "";
 
+    std::cout << "Getting TX receipt for: " << txnhash << std::endl;
```
valeryz, line 1327: Should we log to stdout, or use LOG_GENERAL?

n-hutton: This is just debug to assist saeed and me (only a draft pr now for collaboration)

src/libData/AccountData/AccountStoreSC.tpp
n-hutton: steves PR has been merged already - do you want me to exclude this change from master?

PR #2901, opened 2022-08-24
[master] Zil-4672 128 256 discrepancies in EVM cleanup

moboware:
## Description
#2837 ZIL-4672
be able to handle 256 bit balances to/from the evm and store them in the Account data.

## Backward Compatibility

## Review Suggestion
Most work is done in the Account.cpp/.h file that will convert balances from 256 to internal 128 bit value.
Extra types are defines in the BaseType.h
### Implementation
- [X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz:
The PR is unfortunately outdated, and most of it touches files not related to 128/256 problem anyways.  I think we solved the problem as a by-product of RFC-11 implementation anyways.

PR #2905, opened 2022-08-24
[master] Fixes after testing with web3

n-hutton:
When testing with saeed and web3, found two bugs, one with TX receipt and another with TX version/chain ID

### DIFF
--- a/src/libEth/Eth.cpp
+++ b/src/libEth/Eth.cpp
@@ -77,8 +77,6 @@ EthFields parseRawTxFields(std::string const &message) {
 
   ret.version = DataConversion::Pack(CHAIN_ID, 2);
 
-  std::cout << "version is: "  << ret.version << std::endl;
-
   // RLP TX contains: nonce, gasPrice, gasLimit, to, value, data, v,r,s
   for (auto it = rlpStream1.begin(); it != rlpStream1.end();) {
     auto byteIt = (*it).operator bytes();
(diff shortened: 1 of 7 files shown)


src/libEth/Eth.cpp
```diff
@@ -32,7 +33,7 @@ Json::Value populateReceiptHelper(std::string const &txnhash, bool success,
 
   ret["transactionHash"] = txnhash;
   ret["blockHash"] = blockHash;
-  ret["blockNumber"] = blockNumber;
+  ret["blockNumber"] = "0x" + blockNumber;
```
n-hutton, line 36: caused web3 to break

src/libEth/Eth.cpp
```diff
@@ -74,7 +75,7 @@ EthFields parseRawTxFields(std::string const &message) {
   int i = 0;
   // todo: checks on size of rlp stream etc.
 
-  ret.version = 65538;
+  ret.version = DataConversion::Pack(CHAIN_ID, 2);
```
n-hutton, line 78: Eth TXs do not know the native chain ID, so need to set it

src/libServer/JSONConversion.cpp
```diff
@@ -128,7 +129,7 @@ const Json::Value JSONConversion::convertTxBlocktoEthJson(
 
   Json::Value retJson;
 
-  retJson["number"] = std::to_string(txheader.GetBlockNum());
+  retJson["number"] = (boost::format("0x%x") % txheader.GetBlockNum()).str();
```
n-hutton, line 132: hex string with 0x

bzawisto (approved):

PR #2906, opened 2022-08-25
[master] Several fixes to support more recent versions of gcc & C++:

yaron-zilliqa:
* Replace std::bind1st which is no longer supported in C++17 by
    an equivalent lambda.
  * Instantiate a Retriever in IsolatedServer to avoid a warning (and
    crash) on a null pointer.

## Backward Compatibility

### Implementation

- [x] **Ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


CSideSteve (approved):
LGTM

PR #2907, opened 2022-08-25
[master] Automatically calculate the eth chain ID from network ID

n-hutton:
This makes the eth chain id in constants.xml obsolete


src/common/Constants.cpp
```diff
@@ -745,7 +745,4 @@ const std::string EVM_SERVER_BINARY{ReadConstantString(
     "EVM_SERVER_BINARY", "node.jsonrpc.", "/usr/local/bin/evm-ds")};
 const std::string EVM_LOG_CONFIG{ReadConstantString(
     "EVM_LOG_CONFIG", "node.jsonrpc.", "/usr/local/etc/log4rs.yml")};
-const std::string ETH_CHAINID{
-    ReadConstantString("ETH_CHAINID", "node.jsonrpc.", "0x814d")};
-const uint64_t ETH_CHAINID_INT{DataConversion::HexStringToUint64Ret(
-    ReadConstantString("ETH_CHAINID", "node.jsonrpc.", "0x814d"))};
\ No newline at end of file
+const uint64_t ETH_CHAINID{ReadConstantNumeric("CHAIN_ID") | 0x8000};
```
valeryz, line 748: I'd say it is better to add (+) than to (|).  The reason is that if we have to change 0x8000 to something other, which more bits set to 1,  it might not be equivalent. Whether we have to change it depends on chainlist.org and whether we'll be able to push that scheme there.

valeryz (approved):

PR #2908, opened 2022-08-25
[release/v8.2] oveddiding boost with v.1.79 by cmake script

art-gor:
## Description

We need a newer boost (1.79) in release/8.2 testnet and we need to be sure that this library upgrade doesn't break anything existing, and let it run on the testnet for some time

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,6 @@ addons:
       # Build dependency
       - build-essential
       - cmake
-#      - libboost-filesystem-dev
-#      - libboost-program-options-dev
-#      - libboost-system-dev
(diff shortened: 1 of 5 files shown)


chetan-zilliqa:
I think we will also need support for external boost in `build.sh` script, some miners compile Zilliqa and scilla codebase and run the mining node.

chetan-zilliqa (approved):
As discussed, We make it part of README build instructions for miners using native build

valeryz (approved):
Rubber stamping

PR #2909, opened 2022-08-25
More 0x fixes against web3

n-hutton:
Note: it is not enough to just prepend 0x - that would then make it seem as hex when it is actually decimal. It must be formatted correctly


CSideSteve (approved):
LGTM

PR #2911, opened 2022-08-25
[master] check first byte of txs to make sure it is supported

n-hutton:
EIP-2718 specifies a new format of TX that is not yet supported, the current TX verification is considered 'legacy'. If we see one of these new style TXs we can at least log it, until we have support for it.


src/libCrypto/EthCrypto.cpp
```diff
@@ -227,6 +227,15 @@ secp256k1_context const* getCtx() {
 // is in line with EIP-155.
 // message shall not contain '0x'
 bytes RecoverECDSAPubSig(std::string const& message, int chain_id) {
+  if (message.size() >= 2) {
+    auto const firstByte = DataConversion::HexStrToUint8VecRet(message)[0];
+    if (!(firstByte >= 0xc0 && firstByte <= 0xfe)) {
```
valeryz, line 232:
```suggestion
    // See https://eips.ethereum.org/EIPS/eip-2718, section "Backward Compatibility"
```

valeryz, line 232:
```suggestion
      // See https://eips.ethereum.org/EIPS/eip-2718 section "Backwards Compatibiltiy"
      if (!(firstByte >= 0xc0 && firstByte <= 0xfe)) {
```

valeryz (approved):

PR #2913, opened 2022-08-26
[master] Fixing receipt for contract deployment

bzawisto:
## Description
This PR fixes transaction receipt in case of contract deployment (contractAddress field should provide created contract address). Also there are a few fixes for returned block json. Tested on isolated server.

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [X] local machine test

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -1312,9 +1312,6 @@ Json::Value LookupServer::GetEthTransactionFromBlockByIndex(
 
 Json::Value LookupServer::GetEthTransactionReceipt(const std::string& txnhash) {
   try {
-    auto const ethResult = GetEthTransactionByHash(txnhash);
-    auto const zilResult = GetTransaction(txnhash);
-
     // Scan downwards looking for the block hash with our TX in it
     const auto txBlock = m_mediator.m_txBlockChain.GetLastBlock();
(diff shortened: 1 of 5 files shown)


src/libServer/LookupServer.cpp
```diff
@@ -1357,17 +1371,17 @@ Json::Value LookupServer::GetEthTransactionReceipt(const std::string& txnhash) {
...
-    }
-
     if (blockHash.size() > 2 && blockHash[0] != '0' && blockHash[1] != 'x') {
       blockHash = std::string("0x") + blockHash;
     }
 
-    auto res = populateReceiptHelper(hashId, success, sender, toAddr, cumGas,
-                                     blockHash, blockNumber);
+    // TODO: check if EVM-DS expects constructor calldata to be passed in 'data'
+    // field.
```
valeryz, line 1379: I can asnwer the question in this TODO:  It does expect calldata to be passed in the data field.

But hey, you cannot distinguish contract deployment or not by just looking at the data field.
You can easily have a contract call with an empty calldata. In fact, this is quite common.

The right check is destination address being zero. See `Transaction::GetTransactionType()`

valeryz (changes requested):

valeryz (approved):

PR #2918, opened 2022-08-29
[master] Zil 4695 eth balance subunit must be in wei 10 18

moboware:
## Description
Return Balance from the Eth api in wei
See also RFC-11

## Backward Compatibility

### Implementation
- [X] **ready for review**
- EvmDsTest is not changed, api should be tested with new test framework
- This change is depended on the EvmDs that supports a ZilScalingFactor
 
- [X] local machine test

- [ ] small-scale cloud test


src/libServer/LookupServer.cpp
```diff
@@ -1183,6 +1183,23 @@ Json::Value LookupServer::GetEthBlockCommon(const TxBlock& txBlock,
...
+Json::Value LookupServer::GetEthBalance(const std::string& address) {
+  const auto balanceStr = this->GetBalance(address, true)["balance"].asString();
+
+  const uint256_t ethBalance =
+      std::strtoll(balanceStr.c_str(), nullptr, 16) * EVM_ZIL_SCALING_FACTOR;
+
+  std::stringstream strm;
+  strm << "0x" << std::hex << ethBalance;
+
+#if 1
```
bzawisto, line 1195: Is it a leftover from testing? If we need these debug calls, can we remove conditional `if 1` otherwise?

bzawisto (approved):

PR #2919, opened 2022-08-29
Move templates from .tpp to .cpp with explicit instantiation (master)

valeryz:
## Description
It is the same change as was already merged to release8.2 branch here (https://github.com/Zilliqa/Zilliqa/pull/2894) but for master

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


CSideSteve (approved):
good to go

PR #2921, opened 2022-08-29
[master] Add evm_ds_test to the main repository

valeryz:
## Description
Move `evm_ds_test` to the main repository. After this is merged, the `evm_ds_test` repo will be deprecated.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/license_checker.sh
+++ b/scripts/license_checker.sh
@@ -48,6 +48,8 @@ scope=$(find . -type f \( \
             ! -path "./scilla/*" \
             ! -path "./cmake-build-debug/*" \
             ! -path "./evm-ds/*" \
+            ! -path "./evm-ds/*" \
+            ! -path "./tests/EvmAcceptanceTests/node_modules/*" \
             ! -path "./src/depends/*" \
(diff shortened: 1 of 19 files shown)


its-saeed (approved):
Just one note. Folder names in `tests` folder are in camelCase format, not snake_case. so maybe it's better to name `evm_ds_tests`, to something like `evmDsTests`

valeryz:
> Just one note. Folder names in `tests` folder are in camelCase format, not snake_case. so maybe it's better to name `evm_ds_tests`, to something like `evmDsTests`

I appreciate the desire for consistency.  But the fact that it is unlike others makes it stand out and look weird. This actually is good, because it seems it doesn't belong there :)  Which is true.

Maybe we should rearrange, and put unit tests into another subdirectory, acceptance (though I'd rather call them integration tests) separately, and then other test categories etc. etc.   Not just have everything in one test subdir?

valeryz:
I'll submit it now, untill we decide how to rearrange.

PR #2925, opened 2022-08-30
[master] feat: add Jenkins pipeline

frankmeds:
## Backward Compatibility

### Implementation

- [X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,50 @@
+def podDefinition = """
+apiVersion: "v1"
+kind: "Pod"
+spec:
+  nodeSelector:
+    kubernetes.io/os: "linux"
+  containers:
+  - name: "scilla"
+    image: "zilliqa/scilla:v0.11.0"
+    command:
+    - cat
+    tty: true
+  - name: "ubuntu"
+    image: "ubuntu:bionic"
+    command:
+    - cat
+    tty: true
+"""
+
+podTemplate(yaml: podDefinition) {
+  node(POD_LABEL) {
+    stage('Checkout scm') {
+        checkout scm
+    }
+    stage('Import Scilla') {
+        container('scilla') {
+            sh "mkdir ./moving_folder"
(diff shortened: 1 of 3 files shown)


pipeline.sh
```diff
@@ -0,0 +1,40 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+# This script is dedicated for CI use
+#
+
+set -x
+set -e
+
+export scilla_container_id=$(docker run -d --name zilliqa_scilla_$HOSTNAME zilliqa/scilla:v0.11.0 sleep infinity)
```
valeryz, line 23: Are we sure we'll be using Scilla v 11.0 for master?    I thought it was more for 8.2.  Maybe that's correct, just pointing out that it would make sense to extract it into some variable so that configuration becomes more obvious.

valeryz (approved):

pipeline.sh
frankmeds: I implemented in Jenkins just a replica of the current process in Travis. I am not sure of what should be the right version, but a good time to change if needed.

Reference: https://github.com/Zilliqa/Zilliqa/blob/master/.travis.yml#L61

CSideSteve: The current build of evm does infact use version 0.11.0, this caused some issues when we moved to ubuntu 20.04 as scilla team scrambled to get some imprtant fixes from the latest version into 0.11.0 ubuntu version 20.04.

I will add a ticket into jira for investigating upgrading to what scilla team have in production and make sure we are correct..

CSideSteve (approved):

pipeline.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
n-hutton, line 1: Minor comment, but I would consider moving ci_script and pipeline.sh to the scripts directory

PR #2929, opened 2022-08-31
don't licence check local files

n-hutton:


scripts/ci_build.sh
```diff
@@ -70,7 +70,7 @@ then
     ./scripts/ci_xml_checker.sh constants_local.xml || exit 1
     ./scripts/license_checker.sh || exit 1
     ./scripts/depends/check_guard.sh || exit 1
-    cmake --build ${dir} --target clang-format || exit 1
+    #cmake --build ${dir} --target clang-format || exit 1
```
its-saeed, line 73: is it intentional?

n-hutton: Yes - jenkins cannot find clang-format and it fails so this is a temp fix

its-saeed (approved):

PR #2932, opened 2022-09-01
[master] Remove dependency on pyzil. Normalize amounts in tests

valeryz:
## Description
`evm_ds_test` now only depends on the web3 (eth compatible) APIs. No Zilliqa APIs should be used.

To make this work in your local environment,  replace the address `381f4008505e940ad7681ec3468a719060caf796` with `f0cb24ac66ba7375bf9b9c4fa91e208d9eaabd2e`. This is the address correspoding to the private key `d96e9eb5b782a80ea153c937fa83e5948485fbfc8b7e7c069d7b914dbc350aba`.

Correspondingly, the same address is replaced in the genesis accounts fed to the devnet deployment scripts with ` --genesis-address-file` argument:

`s/381F4008505E940AD7681EC3468A719060CAF796/F0CB24AC66BA7375BF9B9C4FA91E208D9EAABD2E/`

The same account in the Internal EVM Testnet has been generously funded.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/evm_ds_test/tests/conftest.py
+++ b/tests/evm_ds_test/tests/conftest.py
@@ -19,6 +19,7 @@ import os
 import numpy
 import random as rand
 
+
 @pytest.fixture(scope="session", autouse=True)
 def random():
     seed = int(os.environ.get("TESTSEED", "0"))
--- a/tests/evm_ds_test/evm_ds_test/api.py
+++ b/tests/evm_ds_test/evm_ds_test/api.py
@@ -13,4 +13,3 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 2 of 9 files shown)


tests/evm_ds_test/tests/conftest.py
```diff
@@ -19,6 +19,7 @@
 import numpy
 import random as rand
 
+
```
its-saeed, line 22: unnecessary

its-saeed (approved):

tests/evm_ds_test/tests/conftest.py
valeryz: it is coming from 'black', the formatting tool

PR #2934, opened 2022-09-01
[master] block timestamp is in seconds since unix epoch

SashaZezulinsky:
## Description

JSON-RPC API calls and Solidity expects block timestamp in seconds since unix epoch

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/TimeUtils.cpp
+++ b/src/libUtils/TimeUtils.cpp
@@ -56,7 +56,7 @@ std::string microsec_timestamp_to_readable(const uint64_t timestamp) {
 
 long int microsec_to_sec(const uint64_t timestamp) {
   std::chrono::microseconds micro(timestamp);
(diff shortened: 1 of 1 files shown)


valeryz (approved):

src/libServer/JSONConversion.cpp
```diff
@@ -158,11 +160,13 @@ const Json::Value JSONConversion::convertTxBlocktoEthJson(
 
   bytes serializedTxBlock;
   txblock.Serialize(serializedTxBlock, 0);
+  auto timestamp =
+      std::chrono::duration_cast<std::chrono::seconds>(txblock.GetTimestamp());
```
valeryz, line 164: pls check the CI, it looks like it doesn't build at this change.

valeryz (approved):

PR #2935, opened 2022-09-01
[master] Different eth tx validation

bzawisto:
This PR introduces another path in isolated/lookup server for validating transactions created in eth-way.
I also picked up an arbitrary value of 21000 in wei to be minimal gas price for eth transactions.

Update: Gas wrapper will come in separate PR.

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [] local machine test

- [ ] small-scale cloud test


constants_local.xml
```diff
@@ -103,6 +103,7 @@
         <REMOVENODEFROMBLACKLIST_DELAY_IN_SECONDS>1</REMOVENODEFROMBLACKLIST_DELAY_IN_SECONDS>
     </epoch_timing>
     <gas>
+        <MIN_ETH_GAS_PRICE_IN_WEI>21000</MIN_ETH_GAS_PRICE_IN_WEI>
```
valeryz, line 106: This suspiciously looks like min txn gas, rather than its price. Are you sure about it?

src/libServer/IsolatedServer.cpp
```diff
@@ -443,6 +443,52 @@ bool IsolatedServer::ValidateTxn(const Transaction& tx, const Address& fromAddr,
   return true;
 }
 
+bool IsolatedServer::ValidateEthTxn(const Transaction& tx,
```
valeryz, line 446: Can we extract this method in a separate class in a separate file etc.?   I'd be great to eventually remove many of our modifications from IsolatedServer.cpp to a separate file, so that we exactly know the diff of EVM work and it is small in the preexisting code.

src/libServer/IsolatedServer.cpp
```diff
@@ -443,6 +443,52 @@ bool IsolatedServer::ValidateTxn(const Transaction& tx, const Address& fromAddr,
...
+  }
+
+  if (IsNullAddress(fromAddr)) {
+    throw JsonRpcException(ServerBase::RPC_INVALID_ADDRESS_OR_KEY,
+                           "Invalid address for issuing transactions");
+  }
+
+  if (sender == nullptr) {
+    throw JsonRpcException(ServerBase::RPC_INVALID_ADDRESS_OR_KEY,
+                           "The sender of the txn has no balance");
```
valeryz, line 479: The message doesn't correspond to the code.

src/libServer/IsolatedServer.cpp
```diff
@@ -698,29 +738,26 @@ Json::Value IsolatedServer::CreateTransactionEth(EthFields const& fields,
 
       const Account* sender = AccountStore::GetInstance().GetAccount(fromAddr);
 
-      if (!ValidateTxn(tx, fromAddr, sender, m_gasPrice)) {
+      if (!ValidateEthTxn(tx, fromAddr, sender)) {
         return ret;
       }
 
+      senderBalance = uint256_t{sender->GetBalance()} * EVM_ZIL_SCALING_FACTOR;
```
valeryz, line 745: @moboware could you take a look if this makes sense and should be done like this?

src/libServer/IsolatedServer.cpp
valeryz: perhaps in another PR though

src/libServer/LookupServer.h
```diff
@@ -391,11 +391,9 @@ class LookupServer : public Server,
     auto fields = parseRawTxFields(rawTx);
 
     auto shards = m_mediator.m_lookup->GetShardPeers().size();
-    auto currentGasPrice =
-        m_mediator.m_dsBlockChain.GetLastBlock().GetHeader().GetGasPrice();
 
-    auto resp = CreateTransactionEth(fields, pubKey, shards, currentGasPrice,
-                                     m_createTransactionTarget);
+    auto resp =
+        CreateTransactionEth(fields, pubKey, shards, m_createTransactionTarget);
```
valeryz, line 396: Are you sure it is correct to remove the current Gas price and only check against the minimum?

Btw, I'm happy with the change as it is, and we might actually do it like that until we figure out our gas approach.  But please leave a TODO comment about that.

src/libServer/IsolatedServer.cpp
bzawisto: Sure, will do.

src/libServer/IsolatedServer.cpp
bzawisto: Yep, will fix.

src/libServer/LookupServer.h
bzawisto: I think you're right, I misunderstood that with minimum price for legacy transactions for some reason. Probably we need to check the price against `max(currentGasPrice * EVM_ZIL_SCALING_FACTOR, MIN_ETH_GAS_PRICE_IN_WEI)`. What do you think?

valeryz: Theoretically, yes.  But we also simultaneously need a separate gas price for EVM. (See https://zilliqa-jira.atlassian.net/browse/ZIL-4718).

valeryz (approved):

PR #2940, opened 2022-09-02
[master] Port ethereum RPC tests to our testing framework

its-saeed:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/helper/Utils.js
+++ b/tests/EvmAcceptanceTests/helper/Utils.js
@@ -1,0 +1,62 @@
+var axios = require('axios'),
+    _ = require('underscore')
+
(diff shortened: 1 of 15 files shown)


valeryz:
Why do we have to port them? Why can't we just run them directly against our network?

CSideSteve (approved):
33 passing (601ms)
  13 pending
  11 failing

I am not close enough to the contract testing to cinfrm everyline is a good change but its a test tool for other to enhance so I grant my approval on the basis it does what you told me it should do and infact my version has more passes than the guidance you gave.

PR #2942, opened 2022-09-03
Stability improvements: revert #2854 and bump CONNECTION timeout to 1 minute

valeryz:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz:
So this has been deployed on evmdev, and shows reasonable performance.  Seepubs are busylooping quite infrequently, CPU usage is about 1 core when it is quiet, and up to 6-7 cores when we run the test suite. (Seemingly those cores are busy with the persistence DB, not connecitons).

The reason why timeouts affect this, is MHD getting into the state MHD_EVENT_LOOP_INFO_BLOCK (see daemon.c in microhttpd source, which causes MHD to think there's more data already pending on some socket, which causes it to set epoll_wait timeout to zero).  By increasing timeout, we are reducing the instances of this problem.  (And why would a timeout on socket connection be set to just 1 second anyways?)

valeryz:
https://grafana-zilevm20220804.dev.z7a.xyz/d/6581e46e4e5c7ba40a07646395ef7b23/kubernetes-compute-resources-pod?orgId=1&var-datasource=Prometheus&var-cluster=&var-namespace=default&var-pod=evmdev-seedpub-0&from=1662319666566&to=1662343166002

PR #2944, opened 2022-09-05
Adding logs and logsBloom to receipt

bzawisto:
This PR exposes logs and logsBloom generated in transaction receipt. 
## Backward Compatibility

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test


src/libEth/Eth.cpp
```diff
@@ -45,7 +43,7 @@ Json::Value populateReceiptHelper(std::string const &txnhash, bool success,
   ret["cumulativeGasUsed"] = gasUsed.empty() ? "0x0" : gasUsed;
   ret["from"] = from;
   ret["gasUsed"] = gasUsed.empty() ? "0x0" : gasUsed;
-  ret["logs"] = Json::arrayValue;
+  ret["logs"] = logs;
```
valeryz, line 46: Have you checked that the logs are coming in the format expected by the ETH APIs?

src/libEth/Eth.cpp
```diff
@@ -45,7 +43,7 @@ Json::Value populateReceiptHelper(std::string const &txnhash, bool success,
...
   ret["logsBloom"] =
```
valeryz, line 47: Bloom filter all zeros made sense when all logs are always empty.  When you are returning someting in "logs", an empty bloom filter will just prevent client from looking at that logs member.  I.e. the client just looks in the bloom filter first, then if successful, tries to find a log entry.

I.e. if not implemented, rather put all 1's (0xffff...) zeros mean there are no logs at all.

But better implement the calculation of the bloom filter, it is not hard at all, just set the bit in the vector corresponding to the value of the hash of the entry.  But please read up on specifics of the endiannes.

valeryz: Maybe test is with some client with all ffff's in the bloom first - whether the "logs" work in JSON.

bzawisto: I think it's better to implement it in a proper way, will see how that should be done by checking ethereum yellow paper.

src/libUtils/BlockTransactionsHelper.h
```diff
@@ -0,0 +1,33 @@
...
+#define ZILLIQA_SRC_LIBUTILS_BLOCKTRANSACTIONSHELPER_H_
+
+#include <boost/optional.hpp>
+#include <string>
+
+class TxBlockChain;
+
+class BlockTransactionsHelper {
+ public:
+  static boost::optional<uint32_t> GetTransactionIndexInBlock(
```
valeryz, line 28: nit: you don't need a separate class for a utility function. Especially when it is used in one .cpp file - can keep it there as 'static' or in an anonymous namespace.

valeryz (approved):

PR #2948, opened 2022-09-06
[master] Adding gas converter

bzawisto:
## Description
This PR adds support for scaling gas units. Assumed gas price is based on sputnik VM gas costs. The goal is to have consistent gas representation in scilla and evm, still allowing ethereum based client libraries to use 'traditional' gas prices users use with other ethereum networks also at the same time preventing txns from being rejected by shards because of too high gas values. 

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test


src/libData/AccountData/Transaction.cpp
```diff
@@ -227,7 +228,14 @@ const uint128_t Transaction::GetGasPriceWei() const {
...
 
-const uint64_t& Transaction::GetGasLimit() const { return m_coreInfo.gasLimit; }
+uint64_t Transaction::GetGasLimit() const {
+  if (IsEth()) {
+    return GasConv::GasUnitsFromEthToCore(m_coreInfo.gasLimit);
+  }
+  return m_coreInfo.gasLimit;
+}
+
+uint64_t Transaction::GetGasLimitRaw() const { return m_coreInfo.gasLimit; }
```
valeryz, line 238: It would be helpful if you defined `GetGasLimitEth` and `GetGasLimitZil` for readability.  It is hard to understand.  Why would I want to use 'Raw' for Eth, for instance?

Then you may need to convert in the other direction.  If it is an non-Eth transaction, but is somehow used in an Eth calculation context, you'll scale it too.  (There's a use case for that in the future).

Then the only place where you use raw is in signatures.  OTherwise you always use either *Zil or *Eth, and there's no need to guess.

src/libEth/Eth.cpp
```diff
@@ -158,9 +159,9 @@ bool ValidateEthTxn(const Transaction &tx, const Address &fromAddr,
                                gasPriceWei.convert_to<std::string>());
   }
 
-  if (tx.GetGasLimit() < MIN_ETH_GAS) {
+  if (tx.GetGasLimitRaw() < MIN_ETH_GAS) {
```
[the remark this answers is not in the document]
valeryz: Sorry, I re-read the above, and it looks like it needs to be Raw.  But I'd prefer to use 'Raw' only in the signature calculation code (and even then it will probably use m_coreInfo directly).

src/libServer/IsolatedServer.cpp
```diff
@@ -689,7 +689,7 @@ Json::Value IsolatedServer::CreateTransactionEth(Eth::EthFields const& fields,
     // Sender's balance should be higher than value sent in the transaction +
     // max gas to be used by contract action
     const uint256_t requiredGas =
-        uint256_t{tx.GetGasPriceWei()} * uint256_t{tx.GetGasLimit()};
+        uint256_t{tx.GetGasPriceWei()} * uint256_t{tx.GetGasLimitRaw()};
```
valeryz, line 692: Same here, probably "Eth" suffic would be more readable.

valeryz (approved):

PR #2959, opened 2022-09-08
Eth Tx hash generation

n-hutton:
There are two differences wrt. hashing that have not been covered yet: how TX hashes are generated and how contract addresses are derived. Ether.js expects the TX hash to conform and so complains if it is done the zil way.

Solution: TX hashes are now created the Eth way if they are an eth TX (version 2). However in order to do this there were some important trade-offs which I will describe below.

The TX hash in ethereum is a hash of the serialized TX (RLP). This means that from the TX class in the codebase we need to get back to what RLP was used to create the TX (imagine you recieve the TX via gossip, you will not have the RLP the seed node saw).

This has a few issues:
- Technically, multiple different RLP could map to the same TX in the future (legacy/new TXs), although we only support one way for now
- In Ethereum transactions, they store (V, R, S) cryptographic numbers. The Schnorr library only allows for (R, S). So, we do not know what V was when trying to recreate the RLP
- ^ Solution/hack - V can only be 0 or 1 so we can try both and see which one successfully recreates the public key in the TX

Note: the 'better' way would be to modify the schnorr library to accept this but it would be a pretty big code change.

Note: the schnorr library locks a mutex every time it converts bignums to strings etc. Which seems very inefficient.

I ran a sample script that deploys a contract successfully, so it does not appear to mind that contract address is not eth-style generated (although I will also get this in).

Note: a few tests are disabled here until I can fix them - it will be time consuming so to get this feature in as more Eth functionality needs to be replicated in the codebase, the tests will be in a follow-up PR.


src/libCrypto/EthCrypto.cpp
```diff
@@ -52,6 +53,44 @@ auto epFree = [](EC_POINT* b) { EC_POINT_free(b); };
 auto esFree = [](ECDSA_SIG* b) { ECDSA_SIG_free(b); };
 auto ekFree = [](EC_KEY* b) { EC_KEY_free(b); };
 
+bool PubKeysSame(bytes const& pubkA, PubKey const& pubkB) {
```
n-hutton, line 56: Moved this code to a free function for cleanliness

src/libData/AccountData/Transaction.cpp
```diff
@@ -265,16 +245,48 @@ bool Transaction::IsSignedECDSA() const {
                               sigString, pubKeyStr);
 }
 
+// Set what the hash of the transaction is, depending on its type
+bool Transaction::SetHash(bytes const& txnData) {
```
n-hutton, line 249: Here we switch the way that the TX hash is calculated

src/libMessage/Messenger.cpp
```diff
@@ -1240,24 +1240,19 @@ bool ProtobufToTransaction(const ProtoTransaction& protoTransaction,
     return false;
   }
 
-  SHA2<HashType::HASH_VARIANT_256> sha2;
```
n-hutton, line 1243: The code for checking TX hash was outside of TX construction here - that seemed wrong/unclean. Let the TX create its own hash then compare that to what we think it should be.

tests/EvmLookupServer/Test_EvmLookupServer.cpp
```diff
@@ -650,7 +650,8 @@ BOOST_AUTO_TEST_CASE(test_eth_get_balance) {
                     "0xe8d4a51000");
 }
 
-BOOST_AUTO_TEST_CASE(test_eth_get_block_by_number) {
+BOOST_AUTO_TEST_CASE(test_eth_get_block_by_number,
```
n-hutton, line 653: tests disabled until a later PR - the TXs now need to be well formed/signed and we know the RLP in the case they are Eth TXs - which is going to be a little time consuming.

valeryz:
A general comment: why are we using the Schnorr library for the ECDSA transactions?  Should we get rid of that dependency? Too big of a code change @n-hutton? (not asking you to do it though - compromises to move faster are fine at this point)

src/libCrypto/EthCrypto.cpp
```diff
@@ -435,3 +504,18 @@ bytes StripEVM(bytes const& in) {
...
+  auto const asBytes = DataConversion::HexStrToUint8VecRet(rawTx);
+
+  auto const hash = ethash::keccak256(asBytes.data(), asBytes.size());
+
+  bytes hashBytes;
+
+  hashBytes.insert(hashBytes.end(), &hash.bytes[0], &hash.bytes[32]);
+
+  auto asString = DataConversion::Uint8VecToHexStrRet(hashBytes);
+  boost::algorithm::to_lower(asString);
```
valeryz, line 518: Why do you need this lowercase string here if hashBytes are then returned.

valeryz (approved):
LGTM, with one fix required for R, S padding (or rather padding removal) - they are bignums, not byte arrays, makes a difference.

src/libCrypto/EthCrypto.cpp
n-hutton: good spot! that was just used for debug and can be deleted.

PR #2961, opened 2022-09-08
[master] fix getBalance for Metamask compatibility

valeryz:
## Description
We need to return a result in eth_getBalance if/when we get a block number, even if can only serve the latest one.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz:
This PR also fixes a few errors in the previous implementation, and removes a redundant method.

Currently deployed on evmdev.  Please check your MM wallets.

moboware (changes requested):
Make a unit test to check the scaling on 'overflow' number works

valeryz:
@moboware thinking about this Unit test. It cannot really overflow, as accounts are 128 bit and are only multiplied by 1,000,000.     
Using SafeMath there is a bit of defensive programming here.

n-hutton (approved):

valeryz:
> Make a unit test to check the scaling on 'overflow' number works

I don't think I can make a unit test for a scenario that never happens. Defensive programming ...

PR #2890, opened 2022-08-18
[master] feat/ZIL-4688: extract dependencies out to vcpk

yaron-zilliqa:
Most dependencies will be compiled by vcpkg automatically as part of manifest mode (a prerequisites is to install vcpkg, following the instruction at https://vcpkg.io/en/getting-started.html).

When configuring with CMake, the following has to be passed: -DCMAKE_TOOLCHAIN_FILE=_/path/to/vcpkg_/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-linux-dynamic

When using build.sh, VCPKG_ROOT must be defined in the environment.

Dependencies are taken from the 2022.07.25 of vcpkg.

PR #2967, opened 2022-09-12
[master] compile dependencies using vcpk

yaron-zilliqa:
_(This PR is identical to https://github.com/Zilliqa/Zilliqa/pull/2890 - had to reraise due to a Travis rule preventing the check from going through when the PR is raised from a fork)._

Most dependencies will be compiled by vcpkg automatically as part of manifest mode (a prerequisites is to install vcpkg, following the instruction at https://vcpkg.io/en/getting-started.html).

When configuring with CMake, the following has to be passed: -DCMAKE_TOOLCHAIN_FILE=/path/to/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-linux-dynamic

When using build.sh, VCPKG_ROOT must be defined in the environment.

Dependencies are taken from the 2022.07.25 of vcpkg.


valeryz (approved):
I think it is safe to stamp it, as we had the previous PR.

PR #2970, opened 2022-09-12
[master] Actual contract addr generation (test code)

n-hutton:
Add the code that generates eth style contract addresses


valeryz (approved):
but you'll hook it up to contract creation in the next PR?

I think there are quite many places where it is being inferred.

PR #2979, opened 2022-09-13
[master] On isolated server, return a TX hash even if it fails

n-hutton:
as in title


src/libServer/IsolatedServer.cpp
```diff
@@ -25,6 +25,9 @@
 using namespace jsonrpc;
 using namespace std;
 
+const char* ZEROES_HASH =
+    "0x0000000000000000000000000000000000000000000000000000000000000";
```
n-hutton, line 29: If things go fubar, return a 0s hash as specified in eth api

src/libUtils/DataConversion.h
```diff
@@ -153,6 +153,9 @@ class DataConversion {
   /// Normalize alphanumeric hex string to lower and remove prefix "0x"
   static bool NormalizeHexString(std::string& s);
 
+  /// Add '0x' to string if it does not exist already
+  static std::string AddOXPrefix(std::string&& s);
```
n-hutton, line 157: convenience fn. Useful?

valeryz (approved):

PR #2995, opened 2022-09-15
[master] Nonce fix

n-hutton:
Zil nonces work as follows: the account starts with a nonce of 0. Incoming TXs have to have a nonce higher than this. So, the first nonce is 1. In Eth, the first nonce would be 0.

To rectify this, after the TX has arrived, increment the nonce held in the TX class. The nonce must be decremented to get back to what was sent or when querying via the ethereum API.


bzawisto:
Can we pack all logic into `tx.GetNonce()` so it knows what value should be returned based on tx type? Currently increments/decrements are done in a couple of places which is risky in case someone wants to do some other operations and it'd hard to figure out what nonce (either identity or +1) should be used.

valeryz (approved):

valeryz:
> Can we pack all logic into `tx.GetNonce()` so it knows what value should be returned based on tx type?

I'd agree, but if we look carefully at the PR,  most of the increments / decrements are done within implemenation of the Tx anyways. One place is outside of it, and one place has to do with the account nonce, not transaction nonce.  So I wouldn't bother. @n-hutton your call.

n-hutton:
@bzawisto I would say that ultimately, you have to do some incrementing and decrementing outside of getNonce either way (for account checking) - I don't mind changing it so that getNonce is aware of the difference.

n-hutton:
@bzawisto @valeryz on second thought it does not appear to be cleaner to hide the increment behind GetNonce - because the nonce is in the transaction core info which gets passed around and accessed directly as a struct

PR #3000, opened 2022-09-16
[master] Fix gas scaling

valeryz:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/Transaction.cpp
+++ b/src/libData/AccountData/Transaction.cpp
@@ -207,7 +207,7 @@ const uint128_t& Transaction::GetGasPriceRaw() const {
 
 const uint128_t Transaction::GetGasPriceQa() const {
   if (IsEth()) {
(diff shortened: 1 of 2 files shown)


valeryz:
@bzawisto could you please check whether gas price scaling at the class level could conflict with scaling in other places.

bzawisto:
> @bzawisto could you please check whether gas price scaling at the class level could conflict with scaling in other places.

@valeryz  I think we probably need to re-introduce getGasPrice and getGasLimit methods and used them throughout the code and make the ones with Wei/Qa or Zil/Eth suffixes private because it has become a bit complicated now - this is just a transaction that should know how much gas it should advertise - all conditional logic should be hidden inside one function. 
What do you think?

bzawisto (approved):

valeryz:
@bzawisto I think I agree with your proposal.

PR #3010, opened 2022-09-20
[master] vrs add to signature

n-hutton:
Re-use the functions in EthCrypto. Tests to follow.

### DIFF
--- a/src/libCrypto/EthCrypto.cpp
+++ b/src/libCrypto/EthCrypto.cpp
@@ -17,6 +17,7 @@
 
 #include "EthCrypto.h"
 #include <boost/algorithm/string.hpp>
+#include <boost/format.hpp>
 #include "libData/AccountData/Address.h"
 #include "libUtils/DataConversion.h"
 #include "libUtils/Logger.h"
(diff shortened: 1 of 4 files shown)


src/libCrypto/EthCrypto.cpp
```diff
@@ -415,7 +416,7 @@ bytes GetOriginalHash(TransactionCoreInfo const& info, uint64_t chainId) {
 
 // From a zilliqa TX, get the RLP that was sent to the node to create it
 std::string GetTransmittedRLP(TransactionCoreInfo const& info, uint64_t chainId,
-                              std::string signature) {
+                              std::string signature, uint64_t& recid) {
```
n-hutton, line 419: make it such that we can extract the recovery ID when we get the transmitted RLP

src/libServer/JSONConversion.cpp
```diff
@@ -693,25 +690,25 @@ const Json::Value JSONConversion::convertTxtoEthJson(
...
+  if (!tx.GetCode().empty() && IsNullAddress(tx.GetToAddr())) {
     retJson["contractAddress"] =
         "0x" +
-        Account::GetAddressForContract(txn.GetTransaction().GetSenderAddr(),
-                                       txn.GetTransaction().GetNonce() - 1)
+        Account::GetAddressForContract(tx.GetSenderAddr(), tx.GetNonce() - 1)
             .hex();
   }
   retJson["type"] = "0x0";
+  std::string sig{tx.GetSignature()};
```
n-hutton, line 708: The fix is here

PR #3014, opened 2022-09-20
[master] initial commit of int test

n-hutton:
This test adds an integration test to the CI and fixes a bug it found. Comments are inline. Jenkins is the only CI that runs this (not travis)


scripts/integration_test.sh
```diff
@@ -17,65 +17,98 @@
...
-
-    # Modify constants.xml for use by isolated server
-    cp constants.xml constants_backup.xml
-    sed -i 's/.LOOKUP_NODE_MODE.false/<LOOKUP_NODE_MODE>true/g' constants.xml
-    sed -i 's/.ENABLE_EVM>.*/<ENABLE_EVM>true<\/ENABLE_EVM>/g' constants.xml
-    sed -i 's/.EVM_SERVER_BINARY.*/<EVM_SERVER_BINARY>\/home\/travis\/build\/Zilliqa\/Zilliqa\/evm-ds\/target\/release\/evm-ds<\/EVM_SERVER_BINARY>/g' constants.xml
-    sed -i 's/.EVM_LOG_CONFIG.*/<EVM_LOG_CONFIG>\/home\/travis\/build\/Zilliqa\/Zilliqa\/evm-ds\/log4rs.yml<\/EVM_LOG_CONFIG>/g' constants.xml
+# Firstly determine if the user is running this locally, or is this in the CI
+
+RUNNING_LOCALLY=1
```
n-hutton, line 22: determine if running locally since users will want to run this script locally if it fails on CI

src/libCrypto/EthCrypto.cpp
```diff
@@ -389,7 +389,7 @@ bytes GetOriginalHash(TransactionCoreInfo const& info, uint64_t chainId) {
   rlpStreamRecreated << info.nonce;
   rlpStreamRecreated << info.gasPrice;
   rlpStreamRecreated << info.gasLimit;
-  bytes toAddr{};
+  bytes toAddr(20, 0);  // Default is empty address
```
n-hutton, line 392: For contract creation, the tx hash generation would not work because it didn't reconstruct the RLP correctly (it needs 40 0s here.

I do not think this is an issue, but we will get into trouble if leaving the address blank rather than a null address is also a valid way to create a contract.

valeryz: I think this is not correct.  I had so much debugging and confusion in the past, I don't think you should revert it.

image

valeryz (commented):
Please don't change the handling of empty address for contract creation txns.  This is non-intuitive and was hard to understand why it didn't work :)

valeryz (approved):

PR #3015, opened 2022-09-20
[master] Fix the problem with JSON parser crashing

valeryz:
## Description
Multiple changes:

1. Upgrade the nlohmann parser to the latest version (doesn't help by itself, but hopefully useful to keep it current).
2. Log the return value from EVM for future debugging.
3. Remove to_string() in the code handling the exit_reason. It is not necessary, and in fact the branch above doesn't use it.
4. Remove multithreading wrapper in the handling of eth_call

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


moboware (commented):
ok

moboware (approved):

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -254,15 +237,23 @@ uint64_t AccountStoreSC<MAP>::InvokeEvmInterpreter(
 template <class MAP>
 bool AccountStoreSC<MAP>::ViewAccounts(EvmCallParameters& params, bool& ret,
                                        std::string& result) {
-  TransactionReceipt rcpt;
   uint32_t evm_version{0};
-  evmproj::CallResponse response;
+  evmproj::CallResponse response{};
+
+#if 0
```
valeryz, line 243: Here we have two problems:
1. The eth_call doesn't have anything to do with the transaction, so even the fact that TransactionReceipt is involved indicates that calling through EvmCallRunner is superfluous here.
2. Reintroducing multithreading, even though via std::async might mean reintroducing the problem of the crash again (since we don't know exactly why it was crashing, and it was one of the hypotheses).

@CSideSteve PTAL

CSideSteve (approved):
Prettier and safer now :-)

PR #3023, opened 2022-09-22
[master] Fix broken test in master

n-hutton:
One of the tests in evm server fails and was merged to master - comments are inline

### DIFF
--- a/scripts/integration_test.sh
+++ b/scripts/integration_test.sh
@@ -76,6 +76,7 @@ else
     fi
 
     if [[ -d /home/jenkins ]]; then
+        pwd
         ls /home/jenkins/agent/workspace/ZilliqaCIJenkinsfile_PR-*/evm-ds/target/release/evm-ds
 
         # For convenience move the required files to tmp directory
--- a/scripts/integration_test.sh
+++ b/scripts/integration_test.sh
@@ -80,8 +80,8 @@ else
         ls /home/jenkins/agent/workspace/ZilliqaCIJenkinsfile_PR-*/evm-ds/target/release/evm-ds
 
         # For convenience move the required files to tmp directory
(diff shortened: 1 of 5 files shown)


scripts/integration_test.sh
```diff
@@ -76,6 +76,7 @@ else
     fi
 
     if [[ -d /home/jenkins ]]; then
+        pwd
```
n-hutton, line 79: this will be used to fix the integration test on master

tests/EvmLookupServer/Test_EvmLookupServer.cpp
```diff
@@ -69,10 +69,17 @@ std::unique_ptr<LookupServer> getLookupServer() {
   return lookupServer;
 }
 
-// Convenience fn only used to test Eth TXs
+// Convenience fn only used to test Eth TXs.
```
n-hutton, line 72: The transaction and its receipt are stored in this test. The TX receipt needs to have the epoch number in it so that when retrieving it later it doesn't just fail.

tests/EvmLookupServer/Test_EvmLookupServer.cpp
```diff
@@ -1007,19 +1014,24 @@ BOOST_AUTO_TEST_CASE(test_eth_get_transaction_by_hash) {
...
   for (const auto& transaction : transactions) {
     bytes body;
     transaction.Serialize(body, 0);
     BlockStorage::GetBlockStorage().PutTxBody(
         EPOCH_NUM, transaction.GetTransaction().GetTranID(), body);
   }
 
+  // Need to create a block with our TXs in since that is referenced in the
+  // transaction receipt (tx index)
+  buildCommonEthBlockCase(mediator, EPOCH_NUM, transactions, pairOfKey);
```
n-hutton, line 1033: Similarly, there needs to be a block matching the block number that the TX receipt is of. This is used when getting TX by hash etc since the TX receipt needs to know its index in the block etc.

valeryz (approved):

PR #3026, opened 2022-09-23
[master] upgrade lohmann parser to 3.11.2

moboware:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


CSideSteve (approved):

vcpkg.json
```diff
@@ -26,10 +29,25 @@
   ],
   "builtin-baseline": "f93ba152d55e1d243160e690bc302ffe8638358e",
   "overrides": [
-    { "name": "jsoncpp", "version-string": "1.8.1-1" },
-    { "name": "openssl", "version-string": "1.1.1n" },
-    { "name": "protobuf", "version": "3.9.1" },
-    { "name": "libmicrohttpd", "version-string": "0.9.63#7" }
+    {
+      "name": "nlohmann-json",
+      "version-string": "3.11.2"
```
yaron-zilliqa, line 34: vcpkg doesn't have 3.11.2. The highest version is 3.10.5.

yaron-zilliqa (changes requested):

vcpkg.json
moboware: Did you run the code? Because I have that version downloaded in vcpck

yaron-zilliqa: Here's what I got when running cmake:

Error: No version entry for nlohmann-json at version 3.11.2. This may be fixed by updating vcpkg to the latest master via `git pull`.
Available versions:
    3.10.5#3
    3.10.5#2
    3.10.5#1
    3.10.5
    3.10.4
    3.10.2
    3.9.1
    3.9.0
    3.8.0#2
    3.8.0
    3.7.3
    3.7.0
    3.6.1
    3.6.0
    3.5.0-5
    3.5.0
    3.4.0
    3.3.0
    3.2.0
    3.1.2
    3.1.0
    3.0.1
    3.0.0
    2.1.1-1
    2.1.1
`

valeryz: Our upgrade attempt was to solve a particular bug, but nlohmann version doesn't seem to affect that.   FWIW, we can keep the older version if it is easier/cleaner for vcpkg.

moboware: We can use 3.10.5 as long it is clear from the vcpkg.json file which version we are importing. 
And that should be a general thing to do in the vcpkg.json file for all packages.

yaron-zilliqa (approved):

PR #3028, opened 2022-09-23
[master] ZIL-4688: fix missing pip3

yaron-zilliqa:
2. Don't re-compile cryptoutils in dependencies each time.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -191,5 +191,9 @@ ENV LD_LIBRARY_PATH=${INSTALL_DIR}/lib
 RUN python3 -m pip install --upgrade pip setuptools wheel \
   && python3 -m pip install --no-cache-dir -r requirements3.txt
 
+RUN echo '#!/bin/bash\npython3 -m pip "$@"' > /usr/bin/pip3 \
+  && chmod a+x /usr/bin/pip3 \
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -191,5 +191,9 @@ ENV LD_LIBRARY_PATH=${INSTALL_DIR}/lib
 RUN python3 -m pip install --upgrade pip setuptools wheel \
   && python3 -m pip install --no-cache-dir -r requirements3.txt
 
+RUN echo '#!/bin/bash\npython3 -m pip "$@"' > /usr/bin/pip3 \
```
valeryz, line 194: I'm surprised the binary isn't installed, maybe it lives in a different path somewhere? That wrapper works too of course, if nothing else helps ...

valeryz (approved):

docker/Dockerfile
yaron-zilliqa: It isn't installed because I explicitly uninstalled the default python environment. We need to use the vcpkg compiled python and it's proved tricky to have both environments so I'd rather have just the one we need to prevent any collisions in the future.

PR #3029, opened 2022-09-23
[master] Fix travis test issue

n-hutton:
Travis always passes. This makes it such that it does not always pass if your tests fail. You can see here if you look at the commits I make travis fail when the test fails (when before it passes). Changing ; to && fixes this


valeryz (approved):
nice find, thanks!

PR #3032, opened 2022-09-25
[master] Extend basic integration tests

n-hutton:
add integration tests for most of the API, except parts which are not on the isolated server


scripts/integration_test.sh
```diff
@@ -92,7 +92,7 @@ else
     fi
 
     echo "Starting isolated server"
-    ./build/bin/isolatedServer -f isolated-server-accounts.json -u 999 -t 3000 &
+    ./build/bin/isolatedServer -f isolated-server-accounts.json -u 999 &
```
n-hutton, line 95: Don't run it in block generation mode

src/libServer/JSONConversion.cpp
```diff
@@ -166,7 +166,7 @@ const Json::Value JSONConversion::convertTxBlocktoEthJson(
   retJson["timestamp"] = (boost::format("0x%x") % timestamp).str();
   retJson["version"] = (boost::format("0x%x") % txheader.GetVersion()).str();
   // Required by ethers
-  retJson["extraData"] = "0x";
+  retJson["extraData"] = "0x0";
```
n-hutton, line 169: needs to be valid hex

its-saeed (approved):

PR #3033, opened 2022-09-26
[master] Adding contract interaction tests using web3js

bzawisto:
## Description
PR adds missing coverage with web3-js.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/EvmAcceptanceTests/helper/Web3Helper.js
```diff
@@ -29,6 +29,15 @@ class Web3Helper {
 
         return deployedContract;
     }
+
+    async getCommonOptions(base = {}) {
```
its-saeed, line 33: One note. I found it more idiomatic in javascript to have a variable and extend it with properties and methods instead of writing a class for it. something like `GeneralHelper.js` what do you think? It makes more sense because most of the functions are static functions and they don't have any related state. If you agree, please revert web3js helper to its previous state.

bzawisto: Sure, will do. What about making these methods static instead?

its-saeed: Sounds good. I guess all of them become static including the variables. making them static, or what I suggested, eliminate the need of instantation of Web3Helper class.

tests/EvmAcceptanceTests/test/ContractDeployment.js
```diff
@@ -364,15 +358,13 @@ describe("Contract Deployment", function () {
             })
         });
 
-        describe("When web3.js is used", function () {
+        describe("When web3.js is used 1234", function () {
```
its-saeed, line 361: please remove this

tests/EvmAcceptanceTests/test/ContractInteraction.js
```diff
@@ -105,6 +106,51 @@ describe("Contract Interaction", function () {
...
+                expect(await contract.methods.number().call()).to.be.eq(web3.utils.toBN(TO_BE_SET))
+            })
+
+            it("Should set uint256 internal state correctly with external method", async function() {
+                const TO_BE_SET = 100;
+                expect(await contract.methods.setNumberExternal(TO_BE_SET).send(options)).to.be.not.null;
+                expect(await contract.methods.number().call()).to.be.eq(web3.utils.toBN(TO_BE_SET))
+            })
+
+
```
its-saeed, line 130: one of the empty lines is useless

its-saeed (approved):

PR #3039, opened 2022-09-27
[master] remove Travis builds and enable CODEOWNERS

frankmeds:
## Backward Compatibility

### Implementation

- [X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


mauromedda (approved):

pavlops (approved):

.travis.yml.disabled
```diff
@@ -23,4 +23,4 @@ script:
   - if [[ "$TRAVIS_BRANCH" == "master" ]]; then docker run --rm -it -v /scilla:/scilla zilliqa_build:$TRAVIS_COMMIT -c './scripts/ci_build.sh && ./scripts/ci_report_coverage.sh'; fi
 
 env:
-  - VCPKG_ROOT=/vcpkg
+  - VCPKG_ROOT=/vcpkg
```
valeryz, line 26: nit: why remove end of line?

Jenkinsfile
```diff
@@ -17,42 +17,55 @@ spec:
...
-                sh "scripts/integration_test.sh --setup-env"
-            }
-            stage('Report coverage') {
-                // Code coverage is currently only implemented for GCC builds, so OSX is currently excluded from reporting
-                sh "./scripts/ci_report_coverage.sh"
-            }
+        try {
+          stage('Checkout scm') {
+              checkout scm
+              def skipci = sh(script: "git log -1 --pretty=%B | fgrep -ie '[skip ci]' -e '[ci skip]' | wc -l", returnStdout: true).trim()
```
valeryz, line 29: So, how does this work for merge commits of PR with master that are being checked?  The problem is that even if I put [skip-ci] in the history, the merge commit will have a commit message "merge ... blah blah," and will likely not mention the commit message of the original commit.

We either need to check the child commit of the merge commit *that is not master*, but also exlcude this logic if this commit is *not* merge commit (if we fail to do the latter, [skip ci] might have an effect on follow-on PRs.

Therefore my suggestion was to look at the PR title.

mauromedda: got it @valeryz, thanks for the feedback. Reviewing it

yaron-zilliqa (approved):

Jenkinsfile
mauromedda: I included in the condition the check of the PR title as per PR 3034. @valeryz @frankmeds

valeryz (approved):

PR #3049, opened 2022-09-28
[master] Moved lookupServer code into adjacent files so that functionality can…

CSideSteve:
Sepertate the ETH/EVM changes from the original 8.2 core allowing us to disable by switch and rung git diff with ease against 8.2 binaries.

## Description

Preserver lineage of code in repository and reduce risk upon release.

## Backward Compatibility

- [ x] This is not a breaking change

## Review Suggestion

do a git diff of lookupserver against 8.2 version

## Status
Fairly expeditious as many people make chanegs to lookupserver.
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/LookupServer.h
+++ b/src/libServer/LookupServer.h
@@ -18,7 +18,6 @@
 #ifndef ZILLIQA_SRC_LIBSERVER_LOOKUPSERVER_H_
 #define ZILLIQA_SRC_LIBSERVER_LOOKUPSERVER_H_
 
-
 #include "EthRpcMethods.h"
 #include "Server.h"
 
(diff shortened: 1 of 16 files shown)


src/libServer/LookupServer.h
```diff
@@ -19,18 +19,17 @@
...
+
+
 
 class Mediator;
 
 typedef std::function<bool(const Transaction& tx, uint32_t shardId)>
     CreateTransactionTargetFunc;
 
 class LookupServer : public Server,
+                     public EthLookupServer,
```
valeryz, line 32: This way of inheriting could surely confuse future readers.
A LookupServer is certainly not a kind of EthLookupServer.  It is the other way around, and  I think if you swap the inheritance line, this PR should also "just work" (well, maybe ..)

CSideSteve: I just msgd you, the inheritance isn't important, I think composition is easy now the way things operate.

CSideSteve: Upon reflection the name EthLookupServer was a bad choice, the class is in fact the owner of the additional RPC methods that should be added to LookupServer when EVM is enabled.

The class is now renamed to EthRPCMethods and I think makes more sense.

The inheritance provides the exact funtionality required as if the methods were inlined in the lookupserver class, no changes to any existing users of LokkupServer were required the Test Mock routines and IsolatedServer that utilises the code directly is unaware the shape of the class has changed.

valeryz (approved):
Obviously I didn't check every line here, as it is mostly just shuffling around.
LGTM.

valeryz:
Btw, I'm surprised the IsolatedServer doesn't have to be touched. I'm sure we added a lot there.

CSideSteve:
IsolatedServer didn't require any changes to allow these files to be split, however, isolatedServer may require some additional TLC to isolate changes for EVM if in fact changes have been done.

PR #3050, opened 2022-09-28
Fixing receipt logs

bzawisto:
## Description
This PR fixes some issues with logs attached to receipts. Evm-ds returns data as array of bytes whereas it should be exposed as hex string instead. Adding also two tests checking for generated events. 

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [X] local machine test

- [ ] small-scale cloud test


tests/EvmAcceptanceTests/test/ContractInteraction.js
```diff
@@ -149,6 +149,17 @@ describe("Contract Interaction", function () {
         expect(await contract.methods.setAddress(ADDRESS).send(options)).to.be.not.null;
         expect(await contract.methods.someAddress().call()).to.be.eq(ADDRESS);
       });
+
+      it("Calling a method on a contract that generates event without parameters", async function () {
```
its-saeed, line 153: please change `it` description based on our style. we already have `Should have the event argument in returned object when arg type is uint256` `Should have the event name in returned object` and ...

valeryz:
I think we should better make evm-ds return a hex string or something. it is rather easy.

tests/EvmAcceptanceTests/test/ContractInteraction.js
bzawisto: @its-saeed Done.

bzawisto:
> 

Moved the fix from cpp to rust code.

its-saeed (approved):

CSideSteve (approved):

PR #3052, opened 2022-09-29
[release/v8.2] Compress specific leveldb in persistence

chetan-zilliqa:
## Description

same as #2843 
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/cmd/compress_persistence.cpp
+++ b/src/cmd/compress_persistence.cpp
@@ -24,7 +24,7 @@
 /// Should be run with working directory where folder "persistence" consisting
 /// of the individual dbs exists.
 
-enum class STATUS {
+enum STATUS {
   ERROR_UNEXPECTED = -3,
   ERROR_UNHANDLED_EXCEPTION = -2,
   ERROR_IN_COMMAND_LINE = -1,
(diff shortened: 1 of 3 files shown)


src/cmd/compress_persistence.cpp
```diff
@@ -0,0 +1,75 @@
...
+      std::cerr << "ERROR: " << e.what() << std::endl << std::endl;
+      std::cout << desc;
+      return ERROR_IN_COMMAND_LINE;
+    } catch (boost::program_options::error& e) {
+      std::cerr << "ERROR: " << e.what() << std::endl << std::endl;
+      return ERROR_IN_COMMAND_LINE;
+    }
+
+    LOG_GENERAL(INFO, "Begin compression of " << dbname);
+    auto dbptr = std::make_shared<LevelDB>(dbname);
```
yaron-zilliqa, line 65: Why std::make_shared? std::make_unique is definitely preferable (and cheaper) in cases such as this.

yaron-zilliqa (changes requested):
What's the point of this tool? is it really required at the moment?

chetan-zilliqa:
This was a tool initially written by Sandip to see if compression of level DB can reduce the persistence size. I added it as it's good to have for our reference.
https://github.com/Zilliqa/Zilliqa/pull/2732

chetan-zilliqa:
> What's the point of this tool? is it really required at the moment?

This was a tool initially written by Sandip to see if compression of level DB can reduce the persistence size. I added it as it's good to have for our reference.
https://github.com/Zilliqa/Zilliqa/pull/2732

src/cmd/compress_persistence.cpp
```diff
@@ -0,0 +1,77 @@
...
+  ERROR_UNHANDLED_EXCEPTION = -2,
+  ERROR_IN_COMMAND_LINE = -1,
+  SUCCESS = 0,
+};
+
+using namespace std;
+namespace po = boost::program_options;
+
+int main(int argc, const char* argv[]) {
+  string dbName;
```
yaron-zilliqa, line 38: No reason to be here; can just be inside the try.
Then you can consider changing:
```
int main(int argc, const char* argv[]) try {
  ...
  return static_cast<int>(STATUS::SUCCESS);
}
catch (std::exception& e) {
  ...
  return static_cast<int>(STATUS::ERROR_UNHANDLED_EXCEPTION);
}
```

src/cmd/compress_persistence.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <iostream>
+
+#include <boost/program_options.hpp>
+#include "depends/libDatabase/LevelDB.h"
+#include "libUtils/FileSystem.h"
+
+/// Should be run with working directory where folder "persistence" consisting
+/// of the individual dbs exists.
+
+enum class STATUS {
```
yaron-zilliqa, line 27: I think (as opposed to my previous comment), it'd be nicer to just make it a regular enum (i.e. remove the 'class').
This way we can skip the ugly explicit cast to ints below (static_cast<int>(...)).

yaron-zilliqa (changes requested):

PR #3062, opened 2022-09-30
[master] Disable EVM functionality according to flag in config

CSideSteve:
## Description

Disable RPC methods for eth when EVM is disabled.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -191,7 +191,7 @@
         <!-- Timeout in seconds for ONLY connection that reach our callback function, 0 means no timeout-->
         <CONNECTION_CALLBACK_TIMEOUT>0</CONNECTION_CALLBACK_TIMEOUT>
         <ENABLE_EVM>true</ENABLE_EVM>
-        <EVM_SERVER_BINARY>/usr/local/bin/evm-ds</EVM_SERVER_BINARY>
+        <EVM_SERVER_BINARY>/home/stephen/versions/Zilliqa/target/release/evm-ds</EVM_SERVER_BINARY>
         <EVM_SERVER_SOCKET_PATH>/tmp/evm-server.sock</EVM_SERVER_SOCKET_PATH>
         <EVM_LOG_CONFIG>/usr/local/etc/log4rs.yml</EVM_LOG_CONFIG>
         <!-- eth chain id : Test net = 0x814d, main net: 0x8001 -->
@@ -288,7 +288,7 @@
     </recovery>
     <smart_contract>
         <ENABLE_SC>true</ENABLE_SC>
-        <SCILLA_ROOT>/scilla</SCILLA_ROOT>
+        <SCILLA_ROOT>/home/stephen/run/scilla</SCILLA_ROOT>
         <SCILLA_CHECKER>bin/scilla-checker</SCILLA_CHECKER>
         <SCILLA_BINARY>bin/scilla-runner</SCILLA_BINARY>
         <SCILLA_FILES>scilla_files</SCILLA_FILES>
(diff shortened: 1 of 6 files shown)


src/libServer/IsolatedServer.cpp
```diff
@@ -126,96 +126,98 @@ IsolatedServer::IsolatedServer(Mediator& mediator,
...
-                         jsonrpc::JSON_STRING, "param01", jsonrpc::JSON_OBJECT,
-                         NULL),
-      &LookupServer::GetEthCallZilI);
-
-  AbstractServer<IsolatedServer>::bindAndAddMethod(
-      jsonrpc::Procedure("eth_call", jsonrpc::PARAMS_BY_POSITION,
-                         jsonrpc::JSON_STRING, "param01", jsonrpc::JSON_OBJECT,
-                         "param02", jsonrpc::JSON_STRING, NULL),
-      &LookupServer::GetEthCallEthI);
+  if (ENABLE_EVM) {
```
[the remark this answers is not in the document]
CSideSteve: You could raise a new ticket for this although it is less critical as isolatedServer is a test and development tool and does not run on the mainnet.

The purpose of this switch is to make sure that the first release does not expose EVM functionality on Mainnet, if there is that possibility then the code cannot be merged onto the mainnet release for November.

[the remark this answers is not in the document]
CSideSteve: Also see changes made to LookupServer which is on mainnet with the EthRcpMethods plug in that does exactly this for the RPC calls.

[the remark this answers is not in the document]
CSideSteve: Resolved in latest commit

src/libServer/IsolatedServer.cpp
```diff
@@ -248,114 +158,214 @@ IsolatedServer::IsolatedServer(Mediator& mediator,
...
   }
+  BindAlllEvmMethods();
+}
 
-  // Add the JSON-RPC eth style methods
-  AbstractServer<IsolatedServer>::bindAndAddMethod(
-      jsonrpc::Procedure("eth_blockNumber", jsonrpc::PARAMS_BY_POSITION,
-                         jsonrpc::JSON_STRING, NULL),
-      &IsolatedServer::GetEthBlockNumberI);
+void IsolatedServer::BindAlllEvmMethods() {
```
moboware, line 164: Alll with 2 l's

moboware (commented):
one typo

src/libServer/IsolatedServer.cpp
CSideSteve: Corrected

constants.xml
```diff
@@ -191,7 +191,7 @@
         <!-- Timeout in seconds for ONLY connection that reach our callback function, 0 means no timeout-->
         <CONNECTION_CALLBACK_TIMEOUT>0</CONNECTION_CALLBACK_TIMEOUT>
         <ENABLE_EVM>true</ENABLE_EVM>
-        <EVM_SERVER_BINARY>/usr/local/bin/evm-ds</EVM_SERVER_BINARY>
+        <EVM_SERVER_BINARY>/home/stephen/versions/Zilliqa/target/release/evm-ds</EVM_SERVER_BINARY>
```
moboware, line 194: remove your name from here

constants.xml
```diff
@@ -288,7 +288,7 @@
     </recovery>
     <smart_contract>
         <ENABLE_SC>true</ENABLE_SC>
-        <SCILLA_ROOT>/scilla</SCILLA_ROOT>
+        <SCILLA_ROOT>/home/stephen/run/scilla</SCILLA_ROOT>
```
moboware, line 291: and here too

CSideSteve: This should not be committed, good catch, no changes to constants.xml required. We are looking at changing the way we use it locally so you don't need to keep editing.

moboware (approved):

PR #3067, opened 2022-09-30
Add debug flag to tests so that log messages will be printed only if this flag is set. [skip ci]

its-saeed:
## Description
Test results should be as cleanest as we can. just failure or success of tests. I added `--debug` flag so that if it's passed to `npx hardhat test` then logs will be appeared on the screen.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/README.md
+++ b/tests/EvmAcceptanceTests/README.md
@@ -3,6 +3,11 @@
 ```bash
     npm install
     npx hardhat test    # to run tests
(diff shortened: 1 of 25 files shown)


bzawisto (approved):

moboware (commented):
Would it not better to create LogDebug method where check the debug flag? Saves you from many if(debug).... implementations and it is more clear

moboware (approved):
using hre is a better solution!

PR #3069, opened 2022-10-03
[master] Zil 4769 add contract destruction web 3 js tests

bzawisto:
## Description
Adding contract destruction tests and fixing node crash after receiving selfdestruct callback from evm.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/EvmAcceptanceTests/test/ContractDestruction.js
```diff
@@ -1,13 +1,50 @@
...
+          .returnToSenderAndDestruct(destAccount)
+          .send({gasLimit: 1000000, from: web3_helper.getPrimaryAccountAddress()})
+      ).to.be.not.null;
+      const newBalance = await web3.eth.getBalance(destAccount);
+      // Dest Account should have prevBalance + amountPaid
+      expect(web3.utils.toBN(newBalance)).to.be.equal(web3.utils.toBN(prevBalance).add(amountPaid));
+    });
+  });
+
+  describe("When a method call happens through another contract", function () {
```
its-saeed, line 32: one general note for destruction. isn't there any way to test if the contract does not exist anymore after the destruction?

its-saeed (approved):

tests/EvmAcceptanceTests/test/ContractDestruction.js
bzawisto: So far it's not, we have to check how it's done in other networks and decide what would be suitable for our blockchain (either empty account or non-existing one).

PR #3072, opened 2022-10-03
[master] Zil 4848 account store sc evm view accounts doesnt call set bc info prior to calling evm

CSideSteve:
## Description
The query sent to the scillaipcserver was not handled in the test therefore leaving an unititlaised smartpointer.

The question is the second case for the ds blockSharedPtr is that rleavant too or is it just the first case in error.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libServer/ScillaIPCServer.cpp
```diff
@@ -245,14 +245,18 @@ bool ScillaIPCServer::fetchBlockchainInfo(const std::string &query_name,
...
+  if (query_name == "BLOCKHASH" || query_name == "BLOCKCOINBASE" || query_name == "BLOCKDIFFICULTY" ||
       query_name == "BLOCKGASPRICE") {
     if (!BlockStorage::GetBlockStorage().GetDSBlock(blockNum,
                                                     dsBlockSharedPtr)) {
       LOG_GENERAL(WARNING, "Could not get blockNum DS block " << blockNum);
       return false;
     }
   }
+  if (not dsBlockSharedPtr || not txBlockSharedPtr) {
+    LOG_GENERAL(WARNING, "Smart pointers work better when Initialized ");
```
n-hutton, line 257: :)

n-hutton (approved):

src/libServer/ScillaIPCServer.cpp
CSideSteve: I will make that more sensible before closing.

valeryz (commented):
Good catch on the unhandled query. I tested this by calling multiple 'get*' methods on the `bcinfo.sol` contract.   I codified just one example of failing methods, and it goes way beyond   just BLOCKHASH.  

So I can't really understand how m_BCInfo  on line 246 is not nullptr in eth_call...

src/libServer/ScillaIPCServer.cpp
```diff
@@ -245,14 +245,18 @@ bool ScillaIPCServer::fetchBlockchainInfo(const std::string &query_name,
   // block.
   blockNum = m_BCInfo->getCurDSBlockNum();
   DSBlockSharedPtr dsBlockSharedPtr;
-  if (query_name == "BLOCKCOINBASE" || query_name == "BLOCKDIFFICULTY" ||
+  if (query_name == "BLOCKHASH" || query_name == "BLOCKCOINBASE" || query_name == "BLOCKDIFFICULTY" ||
```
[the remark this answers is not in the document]
valeryz: that is, defensively getting both tx and ds block in each query might be reliable, but this is also slow for an EVM instruction.

src/libServer/ScillaIPCServer.cpp
```diff
@@ -232,7 +232,7 @@ bool ScillaIPCServer::fetchBlockchainInfo(const std::string &query_name,
   }
 
   TxBlockSharedPtr txBlockSharedPtr;
-  if (query_name == "BLOCKCOINBASE" || query_name == "BLOCKTIMESTAMP" ||
+  if (query_name == "BLOCKHASH" || query_name == "BLOCKCOINBASE" || query_name == "BLOCKTIMESTAMP" ||
```
moboware, line 235: This part of code needs more protective programming (refactor), in case a new or unknown query request enters here it is caught at all.

PR #3074, opened 2022-10-03
[master] revert block num start index as fails tests

n-hutton:
There is some error with starting the isolated server from block 0 that I see locally - this reverts @moboware change of starting it from 0. The behaviour is such that the tx receipt cannot be found via the regular zil api once the tx is submitted

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -74,7 +74,7 @@ void help(const char* argv[]) {
 int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
-  string blocknum_str{"0"};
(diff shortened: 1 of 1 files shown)


src/cmd/isolated_server.cpp
```diff
@@ -74,7 +74,7 @@ void help(const char* argv[]) {
 int main(int argc, const char* argv[]) {
   string accountJsonFilePath;
   uint port{5555};
-  string blocknum_str{"0"};
+  string blocknum_str{"1"};
```
moboware, line 77: Be aware that reverting this will break the js test getBlockByNumber. So far I understand blocks, also in production, start at '0'.

PR #3078, opened 2022-10-04
Adding test for ecrecover precompile

bzawisto:
## Description
This PR adds acceptance test for `ecrecover` rust evm precompile. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
+++ b/tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
@@ -1,0 +1,28 @@
+const {expect} = require("chai");
+const {web3} = require("hardhat");
+const web3_helper = require("../../helper/Web3Helper");
+const general_helper = require("../../helper/GeneralHelper");
+
+describe("Precompile tests with web3.js", function () {
+  describe("When recover function is used", function () {
+    let contract;
+    let options;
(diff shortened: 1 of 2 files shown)


tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
```diff
@@ -0,0 +1,28 @@
+const {expect} = require("chai");
+const {web3} = require("hardhat");
+const web3_helper = require("../../helper/Web3Helper");
+const general_helper = require("../../helper/GeneralHelper");
+
+describe("Precompile tests with web3.js", function () {
+  describe("When recover function is used", function () {
```
its-saeed, line 7: seems like this can be merged into the `it` block. we use `describe` when there are at least 2 related tests to that `describe`. If you can add more tests to `When recover function is used` I'm OK with it.

its-saeed (approved):
one note...

tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
bzawisto: Thanks, done.

PR #3093, opened 2022-10-06
SHA256 and Ripemd160 precompiles

bzawisto:
## Description
This PR adds two new precompiles as stated in the title. Based on previous branch hence some irrelevant changes are also present.

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
```diff
@@ -11,7 +11,7 @@ describe("Precompile tests with web3.js", function () {
     contract = await web3_helper.deploy("Precompiles", options);
   });
 
-  it("Should return signer address when recover function is used", async function () {
+  it("When recover function is used it should return signer address", async function () {
```
its-saeed, line 14: `it` block already has **it** for the start of the sentence. :D

tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
```diff
@@ -33,4 +33,20 @@ describe("Precompile tests with web3.js", function () {
     const readValue = await contract.methods.idStored().call({gasLimit: 50000});
     expect(readValue).to.be.eq(hash);
   });
+
+  it("When SHA2-256 function is used it should return correct hash", async function () {
```
its-saeed, line 37: Please fix the description based on the convention we use. we read `it` blocks from the beginning of it. so `it("should return correct hash when SHA2-256 function is used")` is read like this: it should return correct hash when SHA2-256 function is used. that's why the called it `it` :D

tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
```diff
@@ -22,7 +22,7 @@ describe("Precompile tests with web3.js", function () {
     expect(result).to.be.eq(accountAddress);
   });
 
-  it("Should return input value when identity function is used", async function () {
+  it("When identity function is used it should return input value", async function () {
```
its-saeed, line 25: please revert these changes.

tests/EvmAcceptanceTests/test/precompiles/EvmPrecompiles.js
bzawisto: Sure sure, fixing.

evm-ds/src/main.rs
```diff
@@ -34,6 +33,8 @@ use jsonrpc_server_utils::codecs;
 use primitive_types::*;
 use scillabackend::{ScillaBackend, ScillaBackendConfig};
 
+use crate::precompiles::get_precompiles;
```
valeryz, line 36: Thanks for organizing this nicely!

valeryz (approved):

PR #3094, opened 2022-10-06
[master] add jenkins timeout

n-hutton:
Add upper bound on CI duration so that pods don't hang around forever if there is a hang in the build etc.


frankmeds:
LGTM

PR #3102, opened 2022-10-11
[master] ZIL-4894-unit-tests-hang-on-latest-master

CSideSteve:
## Description
Send die command to evm to tell evm to close down gracefully, if this collects an exception then resort to brute force approach of killing the task with a pkill as we did in previous generations.

Try to gracefully close the evm engine, if this fails then shoot it with a pkill

fix the unit tests from hanging in 8.3 release candidate,

## Backward Compatibility

- [ x] This is a breaking change

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/EvmClient.cpp
+++ b/src/libData/AccountData/EvmClient.cpp
@@ -32,17 +32,14 @@ void EvmClient::Init() {
 }
 
 EvmClient::~EvmClient() {
-
-
   std::lock_guard<std::mutex> g(m_mutexMain);
   Json::Value _json;
   LOG_GENERAL(DEBUG, "Call evm with die request:" << _json);
   // call evm
   try {
     const auto oldJson = m_clients.at(0)->CallMethod("die", _json);
-  } catch (std::exception& e) {
-    LOG_GENERAL(WARNING,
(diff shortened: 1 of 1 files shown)


moboware (approved):

src/libData/AccountData/EvmClient.cpp
```diff
@@ -32,18 +32,30 @@ void EvmClient::Init() {
...
+      if (!SysCommand::ExecuteCmd(SysCommand::WITHOUT_OUTPUT, cmdStr)) {
+        LOG_GENERAL(WARNING, "ExecuteCmd failed: " << cmdStr);
+      }
+    } catch (const std::exception& e) {
+      LOG_GENERAL(WARNING,
+                  "Exception caught in SysCommand::ExecuteCmd: " << e.what());
+    } catch (...) {
+      LOG_GENERAL(WARNING, "Unknown error encountered");
+    }
   }
```
yaron-zilliqa, line 59: Should there also be a catch(...) here? if there's an uncaught exception from a d-tor the process will terminate.

yaron-zilliqa (approved):

PR #3111, opened 2022-10-13
[master] ZIL-4676: upgrade to C++17

yaron-zilliqa:
## Description

1. Had to patch g3log to fix compilation with C++17.
2. Our codebase currently uses 'byte' as a typedef but this is a reserved by C++17 (std::byte). As a result, the following renaming have been done:
  * byte -> zbyte
  * bytes -> zbytes
  * bytesRef -> zbytesRef
  * bytesConstRef -> zbytesConstRef.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


bzawisto (approved):

chetan-zilliqa (approved):

src/depends/common/Common.h
```diff
@@ -109,9 +109,9 @@ namespace dev
 
     // Map types.
     using StringMap = std::map<std::string, std::string>;
-    using BytesMap = std::map<bytes, bytes>;
+    using BytesMap = std::map<zbytes, zbytes>;
```
moboware, line 112: why not zBytesMap?

src/depends/common/Common.h
```diff
@@ -109,9 +109,9 @@ namespace dev
...
     using u256Map = std::map<u256, u256>;
-    using HexMap = std::map<bytes, bytes>;
+    using HexMap = std::map<zbytes, zbytes>;
```
moboware, line 114: zHexMap?

yaron-zilliqa: The idea is to distinguish names that have 'byte(s)' in them to zbyte(s) to stress this is Zilliqa's definition of byte and not std::byte. I'll skip this one as this isn't required.

PR #3117, opened 2022-10-14
Fixing ParentChild tests

bzawisto:
## Description
This PR fixes an issue with json parsing that skipped delete action returned from evm. 
Part of a bigger json parsing code has been moved to different function too.
It also seems that RemoveAccount doesn't work - for now the balance of the destructed contract is deliberately set to 0 in case parent contract destructs child contract and - as a result - receives child's funds over and over.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz (commented):
@CSideSteve PTAL

CSideSteve (approved):
LGTM

Parse Apply looks earily like the scilla coding, but its parsing so has to be done, at least its all in one place.

PR #3125, opened 2022-10-16
[master] Fix segv

n-hutton:
The code can segv if you have lookup set to false and you forget EnsureEvmAndLookupEnabled() because is does not initialize a DB for structures that are only used during lookup


CSideSteve (approved):
You can check the value of LOOKUP_NODE_MODE if it is isolatedServer.

PR #3128, opened 2022-10-18
[master] ZIL-4921 account store scilla IPC server pointer fix

moboware:
## Description
the scilla ipc svr pointer used in account store is shadowed by another scillaIPCSvr member and was due to that in the AccountStoreSc still a nullptr when ViewAccounts was called. The should one be one definition of the pointer that is shared over the users. This change encapsulate the scilla ipc pointer, and shares it over the users.
The scilla ipc pointer is shared with the AccountStoreTemp via a setter, but could be passed on at construction of the AccountStoreTemp instance, this is will be fixed in different PR and the SetScillaIPCServer method can be removed.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libData/AccountData/AccountStore.h
+++ b/src/libData/AccountData/AccountStore.h
@@ -91,9 +91,6 @@ class AccountStore
   std::condition_variable_any m_writeCond;
   static constexpr int NUM_OF_WRITERS_IN_QUEUE = 1;
 
-  /// Scilla IPC server related
-  std::unique_ptr<jsonrpc::UnixDomainSocketServer> m_scillaIPCServerConnector;
-
   AccountStore();
   ~AccountStore();
 
--- a/src/libData/AccountData/AccountStore.h
+++ b/src/libData/AccountData/AccountStore.h
@@ -68,8 +68,7 @@ class AccountStoreTemp : public AccountStoreSC<std::map<Address, Account>> {
 
 // Singleton class for providing interface related Account System
 class AccountStore
-    : public AccountStoreTrie<std::unordered_map<Address, Account>>,
-      Singleton<AccountStore> {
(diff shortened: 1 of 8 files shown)


src/libData/AccountData/AccountStore.h
```diff
@@ -68,8 +68,7 @@ class AccountStoreTemp : public AccountStoreSC<std::map<Address, Account>> {
 
 // Singleton class for providing interface related Account System
 class AccountStore
-    : public AccountStoreTrie<std::unordered_map<Address, Account>>,
```
valeryz, line 71: I also don't like that Singleton, but is removing it absolutely necessary in this PR?  Would it not cause some unknown problems somewhere where AccountStore singleton is used?

src/libData/AccountData/AccountStoreSC.cpp
```diff
@@ -1610,6 +1610,16 @@ void AccountStoreSC<MAP>::CleanNewLibrariesCache() {
   m_newLibrariesCreated.clear();
 }
 
+template <class MAP>
```
valeryz, line 1613: Doesn't it mean that because we have two instances of this, one with `std::map`, one with `std::unordered_map`, we'll get two instances of this method, eventually called twice, into two different members variables (so far everything is good),  and ... attempting to bind to the same UNIX domain socket (which will likely fail for the 2nd instance).

src/libData/AccountData/AccountStore.h
moboware: As discussed, it is not used and overwritten by the AccountSore::GetInstance

CSideSteve:
This branch can now be closed as the change is no longer required.

PR #3129, opened 2022-10-18
[master] New websocket server and subscribe/unsubscribe [skip ci]

art-gor:
## Description

eth_subscribe/eth_unsubscribe, new websocket server, new http backend

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/APIServerImpl.h
+++ b/src/libServer/APIServerImpl.h
@@ -38,11 +38,11 @@ class APIServerImpl : public APIServer,
   using ConnectionId = uint64_t;
 
   /// Ctor
-  APIServerImpl(std::shared_ptr<AsioCtx> asio);
+  APIServerImpl(std::shared_ptr<AsioCtx> asio, Options options);
 
   /// Called from server's owner. Cannot put everything into the ctor because of
   /// shared_from_this() usage
-  bool Start(Options options);
+  bool Start();
 
   /// Called from a connection on websocket upgrade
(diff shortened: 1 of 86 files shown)


src/libServer/APIServerImpl.h
```diff
@@ -0,0 +1,110 @@
...
+  std::shared_ptr<APIThreadPool> m_threadPool;
+
+  /// Listening socket
+  boost::optional<tcp::acceptor> m_acceptor;
+
+  /// Incremental counter
+  ConnectionId m_counter = 0;
+
+  /// Active connections
+  std::unordered_map<ConnectionId, std::shared_ptr<Connection>> m_connections;
```
valeryz, line 105: I don't understand the interplay of threads here. Just pointing out: are you sure you don't need a mutex to access m_connections?

valeryz (approved):
Thanks, this looks great. Some minor comments.
(well, one about the mutex is not minor, but I'm probably wrong about this - still PTAL).

CSideSteve (approved):
I am liking what I see in some good modern C++ and some long overdue asio to enter the product.
This pattern can most definitely replace some other hairy bits of our existing platform.
The addition of namespaces and using statements make things clear and safe.


This code looks OK to me (OK,Very good) Obviously there is more code needed but this is a clear starting point and should be easy to understand.

PR #3131, opened 2022-10-18
Zil 4871 implement eth call with the correct semantics of quantity tag parameter

moboware:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK:
I believe this PR to be too old to be of any use so it is been closed, the author left almost a year ago.

PR #3133, opened 2022-10-19
[master] Adding eth_estimateGas

bzawisto:
## Description
Adding eth_estimateGas functionality (that reverts in case of calling failing contract methods). 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/contracts/ERC20.sol
+++ b/tests/EvmAcceptanceTests/contracts/ERC20.sol
@@ -91,7 +91,7 @@ contract ZEVMToken is ERC20Interface, SafeMath {
  
     function transferFrom(address from, address to, uint tokens) external returns (bool success) {
         balances[from] = safeSub(balances[from], tokens);
-        allowed[from][msg.sender] = safeSub(allowed[from][msg.sender], tokens);
+        allowed[from][to] = safeSub(allowed[from][to], tokens);
         balances[to] = safeAdd(balances[to], tokens);
         emit Transfer(from, to, tokens);
         return true;
--- a/tests/EvmAcceptanceTests/test/Erc20.js
+++ b/tests/EvmAcceptanceTests/test/Erc20.js
@@ -45,14 +45,17 @@ describe("ERC20 functionality", function () {
     });
     // TODO: Add event related test(s), e.g. for Transfer event
 
-    // FIXME: In ZIL-4879
-    xit("Should not be possible to move more than available tokens to some address [@transactional]", async function () {
+    it("Should not be possible to move more than available tokens to some address [@transactional]", async function () {
       const [owner, receiver] = await ethers.getSigners();
       const prevBalance = await contract.balanceOf(owner.address);
 
       // Move one coin more than available coins.
-      await contract.transfer(receiver.address, prevBalance + 1);
-
+      // Workaround below since for some reason chai assertions don't handle exceptions properly,
+      // e.g. expect(cond).to.throw() doesn't work here
+      try {
+        await contract.transfer(receiver.address, prevBalance + 1);
+        expect(false).to.be(true);
+      } catch (_) {}
       expect(await contract.balanceOf(owner.address)).to.eq(prevBalance);
     });
   });
@@ -73,8 +76,7 @@ describe("ERC20 functionality", function () {
   });
 
   describe("Transfer From", function () {
-    // FIXME: In ZIL-4879
-    xit("Should be possible to transfer from one account to another [@transactional]", async function () {
+    it("Should be possible to transfer from one account to another [@transactional]", async function () {
       const [owner, _, sender, spender] = await ethers.getSigners();
 
       // Fund the 2nd account first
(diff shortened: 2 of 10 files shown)


tests/EvmAcceptanceTests/test/GasEstimations.js
```diff
@@ -0,0 +1,34 @@
+const {expect} = require("chai");
+const {web3} = require("hardhat");
+const web3_helper = require("../helper/Web3Helper");
+const zilliqa_helper = require("../helper/ZilliqaHelper");
+
+describe("Gas estimation with web3.js", function () {
+  let contract;
+  const tranferFundMinGas = 21000;
```
its-saeed, line 8: we prefer MACRO_CASE for const variables.

tests/EvmAcceptanceTests/test/GasEstimations.js
```diff
@@ -0,0 +1,34 @@
...
+  const create2MinGas = 32000;
+
+  describe("When a fund transfer is made", function () {
+    it("should return proper estimation [@transactional]", async function () {
+      const gasAmount = await web3.eth.estimateGas({
+        to: zilliqa_helper.getSecondaryAccountAddress(),
```
its-saeed, line 14: It's better to get all of your requirements from one of `zilliqa_helper` or `web3_helper`. Web3 helper does have this function for `primary account`. Just add another for `secondary` and use it.

tests/EvmAcceptanceTests/test/Erc20.js
```diff
@@ -45,14 +45,17 @@ describe("ERC20 functionality", function () {
...
       const [owner, receiver] = await ethers.getSigners();
       const prevBalance = await contract.balanceOf(owner.address);
 
       // Move one coin more than available coins.
-      await contract.transfer(receiver.address, prevBalance + 1);
-
+      // Workaround below since for some reason chai assertions don't handle exceptions properly,
+      // e.g. expect(cond).to.throw() doesn't work here
+      try {
+        await contract.transfer(receiver.address, prevBalance + 1);
```
its-saeed, line 56: if it reverts the transaction you can use [revert assertions](https://hardhat.org/hardhat-chai-matchers/docs/overview#reverts)

valeryz (approved):
LGTM!
(please address Saeed's comments).

I think we need to deploy evmdev after this change and play a lot with it!
Thanks!

tests/EvmAcceptanceTests/test/Erc20.js
bzawisto: Tried that, it actually doesn't revert but throws an exception during gas estimation.

tests/EvmAcceptanceTests/contracts/ERC20.sol
```diff
@@ -91,7 +91,7 @@ contract ZEVMToken is ERC20Interface, SafeMath {
  
     function transferFrom(address from, address to, uint tokens) external returns (bool success) {
         balances[from] = safeSub(balances[from], tokens);
-        allowed[from][msg.sender] = safeSub(allowed[from][msg.sender], tokens);
+        allowed[from][to] = safeSub(allowed[from][to], tokens);
```
its-saeed, line 94: seems like we shouldn't touch it. see (this)[https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.7.3/contracts/token/ERC20/ERC20.sol#L158]

bzawisto: Hm, what would be the solution in this case? Either the test is wrong or the contract code is wrong I guess?

its-saeed: yep. after reading openzepplin, I realized that it should be like this:

```solidity
    it("Should be possible to transfer from one account to another [@transactional] xyz", async function () {
      const [owner, _, sender, spender] = await ethers.getSigners();

      // Fund the 2nd account first
      await contract.transfer(sender.address, 5_000);

      // Approve 30_000 for withdrawal by Acc #3
      await contract.connect(sender).approve(owner.address, 3_000);

      // Try transferring allowed amount
      expect(await contract.transferFrom(sender.address, spender.address, 2_999))
        .to.changeTokenBalances(contract, [sender.address, spender.address], [-2_999, +2_999])
        .emit(contract, "Transfer")
        .withArgs(sender.address, spender.address, 1_999);

      expect(await contract.allowance(sender.address, owner.address)).to.be.eq(1);
    });
```

bzawisto: I checked your version and it still fails during eth_estimateGas (this is before transaction is made).

its-saeed: LMTAL

its-saeed: anyway, this change to the contract shouldn't have been committed. [It's not correct](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.7.3/contracts/token/ERC20/ERC20.sol#L158).

PR #3144, opened 2022-10-22
fireblocks tracing v1

n-hutton:
Add the framework for transaction tracing into the codebase (although the rust side needs to be extended to add additional information). Transaction tracing is a json struct of every instruction executed by the EVM and its state at each point. Currently the EVM only emits events like this when there is contract to contract calls, so I put the test for this in the child parent JS test, even though it is RPC.

WRT tracing lookup, rather than try to revert the blockchain to a particular block and execute every transaction up until the TX you want the trace for, we can store every TXs trace. To this end there is a database of tx hash to trace.


tests/EvmAcceptanceTests/test/ParentChild.js
```diff
@@ -42,10 +45,20 @@ describe("Parent Child Contract Functionality", function () {
...
     });
 
+    it("Should create a transaction trace after child creation", async function () {
+      const METHOD = "debug_traceTransaction";
+
+      await helper.callEthMethod(METHOD, 1, [installedChild.hash], (result, status) => {
+        hre.logDebug(result);
+
+        assert.equal(status, 200, "has status code");
+        assert.isString(result.result, "Expected to be populated");
```
its-saeed, line 58: couldn't it be more specific? using [method_property](https://www.chaijs.com/api/assert/#method_property) for example?

its-saeed (approved):
JSPLGTM (Javascript part looks good to me :D)

PR #3150, opened 2022-10-26
[master] Fix contract size check

valeryz:
## Description
What it says in the title. We still store contracts as HEX with EVM prefix.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libEth/Eth.cpp
+++ b/src/libEth/Eth.cpp
@@ -155,7 +155,9 @@ bool ValidateEthTxn(const Transaction &tx, const Address &fromAddr,
             std::to_string(DataConversion::UnpackB(tx.GetVersion())));
   }
 
-  if (tx.GetCode().size() > MAX_EVM_CONTRACT_SIZE_BYTES) {
+  // While checking the contract size, account for Hex representation
+  // with the 'EVM' prefix.
+  if (tx.GetCode().size() > 2 * MAX_EVM_CONTRACT_SIZE_BYTES + 3) {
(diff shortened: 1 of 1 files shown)


bzawisto (approved):

src/libEth/Eth.cpp
```diff
@@ -155,7 +155,9 @@ bool ValidateEthTxn(const Transaction &tx, const Address &fromAddr,
             std::to_string(DataConversion::UnpackB(tx.GetVersion())));
   }
 
-  if (tx.GetCode().size() > MAX_EVM_CONTRACT_SIZE_BYTES) {
+  // While checking the contract size, account for Hex representation
+  // with the 'EVM' prefix.
+  if (tx.GetCode().size() > 2 * MAX_EVM_CONTRACT_SIZE_BYTES + 3) {
```
its-saeed, line 160: wouldn't it be better to replace that 3 with a `constexpr` to improve readability?

its-saeed: or maybe strlen("EVM") or something instead of comment?

its-saeed (approved):

src/libEth/Eth.cpp
valeryz: I started with a constexpr, but it is more code to read actually. I don't know how you can compile strlen or analog at compile time without creating another function elsewhere, but I'm bad at C++.  Seems more complicated than necessary to me (which doesn't help readability!)

PR #3151, opened 2022-10-26
[master] Fixing temporarily stoul usage

bzawisto:
## Description
Wrapping with try/catch blocks. In other cases, converting directly to uint_xxx types (which don't throw due to `unchecked` parameter passed to a template).

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/EthRpcMethods.cpp
+++ b/src/libServer/EthRpcMethods.cpp
@@ -664,14 +664,14 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
   uint256_t value = 0;
   if (json.isMember("value")) {
     const auto valueStr = json["value"].asString();
-    value = stoull(valueStr.c_str(), nullptr, 0);
+    value = uint256_t{valueStr};
   }
 
   uint256_t gasPrice = GetEthGasPriceNum();
   if (json.isMember("gasPrice")) {
     const auto gasPriceStr = json["gasPrice"].asString();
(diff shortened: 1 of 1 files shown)


src/libServer/EthRpcMethods.cpp
```diff
@@ -664,14 +664,14 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
...
-    value = stoull(valueStr.c_str(), nullptr, 0);
+    value = uint256_t{valueStr};
   }
 
   uint256_t gasPrice = GetEthGasPriceNum();
   if (json.isMember("gasPrice")) {
     const auto gasPriceStr = json["gasPrice"].asString();
-    const uint256_t gasPriceNum = stoull(gasPriceStr.c_str(), nullptr, 0);
-    gasPrice = max(gasPrice, gasPriceNum);
+    uint256_t inputGasPrice{gasPriceStr};
```
valeryz, line 673: Does this constructor throw?

bzawisto: Nope, since we use `unchecked` parameter for template it handles errors as specified in the table:
https://www.boost.org/doc/libs/1_80_0/libs/multiprecision/doc/html/boost_multiprecision/tut/ints/cpp_int.html

bzawisto: Ok, seems it throws unfortunately if arbitrary string is provided, will fix it.

valeryz (approved):

PR #3153, opened 2022-10-26
[master] Adding strToInt helpers

bzawisto:
## Description
Adding two helpers for basic unsinged and boost types. They allow for optional fallback otherwise throw. 
Only replaced usage in places we touched. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


CSideSteve (approved):
Nice

PR #3158, opened 2022-10-27
[master] Remove the incorrect handling of eth_coinbase.  Do what Alchemy/Infura do

valeryz:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


its-saeed:
`eth_coinbase.js` should change as well.

valeryz:
> We shouldn't remove `eth_coinbase.js` we should test it and expect it to return METHOD_NOT_FOUND. something like `eth_signTransaction.js`

PTAL, @its-saeed

its-saeed (approved):

PR #3167, opened 2022-10-31
ZIL-4957: remove BloomFilters

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz (approved):
Maybe @bzawisto you could look if you regret these (as we have bloom filters in Eth logs APIs, but apparently not this code.

valeryz:
Just noticed it is 8.3, not master, shouldn't we do it in master first and then cherry-pick?

yaron-zilliqa:
Not necessarily; I'd rather remove it here and the merge v8.3 -> master will take of this as well as other fixes/changes in v8.3.

PR #3168, opened 2022-10-31
[master] Run cargo fmt on evm-ds

its-saeed:
## Description
Running `cargo fmt` on evm-ds

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz (approved):
Thanks, indeed should be on git hook.

PR #3179, opened 2022-11-01
[master] Make build macos compatible

n-hutton:
naming the threads does not work/build on macos - the other option than deleting it is to have a macro/ifdef


src/libServer/EthRpcMethods.cpp
```diff
@@ -930,16 +930,16 @@ Json::Value EthRpcMethods::GetEthTransactionByHash(
                            "Sent to a non-lookup");
   }
   try {
-    TxBodySharedPtr transactioBodyPtr;
+    TxBodySharedPtr transactionBodyPtr;
```
n-hutton, line 933: unrelated spelling fix

PR #3184, opened 2022-11-02
[master] feat: evm-ds log4rs json format

mauromedda:
## Description
This PR will change the default format of the evm-ds log4rs from plain text to json format.
The json logs produced can be directly indexed by OpenSearch and searchable through Kibana without any additional/custom manipulation or filter.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/evm-ds/log4rs.yml
+++ b/evm-ds/log4rs.yml
@@ -1,13 +1,15 @@
 appenders:
-  stderr:
+  stdout:
     kind: console
     target: stderr
+    encoder:
+      kind: json
 
   filelog:
     kind: rolling_file
     path: "/run/zilliqa/evm.log"
     encoder:
-      pattern: "{d} {l} - {m}{n}"
+      kind: json
     policy:
       trigger:
         kind: size
@@ -22,4 +24,3 @@ root:
   appenders:
     - stderr
     - filelog
-


JamesHinshelwood (approved):

evm-ds/log4rs-local.yml
```diff
@@ -1,17 +1,26 @@
 appenders:
-  stderr:
+  stdout:
```
valeryz, line 2: I think stdout from evm-ds might be eaten by the node?
(although @CSideSteve recently changed the invocation mechanism and this may no longer be true).

evm-ds/log4rs-local.yml
```diff
@@ -1,17 +1,26 @@
...
     kind: console
-    target: stderr
+    target: stdout
+    encoder:
+      kind: json
```
valeryz, line 6: Why do we need json in the "local" configuration?

evm-ds/log4rs-local.yml
CSideSteve: The output is not currently redirected, it used to be sent to /dev/null , the facility exists to redirect it to a specific place but currently stdout and stderr are not redirected for evm-ds.

mauromedda: Yes, the stdout is eaten by a tail -f /dev/null in the pod configuration. we are streaming the evm-ds logs to a file and reading it back. Just set stdout for consistency with pods definition practice.

evm-ds/log4rs-local.yml
valeryz: We are seeing this in the console output, and I'm not sure json is the right format for that.

evm-ds/log4rs-local.yml
valeryz: I think stdout used to be eaten by the parent process. Maybe not anymore. But I feel like stderr is the right outptu for that.  You can e..g feed isolated server logs into one place (stdout), and evm-ds as stderr to another.

I think rather let's not change the local config?

valeryz (approved):
please revert local config, otherwise LGTM.

PR #3187, opened 2022-11-02
[master] ZIL-4967: Fix Zil contract address calculation

JamesHinshelwood:
We also need to decrement the nonce for Zil-style transactions before calculating contract addresses.

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -558,6 +558,12 @@ Json::Value LookupServer::CreateTransaction(
 
     Transaction tx = JSONConversion::convertJsontoTx(_json);
 
+    if (tx.IsEth()) {
+      throw JsonRpcException(RPC_INVALID_PARAMETER,
+                             "Eth txs not supported for CreateTransaction api "
+                             "- use eth_sendRawTransaction");
+    }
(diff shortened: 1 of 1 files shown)


n-hutton (approved):

valeryz:
Is this a backport of 8.3 change?

valeryz:
or, rather, not backport, but forward-port?

src/libServer/LookupServer.cpp
```diff
@@ -558,6 +558,12 @@ Json::Value LookupServer::CreateTransaction(
 
     Transaction tx = JSONConversion::convertJsontoTx(_json);
 
+    if (tx.IsEth()) {
+      throw JsonRpcException(RPC_INVALID_PARAMETER,
+                             "Eth txs not supported for CreateTransaction api "
+                             "- use eth_sendRawTransaction");
+    }
+
```
valeryz, line 566: Yes! It is time! 

We had used this interface for EVM as well initially until we had eth_sendRawTransaction.

yaron-zilliqa:
No need to raise 2 at the same time. You can raise one for the release branch and this will be carried forward by a merge from release -> master.

PR #3189, opened 2022-11-02
[master] Fix simple send of value to an EVM contract

valeryz:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/test/Transfer.js
+++ b/tests/EvmAcceptanceTests/test/Transfer.js
@@ -30,7 +30,6 @@ describe("ForwardZil contract functionality", function () {
     );
   });
 
-  // FIXME: https://zilliqa-jira.atlassian.net/browse/ZIL-4954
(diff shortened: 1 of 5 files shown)


tests/EvmAcceptanceTests/test/Transfer.js
```diff
@@ -31,7 +31,7 @@ describe("ForwardZil contract functionality", function () {
   });
 
   // FIXME: https://zilliqa-jira.atlassian.net/browse/ZIL-4954
```
its-saeed, line 33: this should be removed as well.

CSideSteve (approved):
Makes sense.

tests/EvmAcceptanceTests/test/Transfer.js
valeryz: done

PR #3194, opened 2022-11-04
[master] Fix constant in constants for not starting evm

n-hutton:
Its launch_evm not evm_launch


CSideSteve (approved):
Must have had a brain hickup

PR #3196, opened 2022-11-06
Less JSON, more Protobufs!

valeryz:
## Description
This PR replaces communication mechanism between Zilliqa node and evm-ds from JSON to Protobuf.
It is not yet gRPC, it still uses JSONRPC, but the serialization is simple:

    - Arguments: array with one parameter element, which is `base64(protobuf_serialization(EvmEvalArgs))`
    - Return value: a string which is `base64(protobuf_serialization(EvmResult))`

At some point we'll replace JSON with gRPC, but already now we can enjoy the type safety of protobuf.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


evm-ds/src/convert.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+impl From<U256> for EvmProto::UInt256 {
+    fn from(num: U256) -> Self {
+        let mut result = Self::new();
+        result.set_x3(num.low_u64());
+        result.set_x2(num.shr(64).low_u64());
+        result.set_x1(num.shr(128).low_u64());
+        result.set_x0(num.shr(192).low_u64());
```
JamesHinshelwood, line 34: Would this be clearer?
```suggestion
        // `U256` stores bytes in little-endian order.
        let U256([x3, x2, x1, x0]) = num;
        result.set_x3(x3);
        result.set_x2(x2);
        result.set_x1(x1);
        result.set_x0(x0);
```

evm-ds/src/main.rs
```diff
@@ -268,63 +172,79 @@ async fn run_evm_impl(
         }));
         // Scale back remaining gas to Scilla units (no rounding!).
         let remaining_gas = executor.gas() / gas_scaling_factor;
-        match result {
+        let result = match result {
             Ok(exit_reason) => {
+                match exit_reason {
```
JamesHinshelwood, line 177: FWIW we could probably flatten these `match`es into a single one by e.g. matching on `Ok(evm::ExitReason::Succeed(_)`, etc.

evm-ds/build.rs
```diff
@@ -3,13 +3,28 @@ extern crate protoc_rust;
...
             carllerche_bytes_for_bytes: Some(true),
             carllerche_bytes_for_string: Some(true),
             ..Default::default()
         })
         .run()
-        .expect("Running protoc failed.");
+        .expect("Running protoc failed for ScillaMessage.proto");
+
+    protoc_rust::Codegen::new()
+        .out_dir("src/protos")
```
JamesHinshelwood, line 17: Generally I've seen the convention to generate code into `$OUT_DIR/protos.rs` and the `include!(concat!(env!("OUT_DIR), "/protos.rs"));` in your source (e.g. https://doc.rust-lang.org/cargo/reference/build-script-examples.html#code-generation)

JamesHinshelwood (approved):
LGTM - Excited for some un-hexing :)

evm-ds/src/convert.rs
valeryz: Is little endian order of U256 guaranteed though?

evm-ds/rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2021"
```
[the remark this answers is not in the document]
valeryz: normally it is my edits that add a ton of newlines, this was something random :)

evm-ds/src/convert.rs
JamesHinshelwood: Fair point - I guess I'd argue yes because it says so in the documentation, but maybe you're right that explicitly asking for the low word is better.

> Little-endian large integer type 256-bit unsigned integer.

evm-ds/src/main.rs
valeryz: Good point, but  it would de-emphasize the nesting logic of these cases - why do we need that?

PR #3197, opened 2022-11-07
[release/v8.3] Fix enable scilla when evm enabled is disabled

chetan-zilliqa:
## Description
The following are the changes:

- Add EVM-specific constants in `constants.xml`
- Revert logic for contract creation call to send to normal shard. It's the existing behavior running in the mainnet.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/constants.xml
+++ b/constants.xml
@@ -190,7 +190,7 @@
         <CONNECTION_ALL_TIMEOUT>60</CONNECTION_ALL_TIMEOUT>
         <!-- Timeout in seconds for ONLY connection that reach our callback function, 0 means no timeout-->
         <CONNECTION_CALLBACK_TIMEOUT>0</CONNECTION_CALLBACK_TIMEOUT>
-        <ENABLE_EVM>true</ENABLE_EVM>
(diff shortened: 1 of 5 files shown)


constants.xml
```diff
@@ -190,7 +190,7 @@
         <CONNECTION_ALL_TIMEOUT>60</CONNECTION_ALL_TIMEOUT>
         <!-- Timeout in seconds for ONLY connection that reach our callback function, 0 means no timeout-->
         <CONNECTION_CALLBACK_TIMEOUT>0</CONNECTION_CALLBACK_TIMEOUT>
-        <ENABLE_EVM>true</ENABLE_EVM>
+        <ENABLE_EVM>false</ENABLE_EVM>
```
JamesHinshelwood, line 193: cc @valeryz @its-saeed Do we need to update any configuration in CI if we're changing this default to make sure the tests still pass?

JamesHinshelwood (approved):

PR #3200, opened 2022-11-07
dedup the code around TX creation for eth case

n-hutton:


CSideSteve (approved):
This looks nice, be aware that there are changes coming in with the protobuf change from Valeriy that obviates the need for hex and also the change that I am doing to eliminate ViewAccounts and alter the EvmInvocation interface.

PR #3201, opened 2022-11-07
[release/v8.3] ZIL-4972: fix/redesign logging

yaron-zilliqa:
## Description

1. Use g3sinks' LogRotate to implement log rotation instead of relying on *nix's logrotate.
2. Fix confusing "loggers" and uses sinks instead.
3. Uniform behavior when logging FATAL.
4. Consolidate common logging fields into common macros/manipulators.
5. Performance improvements (don't expand the current directory for each log statement).

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz:
@art-gor PTAL.  Maybe your  work on logging all JSONRPC stuff should use this change (needs to be in master too then).

src/cmd/signmultisig.cpp
```diff
@@ -15,6 +15,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+#include <fstream>
```
JamesHinshelwood, line 18: Are these includes used?

vcpkg-registry/ports/g3log/portfile.cmake
```diff
@@ -1,9 +1,13 @@
+set(PATCHES
+     sink_filter.patch)
+
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO KjellKod/g3log
     REF 2fca06ff6da5c67465b591f4d45e8fd14d531142 #v1.3.4
     SHA512 8dba89e5a08e44d585478470725e25e37486685d8fe4d3cb5e97c81013389c95d96bdde658244e425008169bc8a9fc2d34a065b83b110c62e73d3ccab9b2b9e1
     HEAD_REF master
+    PATCHES ${PATCHES}
```
JamesHinshelwood, line 10: It might be helpful to have a comment somewhere describing what these patches are doing and why we need them?

JamesHinshelwood:
Do you have an idea of how difficult outputting JSON would be? Does g3log support this natively at all? IIRC it would make @mauromedda and co.'s lives much easier if we could.

src/cmd/signmultisig.cpp
yaron-zilliqa: Yes.

vcpkg-registry/ports/g3log/portfile.cmake
yaron-zilliqa: Explained in the Jira ticket.

yaron-zilliqa:
Shouldn't be hard to do given that jsoncpp has a StyledStreamWriter which outputs to a std::ostream.
@mauromedda what exactly are you after here?

mauromedda:
> Shouldn't be hard to do given that jsoncpp has a StyledStreamWriter which outputs to a std::ostream. @mauromedda what exactly are you after here?
Independent and atomic field/value so that we don't have to parse the messages to extract the fields that have to be indexed by OpenSearch.
i.e. from log4
{
    "time": "2016-03-20T14:22:20.644420340-08:00",
    "message": "the log message",
    "module_path": "foo::bar",
    "file": "foo/bar/mod.rs",
    "line": 100,
    "level": "INFO",
    "target": "foo::bar",
    "thread": "main",
    "thread_id": 123,
    "mdc": {
        "request_id": "123e4567-e89b-12d3-a456-426655440000"
    }
}

vcpkg-registry/ports/g3log/portfile.cmake
JamesHinshelwood: Thanks, the explanation is good, but I still think it would be useful to have it in the source so future readers can find it.

src/libUtils/Logger.cpp
```diff
@@ -16,325 +16,262 @@
...
+         << ']';
+  return stream;
 }
-}  // namespace
 
-const streampos Logger::MAX_FILE_SIZE =
-    1024 * 1024 * 100;  // 100MB per log file
+std::ostream& logTimestamp(std::ostream& stream, const LogMessage& message) {
+  // The following is taken from g3log's localtime_formatted; we're changing
+  // it here to show our own time format in UTC instead of localtime.
```
JamesHinshelwood, line 71: Add a link and mention the license? We can't accidentally re-license it under GPL :)

src/libUtils/Logger.cpp
```diff
@@ -16,325 +16,262 @@
...
+using Level = LoggerManip<&logLevel>;
+
+std::ostream& logMessageCommon(std::ostream& stream,
+                               const LogMessage& message) {
+  stream << ThreadId(message) << Timestamp(message) << CodeLocation(message)
+         << Message(message) << std::endl;
+  return stream;
+}
+
+// FIXME: make this configurable. This is per LogRotate sink.
```
JamesHinshelwood, line 112: File a ticket for the `FIXME`?

JamesHinshelwood (approved):

src/libUtils/Logger.cpp
yaron-zilliqa: The license is completely free. Can be checked in the repo but I've added the link.

vcpkg-registry/ports/g3log/portfile.cmake
yaron-zilliqa: Ok, I added a short comment.

src/libUtils/Logger.cpp
```diff
@@ -16,325 +16,264 @@
...
+const int MAX_ARCHIVED_LOG_COUNT = 15;
```
chetan-zilliqa, line 115: can we make this configurable?

src/libUtils/Logger.cpp
```diff
@@ -16,325 +16,264 @@
...
+      .wait();
 }
 
-Logger::~Logger() { m_logFile.close(); }
+}  // namespace
 
-void Logger::checkLog() {
-  std::ifstream in(m_fileName.c_str(),
-                   std::ifstream::ate | std::ifstream::binary);
+const int Logger::MAX_FILE_SIZE = 1024 * 1024 * 10;  // 10MB per log file
```
chetan-zilliqa, line 218: can we make it configurable via constants.xml ?

chetan-zilliqa (changes requested):

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -27,6 +27,7 @@
 #include "libUtils/EvmUtils.h"
 #include "libUtils/GasConv.h"
 #include "libUtils/SafeMath.h"
+#include "libUtils/TimeUtils.h"
```
chetan-zilliqa, line 30: Should we have a common `includes.h` that wraps common include ? Just a suggestion, I am fine.

chetan-zilliqa (changes requested):

src/libData/AccountData/AccountStoreSCEvm.cpp
valeryz: actually having each .cpp include exactly what it needs and nothing else is much preferred.

src/libUtils/Logger.cpp
valeryz: +1, unlike many other things that we should make non-configurable, this one deserves it.

src/libUtils/Logger.cpp
```diff
@@ -16,325 +16,264 @@
...
   try {
-    if (!boost::filesystem::create_directory(this->m_logPath)) {
-      if ((boost::filesystem::status(this->m_logPath).permissions() &
+    if (!boost::filesystem::create_directory(logFileRoot)) {
+      if ((boost::filesystem::status(logFileRoot).permissions() &
            boost::filesystem::perms::owner_write) ==
           boost::filesystem::perms::no_perms) {
-        std::cout << this->m_logPath
+        useDefaultLocation = true;
+        std::cout << logFileRoot
```
valeryz, line 190: `std::cerr` would be more appropriate perhaps for error messages? in other places too.

src/libUtils/Logger.cpp
```diff
@@ -16,325 +16,264 @@
...
+template <typename LogRotateSinkT>
+void AddFileSink(LogWorker& logWorker, const std::string& filePrefix,
+                 const boost::filesystem::path& filePath,
+                 int maxFileSize /*= MAX_FILE_SIZE*/) {
+  auto logFileRoot = boost::filesystem::absolute(filePath);
+  bool useDefaultLocation = false;
   try {
-    if (!boost::filesystem::create_directory(this->m_logPath)) {
-      if ((boost::filesystem::status(this->m_logPath).permissions() &
+    if (!boost::filesystem::create_directory(logFileRoot)) {
```
valeryz, line 185: The standard patter (pre-boost) is to just do an mkdir, then bail out with an error if that breaks, just interpreting the errno.

No need to check permissions etc. - just fail loudly.  Makes sense to just crash if things are misconfigured.

valeryz (commented):
Left some comments. The general theme is not trying to be smart, if something is broken, just abort everything. We should be able to count on the envrionment being sanely configured.

src/libUtils/Logger.cpp
yaron-zilliqa: Made the change as per the FIXME comment.

src/libUtils/Logger.cpp
yaron-zilliqa: Made this change in the PR.

src/libUtils/Logger.cpp
yaron-zilliqa: This is now configurable.

chetan-zilliqa (approved):

PR #3202, opened 2022-11-09
[master] minor fixes regarding API

art-gor:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libServer/EthRpcMethods.cpp
```diff
@@ -459,6 +450,14 @@ std::string EthRpcMethods::CreateTransactionEth(
                              "Txn could not be added as database exceeded "
                              "limit or the txn was already present");
     }
+
```
valeryz, line 453: I don't like the idea that it "sometimes works".

PR #3203, opened 2022-11-09
[master] Add delegatecall tests.

its-saeed:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/test/Delegatecall.js
+++ b/tests/EvmAcceptanceTests/test/Delegatecall.js
@@ -11,7 +11,7 @@ describe("Delegatecall functionality", function () {
   });
 
   it("should delegate function call correctly", async function () {
-    const VALUE = 1000;
+    const VALUE = 1000000;
     const NUM = 123;
 
     const [owner] = await ethers.getSigners();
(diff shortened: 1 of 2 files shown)


tests/EvmAcceptanceTests/test/Delegatecall.js
```diff
@@ -0,0 +1,29 @@
...
+  before(async function () {
+    const Contract = await ethers.getContractFactory("Delegatecall");
+    this.delegateContract = await Contract.deploy();
+
+    const TestContract = await ethers.getContractFactory("TestDelegatecall");
+    this.testDelegateContract = await TestContract.deploy();
+  });
+
+  it("should delegate function call correctly", async function () {
+    const VALUE = 1000;
```
valeryz, line 14: This value in Wei will get lost in Zilliqa due to the fact that we round to 1,000,000. 
You need to use a value larger than 1mln.

valeryz (approved):
please try with a larger value. Otherwise LGTM.

tests/EvmAcceptanceTests/test/Delegatecall.js
its-saeed: yep it worked with 1,000,000

PR #3206, opened 2022-11-10
fix: update Rust to version 1.65

frankmeds:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


evm-ds/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "evm-ds"
 version = "0.1.0"
 edition = "2021"
+rust-version = "1.65.0"
```
valeryz, line 5: That's what I reverted in my original PR.

scripts/setup_environment.sh
```diff
@@ -60,3 +59,7 @@ rm cmake-3.24.2-Linux-x86_64.sh
 
 export PATH="$(pwd)/.local/bin:${PATH}"
 cmake --version
+
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
valeryz, line 63: I'm sure we don't want to always have the most recent version of Rust.
I was thinking maybe we have some build image where it is all preinstalled, and we just need to update one.  But installing new Rust on each build is unpreditable and flaky, IMO.

mauromedda (approved):
LGTM

scripts/setup_environment.sh
mauromedda: @valeryz yes, it has been pinned to specific version otherwise as you mentioned the process will be unpredictable and flaky.

pavlops (approved):

PR #3207, opened 2022-11-10
[master] Xxx queued transactions 2

Steve-White-UK:
## Description
This is an implementation of the system that supports a transaction envelope allowing multiple parts of the system to post transactions for processing, which will be utilised in future developments.

Review and help support manufacturing transactions without signature and public key.
Potentially can replace the once-called ViewAccounts implementation and flow all transactions through the same processing chain.

## Backward Compatibility

## Status
Not ready to test, still under development,
### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK:
redundant now AccountStore has advanced beyond this

PR #3211, opened 2022-11-11
ZIL-5000: move to C++20.

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


chetan-zilliqa (approved):

tests/Persistence/Test_TxPersistence.cpp
```diff
@@ -30,6 +30,12 @@
 
 using namespace std;
 
+struct Fixture {
+  Fixture() { INIT_STDOUT_LOGGER() }
```
JamesHinshelwood, line 34: Is there any way we can avoid writing this at the start of every test? It will get tedious quite quickly.

yaron-zilliqa: Maybe, but this is actually less code than having INIT_STDOUT_LOGGER() in every test case.
Some test suites do this in the 1st test but I don't like this because it essentially relies on the order of the tests whereas the fixtures are meant for something of this sort.
Not a huge of this either but we can fix this in a better way later.

PR #3216, opened 2022-11-14
Fix gas estimation

n-hutton:
The gas estimation was pretty broken such that the amount it estimated was based on the gas limit and not gas used.

This was caused by the 'estimate' flag used when passed to the evm. There might be a bug in the EVM but what happens is that when estimating the gas mde during a contract call it creates a 'gas stipend' of 95% of available gas, for the new stack to use (detailed here: https://leftasexercise.com/2021/09/19/q-understanding-the-ethereum-virtual-machine-part-iii/ ) - this is logged as a 'gas cost' to the gasometer and it should recieve a refund of the whole stipend after the contract call, but what in fact happens is in the gas estimate case is this is never reverted. So, the gas used after execution is always 95% of the gas limit!

Since the external behaviour/state writes are the same irrespective of the flag, I suggest we disable it.

There are other quality of life debugging changes. Comments inline. 

Update: the gas estimation was fixed in this PR: https://github.com/Zilliqa/Zilliqa/pull/3228


src/cmd/isolated_server.cpp
```diff
@@ -99,7 +99,12 @@ int main(int argc, const char* argv[]) {
   bool nonisoload{false};
   string uuid;
 
+  ENABLE_EVM = true;
```
n-hutton, line 102: So you do not have to change constants.xml every time.

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -236,7 +236,7 @@ bool AccountStoreSC<MAP>::UpdateAccountsEvm(
   error_code = TxnStatus::NOT_PRESENT;
   const Address fromAddr = transaction.GetSenderAddr();
 
-  uint64_t gasRemained = transaction.GetGasLimitEth();
+  uint64_t gasLimitEth = transaction.GetGasLimitEth();
```
n-hutton, line 239: incorrectly named variable

src/libData/AccountData/AccountStoreSCEvm.cpp
```diff
@@ -585,6 +611,8 @@ bool AccountStoreSC<MAP>::UpdateAccountsEvm(
         return false;
       }
 
+      // TODO: the cum gas might not be applied correctly (should be block
```
n-hutton, line 614: need to make jira ticket

src/libData/AccountData/AccountStoreTrie.cpp
```diff
@@ -71,7 +71,6 @@ Account* AccountStoreTrie<MAP>::GetAccount(const Address& address,
     return account;
   }
 
-  LOG_GENERAL(INFO, "resetRoot:" << resetRoot);
```
n-hutton, line 74: logging noise

src/libServer/EthRpcMethods.cpp
```diff
@@ -631,6 +631,7 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
                              "data argument invalid");
     }
   }
+  std::transform(data.begin(), data.end(), data.begin(), ::toupper);
```
n-hutton, line 634: so the args to evm are always uppercase (not a bugfix)

src/libServer/EthRpcMethods.cpp
```diff
@@ -703,7 +704,7 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
     throw JsonRpcException(ServerBase::RPC_INTERNAL_ERROR,
                            "Failed to get EVM call extras");
   }
-  args.set_estimate(true);
+  args.set_estimate(false);
```
n-hutton, line 707: Bugfix here

tests/EvmAcceptanceTests/test/GasEstimations.js
```diff
@@ -28,6 +37,16 @@ describe("Gas estimation with web3.js", function () {
...
       expect(gasAmount).to.be.at.least(CREATE2_MIN_GAS);
+
+      const result = await contract.methods
+        .installChild(123)
+        .send({gasLimit: 420000000, from: web3_helper.getPrimaryAccountAddress()});
+
+      expect(result).to.be.not.null;
+      const actualGas = result.gasUsed;
+
+      expect(actualGas).to.be.at.least(gasAmount * 0.9);
```
n-hutton, line 48: this test now makes sure the gas estimated is within 10% of what it actually costs to use

src/libServer/EthRpcMethods.cpp
valeryz: somehow this fix is already in master

src/libData/AccountData/AccountStoreSC.h
```diff
@@ -29,7 +29,6 @@
 #include "InvokeType.h"
 #include "libUtils/DetachedFunction.h"
 #include "libUtils/Evm.pb.h"
-#include "libUtils/EvmCallParameters.h"
```
n-hutton, line 32: no longer needed

valeryz (approved):
Thanks, useful cleanup and fixes.

PR #3219, opened 2022-11-14
Improve evm processing interface

CSideSteve:
## Description
Improve the interface to the EvmProcessing so that it is not specific to the legacy Zilliqa transaction.

The interface now supports a directcall interface as well as the transaction interface which should allow 
estimation code to execute via the exact code path that normal evm messages follow and be subject to all applicable
charging.

Also removes the need to program at the evm ptotobuf level but makes receipts and evm returns available to direct users should they need them

For understanding and review only, still some remaining testing and changes on the estimate gas functionality.

## Backward Compatibility

## Review Suggestion

Fully tested for all transactions with unit test and EvmAcceptance tests.

### Implementation

- [x] **ready for review**

- [x] local machine test

- [ ] small-scale cloud test


constants.xml
```diff
@@ -193,7 +193,7 @@
         <!-- Timeout in seconds for ONLY connection that reach our callback function, 0 means no timeout-->
         <CONNECTION_CALLBACK_TIMEOUT>0</CONNECTION_CALLBACK_TIMEOUT>
         <ENABLE_EVM>false</ENABLE_EVM>
-        <EVM_SERVER_BINARY>/usr/local/bin/evm-ds</EVM_SERVER_BINARY>
+        <EVM_SERVER_BINARY>/home/stephen/versions/latest/Zilliqa/evm-ds/target/debug/evm-ds</EVM_SERVER_BINARY>
```
valeryz, line 196: I think having your own copy of constants.xml that you feed into your isolated server runs, and you keep it outside of the workstree, will prevent accidental checkins.

constants.xml
```diff
@@ -294,7 +294,7 @@
     </recovery>
     <smart_contract>
         <ENABLE_SC>true</ENABLE_SC>
-        <SCILLA_ROOT>/scilla</SCILLA_ROOT>
+        <SCILLA_ROOT>/home/stephen/run/scilla13/scilla</SCILLA_ROOT>
```
valeryz, line 297: ditto

src/libData/AccountData/AccountStore.cpp
```diff
@@ -534,8 +534,9 @@ bool AccountStore::UpdateAccountsTemp(
     return false;
   }
   if (isEvm) {
+    EvmProcessContext p(blockNum,transaction,txnExtras);
     return m_accountStoreTemp->UpdateAccountsEvm(
-        blockNum, numShards, isDS, transaction, txnExtras, receipt, error_code);
+        blockNum, numShards, isDS, receipt, error_code, p);
```
valeryz: maybe it deserves a better name than `p`.

src/libData/AccountData/AccountStoreSC.h
```diff
@@ -268,6 +270,8 @@ class AccountStoreSC : public AccountStoreBase<MAP> {
   bool AddAccountAtomic(const Address& address, const Account& account);
 
   bool ViewAccounts(const evm::EvmArgs& args, evm::EvmResult& result);
+  bool EvmProcessMessage(EvmProcessContext& params,
+                                              evm::EvmResult& result) ;
```
valeryz, line 274: formatter will complain here

src/libServer/EthRpcMethods.cpp
```diff
@@ -577,6 +577,11 @@ string EthRpcMethods::GetEthCallEth(const Json::Value& _json,
 std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
   Address fromAddr;
 
+#ifdef NEW_TESTING
```
valeryz, line 579: Hope you remove this before submission?

src/libServer/EthRpcMethods.cpp
```diff
@@ -739,6 +744,146 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
   }
 }
 
+std::string EthRpcMethods::NewGetEthEstimateGas(const Json::Value& json) {
```
valeryz, line 746: I think you don't have to rename the method, you can replace it. 
(And `New...` is not a good name - one day it will be old enough)

tests/Persistence/Test_ExtSeedPubKeys.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include "libPersistence/BlockStorage.h"
 #include "libTestUtils/TestUtils.h"
-//#include "libUtils/TimeUtils.h"
+// #include "libUtils/TimeUtils.h"
```
valeryz, line 25: Please don't comment out things - remove them.  We have git history to undo things.

src/libLookup/Lookup.cpp
```diff
@@ -2934,7 +2934,7 @@ bool Lookup::ProcessSetMinerInfoFromSeed(
 bool Lookup::ProcessSetTxBlockFromSeed(
     const zbytes& message, unsigned int offset, const Peer& from,
     [[gnu::unused]] const unsigned char& startByte) {
-  //#ifndef IS_LOOKUP_NODE
+  // #ifndef IS_LOOKUP_NODE
```
valeryz, line 2937: plese remove

constants.xml
CSideSteve: Yeah, I do that for my tests, its the ./scripts/build.sh that needs a special one in the root directory that is different to the one checked in, I have raised this with platform many times.

src/libServer/EthRpcMethods.cpp
CSideSteve: Only if I can get it to work!! :-)

src/libServer/EthRpcMethods.cpp
CSideSteve: Its like that for this check in as it passes all tests with the old routine, this is the last routine we are now working on.

Hopefully very soon, the version with New at the front will replace the old one.

tests/Persistence/Test_ExtSeedPubKeys.cpp
CSideSteve: This was already like this, i added the space because my super nit picky clang 15 objected to the format of the commented-out code. :-)

src/libLookup/Lookup.cpp
CSideSteve: It looks like these are Artems placeholders marked as for later.

src/libData/AccountData/EvmProcessContext.cpp
```diff
@@ -0,0 +1,585 @@
...
+/*
+ * GetEvmReceipt()
+ */
+
+const TransactionReceipt& EvmProcessContext::GetEvmReceipt() const {
+  return m_evmRcpt;
+}
+
+bool EvmProcessContext::GenerateEvmArgs(evm::EvmArgs& arg) {
+  /*
```
valeryz, line 511: please clean up

src/libData/AccountData/EvmProcessContext.cpp
```diff
@@ -0,0 +1,585 @@
...
+    stringStream.clear();
+    stringStream << "Call to GetEvmExtraValues has failed";
+    m_journal.push_back(stringStream.str());
+    m_status = false;
+    return false;
+  }
+  return true;
+}
+/*
+ * Determine the type of call that is required by Evm Processing
```
valeryz, line 545: ditto

valeryz (commented):
Aside from one bug found,  I worry a bit about too much string manipulation in the diagnostics.

It seems useful, but we should be able to make it conditioned on some future 'control plane' settings as you wanted in the past.  For now, maybe it would be useful to wrap it into some macro/function that could be enable/disabled in constants?

I can easily get more expensive than a typical eth call!

valeryz (dismissed):
Conditionally approving.  Need to fix the gasprice bug, the rest of the comments can be done later.

CSideSteve:
The code still contains some diagnostics that can be removed after proven on staging and testnet, it was useful in development but needs replacing with more substantial diagnostics as discussed in the project roadmap.

CSideSteve:
Work on this branch was temporarily suspended but left as WIP with irrelevant sections removed.
Optimised to remove unnecessary copies all processing done in place.

bzawisto (approved):

PR #3221, opened 2022-11-14
[master] Upgrade evm-ds to latest version 0.37

n-hutton:
Most notably the precompiles want the cost to be returned as well, now


evm-ds/src/convert.rs
```diff
@@ -178,3 +181,153 @@ impl From<(Bytes, Bytes)> for EvmProto::Storage {
         result
     }
 }
+
+fn as_string(opcode: &Opcode) -> String {
```
valeryz, line 185: This should rather be done as `impl fmt::Display for Opcode`

valeryz: which you can as well implement by calling into fmt::Debug imlementation that Rust already provides for enums

evm-ds/src/main.rs
```diff
@@ -3,6 +3,8 @@
 // #![deny(warnings)]
 #![forbid(unsafe_code)]
 
+extern crate core;
```
valeryz, line 6: I'm not sure why this is needed.

valeryz (approved):
Thanks, all seems good, but a few suggestions about the Rust style.

evm-ds/src/main.rs
n-hutton: It's not

evm-ds/src/convert.rs
valeryz: Indeed, as discussed in chat, outside evm crate you can't do it, please ignore the above.

PR #3225, opened 2022-11-15
[master] Point to the new version of jsonrpc Rust libs with a more correct fix

valeryz:
asking @n-hutton to retest


n-hutton (approved):
looks great!

PR #3229, opened 2022-11-16
[master] ZIL-5000: patch websocketpp 0.8.2 to compile on gcc in C++20.

yaron-zilliqa:
## Description

* Add a vcpkg port to patch websocketpp for higher gcc versions with C++20.
* Add a link to the license in README.md (are we required to include the full copy?)
* This might be a temporary solution as there's already a PR in the websocketpp repo for this but for master (and not 0.8.2).

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,8 @@ $ cd /path/to/vcpkg && git checkout 2022.09.27 && ./bootstrap-vcpkg.sh
 $ cd /path/to/zilliqa
 $ export VCPKG_ROOT=/path/to/vcpkg
 ```
-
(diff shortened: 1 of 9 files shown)


JamesHinshelwood (approved):
What a pain :( Maybe we can drop the dependency at some point?

yaron-zilliqa:
> What a pain :( Maybe we can drop the dependency at some point?

Definitely hope so... along with a few other.
Strangely this patch isn't needed in gcc 9 or in clang. It's specifically for higher versions of gcc...

PR #3235, opened 2022-11-18
[master] Fixing subscription tests

bzawisto:
## Description
This should help with non-deterministic behavior and sporadically failing tests (due to e.g. hang websocket connection, etc).

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/test/Subscriptions.js
+++ b/tests/EvmAcceptanceTests/test/Subscriptions.js
@@ -3,12 +3,6 @@ const {ethers} = require("hardhat");
 
 const general_helper = require("../helper/GeneralHelper");
 
-// Ensures all events in current eventloop run are dispatched
-async function waitForEvents(events, timeout = 0) {
-  await new Promise((r) => setTimeout(r, timeout));
-  return events;
-}
-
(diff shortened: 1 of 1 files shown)


tests/EvmAcceptanceTests/test/Subscriptions.js
```diff
@@ -18,13 +12,23 @@ describe("Subscriptions functionality", function () {
...
     provider = new ethers.providers.WebSocketProvider(general_helper.getWebsocketUrl());
     eventsContract = new ethers.Contract(contract.address, contract.interface, provider);
   });
+
   afterEach(async function () {
     eventsContract.removeAllListeners();
   });
 
+  // Ensures all events in current eventloop run are dispatched
+  async function waitForEvents(events, timeout = 5000) {
```
its-saeed, line 27: does it work? shouldn't you change all calls to this function to `this.waitForEvents`?

bzawisto: Nope, it looks that the scope is not private and it works without `this`.

its-saeed (approved):

PR #3237, opened 2022-11-21
[master] Transport dependencies reform

art-gor:
## Description

Removed MHD, jsonrpccpp backends, they all spinlocked and we were unable to detect performance bottlenecks. Now we can
.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,7 +6,7 @@ spec:
     beta.kubernetes.io/os: "linux"
   containers:
   - name: "scilla"
-    image: "zilliqa/scilla:v0.13.0"
+    image: "zilliqa/scilla:v0.13.1-alpha"
     command:
     - cat
     tty: true
@@ -58,11 +58,13 @@ timestamps {
                     sh "apt update -y && apt install -y patchelf chrpath"
                     sh "chmod u+w /scilla/0/_build/install/default/bin/*" 
                     sh "patchelf --set-rpath \"\$(pwd)/build/vcpkg_installed/x64-linux-dynamic/lib\" /scilla/0/_build/install/default/bin/*" 
+                    sh "ldd /scilla/0/_build/install/default/bin/*" 
                 }
             }
             container('ubuntu') {
                 env.VCPKG_ROOT="/vcpkg"
                 stage('Build') {
+                    sh "apt update -y && apt install -y libsecp256k1-dev"
                     sh "update-alternatives --install /usr/bin/python python \$(which python3) 1"
                     sh "git config --global --add safe.directory '*'"
                     // Hack: since Jenkins checks out the branch to a different directory each time and given
@@ -71,6 +73,7 @@ timestamps {
                     sh "git clone file://\"\$(pwd)\" /root/zilliqa"
                     sh "git -C /root/zilliqa checkout \$(git rev-parse HEAD)"
                     sh "ln -s /root/zilliqa/build build"
+                    sh "ldd /scilla/0/_build/install/default/bin/*" 
                     sh "cd /root/zilliqa && VCPKG_ROOT=${env.VCPKG_ROOT} CCACHE_BASEDIR=\"\$(pwd)\" ./scripts/ci_build.sh"
                 }
                 stage('Integration test') {
--- a/docker/Dockerfile.ccache
+++ b/docker/Dockerfile.ccache
@@ -53,5 +53,6 @@ COPY --from=deps  ${VCPKG_ROOT}/triplets   ${VCPKG_ROOT}/triplets
 COPY --from=deps  ${VCPKG_ROOT}/versions   ${VCPKG_ROOT}/versions
 COPY --from=deps  ${VCPKG_ROOT}/.vcpkg-root   ${VCPKG_ROOT}/.vcpkg-root
 COPY --from=deps  ${VCPKG_ROOT}/vcpkg   ${VCPKG_ROOT}/vcpkg
+COPY --from=deps  ${VCPKG_ROOT}/.git   ${VCPKG_ROOT}/.git
 
 WORKDIR /
--- a/src/libServer/EthRpcMethods.cpp
+++ b/src/libServer/EthRpcMethods.cpp
@@ -752,8 +752,7 @@ string EthRpcMethods::GetEthCallImpl(const Json::Value& _json,
     Account* contractAccount =
         AccountStore::GetInstance().GetAccount(addr, true);
     if (contractAccount == nullptr) {
-      throw JsonRpcException(ServerBase::RPC_INVALID_PARAMS,
-                             "Account does not exist");
+      return "0x";
     }
     code = contractAccount->GetCode();
   }


valeryz:
Haven't yet looked in detail at the change itself, but it looks like it should be two separate PRs: 

1) vendoring of jsonrpccpp, (@yaron-zilliqa could you please have a look of this is the best option, and if there's no better option with vcpkg), 
2) removal of MHD, and other reform.

CSideSteve:
Built and tested the code on Ubuntu 22.10 and Apple Mac M2 using clnag 14 C++ (LLVM).
    Both work and perform as expected.

Debugged through the EvmClient code to inspect code path and happy that old modifications can be discarded and the new implementation does not exhibit previous bugs exhibited by jsonrpccpp socket layer.

Number of files in branch seems overly large, I suspect this will reduce substantially when depends files are moved into platform vcpkg or successor.

Overall very pleased with the way this branch performs, need to eliminate some compiler warning on very new compilers, but probably not in this branches code as such.

Would be interested to see cpu profiles on staging and testnet, I can only imagine they are sunstantially flatter  than the the code it is going to supercede.

src/libServer/APIServer.h
```diff
@@ -50,6 +54,9 @@ class APIServer {
     /// Limit in bytes for POST bodies of incoming requests (security)
     size_t inputBodyLimit = 5 * 1024 * 1024;
```
yaron-zilliqa, line 55: A suggestion: use the unit as part of the name (more readable and saves the developer the need to look up the comments): _limitBodyLimitBytes_

src/libServer/GetWorkServer.cpp
```diff
@@ -34,10 +34,22 @@ using namespace jsonrpc;
 
 static ethash_mining_result_t FAIL_RESULT = {"", "", 0, false};
 
+namespace {
+
+std::shared_ptr<rpc::APIServer> GetServerConnector() {
+  rpc::APIServer::Options options;
+  options.threadPoolName = "GetWork";
+  options.numThreads = 2;
```
yaron-zilliqa, line 42: Any specific reason why it's 2? comment?

src/libServer/LocalAPIServer.cpp
```diff
@@ -0,0 +1,184 @@
...
+  m_thread.emplace([this] { WorkerThread(); });
+
+  return true;
+}
+
+bool LocalAPIServer::StopListening() { return DoStop(); }
+
+namespace {
+
+void WakeAcceptor(const boost::asio::ip::tcp::endpoint& endpoint) {
```
yaron-zilliqa, line 84: What's the point of this?

src/libServer/LocalAPIServer.cpp
```diff
@@ -0,0 +1,184 @@
...
+    std::string readBuffer;
+    std::string response;
+
+    while (m_started) {
+      boost::asio::ip::tcp::socket socket(m_asio);
+
+      boost::system::error_code ec;
+      m_acceptor->accept(socket, ec);
+      if (ec && m_started) {
+        throw std::runtime_error(ec.message());
```
yaron-zilliqa, line 120: Consider throwing a system_error constructed from ec.

src/libServer/UnixDomainSocketClient.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    // Write the JsonRpc
+    try {
+      boost::asio::write(unixDomainSocket,
+                         boost::asio::buffer(toSend, toSend.length()));
+    } catch (std::exception& e) {
+      if (LOG_SC) {
+        LOG_GENERAL(INFO, "Exception calling write " << e.what());
+      }
+      throw e;
```
yaron-zilliqa, line 55: Can just be: throw; to rethrow the same exception.

src/libServer/UnixDomainSocketClient.h
```diff
@@ -0,0 +1,40 @@
...
+#include <jsonrpccpp/client.h>
+
+// This is a Custom socket handler using asio
+// for the client connection to evm-ds and scilla server
+
+namespace rpc {
+
+class UnixDomainSocketClient : public jsonrpc::IClientConnector {
+ public:
+  UnixDomainSocketClient(const std::string& path) : m_path(path){};
```
yaron-zilliqa, line 30: explicit?

tests/Server/Test_ScillaIPCServer.cpp
```diff
@@ -68,10 +69,10 @@ BOOST_AUTO_TEST_CASE(test_connection) {
 
 // Simple non-map query.
 BOOST_AUTO_TEST_CASE(test_query_simple) {
-  UnixDomainSocketServer s(SCILLA_IPC_SOCKET_PATH);
+  rpc::UnixDomainSocketServer s(SCILLA_IPC_SOCKET_PATH);
```
yaron-zilliqa, line 72: Perhaps move this to some fixtures class so it doesn't have to be repeated for each test?

src/libData/AccountData/ScillaClient.cpp
```diff
@@ -137,10 +138,9 @@ bool ScillaClient::CheckClient(uint32_t version, bool enforce) {
     return false;
   }
 
-  std::shared_ptr<jsonrpc::UnixDomainSocketClient> conn =
-      std::make_shared<jsonrpc::UnixDomainSocketClient>(
-          SCILLA_SERVER_SOCKET_PATH +
-          (ENABLE_SCILLA_MULTI_VERSION ? ("." + std::to_string(version)) : ""));
+  auto conn = std::make_shared<rpc::UnixDomainSocketClient>(
```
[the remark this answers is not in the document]
yaron-zilliqa: That point is that std::unique_ptr is "lighter" than std::shared_ptr, so I would default to using this unless necessary.

src/libServer/LocalAPIServer.cpp
yaron-zilliqa: It seems to be called from DoStop() so I'm not sure my guess would be correct.

CSideSteve (dismissed):
As long as platform approve or change the packaging of jsonrpc I am happy to approve this version after it has run on evmdev. the src/depends directory is on its way out, would be sad to add to it.

As we will have to support jsonrpc going forward it would be beneficial to remove our dependency on it completely and use either the new rpccpp from same author or a suitable alternative as this will reduce code size and modernise source in a few areas.

valeryz:
Ack, @art-gor please don't submit without approval from Platform regarding the jsonrpc package vs. vendoring in 'depends'.

src/libServer/GetWorkServer.cpp
valeryz: I believe the question here was to just put a comment in the code, not to make it a magic constant, @art-gor

yaron-zilliqa (approved):

PR #3239, opened 2022-11-21
Increase logging for evm calls

n-hutton:
Basically when you put on the LOG_SC flag, this should be more informative about what is going to and coming from the EVM, and not writing bytecode to the terminal and so having it printed as broken utf


CSideSteve (commented):
I think we have hijacked the LOG_SC flag from Scilla maybe we should rename it to VERBOSE once and for all.

CSideSteve (approved):
I think we have hijacked the LOG_SC flag from Scilla maybe we should rename it to VERBOSE once and for all.

PR #3244, opened 2022-11-24
[master] Remove one XML constant: EVM_ZIL_SCALING_FACTOR

valeryz:


src/libUtils/EvmUtils.h
```diff
@@ -28,6 +28,8 @@
 #include "libUtils/Evm.pb.h"
 #include "libUtils/TxnExtras.h"
 
+const int EVM_ZIL_SCALING_FACTOR = 1'000'000;
```
yaron-zilliqa, line 31: Not a great idea. This will be included in each translation unit that #include-s this header (and have static linkage). We should stop putting variables in the global namespace.

yaron-zilliqa (changes requested):

src/libUtils/EvmUtils.h
valeryz: I believe const gives it internal linkage, doesn't it?

Namespace pollution - well, I probably agree with you, but this one doesn't change the whole picture.

JamesHinshelwood (approved):

src/libUtils/EvmUtils.h
yaron-zilliqa: Of course - internal linkage (both static & const do... :).
It's not just about namespace pollution - far worse, it's about the order of initialization which is implementation defined and we can't control. Many of these constants are accessed through singletons before main() starts executing and can easily be accessed when uninitialized yet.

valeryz: I'm very very far from claiming being a C++ expert. Could you pls provide the correct idiomatic way of handling this?

valeryz:
This PR is now old, and I think it won't merge with master. In the future need to remove older constants.

PR #3246, opened 2022-11-24
[master] ZIL-4799: Simplify and improve `download_incr_DB.py`

JamesHinshelwood:
These changes come from the following understanding for the behaviour of the existing implementation.

The script downloads all objects from the `incremental/` and `statedelta/` directories from an S3 bucket into a specified local directory. Before starting the download, the script clears the local directory.

The S3 bucket contains a `.currentTxBlk` file specifying the TX block at which the data was uploaded. The script checks this file before and after downloading objects. If there has been newer TX blocks since the download, the script attempts to catch up by downloading what it assumes are the new files. There is some extra logic to restart the download completely if a DS epoch has been crossed during the download period.

Finally, there is some logic to extract `.tar.gz` files after they have been downloaded.

I have changed the script to instead perform this by shelling out to `aws s3 sync ...`. I claim the end-result is the same, despite some differences in the internal logic. The notable differences are:

1. We don't clear the existing persistence folders on startup. Instead, we pass the `--delete` flag to `aws s3 sync` which tells it to delete local files in the target directory that no longer exist in the S3 bucket. I claim this accomplishes the same as deleting and redownloading from scratch, but in much shorter time since we only download the delta between the old and new directories.

2. We don't attempt to catch up to newer TX blocks by downloading individual files. Instead, we just keep invoking `aws s3 sync` until we've managed to do so without crossing a TX block. (Note that both the old and new scripts have a race condition here where a new TX block might occur between the script terminating and the main process joining the network)

3. We don't restart the download if we cross a DS epoch. Since we're always downloading the latest version of persistence via a delta, I claim this is unnecessary.

Overall, I think this results in a simpler and more performant script. Specifically, there are large performance wins in the cases where the old script would delete the local directory and download everything from scratch.

### DIFF
--- a/scripts/download_incr_DB.py
+++ b/scripts/download_incr_DB.py
@@ -97,6 +97,7 @@ def GetAllObjectsFromS3(url, folderName="", subfolder=None):
 	if subfolder:
 		download_path += f"/{subfolder}"
 	dest = f"./{subfolder}" if subfolder is not None else "."
+	latest_exception = None
 	# We can usually expect a single failure during a large sync, because a file will disappear between the start and
 	# end of the invocation. There's no harm in retrying a few times, since we'll only download the delta between the
 	# each attempt.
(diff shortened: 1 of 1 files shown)


rrw-zilliqa (approved):
Hard for me to know what the implications of this are at the moment, I'm afraid - I'm not as close to the code as I should be - but I say let's try it and see how it goes?

PR #3248, opened 2022-11-25
[master] Merge v8.3 -> v8.4

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


docker/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN mkdir -p /scilla/0/bin2/ && cp -L /scilla/0/bin/* /scilla/0/bin2/ && strip /
 FROM zilliqa/zilliqa-ccache:${ZILLIQA_CCACHE_VERSION} AS builder
 
 # Manually input tag or commit, can be overwritten by docker build-args
-ARG COMMIT_OR_TAG=master
```
chetan-zilliqa, line 28: should be `v8.4.0` , right?

docker/Dockerfile.deps
```diff
@@ -24,7 +24,7 @@ ARG VCPKG_COMMIT_OR_TAG=2022.09.27
 ARG VCPKG_ROOT=/vcpkg
 
 # Manually input tag or commit, can be overwritten by docker build-args
-ARG COMMIT_OR_TAG=master
+ARG COMMIT_OR_TAG=v8.3.0
```
chetan-zilliqa, line 27: same as above.

chetan-zilliqa (changes requested):

docker/Dockerfile
yaron-zilliqa: Yes.. it's usually passed in as a --build-arg though in the command line though.

docker/Dockerfile.deps
yaron-zilliqa: Changed.

docker/Dockerfile
JamesHinshelwood: I don't know if this is correct - `8.4.0` isn't a tag in GitHub. Really I think we should just leave this blank if we're expecting it to be passed as a build-arg?

chetan-zilliqa (approved):

PR #3269, opened 2022-12-06
[master] disable warning so mac build passes

n-hutton:
We see this warning:

```
/Users/nhutton/repos/Zilliqa/src/depends/jsonrpccpp/common/procedure.cpp:27:24: error: passing an object that undergoes default argument promotion to 'va_start' has undefined behavior [-Werror,-Wvarargs]
  va_start(parameters, returntype);
                       ^
/Users/nhutton/repos/Zilliqa/src/depends/jsonrpccpp/common/procedure.cpp:25:33: note: parameter of type 'jsonrpc::jsontype_t' is declared here
                     jsontype_t returntype, ...) {
```

### DIFF
--- a/src/depends/jsonrpccpp/common/procedure.cpp
+++ b/src/depends/jsonrpccpp/common/procedure.cpp
@@ -13,8 +13,10 @@
 #include <cstdarg>
 #include <vector>
 
+#ifdef __clang__
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wvarargs"
+#endif
 using namespace std;
 using namespace jsonrpc;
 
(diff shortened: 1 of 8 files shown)


src/depends/jsonrpccpp/common/procedure.cpp
```diff
@@ -13,6 +13,10 @@
 #include <cstdarg>
 #include <vector>
 
+#ifdef __clang__
```
yaron-zilliqa, line 16: Just one point: please check we're allowed to change the source directly here based on the license.

yaron-zilliqa (approved):

PR #3272, opened 2022-12-06
[master] change evm-ds logging to fit in one line

valeryz:
Change evm-ds logging to fit in one line, so we don't have to look for context
Additionally, provide context, whether this is TXN ID, or estimate or eth_call.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/evm-ds/src/main.rs
+++ b/evm-ds/src/main.rs
@@ -17,7 +17,11 @@ use std::sync::{Arc, Mutex};
 
 use anyhow::Context;
 use clap::Parser;
-use evm::{backend::Apply, executor::stack::{MemoryStackState, StackSubstateMetadata}, tracing};
+use evm::{
+    backend::Apply,
+    executor::stack::{MemoryStackState, StackSubstateMetadata},
+    tracing,
+};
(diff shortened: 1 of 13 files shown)


CSideSteve (approved):

evm-ds/src/main.rs
```diff
@@ -222,7 +228,10 @@ async fn run_evm_impl(
                 result.set_trace(listener.traces.into_iter().map(Into::into).collect());
                 result.set_logs(logs.into_iter().map(Into::into).collect());
                 result.set_remaining_gas(remaining_gas);
-                debug!("Result: {:?}", result);
+                info!(
```
its-saeed, line 231: Is `info` deliberate? the rest are `debug`

its-saeed (approved):
I think you did it, but if not please run `cargo fmt`

evm-ds/src/main.rs
valeryz: yes, deliberate, so that we have all relevant info in one line and no other noise.

PR #3274, opened 2022-12-06
Provide correct gas limit to evm

bzawisto:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz:
Should the same logic be replicated in eth_estimateGas and eth_call, or have we done it correctly there?

bzawisto:
> Should the same logic be replicated in eth_estimateGas and eth_call, or have we done it correctly there?

Yes, they are also handled it this PR.

valeryz (approved):

PR #3283, opened 2022-12-12
[master] Cps for EVM

bzawisto:
## Description
CPS support form EVM. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


Steve-White-UK (approved):
I like the coding style, nothing show off or clever, just clear well laid out readable code.

I have only reviewed the CPP code it makes no sense for me to critique the rust code, there are plenty others to that.

Merged it into local branch tested on local cluster< fine.
built and tested on Kind local cluster just fine.

Running long test now, but happy to approve this.

evm-ds/src/cps_executor.rs
```diff
@@ -91,11 +94,35 @@ impl<'a> CpsExecutor<'a> {
...
+                let offset_len: U256 = U256::from(evm_feedback.get_calldata().get_offset_len());
+                let target_len = min(offset_len, U256::from(runtime.return_data_buffer().len()));
+
+                match runtime.machine_mut().memory_mut().copy_large(
+                    U256::from(evm_feedback.get_calldata().get_memory_offset()),
+                    U256::zero(),
+                    target_len,
+                    evm_feedback.get_calldata().get_data(),
+                ) {
+                    Ok(()) => {
```
valeryz, line 113: I absolutely don't understand what's happening here.  Could you add some comments/documentation.
Why do we push zero on success and one on failure?  If it is copied from EVM internals, maybe point to the source? Otherwise could you explain it in the comments?

evm-ds/src/cps_executor.rs
```diff
@@ -91,11 +94,35 @@ impl<'a> CpsExecutor<'a> {
         feedback: Option<EvmProto::Continuation>,
     ) -> Result<(), evm::ExitError> {
         if let Some(evm_feedback) = feedback {
```
valeryz, line 96: Shouldn't we also check here if cps is enabled? Surely, the feedback argument should always be `None` if it isn't, but a bit of defensive approach wouldn't hurt?

evm-ds/src/cps_executor.rs
```diff
@@ -286,6 +313,41 @@ impl<'a> Handler for CpsExecutor<'a> {
...
+                    .execute(&mut StackExecutorHandle {
+                        executor: &mut self.stack_executor,
+                        code_address,
+                        input: &input,
+                        gas_limit: target_gas,
+                        context: &context,
+                        is_static,
+                    })
+            {
+                return match result {
```
valeryz, line 328: Can a precompile itself cause a Capture::Trap? I think it can, e.g. we may have precompiles that actually stop EVM and yield control to Scilla in some non-default way.

Here I don't think we have that.

evm-ds/src/cps_executor.rs
```diff
@@ -286,6 +313,41 @@ impl<'a> Handler for CpsExecutor<'a> {
...
+                        self.stack_executor
+                            .state_mut()
+                            .metadata_mut()
+                            .gasometer_mut()
+                            .fail();
+                        Capture::Exit((ExitReason::Fatal(exit_status), Vec::new()))
+                    }
+                };
+            }
+
```
valeryz, line 350: GitHub doesn't allow reviewing the code not in the PR, but a comment about line 374 with
```
match result {
    Capture::Exit(s) => Capture::Exit(s)
    Capture::Trap(_) =>   ....
```
I don't understand why we cannot simply pass the result. If CPS is not enabled, why should there be any handling whatsoever, and not just pass `self.stack_executor.call()` directly?  

If it is necessary, please explain in the comments.

The same in `create` method.

valeryz (commented):
This is looking awesome, but I have many questions and I'm only done with the Rust part.
Will get to the C++ part soon-ish.

src/libData/AccountStore/services/evm/EvmProcessContext.h
```diff
@@ -73,7 +73,7 @@ struct EvmProcessContext {
                     const zbytes& code, const zbytes& data, const uint64_t& gas,
                     const uint256_t& amount, const uint64_t& blkNum,
                     const TxnExtras& extras, std::string_view context,
-                    bool estimate = false);
+                    bool estimate, bool direct);
```
valeryz, line 76: I think this requires some documentation in comments. Just as all the members of this class.

valeryz (approved):
Please address the feedback, especially comments/documentation, and merge.
A really nice PR, thanks!

tests/EvmAcceptanceTests/test/Transfer.ts
```diff
@@ -51,4 +52,113 @@ describe("Transfer ethers", function () {
...
+    const batchTransfer = await BatchTransferContract.deploy(addresses, ACCOUNT_VALUE, {
+      value: ACCOUNTS_COUNT * ACCOUNT_VALUE
+    });
+    await batchTransfer.deployed();
+
+    const balances = await Promise.all(accounts.map((account) => account.getBalance()));
+    balances.forEach((el) => expect(el).to.be.eq(ACCOUNT_VALUE));
+  });
+
+  xit("should be possible to batch transfer using a smart contract with full precision", async function () {
```
its-saeed, line 76: ignored tests should have a ticket to track them. We specify them in TODO comment above them.

tests/EvmAcceptanceTests/test/Transfer.ts
```diff
@@ -1,6 +1,7 @@
-import {expect} from "chai";
-import {parallelizer} from "../helpers";
-import {ethers} from "hardhat";
+import { expect } from "chai";
```
its-saeed, line 1: Please run lint on the code. {} spaces are not consistent throughout the code.

tests/EvmAcceptanceTests/test/Transfer.ts
```diff
@@ -51,4 +52,113 @@ describe("Transfer ethers", function () {
...
+    expect(initialBal).to.be.eq(0);
+
+    const [owner] = await ethers.getSigners();
+    let InitialOwnerbal = await ethers.provider.getBalance(owner.address);
+
+    // check enough funds + gas
+    expect(InitialOwnerbal).to.be.at.least(TRANSFER_VALUE * 1.1);
+
+    // Deploy with no funds at contract address, but send value
+    const SingleTransferContract = await ethers.getContractFactory("SingleTransfer");
```
its-saeed, line 113: to have parallelism please do not use ethers.js directly to deploy a contract.

evm-ds/src/cps_executor.rs
```diff
@@ -91,11 +94,35 @@ impl<'a> CpsExecutor<'a> {
         feedback: Option<EvmProto::Continuation>,
     ) -> Result<(), evm::ExitError> {
         if let Some(evm_feedback) = feedback {
+            // Pop placeholder placed on a stack by evm before trap
```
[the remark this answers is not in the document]
bzawisto: Well, I think it should be actually a `feedback_to_continuation` (it's embedded inside continuation and provided to resumable run). I'm happy to change the name.

evm-ds/src/cps_executor.rs
```diff
@@ -91,11 +94,35 @@ impl<'a> CpsExecutor<'a> {
...
+            runtime.machine_mut().stack_mut().pop()?;
```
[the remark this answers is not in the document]
bzawisto: It's guaranteed that it's there, sputnik pushes a placeholder on a stack before returning a trap.

evm-ds/src/cps_executor.rs
bzawisto: Yep I can fix that.

evm-ds/src/cps_executor.rs
bzawisto: Well, it's been a copy of code that handles precompiles from stack executor. I'll add comment where it comes from.

evm-ds/src/cps_executor.rs
bzawisto: It's possible to do that but it's not implemented yet (deliberately to reduce the scope of changes). We can address that at the time of implementing interop.

evm-ds/src/cps_executor.rs
bzawisto: I had to do repacking here because StackExecutor returns StackExecutor::CreateInterrupt which is infallible whereas CpsExecutor has different Trap type for Create and Call.

src/libData/AccountData/Account.cpp
```diff
@@ -582,6 +582,9 @@ const zbytes Account::GetInitData() const {
 }
 
 bool Account::SetImmutable(const zbytes& code, const zbytes& initData) {
+  LOG_GENERAL(WARNING,
```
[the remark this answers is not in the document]
bzawisto: Well, it's a leftover and should be removed, will clean this up.

src/libData/AccountStore/services/evm/EvmProcessContext.h
bzawisto: Ok, will do.

tests/EvmAcceptanceTests/test/Transfer.ts
bzawisto: Will do thanks.

tests/EvmAcceptanceTests/test/Transfer.ts
bzawisto: Roger.

tests/EvmAcceptanceTests/test/Transfer.ts
bzawisto: Will do.

src/libCps/CpsRunEvm.cpp
```diff
@@ -0,0 +1,568 @@
...
+    receipt.AddJsonEntry(entry);
+  }
+
+  Address thisContractAddress = ProtoToAddress(mProtoArgs.address());
+  Address accountToRemove;
+  Address fundsRecipient;
+  Amount funds;
+
+  // parse the return values from the call to evm.
+  for (const auto& it : result.apply()) {
```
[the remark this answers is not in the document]
bzawisto: I think it has more than one entry in case of selfdestruct (then we have a contract that should be removed and fund recipient). I left that in a loop otherwise we should specify another counter (not sure if result.apply() has known size).

evm-ds/src/cps_executor.rs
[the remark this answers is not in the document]
valeryz: ok, that sounds worse, let's leave feedaback then.

evm-ds/src/cps_executor.rs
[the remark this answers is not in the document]
valeryz: maybe add a comment for someone who like me might wonder.

evm-ds/src/cps_executor.rs
valeryz: good, I agree with that approach.  Leaving a TODO might help.

src/libCps/CpsRunEvm.cpp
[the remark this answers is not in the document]
valeryz: SG. document this please.

PR #3288, opened 2022-12-13
[master] Make sure evm knows the chain id

n-hutton:
The chain id is used for contracts like NFTs where the contract wants to verify the owner has signed some data.

So you can have 

```
		signature = await seller._signTypedData(
			{name: 'XXX', version: '1', 
			 chainId: chainid, 
			 verifyingContract: trading.address
			}, 
			{verify: [
				{ name: 'seller', type: 'address' },
				{ name: 'buyer', type: 'address' },
				{ name: 'price', type: 'uint256' },
			]}, 
			{seller: seller.address,
			 buyer: "0x0000000000000000000000000000000000000000", // buyer.address,
			 price: 1000,
			}
		);
```

And, in the contract:

```
		bytes32 hash = _hashTypedDataV4(keccak256(abi.encode(
			TYPE_HASH_VERIFY,
			seller,
			address(0), // unless the buyer makes an offer and seller signs the buyer's address which should be msg.sender
			price,
		)));

		require(hash.recover(signature) == seller, "Offer was not signed properly by the seller");

```

Although it is not immediately clear, _hashTypedDataV4 hashes in a 'domain' which is a unique combination of strings including the chain ID. So, this failed as the recovered signature was incorrect. 

See: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol and https://docs.openzeppelin.com/contracts/3.x/api/drafts


valeryz (approved):
Nice catch, thanks!

PR #3293, opened 2022-12-15
[master] Fix estimating gas when contract sends funds

n-hutton:
Estimating gas failed for this simple contract because the contract address does not have any 'funds' when the estimate gas call is made.

For the real call, the contract account is created and given balance, so it does work if you skip out on estimating gas.


src/libScilla/ScillaIPCServer.cpp
```diff
@@ -22,9 +22,18 @@
...
 
+#include "libData/AccountData/Account.h"
 #include "libPersistence/BlockStorage.h"
 #include "libPersistence/ContractStorage.h"
 #include "libUtils/DataConversion.h"
+#include <algorithm>
+#include <cctype>
+#include <string>
+
+#ifndef __APPLE__
```
valeryz, line 33: Why do we need bits/stdc++?

Could you pls also move standard headers to the top of the file, and make local includes follow them?

bzawisto:
It looks okish but a bit hacky. I have an impression that we should do a regular updateAccounts in this case too but always revert, otherwise I anticipate some pain in disassembling such workarounds when time comes for CPS.

src/libData/AccountData/AccountStore.h
```diff
@@ -254,6 +253,9 @@ class AccountStore
   std::condition_variable_any& GetPrimaryWriteAccessCond() {
     return m_writeCond;
   }
+
+  // temp public
+  std::shared_ptr<ScillaIPCServer> m_scillaIPCServer;
```
n-hutton, line 258: make this public so we can insert a 'shim'

src/libPersistence/ContractStorage.cpp
```diff
@@ -403,7 +403,7 @@ bool ContractStorage::FetchStateValue(const dev::h160& addr,
 bool ContractStorage::FetchExternalStateValue(
     const dev::h160& /*caller*/, const dev::h160& target, const zbytes& src,
     unsigned int s_offset, zbytes& dst, unsigned int d_offset, bool& foundVal,
-    string& type, uint32_t caller_version) {
+    string& type, uint32_t caller_version, Account* injected) {
```
n-hutton, line 406: add the option to 'inject' an account when getting the state value (since there is a lot of protobuf logic in here)

src/libScilla/ScillaIPCServer.cpp
n-hutton: ah, accidentally included these

PR #3296, opened 2022-12-15
[master] fix telemetry build for apple

n-hutton:


CSideSteve:
OberverType is too wide stretching and may not always be a long value it could be a  type originally in the example for longs it was defined as uint64_t but gave a problem this, which is why I left a comment near it that said looks like confusion or a bug in the typing of code.

I would rename it with 64 bitness in mind as other routines will use the double variant of the type.

PR #3299, opened 2022-12-16
[master] Simplify Dockerfiles

JamesHinshelwood:
This removes all the intermediate docker layers and builds Zilliqa in a single Dockerfile. Within that dockerfile we use separate stages for building and packaging the final image.

This also changes the Dockerfile to `COPY` source files from the local directory rather than cloning a copy of the repository from Github. This means we can build docker images without requiring the user to push their changes to Github.

Some effort is put into ordering layers to utilise caching effectively. For example, we first copy the vcpkg configuration, run vcpkg install, then copy in the rest of our source files and build Zilliqa. This should ensure that vcpkg dependencies are cached. However, we should definately put more effort into more effective caching, as builds can still take a while after changing the source.


rrw-zilliqa (approved):
Looks good to me :-) (with my limited idea of what constitutes "good")

docker/Dockerfile
```diff
@@ -13,47 +13,129 @@
...
+
+ENV INSTALL_DIR=/usr/local
 
 # Build & install Zilliqa
-RUN (( git clone -b ${COMMIT_OR_TAG} --depth 1 ${REPO} ${SOURCE_DIR} || git clone ${REPO} ${SOURCE_DIR} )) \
-  && git -C ${SOURCE_DIR} checkout ${COMMIT_OR_TAG} \
-  && cd ${SOURCE_DIR} && CCACHE_BASEDIR="${SOURCE_DIR}" ./build.sh ninja \
-  && cmake --build "${BUILD_DIR}" --target install \
-  && cd ${SOURCE_DIR}/evm-ds && PATH=${BUILD_DIR}/vcpkg_installed/x64-linux-dynamic/tools/protobuf:$PATH cargo build --release --package evm-ds
+RUN CCACHE_BASEDIR="/zilliqa" ./build.sh ninja \
```
yaron-zilliqa, line 118: Placing the code in /zilliqa used to cause issues with ccache which is why it was moved to /root/zilliqa (so it wouldn't be directly under /).
Might not make much of a difference in this docker file since the previous benefits from ccache are pretty much out the window.

docker/Dockerfile
```diff
@@ -13,47 +13,129 @@
...
+# If COMMIT_OR_TAG is a branch name or a tag, clone a shallow copy which is
+# faster; if this fails, just clone the full repo and checkout the commit.
+RUN git clone https://github.com/microsoft/vcpkg ${VCPKG_ROOT} \
+  && git -C ${VCPKG_ROOT} checkout ${VCPKG_COMMIT_OR_TAG} \
+  && ${VCPKG_ROOT}/bootstrap-vcpkg.sh
 
-# Manually input tag or commit, can be overwritten by docker build-args
-ARG COMMIT_OR_TAG=v8.4.0
-ARG REPO=https://github.com/Zilliqa/Zilliqa.git
-ARG SOURCE_DIR=/root/zilliqa
```
yaron-zilliqa, line 30: Not sure it was a good idea to get rid of all of these. See the comment below about ccache & its base dir.

docker/Dockerfile
```diff
@@ -73,51 +155,98 @@ RUN strip /usr/local/bin/grepperf \
...
    /usr/local/lib/libCommon.so \
    /usr/local/lib/libTrie.so
 
-FROM zilliqa/zilliqa-base:${ZILLIQA_BASE_VERSION}
+FROM zilliqa/scilla:${SCILLA_VERSION} AS scilla
+
+# run a copy -L to unfold the symlinkes, and strip all exes
+RUN mkdir -p /scilla/0/bin2/ && cp -L /scilla/0/bin/* /scilla/0/bin2/ && strip /scilla/0/bin2/*
+
+FROM ubuntu:20.04
```
yaron-zilliqa, line 171: Easily forgotten when upgrading the base image. Consider an argument. See the next comment for a similar issue of repeating the list of packages.

docker/Dockerfile
```diff
@@ -73,51 +155,98 @@ RUN strip /usr/local/bin/grepperf \
...
 
-ENV LD_LIBRARY_PATH=${INSTALL_DIR}/lib
+ENV LD_LIBRARY_PATH=/usr/local/lib
 
 RUN python3 -m pip install --upgrade pip setuptools wheel \
   && python3 -m pip install --no-cache-dir -r requirements3.txt
 
 RUN echo '#!/bin/bash\npython3 -m pip "$@"' > /usr/bin/pip3 \
   && chmod a+x /usr/bin/pip3 \
+  && rm /usr/bin/pip \
```
yaron-zilliqa, line 249: Why is this needed? if this is a remainder from a previous apt package, it's better to remove using apt and not with rm.

docker/Makefile
```diff
@@ -1,61 +1,14 @@
...
-
-release: check-commit
-	docker build -t zilliqa:v$(major).$(minor).$(fix) \
-		--build-arg ZILLIQA_BASE_VERSION="v$(major).$(minor).$(fix)" \
-		--build-arg ZILLIQA_CCACHE_VERSION="v$(major).$(minor).$(fix)" \
-		--build-arg COMMIT_OR_TAG="$(COMMIT_OR_TAG)" \
-		--build-arg EXTRA_CMAKE_ARGS="$(cmake_extra_args)" -f Dockerfile .
+tag="v$(major).$(minor).$(fix)"
 
+release: docker build -t zilliqa:$tag --build-arg EXTRA_CMAKE_ARGS="$(cmake_extra_args)" -f Dockerfile ..
```
yaron-zilliqa, line 14: You're relying on the value of the environment variable EXTRA_CMAKE_ARGS. If this includes "-DCMAKE_BUILD_TYPE=Debug" this wouldn't be a release build.
In my opinion, two options: one is to change the target name from _release_ to something else (like default or build). Another is to set -DCMAKE_BUILD_TYPE=RelWithDebInfo explicitly here and also provide a 'dev' target which sets -DCMAKE_BUILD_TYPE=Debug.

yaron-zilliqa (changes requested):
A single file is good but please go over the comments for the list of issues that arise from this change. Multiple Dockerfiles are by no means ideal but unfortunately, you've undone some of the benefits that they do.
Another important point is that currently we have to "paths" in our CI:
1. Through Jenkins pipeline  which builds inside a Docker image and invokes the build script directly (see Jenkinsfile).
2. Through a "standard" job which invokes docker build through the Makefile.

Have you tested that both work properly?

docker/Dockerfile
```diff
@@ -73,51 +155,98 @@ RUN strip /usr/local/bin/grepperf \
...
 
-FROM zilliqa/zilliqa-base:${ZILLIQA_BASE_VERSION}
+FROM zilliqa/scilla:${SCILLA_VERSION} AS scilla
+
+# run a copy -L to unfold the symlinkes, and strip all exes
+RUN mkdir -p /scilla/0/bin2/ && cp -L /scilla/0/bin/* /scilla/0/bin2/ && strip /scilla/0/bin2/*
+
+FROM ubuntu:20.04
+
+RUN apt-get update \
```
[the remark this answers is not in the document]
JamesHinshelwood: Is there an argument that it is better to leave them separate, because the packages required by the build are very unlikely to be the same as the packages required by the runtime and so, to minimise the size of the final image, we should have separate lists of packages?

Obviously, that requires a bit of work to find out what is required for each, but it shouldn't take too long - I'm happy to do that if you don't think its a bad idea?

Otherwise, I guess we could have a separate stage earlier on which installs these packages, used as the base by both the build and runtime stages.

docker/Dockerfile
```diff
@@ -13,47 +13,129 @@
...
+
 ARG SCILLA_VERSION=v0.13.1-alpha
-ARG ZILLIQA_BASE_VERSION=
-ARG ZILLIQA_CCACHE_VERSION=
 
-FROM zilliqa/scilla:${SCILLA_VERSION} AS scilla
+FROM ubuntu:20.04 AS builder
+
+# Format guideline: one package per line and keep them alphabetically sorted
+RUN apt-get update \
```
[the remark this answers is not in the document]
JamesHinshelwood: It didn't take very long to be honest. But since its very early on in the Dockerfile, it is likely to be cached 99% of the time.

docker/Dockerfile
```diff
@@ -13,47 +13,129 @@
...
-ARG BUILD_DIR=/root/zilliqa/build
-ARG INSTALL_DIR=/usr/local
-ARG VCPKG_ROOT=/vcpkg
+WORKDIR /zilliqa
+
+COPY vcpkg-registry/ vcpkg-registry
+COPY vcpkg-configuration.json .
+COPY vcpkg.json .
+
+RUN ${VCPKG_ROOT}/vcpkg install --triplet=x64-linux-dynamic
```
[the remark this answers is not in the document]
JamesHinshelwood: > Compiling the dependencies takes around 20-30 mins

Is there any way to reduce this time? I guess I'll try the `RUN` cache here too, but IMO its less important since our dependency don't change as often as our source code.

> They use a subset of those, but the whole idea is to make sure they use the same base image & dependencies (i.e. environment) as we do.

My idea would be to decouple the dependencies. I don't see why Scilla inherently needs to have the same dependencies as Zilliqa? If anything, as a Scilla developer it would probably annoy me a lot having to built a load of C++ files that I don't use. @jjcnn thoughts?

docker/Dockerfile
JamesHinshelwood: Yeah - Good to know thanks, lets see what happens to ccache..

docker/Dockerfile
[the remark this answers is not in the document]
yaron-zilliqa: The argument is exactly what I mentioned: you're duplicating the same thing twice which we will need to maintain twice and run twice (unless indeed it's cached). Just the maintenance issue is a problem because it's happened several times before that something was added in one place but forgotten in another.

docker/Dockerfile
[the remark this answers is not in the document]
yaron-zilliqa: "Didn't take very long" being re-run each build (and possibly twice) because long and more annoying given that you can get away with this.

docker/Dockerfile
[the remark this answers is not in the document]
yaron-zilliqa: > Is there any way to reduce this time? I guess I'll try the RUN cache here too, but IMO its less important since our dependency don't change as often as our source code.

As I've explained above: the set of dependencies are not stable yet and we've just gone through a phase of updating them pretty frequently. They are quite likely to also change and updated in the future. This would cause a recompilation of this whole thing time and again!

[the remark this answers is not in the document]
yaron-zilliqa: > My idea would be to decouple the dependencies. I don't see why Scilla inherently needs to have the same dependencies as Zilliqa? If anything, as a Scilla developer it would probably annoy me a lot having to built a load of C++ files that I don't use. @jjcnn thoughts?

What you're doing is the exact opposite of the whole point of the changes we've introduced in the last couple of months!

Having different dependencies, *especially* when we're using shared libraries all over the place is a pretty bad idea. We've already gone through a phase where Scilla was using different versions of our dependencies (many introduced as old apt packages). On top of one project having a completely different version of something like CryptoUtils it will also have been compiling with lower-level dependencies such as boost, libffi, etc..

I also don't get your argument about "being annoyed as a developer". You have no problem having our docker build for an extra 20-30 mins when our dependencies change (more often) but you find it annoying that for a Scilla developer this compilation happens once (for a reduced subset of those dependencies) and barely ever changes.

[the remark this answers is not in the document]
jjcnn: Just a few comments, without fully understanding what the PR does:
- Building Scilla from scratch takes maybe 20-30 seconds on my machine, so it really isn't prohibitively slow. It depends on the machine of course, but Scilla build time is not something I would consider an issue as of right now.
- Originally the ambition for Scilla was to make it as blockchain-independent as possible, which suggests that the dependencies should be decoupled as much as possible from those of the underlying blockchain. However, as of right now it looks unlikely that Scilla will be adopted by anyone other than us, so that's less of a concern right now. What is a concern right now is manpower shortage, so anything we can do to limit the amount of maintenance required is good by me. If at some point we need to port Scilla to another system, then we can revisit the problem at that time.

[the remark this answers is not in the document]
yaron-zilliqa: @jjcnn , @JamesHinshelwood 's argument is mainly about the dependencies (which can take longer to compile, but especially for you guys is really a one-off). 
It might be possible, using static libraries and only relying on the package manager for the underlying OS. This means writing more CMake find scripts and shifting to linking everything using static libraries.
At the moment, given that we use shared libraries all over the place and that we'd like to be able to upgrade the OS / compilers more frequently this isn't easy to do, ensuring the same shared libraries are used by all binaries.

docker/Dockerfile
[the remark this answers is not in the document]
JamesHinshelwood: Since its the 2nd step it will almost always be cached though, unless the list changes?

docker/Dockerfile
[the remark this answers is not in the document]
JamesHinshelwood: But my point is the actual required packages are probably different, so doesn't it make sense to have separate lists?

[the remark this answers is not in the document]
yaron-zilliqa: Compare this one and the one later in the file and you'll see that the two lists are almost identical. There's basically a list of basic apt packages that should be installed anyway + extra packages (for dependencies & environment utility) that are needed at a later stage. You're re-installing everything again at a later build stage. Takes both longer, and if you want to remove or add a common package, you'll have to do it twice. Not sure what the misunderstanding is.

JamesHinshelwood:
@yaron-zilliqa Could you have a look at my latest set of changes? We're utilising ccache again and sharing the list of dependencies between each stage. I hope that addresses at least some of your concerns? If you're happy, I'll start looking at CI.

docker/Dockerfile
```diff
@@ -13,47 +13,140 @@
...
+  && ${VCPKG_ROOT}/bootstrap-vcpkg.sh
 
-# Manually input tag or commit, can be overwritten by docker build-args
-ARG COMMIT_OR_TAG=v8.4.0
-ARG REPO=https://github.com/Zilliqa/Zilliqa.git
-ARG SOURCE_DIR=/root/zilliqa
-ARG BUILD_DIR=/root/zilliqa/build
-ARG INSTALL_DIR=/usr/local
-ARG VCPKG_ROOT=/vcpkg
+WORKDIR /zilliqa
```
yaron-zilliqa, line 105: As mentioned, not great to put this under /

docker/Dockerfile
```diff
@@ -13,47 +13,140 @@
...
+
+COPY vcpkg-registry/ vcpkg-registry
+COPY vcpkg-configuration.json .
+COPY vcpkg.json .
+
+RUN --mount=type=cache,target=/root/.cache/vcpkg/ ${VCPKG_ROOT}/vcpkg install --triplet=x64-linux-dynamic
```
yaron-zilliqa, line 111: Comment about this?

docker/Dockerfile
```diff
@@ -13,47 +13,140 @@
...
+COPY cmake/ cmake
+COPY daemon/ daemon
+COPY scripts/ scripts
+COPY constants.xml .
+COPY constants_local.xml .
+COPY build.sh .
+COPY CMakeLists.txt .
+
+ENV INSTALL_DIR=/usr/local
+ENV CCACHE_DIR="/ccache"
```
yaron-zilliqa, line 123: As above. Consider an alternative location.

docker/Dockerfile
```diff
@@ -13,47 +13,140 @@
...
+COPY evm-ds/ evm-ds
+RUN --mount=type=cache,target=/sccache/ cd evm-ds && PATH=/zilliqa/vcpkg_installed/x64-linux-dynamic/tools/protobuf:$PATH cargo build --release --package evm-ds && sccache --show-stats
+
+RUN ls -l /zilliqa/evm-ds/target
 
 # Make sure all DLLs / shared libraries are installed
-RUN ldd ${INSTALL_DIR}/bin/zilliqa | grep vcpkg_installed | gawk '{print $3}' | xargs -I{} cp {} ${INSTALL_DIR}/lib \
-  && cp ${BUILD_DIR}/vcpkg_installed/x64-linux-dynamic/lib/libffi.so ${INSTALL_DIR}/lib 
-RUN echo "built files:" && ls -lh ${BUILD_DIR} && echo "installed libs:" && ls -lh ${INSTALL_DIR}/lib
+RUN ldd /usr/local/bin/zilliqa | grep vcpkg_installed | gawk '{print $3}' | xargs -I{} cp {} /usr/local/lib \
```
yaron-zilliqa, line 136: INSTALL_DIR still defined above. Did you forget to revert this?

docker/Dockerfile
yaron-zilliqa: Forgot to ask previously: upon changing the dependencies which would trigger this; how does this work in parallel builds?

yaron-zilliqa (changes requested):

docker/Dockerfile
JamesHinshelwood: I tested quite a few times and it seemed to work fine. What was the issue you were thinking of?

docker/Dockerfile
JamesHinshelwood: It mounts the `/root/.cache/vcpkg` directory so it is shared between runs of `vcpkg install`. 

re parallel builds - https://docs.docker.com/engine/reference/builder/#run---mounttypecache
By default, it is used concurrently by multiple writers. I don't see a good reason to change this?

docker/Dockerfile
yaron-zilliqa: The ones I explained previously. Setting / as the basedir can have issues matching the correct file in the cache. In addition, it's explicitly stated in the ccache documentation:

_Don’t use / as the base directory since that will make ccache also rewrite paths to system header files, which typically is contraproductive_.

When you said you tried and it worked, did you try this with a warm cache? If so, what cache hit rate did you get?

docker/Dockerfile
yaron-zilliqa: If it works as expected and doesn't cause any issues I don't see a reason to change anything. I'm not familiar or sure how this works in the case of parallel builds and hope it wouldn't cause a recompilation of all the dependencies in each of those builds, hence the question.

docker/Dockerfile
JamesHinshelwood: The base dir here is `/zilliqa` though so surely that caveat doesn't apply?

Here's the logs from a run with a warm cache after I change a source file:
```
#32 0.349 Statistics zeroed
#32 0.349 ccache status
#32 0.350 cache directory                     /ccache
#32 0.350 primary config                      /ccache/ccache.conf
#32 0.350 secondary config      (readonly)    /etc/ccache.conf
#32 0.350 stats updated                       Thu Jan  5 15:37:35 2023
#32 0.350 stats zeroed                        Thu Jan  5 15:37:35 2023
#32 0.350 cache hit (direct)                     0
#32 0.350 cache hit (preprocessed)               0
#32 0.350 cache miss                             0
#32 0.350 cache hit rate                      0.00 %
#32 0.350 cleanups performed                     0
#32 0.350 files in cache                       957
#32 0.350 cache size                           4.1 GB
#32 0.350 max cache size                       5.0 GB
...
#32 37.16 ccache status
#32 37.17 cache directory                     /ccache
#32 37.17 primary config                      /ccache/ccache.conf
#32 37.17 secondary config      (readonly)    /etc/ccache.conf
#32 37.17 stats updated                       Thu Jan  5 15:38:03 2023
#32 37.17 stats zeroed                        Thu Jan  5 15:37:35 2023
#32 37.17 cache hit (direct)                   169
#32 37.17 cache hit (preprocessed)              13
#32 37.17 cache miss                             0
#32 37.17 cache hit rate                    100.00 %
#32 37.17 called for link                       39
#32 37.17 called for preprocessing              13
#32 37.17 no input file                          6
#32 37.17 cleanups performed                     0
#32 37.17 files in cache                       957
#32 37.17 cache size                           4.1 GB
#32 37.17 max cache size                       5.0 GB
```

JamesHinshelwood: Why would the Jenkins pipeline build from a different location? The Dockerfile is self-contained, so we'll only ever build inside `/zilliqa`. I think it would work from a separate location since the ccache is shared, but I don't see why that would ever matter.

yaron-zilliqa: Because the CI Jenkinsfile job works differently. Always has. You can see the Jenkinsfile for details and also some of the logs to get an idea (e.g. https://ci.pipeline.z7a.xyz/blue/rest/organizations/jenkins/pipelines/ZilliqaCIJenkinsfile/branches/valeryz-rc-3/runs/2/log/?start=0). It essentially builds by invoking the build script inside a container and doesn't simply builds a docker image the same way that Zilliqa Private Image Manual Trigger does.

Of course ccache is shared, but previously it used to have issues matching the same file if it was from a different location. I suggest you give it a try. If it works - great. If not, will need to be fixed.

JamesHinshelwood: Hmm, looking at the `Jenkinsfile` and those logs, I don't think we're using docker at all for that CI pipeline. We invoke `./ci_build.sh` inside a container, but that image is `zilliqa/zilliqa-ccache:v0.0.0`.

Frankly, I think we should throw away our CI soon and rebuild it so it isn't intractable.

The other Jenkins pipelines seem to just build an image with `docker build` (via `make`) so I think they should work fine?

yaron-zilliqa: That's precisely what I've been saying. You can read again one of my first comments here (https://github.com/Zilliqa/Zilliqa/pull/3299#pullrequestreview-1221685848) which was precisely about this and why I asked whether you'd tested both scenarios.

Both ci_build.sh & the Dockerfile (through docker build) eventually invoke build.sh but this will occur from different locations. The whole point of the ccache image was to provide a warm(er) ccache to speed up these jobs as well. So the question of whether this works as expected after your changes.

I don't like how we have these different ways of building things (Jenkinsfile vs. Dockerfile) any more than you do but it's a given. We can't break things or make them worse. 
@mauromedda has been working at improving this so indeed we should simplify this going forward. Another idea for improvement would be to provide a shared ccache to all builds. This however requires more work and coordination with the DevOps team.

Anyway, it feels a bit like we're going in circles here.

ccache currently doesn't provide a huge benefit for the Jenkinsfile builds (as there's no shared cache) but it does provide some and beside being the ground work for future improvements we really want to reduce build times which are still pretty high. Therefore, no reason to deprecate this.

JamesHinshelwood: The Jenkins job works - https://ci.pipeline.z7a.xyz/blue/organizations/jenkins/ZilliqaCIJenkinsfile/detail/PR-3299/4/pipeline/61

This PR didn't chance any of the build scripts, so it shouldn't change anything outside of Docker builds.

JamesHinshelwood: Ah sorry, I was missing the point, thank you :laughing: 

So I guess the open question is what to use as the base image for the native Jenkins build? It currently uses `zilliqa-ccache`, so we should either keep the Dockerfile for that around, or change the base image to something else?

With a static base image containing the cache, I'd assume that caching performance will go down with time, because the cache will get more and more stale, without being updated. Ideally, the cache directory would be mounted as a persistent volume, so that it gets updated by each subsequent build. cc @mauromedda ?

Is this something better to fix in a subsequent PR though? This shouldn't break anything, since the existing image will still exist in Dockerhub?

PR #3317, opened 2023-01-11
[core_staging] Send fns in P2PComm include distributed trace option, and placeholde…

art-gor:
…rs for possible traces

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


CSideSteve (approved):
Tested against core_staging and passes local 8.5 tests

PR #3318, opened 2023-01-11
[master] ZIL-5086: refactor & combine block & block headers into one library

yaron-zilliqa:
## Description

* Add unit tests for block & block header classes.
* Remove redundant code in those classes.
* Combine into one library in its own separate directory.
* Improve dependencies of the library.
* Move serialization code of those classes to the library itself.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


chetan-zilliqa (approved):
No harm in changing the directory from `libData` to `libBlockchain`, However, the contents of the folder is still deals with blockchain data structures. maybe we can name the folder `libBlockchainData`. Just suggestion but no complaints.

Steve-White-UK (approved):
Merged it local and ran tests against isolated server which pass 100% in integration and EvmAcceptanceTest

PR #3328, opened 2023-01-13
[master] Make AWS URL configurable in persistence scripts

JamesHinshelwood:
If the `AWS_ENDPOINT_URL` environment variable is set, the scripts will use the provided URL instead of the default.

This can be used with (for example) localstack to mock AWS locally by setting `AWS_ENDPOINT_URL=http://localhost:4566`.


mauromedda (approved):
From a high level view, looks good to me. The references have been updated in alignment with the performed changes.

PR #3330, opened 2023-01-13
[master] Fix TX traces

n-hutton:
Fixes TX tracing to work with CPS and on seed nodes (tested with local network) (callTracer only). Comments are inline.

For the CPS, I took the approach of having a string in the executor which is always passed into any evm run, and always gets set after any evm run. This is the JSON call stack/tx trace, and so it gets deserialized with the lower level call information on a continuation.

I would propose a similar scheme for the event logs, since AFAIK they are not collected and returned from subcalls currently.

Scilla calls will appear invisible as they are not implemented.

In order to achieve this, the seed pubs, when enabled, will wait until they have the TX block and all of the TXs. This isn't guaranteed to not stall the seed forever as usually it just applied the state delta as soon as it got the block hash.

The storage of TX traces to disk is just done using the TX hash as the lookup (so retains all). In a subsequent PR I will come up with a scheme to keep the last 100 blocks' worth of TXs.


evm-ds/src/main.rs
```diff
@@ -64,13 +66,86 @@ struct Args {
...
+            call_stack: Default::default(),
+            enabled: enabled,
+        }
+    }
 }
 
 impl tracing::EventListener for LoggingEventListener {
     fn event(&mut self, event: tracing::Event) {
-        self.traces.push(format!("{:?}", event));
+        println!("recvd event: {:?}", event);
```
n-hutton, line 124: Note: this never happens now due to trap call

src/libData/AccountStore/AccountStoreSCEvm.cpp
```diff
@@ -327,6 +329,21 @@ bool AccountStoreSC::UpdateAccountsEvm(const uint64_t &blockNum,
     libCps::CpsExecutor cpsExecutor{acCpsInterface, receipt};
     const auto cpsRunResult = cpsExecutor.Run(evmContext);
     error_code = cpsRunResult.txnStatus;
+
+    if (!cpsRunResult.evmResult.tx_trace().empty() && ARCHIVAL_LOOKUP_WITH_TX_TRACES) {
+      LOG_GENERAL(INFO, "Putting in TX trace for: " << evmContext.GetTranID());
+
+      if(!evmContext.GetTranID()) {
+        LOG_GENERAL(INFO, "hash is all zeroes, do nothing!");
```
n-hutton, line 337: Will not save TX traces from estimate gas calls

src/libServer/EthRpcMethods.cpp
```diff
@@ -749,6 +749,8 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value &json) {
     }
     LOG_GENERAL(WARNING, "Gas estimated: " << retGas);
 
+    retGas *= 2;
```
n-hutton, line 752: ! Here I double estimated gas returned because of a gas estimation bug. Either this PR can disable the TX trace test or it has some hack.

src/libServer/EthRpcMethods.cpp
```diff
@@ -1748,12 +1734,43 @@ Json::Value EthRpcMethods::GetDSLeaderTxnPool() {
...
+        BlockStorage::GetBlockStorage().GetTxTrace(tranHash, trace);
+
+    if (!isPresent) {
+      LOG_GENERAL(INFO, "Trace request failed! ");
+      return Json::nullValue;
+    }
+
+    Json::Value trace_json;
+    JSONUtils::GetInstance().convertStrtoJson(trace, trace_json);
+    auto const item = trace_json["call_stack"][0];
```
n-hutton, line 1766: The saved trace is JSON and has the map {"call_stack": [...]} - this makes it easier to later add other tracers if we desire (trace with full stack and memory, opcodes etc.)

bzawisto (approved):

PR #3333, opened 2023-01-13
[master] Zil/add more metrics

Steve-White-UK:
## Description
As the names suggests more an more metrics for the platform

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libUtils/Metrics.cpp
```diff
@@ -44,13 +48,87 @@ void Metrics::Init() {
...
+    std::unique_ptr<metrics_sdk::MeterSelector> observable_meter_selector{
+            new metrics_sdk::MeterSelector(name, version, schema)};
+    std::unique_ptr<metrics_sdk::View> observable_sum_view{
+            new metrics_sdk::View{name, "test_description", metrics_sdk::AggregationType::kSum}};
+    p->AddView(std::move(observable_instrument_selector), std::move(observable_meter_selector),
+               std::move(observable_sum_view));
+
+    // histogram view
+    std::string histogram_name = name + "_histogram";
+    std::unique_ptr<metrics_sdk::InstrumentSelector> histogram_instrument_selector{
```
bzawisto, line 115: Could it be implemented as std::make_unique<> ?

src/libData/AccountStore/AccountStore.cpp
```diff
@@ -518,6 +521,8 @@ bool AccountStore::UpdateAccountsTemp(
   LOG_GENERAL(WARNING, "[AS] Finished Processing <"
                            << transaction.GetTranID() << "> ("
                            << (status ? "Successfully)" : "Failed)"));
+
+  AccountStoreSC::m_transactionLatency = ((double)r_timer_end(tpStart)/1000.0);
```
bzawisto, line 525: better to use static_cast<>.

bzawisto (approved):

src/libData/AccountStore/AccountStore.cpp
Steve-White-UK: Yes, that's the old c programmer coming out, he is still in there somewhere, again this functionality is temporary as traces and spans carry the timing information, this is just to get a metric onto the dashboard this week. I will mark the code.

Steve-White-UK: changed in code anyway.

PR #3337, opened 2023-01-17
[master] Feature/scilla ext reads contract params

jjcnn:
## Description
Scilla needs to be able to query data from init files using the external reads system. This PR implements the changes needed on the C++ side of the queries and responses. 

The PR is coupled to https://github.com/Zilliqa/scilla/pull/1220. The changes in this PR and the changes in the Scilla PR are logically linked, and must be tested and released together.

## Backward Compatibility
- [x ] This is a breaking change

### Implementation

ProtoScillaQuery has been updated.

The actual reading of init data is currently only available in pseudo-code.
- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/evm-ds/src/scillabackend.rs
+++ b/evm-ds/src/scillabackend.rs
@@ -127,6 +127,7 @@ impl ScillaBackend {
             address, query_name, key, use_default
         );
         let mut query = ScillaMessage::ProtoScillaQuery::new();
+        query.set_is_mutable(true);
         query.set_name(query_name.into());
         if let Some(key) = key {
             query.set_indices(vec![bytes::Bytes::from(format!("{:X}", key))]);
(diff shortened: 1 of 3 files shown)


valeryz:
* interface is the same as for storage reads
* reads are from the init json, rather than key-value store
* pay attention to which types are returned.

jjcnn:
The attached zip file

 [remote_reads_contract_params_tests.zip](https://github.com/Zilliqa/Zilliqa/files/10550434/remote_reads_contract_params_tests.zip)

 contains a set of suggested regression tests, to be used as follows:

- Deploy `remote_reads_support.scilla` using the init file `support_init.json`. Remember to change the `_this_address` entry to the address that the contract is deployed at.
- Deploy `remote_reads_test.scilla` using the init file `test_init.json`. Remember to change the `_this_address` entry to the address that the contract is deployed at, and the entry `support_contract_init` to the address that the support contract is deployed at.
- Invoke the transitions `Test1`, `Test2`, ..., `Test18` on the test contract. None of them take any parameters. The expected behavior of all of them is that they terminate without raising a Scilla exception. If they terminate with a Scilla exception, then something has gone wrong in the implementation.

Note that I have not been able to test the contracts, so it is possible that I have made a mistake when writing them. If you encounter an exception, then the first thing to do is to make sure that the Scilla code actually does what it's supposed to.

rrw-zilliqa:
This is ancient and is likely never to be merged; closing.

PR #3341, opened 2023-01-17
[master] Zil 5090 load iso server accounts for quick testing

Steve-White-UK:
## Description
Here are the changes I used to get the node working with isolated server accounts and make testing pass.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libNode/Node.cpp
```diff
@@ -279,8 +312,15 @@ void Node::Init() {
   const auto& dsBlock = m_mediator.m_dsBlockChain.GetBlock(0);
   m_mediator.m_blocklinkchain.AddBlockLink(0, 0, BlockType::DS,
                                            dsBlock.GetBlockHash());
+#ifndef PRODUCTION_VERSION
```
yaron-zilliqa, line 315: Is this meant to be PRODUCTION_BUILD?

src/libNode/Node.cpp
```diff
@@ -60,6 +60,39 @@ const unsigned int MIN_CHILD_CLUSTER_SIZE = 2;
 
 #define IP_MAPPING_FILE_NAME "ipMapping.xml"
 
+#ifndef PRODUCTION_BUILD
+namespace {
+// testing version
+[[maybe_unused]] int readAccountJsonFromFile(const string& path) {
```
yaron-zilliqa, line 66: I suggest using std::filesystem::path for paths.

src/libNode/Node.cpp
```diff
@@ -60,6 +60,39 @@ const unsigned int MIN_CHILD_CLUSTER_SIZE = 2;
...
+  ifstream in(path.c_str());
+
+  if (!in) {
+    cerr << "Cannot open file \n" << path << endl;
+    return -1;
+  }
+  Json::Value _json;
```
yaron-zilliqa, line 73: Identifier prefixed with _ or double _ are reserved actually.

Steve-White-UK: Copied from isolated server code untouched by Yorkshire fingers.

src/libNode/Node.cpp
Steve-White-UK: Again if we find this useful then we should clean it up, its from our own code base.

src/libNode/Node.cpp
Steve-White-UK: I put this in as an example to show how to read in the same accounts we use in isolated-server for testing, it means then we have a stable set of results to test against whilst playing with this rig.

If people like this facility then it should be implemented as a command line option, there is some already but with input in other formats, this uses the exact same file as the isolated server.

JamesHinshelwood (commented):
IMO we should try to make this work with the genesis accounts, rather than adding code which will only be useful for development and actively harmful in production. We can compile it out in production builds, but it still adds unnecessary complexity in my view.

Steve-White-UK:
Closed as not relavant now

PR #3348, opened 2023-01-18
[master] AccountStore refactoring: step 2

valeryz:
Removing AccountStoreTrie class and merging all into AccountStore.

All new code in AccountStore is just moved methods of AccountStoreTrie.  New data members of AccountStore are marked in comments when they come from AccountStoreTrie.


src/libData/AccountStore/AccountStore.cpp
```diff
@@ -606,3 +692,124 @@ bool AccountStore::RevertCommitTemp() {
...
+    t_account.SetAddress(address);
+  }
+
+  account = std::move(t_account);
+
+  return true;
+}
+
+bool AccountStore::UpdateStateTrie(const Address& address,
+                                            const Account& account) {
```
bzawisto, line 753: Does it required clang-format?

bzawisto (approved):

src/libData/AccountStore/AccountStore.cpp
valeryz: Running clang-format now messes up the whole codebase, probably as a result of it being disabled for a while?

PR #3349, opened 2023-01-19
[master] add batch transfer test

n-hutton:
Arrange it such that gas estimation follows the exact same path as  normal execution, only it doesn't make any changes.

Previously, it just did the same as eth_call basically.

The evm process context needed to be updated to hide whether it held an actual transaction, or just the information of one, as the account store was doing evmProcessContext.GetTransaction() - but in the gas estimation case, there is no transaction.


src/cmd/isolated_server.cpp
```diff
@@ -105,6 +105,7 @@ int main(int argc, const char* argv[]) {
   bool nonisoload{false};
   string uuid;
 
+  LOG_SC = true;
```
n-hutton, line 108: always extra logs on during isolated server mode

src/libServer/EthRpcMethods.cpp
```diff
@@ -692,18 +694,20 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
     std::swap(data, code);
   }
 
-  uint64_t gas = GasConv::GasUnitsFromCoreToEth(2 * DS_MICROBLOCK_GAS_LIMIT);
+  uint64_t gas = GasConv::GasUnitsFromCoreToEth(DS_MICROBLOCK_GAS_LIMIT / 2);
```
n-hutton, line 697: some of the tests fail because they don't have enough funds to estimate this much gas - could/should this value be calculated as account max((account_balance - tx_value), DS_MICROBLOCK_GAS_LIMIT)?

src/libServer/EthRpcMethods.cpp
```diff
@@ -712,16 +716,23 @@ std::string EthRpcMethods::GetEthEstimateGas(const Json::Value& json) {
   uint64_t blockNum =
       m_sharedMediator.m_txBlockChain.GetLastBlock().GetHeader().GetBlockNum();
 
+  if (!code.empty() && !data.empty()) {
```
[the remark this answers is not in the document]
n-hutton: In the old code it never had to use evmContext.GetTransactionType

tests/EvmAcceptanceTests/test/BlockchainInstructions.ts
```diff
@@ -1,18 +1,18 @@
...
-  it("Should be deployed successfully", async function () {
-    expect(this.contract.address).to.be.properAddress;
-  });
-
-  it("Should return the owner address when getOrigin function is called", async function () {
-    const owner = this.contract.signer;
-    expect(await this.contract.getOrigin()).to.be.eq(owner.address);
-  });
-});
+//// TODO: Change the description to something more meaningful.
```
n-hutton, line 4: this only fails when running in parallel (during the 'before') - need @its-saeed to help look at it maybe

tests/EvmAcceptanceTests/test/GasEstimations.ts
```diff
@@ -35,7 +35,7 @@ describe("Gas estimation with web3.js", function () {
       this.contract = await parallelizer.deployContract("ParentContract", {value: amountPaid});
     });
 
-    it("Should return proper gas estimation [@transactional]", async function () {
+    xit("Should return proper gas estimation [@transactional]", async function () {
```
n-hutton, line 38: also fails only when run in parallel

PR #3367, opened 2023-01-24
[master] Update docs for local network.

rrw-zilliqa:
## Description

Trivial doc update.

## Backward Compatibility

- [x] This is not a breaking change
- [ ] This is a breaking change

PR #3374, opened 2023-01-27
[master] Simplify local development

rrw-zilliqa:
## Description

A script to build Zilliqa and Scilla together, and also to bodge python so that you don't need to download packages again every time you change the source. This is horribly unsafe for release, but works OK for local development and shortens the modify-compile-run cycle.

Note that you will want to merge https://github.com/Zilliqa/Zilliqa/pull/3367 before this, probably - it updates documentation and we rely on this here.

## Backward Compatibility
- [  ] This is a breaking change

## Review Suggestion

Look at docs/local-network.md, run the localdev script and verify that it performs as advertised. Look at docker/Dockerfile.dev and check that, whilst unsafe, it caches properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,11 @@ isolated-server.logs
 
 *.pyc
 .husky/
+
+# Localdev's working directory
+_localdev/
+
+# Emacs backups
+*~
+\#*
+.\#*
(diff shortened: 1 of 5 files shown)


rrw-zilliqa:
I have a shinier version of this - will resubmit.

PR #3393, opened 2023-02-06
ZIL-5111: remove unnecessary utilities

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


JamesHinshelwood (approved):
LGTM, but it might be safer if we just merged this into master and released in later?

yaron-zilliqa:
> LGTM, but it might be safer if we just merged this into master and released in later?

Don't think it's high risk at all but I've changed the base and will merge to master first.

PR #3396, opened 2023-02-06
[release/v8.6] Update software tag

chetan-zilliqa:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -4,6 +4,8 @@
 6
 === [Zilliqa]Fix Version ===
 0
+=== [Zilliqa]rc version ===
+0
 === [Zilliqa] Expected DS Epoch===
 0
 === [Scilla] Expected DS Epoch===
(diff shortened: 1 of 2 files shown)


VERSION
```diff
@@ -4,6 +4,8 @@
 6
 === [Zilliqa]Fix Version ===
 0
+=== [Zilliqa]rc version ===
+0
```
JamesHinshelwood, line 8: @yaron-zilliqa Can you take a look at this one? Do we need to update the stuff in `docker/Makefile` to respect this?

yaron-zilliqa: Don't recall these were ever respected. But basically it'd be good to, I suppose.

yaron-zilliqa: Anyway, I think for now it can be ignored or removed.

JamesHinshelwood: Looks like they are now: https://github.com/Zilliqa/Zilliqa/blob/4e74fea415feac83a926a0d87cee22cea6c57c1c/docker/Makefile#L9-L11
(I can't remember if I changed this or not)

PR #3413, opened 2023-02-08
[master] feat: add integration tests to the GHA pipeline

frankmeds:
## Backward Compatibility

### Implementation

- [X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


.dockerignore
```diff
@@ -12,7 +12,7 @@ persistence/
 scilla/
 scilla_files/
 config.xml
-isolated-server-accounts.json
+# isolated-server-accounts.json
```
yaron-zilliqa, line 15: Comment why this is commented out?

yaron-zilliqa (approved):

docker/Dockerfile
```diff
@@ -172,6 +173,13 @@ RUN strip /usr/local/bin/zilliqad \
    /usr/local/lib/libCommon.so \
    /usr/local/lib/libTrie.so
 
+EXPOSE 5555
```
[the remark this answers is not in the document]
JamesHinshelwood: This looks a bit sus to me too. Can we instead keep the `Dockerfile` the same and run the tests with something like `docker run ... --expose 5555, etc.` ?

It feels wrong that you should have to run the tests just to build an image.

.dockerignore
frankmeds: I needed to comment this line to be able to copy the isolated-server-accounts.json inside the Dockerfile to run the tests, otherwise it is ignored and not copied

PR #3416, opened 2023-02-09
[master] P2p and distributed tracing

art-gor:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [x] local machine test

- [x] small-scale cloud test


src/libMetrics/Metrics.h
```diff
@@ -22,13 +22,16 @@
 #include <list>
 #include <string>
 
+#ifndef HAVE_CPP_STDLIB
```
Steve-White-UK, line 25: Note for @Yaron:
This was required for the latest open telemetry, not having it caused some very nasty memory issues.

src/common/Constants.cpp
```diff
@@ -766,7 +766,7 @@ const std::string METRIC_ZILLIQA_SCHEMA{
                        "https://opentelemetry.io/schemas/1.2.0")};
 const std::string METRIC_ZILLIQA_SCHEMA_VERSION{ReadConstantString(
```
Steve-White-UK, line 767: On reflection we should probably move these schemas back into the code, the user has no business editing these, it is just opening us up to problems in the future.

Steve-White-UK (approved):
We need to document the p2p changes, and the mechanism for defining filter masks otherwise it will always be set to all or None and nobody will know it exists.

Most of the notes are notes to self whilst reviewing the code and shouldn't necessarily prevent merge , I can pick them up in a couple of days.

src/libMetrics/internal/mixins.h
```diff
@@ -124,84 +114,91 @@ class DoubleHistogram {
 
 class DoubleGauge {
  public:
-  DoubleGauge(const std::string &name, const std::string &description,
-              const std::string &units, bool obs)
-      : m_theGauge(Metrics::GetInstance().CreateDoubleGauge(
-            zil::metrics::GetFullName(METRIC_FAMILY, name), description,
-            units)) {}
+  DoubleGauge(zil::metrics::FilterClass fc, const std::string &name, const std::string &description, const std::string &units,
+              bool)
```
yaron-zilliqa, line 118: Is this bool neeeded?

yaron-zilliqa (changes requested):

yaron-zilliqa (approved):

src/libMetrics/Metrics.h
Steve-White-UK: OK then please do the official way after this is merged, life was very painful with this bug.

src/libMetrics/Tracing.h
```diff
@@ -18,93 +18,196 @@
...
+  T(LOOKUP_SERVER)              \
+  T(QUEUE)                      \
+  T(ACC_EVM)                    \
+  T(NODE)                       \
+  T(ACC_HISTOGRAM)
+
+namespace zil::trace {
+
+enum class FilterClass {
+#define ENUM_FILTER_CLASS(C) C,
```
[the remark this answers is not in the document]
Steve-White-UK: This is a clever way of defining an enum with each value representing the value of the bit, the programmer does not need to worry about what the hex value is he simply adds the next nemonic and it takes the next value 1,2,4,8,16,32,64 etc.

The clever bit to set the mask in the config you do not need to calculate the hex value of 10001001110 to represent which bits are turned on.

The Mask is set by   CPS,EVM_PROCESSING,SCILLA  a simple list to represent the bits.

The actually test at runtime is efficient it is simply a single AND of two 64 bit values to test of the bit is on, this is the permissions system.

Programmer friendly, user friendly and as about efficient filter as you can get.

src/libMetrics/internal/mixins.h
Steve-White-UK: There might be a better way, is simply delineates ambiguity between the other templates, it is behind the scenes but helps the compiler generate the correct code for the type as there are three possibilities it really signifies does this routine have a callback (Asynchronous or Synchronous) as it is never tested we removed the name so the compiler doesn't complain about the unused variable. @Artem could probably change this if it's offensive.

Steve-White-UK:
We have taken on board all of the comments in this PR and will address the majority in tomorrow's PR by EOD.

PR #3417, opened 2023-02-09
[master] ZIL-5112: Fix account delta sorting

JamesHinshelwood:


src/libMessage/Messenger.cpp
```diff
@@ -1645,7 +1645,7 @@ bool Messenger::SetAccountStoreDelta(zbytes& dst, const unsigned int offset,
   }
 
   std::sort(std::begin(accountsToSerialize), std::end(accountsToSerialize),
-            [&](const auto& l, const auto& r) { return l.first < l.first; });
+            [&](const auto& l, const auto& r) { return l.first < r.first; });
```
yaron-zilliqa, line 1648: I suggest removing the & from the capture. No need for this.

yaron-zilliqa (approved):

PR #3437, opened 2023-02-20
[master] Interop

bzawisto:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


evm-ds/src/precompiles/scilla_call.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::precompiles::scilla_common::get_contract_addr_and_name;
+use crate::precompiles::scilla_common::substitute_scilla_type_with_sol;
+use ethabi::decode;
+use ethabi::param_type::ParamType;
+use ethabi::token::Token;
+use hex::ToHex;
+use serde_json::{json, Value};
+
+const BASE_COST: u64 = 15;
+const PER_BYTE_COST: u64 = 3;
```
rrw-zilliqa, line 15: Leave a @todo here saying that we should move these to the config file. How did you arrive at them?

evm-ds/src/precompiles/scilla_read.rs
```diff
@@ -0,0 +1,252 @@
...
+                exit_status: ExitError::OutOfGas,
+            });
+        }
+    }
+
+    let (code_address, passed_field_name) = get_contract_addr_and_name(input)?;
+
+    let code = backend.code_as_json(code_address);
+    if code.is_empty() {
+        return Err(PrecompileFailure::Error {
```
rrw-zilliqa, line 46: Technically this is probably legal - all state will fail/default out, but ..

evm-ds/src/precompiles/scilla_call.rs
```diff
@@ -0,0 +1,132 @@
...
+                exit_status: ExitError::OutOfGas,
+            });
+        }
+    }
+
+    let (code_address, passed_transition_name) = get_contract_addr_and_name(input)?;
+
+    let code = backend.code_as_json(code_address);
+    if code.is_empty() {
+        return Err(PrecompileFailure::Error {
```
rrw-zilliqa, line 41: Why is this an error? (OK, it's an error if this isn't a contract, perhaps) - a call to an empty contract should presumably behave just as though you were calling an ordinary contract, but trying to invoke a transition which doesn't exist - which, it seems, just returns empty?

evm-ds/src/precompiles/scilla_call.rs
```diff
@@ -0,0 +1,132 @@
...
+            };
+            for arg_obj in args {
+                scilla_args.push((
+                    arg_obj["vname"].as_str().unwrap_or_default(),
+                    arg_obj["type"].as_str().unwrap_or_default(),
+                ));
+            }
+            break;
+        }
+    }
```
rrw-zilliqa, line 93: (contrariwise to the previous comment :-) ) - I think it should probably be an error if you call a non-existent transition, because otherwise there seems to be no way to distinguish between calling an empty transition with no arguments and no return value, which succeeds, and calling a transition which doesn't exist in the target contract?

Definitely worth some comments here to indicate what you think the semantics ought to be..

evm-ds/src/precompiles/scilla_read.rs
```diff
@@ -0,0 +1,252 @@
...
+            exit_status: ExitError::Other(Cow::Borrowed("Unable to parse scilla contract code")),
+        });
+    };
+
+    let fields = code["contract_info"]["fields"].to_owned();
+    let field_type = get_field_type_by_name(&passed_field_name, &fields)?;
+    let scilla_field = decode_indices(input, &field_type)?;
+
+    let substate_json =
+        backend.susbtate_as_json(code_address, &passed_field_name, &scilla_field.indices);
```
rrw-zilliqa, line 61: "substate_as_json"

evm-ds/src/precompiles/scilla_read.rs
```diff
@@ -0,0 +1,252 @@
...
+    if chunks.len() == 1 {
+        return Ok(ScillaField {
+            indices: vec![],
+            ret_type: chunks[0].to_string(),
+        });
+    }
+    // We support only maps and maps of maps
+    let mut idx = 0;
+    let mut field_types = vec![ParamType::Address, ParamType::String];
+    while idx < chunks.len() {
```
rrw-zilliqa, line 128: Worth a comment to explain what you're doing here. I don't  _think_ this has bad security implications, but I'm not entirely sure.

evm-ds/src/precompiles/scilla_read.rs
```diff
@@ -0,0 +1,252 @@
...
+        return Ok(Some(Token::Uint(
+            Uint::from_dec_str(value).unwrap_or_default(),
+        )));
+    } else if def.ret_type.starts_with("Int") {
+        if value.is_null() || !value.is_string() {
+            return Ok(Some(Token::Int(Uint::from(0))));
+        }
+        let value = value.as_str().unwrap();
+        return Ok(Some(Token::Int(
+            Uint::from_dec_str(value).unwrap_or_default(),
```
rrw-zilliqa, line 221: Does this not result in -ve numbers either being wrapped or ignored?

evm-ds/src/precompiles/scilla_read.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    encode_result_type(&value[def.indices.last().unwrap()], def)
+}
+
+fn encode_result_type(
+    value: &Value,
+    def: &ScillaField,
+) -> Result<Option<Token>, PrecompileFailure> {
+    if def.ret_type.starts_with("Uint") {
```
rrw-zilliqa, line 207: I'm not sure I'm entirely happy with the amount of information being destroyed here - though I guess there is no way to return a null value ("something went wrong decoding this and I don't know what the value is")? In particular, defining null to be 0 for ints is not a great idea if you can avoid it?

evm-ds/src/scillabackend.rs
```diff
@@ -300,6 +300,33 @@ impl Backend for ScillaBackend {
...
+            base64::encode(query.write_to_bytes().unwrap()).into(),
+        );
+
+        let Ok(result) = self.call_ipc_server_api("fetchCodeJson", args) else {
+            return Vec::new()
+        };
+        serde_json::to_vec(&result).unwrap_or_default()
+    }
+
+    fn susbtate_as_json(&self, address: H160, vname: &str, indices: &[String]) -> Vec<u8> {
```
rrw-zilliqa, line 319: "substate"

src/libCps/CpsRunEvm.cpp
```diff
@@ -616,7 +697,15 @@ void CpsRunEvm::ProvideFeedback(const CpsRun& previousRun,
       }
     }
   } else {
-    // TODO: allow scilla runner to provide feedback too
+    const auto& scillaResult = std::get<ScillaResult>(results.result);
```
rrw-zilliqa, line 700: Can't tell from reading the code, but does this not zap the logs from the scilla run, so that scilla events/messages don't appear in the transaction receipt?

src/libScilla/ScillaIPCServer.cpp
```diff
@@ -230,6 +241,63 @@ bool ScillaIPCServer::fetchExternalStateValue(const std::string &addr,
   return true;
 }
 
+void ScillaIPCServer::fetchStateJsonI(const Json::Value &request,
+                                      Json::Value &response) {
+  INC_CALLS(GetCallsCounter());
+  const auto address = Address{request["addr"].asString()};
+  const auto vname = request["vname"].asString();
+  if (!request["indices"].isArray()) {
+    return;
```
rrw-zilliqa, line 250: Really? Do you not want to return a Result<> here and actually provide an error?

tests/EvmAcceptanceTests/contracts/BasicInterop.sol
```diff
@@ -0,0 +1,109 @@
...
+pragma abicoder v2;
+
+contract BasicInterop {
+
+    function callSimpleMap(address contract_address, string memory tran_name, address recipient, uint128 amount) public {
+        bytes memory encodedArgs = abi.encode(contract_address, tran_name, recipient, amount);
+        uint256 argsLength = encodedArgs.length;
+        bool success;
+        assembly {
+            success := call(21000, 0x98, 0, add(encodedArgs, 0x20), argsLength, 0x20, 0)
```
rrw-zilliqa, line 14: Maybe some comments about what this is doing, so we can check that it is doing what you think it's doing?
(also because this test is the only docs, and it's a bit obscure how to use the interop from the test!)

rrw-zilliqa (approved):
Approve, on the whole, but worth addressing the issues in the comments ..

evm-ds/src/precompiles/scilla_read.rs
bzawisto: Well, it seems that underneath there's typedef Int = ethereum_types::Uint256 anyway.

src/libCps/CpsRunEvm.cpp
```diff
@@ -318,6 +324,82 @@ CpsExecuteResult CpsRunEvm::HandleCallTrap(const evm::EvmResult& result) {
...
+  const auto strJsonData = callData.call_data();
+  Json::Value jsonData;
+  if (!JSONUtils::GetInstance().convertStrtoJson(strJsonData, jsonData)) {
+    LOG_GENERAL(
+        WARNING,
+        "Error with convert precompile call_data string to json object");
+    return {TxnStatus::INCORRECT_TXN_TYPE, false, {}};
+  }
+
+  jsonData["_origin"] = "0x" + mCpsContext.origSender.hex();
```
n-hutton, line 341: this looks incorrect? _origin and _sender are the same

n-hutton (dismissed):
Can the 'description' of the PR be filled with a breakdown? It is quite a lot of code. It seems like the high level is something like this: adds the ability for EVM to call scilla contracts and read scilla data by adding new precompiles. The precompile gets trapped in the same way contract to contract calls work. There is no support for scilla to call EVM

evm-ds/src/precompiles/blake2.rs
```diff
@@ -228,14 +235,23 @@ mod tests {
         }
     }
 
+    fn new_backend() -> ScillaBackend {
+        let config = ScillaBackendConfig {
+            path: PathBuf::default(),
+            zil_scaling_factor: 1,
```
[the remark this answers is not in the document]
bzawisto: Nope, it's scaled up/down via GAS_SCALING_FACTOR in c++.

evm-ds/src/precompiles/mod.rs
```diff
@@ -52,5 +55,13 @@ pub fn get_precompiles() -> BTreeMap<H160, PrecompileFn> {
             H160::from_str("0000000000000000000000000000000000000009").unwrap(),
             blake2::blake2 as PrecompileFn,
         ),
+        (
+            H160::from_str("0000000000000000000000000000000000000098").unwrap(),
```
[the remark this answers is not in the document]
bzawisto: Nope, since a smart contract user needs to know it in advance - otherwise all contracts using these precompiles will stop working.

valeryz (approved):
A few comments, but nothing critical.  I'll keep reviewing the whole thing, but it is beyond just this PR, so this one LGTM.

evm-ds/src/cps_executor.rs
```diff
@@ -16,7 +17,13 @@ use primitive_types::{H160, H256, U256};
 use crate::scillabackend::ScillaBackend;
 type PrecompileMap = BTreeMap<
     H160,
-    fn(&[u8], Option<u64>, &Context, bool) -> Result<(PrecompileOutput, u64), PrecompileFailure>,
+    fn(
+        &[u8],
+        Option<u64>,
+        &Context,
+        &dyn Backend,
```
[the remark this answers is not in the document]
bzawisto: Yep, could be doable this way. I'll put a comment to rework it one day.

evm-ds/src/precompiles/scilla_call.rs
```diff
@@ -0,0 +1,134 @@
...
+use ethabi::param_type::ParamType;
+use ethabi::token::Token;
+use hex::ToHex;
+use serde_json::{json, Value};
+
+// TODO: revisit these consts
+const BASE_COST: u64 = 15;
+const PER_BYTE_COST: u64 = 3;
+
+pub(crate) fn scilla_call(
```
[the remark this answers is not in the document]
bzawisto: Will do.

evm-ds/src/precompiles/scilla_call.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    if let Some(gas_limit) = gas_limit {
+        if gas_limit < gas_needed {
+            return Err(PrecompileFailure::Error {
+                exit_status: ExitError::OutOfGas,
+            });
+        }
+    }
+
+    let (code_address, passed_transition_name) = get_contract_addr_and_name(input)?;
```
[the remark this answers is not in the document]
bzawisto: There was not abi-encoding production ready code that could be easy to pick up - also - it would've meant even more code in c++.

evm-ds/src/precompiles/mod.rs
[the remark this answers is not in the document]
JamesHinshelwood: Maybe we should put these at an address less likely to collide with Eth/other EVM precompiles?
e.g. have the first 3 bytes of the address be ZIL in ASCII?

PR #3438, opened 2023-02-21
[master] Add g3log sink in validateDb

chetan-zilliqa:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


JamesHinshelwood (approved):
Do we need this in any of the other CLI tools as well?

chetan-zilliqa:
> Do we need this in any of the other CLI tools as well?

I think we added to most of the tools that we required. I will confirm again on it.

PR #3443, opened 2023-02-21
ZIL-5115: Support address array in `eth_subscribe`

JamesHinshelwood:


src/libEth/filters/FiltersUtils.cpp
```diff
@@ -377,9 +395,13 @@ bool InitializeEventFilter(const Json::Value &params, EventFilterParams &filter,
 
 bool Match(const EventFilterParams &filter, const Address &address,
            const std::vector<Quantity> &topics) {
-  if (!filter.address.empty() &&
-      boost::to_lower_copy(address) != filter.address) {
-    return false;
+  if (!filter.address.empty()) {
```
valeryz, line 398: Do you need a condition here?  the `std::find_if` below will do the job on the empty filter, won't it?

valeryz (dismissed):
Thanks, left a cosmetic comment.

src/libEth/filters/FiltersUtils.cpp
JamesHinshelwood: If `filter.address.empty()` then `(std::find_if(v.begin(), v.end(), [&address](const auto &a) { return boost::iequals(address, a); }) == v.end()) == true` and so we'd return `false`, indicating this doesn't match which isn't what we want.

This is because we're using the empty list of filters as something different (match everything) than what it really should mean (match nothing). Maybe we could represent it more neatly to avoid this?

PR #3448, opened 2023-02-22
[master] Tracing v2

n-hutton:
Add full tracing as per @valeryz request - this is every opcode executed etc. This is massive, and so can cause RPC requests to time out on localhost. Need to be careful when enabling TX_TRACE.

Additionally, the PR puts a scheme in place to only keep TX traces for 30 blocks' worth - after this point, for every trace added, one is removed.


evm-ds/Cargo.toml
```diff
@@ -13,7 +13,7 @@ bytes = "1.1.0"
 clap = { version = "3.1.6", features = ["derive"] }
 log4rs = { version = "1.1.1", features = ["all_components", "gzip"] }
 ethereum = "0.12.0"
-evm = { git = "https://github.com/Zilliqa/evm", branch = "make-machine-constructible-from-state-3", version = "0.37.0", features = ["tracing"] }
+evm = { git = "https://github.com/Zilliqa/evm", branch = "tracing", version = "0.37.0", features = ["tracing"] }
```
n-hutton, line 16: I had to make changes of the evm fork (minor) for this PR

evm-ds/src/convert.rs
```diff
@@ -182,156 +181,6 @@ impl From<(Bytes, Bytes)> for EvmProto::Storage {
     }
 }
 
-fn as_string(opcode: &Opcode) -> String {
```
n-hutton, line 185: now we are forking we can use std::display

evm-ds/src/main.rs
```diff
@@ -49,10 +48,6 @@ struct Args {
     #[clap(short = 'p', long, default_value = "3333")]
     http_port: u16,
 
-    /// Trace the execution with debug logging.
-    #[clap(short, long)]
-    tracing: bool,
```
n-hutton, line 54: tracing is passed via evm args

evm-ds/src/main.rs
```diff
@@ -98,44 +93,123 @@ impl CallContext {
...
+    pub gas: u64, // not populated
+    #[serde(rename = "gasCost")]
+    pub gas_cost: u64, // not populated
+    pub op: String,
+    pub pc: usize,
+    pub stack: Vec<String>,
+    pub storage: Vec<String>, // not populated
+}
+
+impl StructLog {
```
valeryz, line 109: Maybe you could use Default derivation on the struct?, and then use `StructLog::default()` instead of `new()`

evm-ds/src/main.rs
```diff
@@ -98,44 +93,123 @@ impl CallContext {
...
-    fn event(&mut self, event: tracing::Event) {
-        println!("recvd event: {:?}", event);
+impl evm::runtime::tracing::EventListener for LoggingEventListener {
+    fn event(&mut self, event: evm::runtime::tracing::Event) {
+
+        if !self.enabled {
+            return;
+        }
+
+        let mut struct_log = StructLog::new();
```
valeryz, line 161:
```suggestion
        let mut struct_log = StructLog { depth: self.call_tracer.len() - 1, ..Default::default() };
```

evm-ds/src/main.rs
```diff
@@ -98,44 +93,123 @@ impl CallContext {
...
+
+        if self.raw_tracer.struct_logs.len() < 5 {
+            self.raw_tracer.struct_logs.push(struct_log);
+        }
     }
 }
 
 impl LoggingEventListener {
 
+    #[allow(dead_code)]
```
valeryz, line 198: If it is dead code, do we need it?

valeryz (approved):
A few Rust comments, please address before merging.

evm-ds/src/main.rs
n-hutton: I like when debugging to switch to pretty strings which is otherwise unused

evm-ds/src/main.rs
n-hutton: Great, I didn't know you could do such patterns :)

evm-ds/src/main.rs
n-hutton: Yup, sounds good!

evm-ds/Cargo.toml
```diff
@@ -13,7 +13,7 @@ bytes = "1.1.0"
 clap = { version = "3.1.6", features = ["derive"] }
 log4rs = { version = "1.1.1", features = ["all_components", "gzip"] }
 ethereum = "0.12.0"
-evm = { git = "https://github.com/Zilliqa/evm", branch = "make-machine-constructible-from-state-3", version = "0.37.0", features = ["tracing"] }
+evm = { git = "https://github.com/Zilliqa/evm", rev = "6876b5b", features = ["tracing"] }
```
n-hutton, line 16: point to master commit

PR #3465, opened 2023-02-27
[master] Add debug trace call

n-hutton:
Should work with something like: 

```
curl http://localhost:5555/ -X POST -H "Content-Type: application/json" --data '{"method":"debug_traceCall","params":[{"from":"0xf0cb24ac66ba7375bf9b9c4fa91e208d9eaabd2e","to":"0xa8cae66f62648529eb6ac2f026893fc436107510","data":"0x12CBB1AD0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000030000000000000000000000009FB20310353B553E99904B2E71CDDADAD76A361200000000000000000000000084D15E9E4B1A134EDAB531DAE33908018FD84A9A000000000000000000000000A8CAE66F62648529EB6AC2F026893FC436107510"}, "latest", {"tracer" : "callTracer"}],"id":1,"jsonrpc":"2.0"}'
```

### DIFF
--- a/src/libServer/EthRpcMethods.cpp
+++ b/src/libServer/EthRpcMethods.cpp
@@ -24,7 +24,6 @@
 #include "JSONConversion.h"
 #include "LookupServer.h"
 #include "common/Constants.h"
-#include "common/Serializable.h"
 #include "json/value.h"
 #include "libCrypto/EthCrypto.h"
 #include "libData/AccountData/Account.h"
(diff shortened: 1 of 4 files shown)


src/libServer/EthRpcMethods.cpp
```diff
@@ -618,6 +623,54 @@ string EthRpcMethods::GetEthCallEth(const Json::Value &_json,
   return this->GetEthCallImpl(_json, {"from", "to", "value", "gas", "data"});
 }
 
+// Convenience fn to extract the tracer - valid types are 'raw' and 'callTracer'
+// This is as the tracer is a JSON which has both types as entries
+string extractTracer(const std::string &tracer, const std::string trace) {
```
valeryz, line 628:
```suggestion
string extractTracer(const std::string &tracer, const std::string &trace) {
```

valeryz (approved):

PR #3466, opened 2023-02-28
[master] Max depth fixes

bzawisto:
## Description
Edge was and is used to determine the depth. We only increase the edge when new level of runs is scheduled.
This is for CPS only.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


JamesHinshelwood (approved):
LGTM

PR #3467, opened 2023-02-28
[master] Cleanup and testing of metrics.

Steve-White-UK:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -168,6 +168,8 @@ int main(int argc, const char* argv[]) {
                                                 __FUNCTION__);
     Metrics::GetInstance().Init();
 
+    TRACE_ERROR("This is a test");
+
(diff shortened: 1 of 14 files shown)


src/cmd/isolated_server.cpp
```diff
@@ -168,6 +168,8 @@ int main(int argc, const char* argv[]) {
                                                 __FUNCTION__);
     Metrics::GetInstance().Init();
 
+    TRACE_ERROR("This is a test");
```
Steve-White-UK, line 171: This should be removed now

Steve-White-UK, line 171: This should be removed

Steve-White-UK (commented):
remove the two test statements and good to go.

PR #3483, opened 2023-03-02
[master] hotfix for logger and trace ids (multithreading related)

art-gor:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libUtils/Logger.cpp
+++ b/src/libUtils/Logger.cpp
@@ -99,8 +99,22 @@ std::ostream& logLevel(std::ostream& stream, const LogMessage& message) {
   return stream;
 }
 
+namespace {
+
+constexpr char TRACE_INFO_MARKER = 0x02;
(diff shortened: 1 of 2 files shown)


src/libUtils/Logger.cpp
```diff
@@ -377,3 +396,18 @@ Logger::ScopeMarker::~ScopeMarker() {
         << " END";
   }
 }
+
+std::string_view LogTracesImpl() {
+  static thread_local char BUFFER[TRACE_INFO_BUFSIZE];
```
bzawisto, line 401: nitpick, std::array<> ?

bzawisto (approved):

src/libUtils/Logger.cpp
Steve-White-UK: I need to take this code immediately, we will flag this and fix it in a new PR.

PR #3489, opened 2023-03-02
Graf only

n-hutton:


Steve-White-UK (approved):
The ugliness of the callback is noted and we are about to release an Api update that internalises the callback and releases a shared object back to you that you manipulate to your hearts content and the internal callback will groom the value for you. We will convert your code for you when it's complete.

Steve-White-UK:
RE: ci integration test failure : Looks like the ephemeral storage for the pod needs enlarging, I used to know how to do this on jenkins but not sure now that we are on GHA, Francesco would be the man to talk to on this topic.

mauromedda (approved):

PR #3494, opened 2023-03-06
[master] Final settings

Steve-White-UK:
## Description
AMend build script to use all but 90% of cpus on machines with more than 10 cpus.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -250,7 +250,7 @@ echo "Build directory: ${build_dir}"
 echo "Install directory: ${install_dir}"
 
 if [ "$n_parallel"  -gt "10" ]; then
-    jobs=$(( n_parallel * 10 / 9 ))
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -249,8 +249,16 @@ echo "Currenct directory: $(pwd)"
 echo "Build directory: ${build_dir}"
 echo "Install directory: ${install_dir}"
 
+if [ "$n_parallel"  -gt "10" ]; then
+    jobs=$(( n_parallel * 10 / 9 ))
```
yaron-zilliqa, line 253: Better write ( n_parallel * 10 ) to make sure it's evaluated first

yaron-zilliqa (approved):

PR #3502, opened 2023-03-10
[master] EM-49: Protect m_txnProcessingFinished with a mutex consistently to prevent race condition

yaron-zilliqa:
## Description

m_txnProcessingFinished wasn't always protected by the mutex (m_mutexCVTxnProcFinished) used by the condition variable (cv_TxnProcFinished) to check whether transaction processing had finished. In addition, waiting wasn't done properly and can be done more correctly in C++20.
This fix solves the situation where WaitUntilTxnProcessingDone() was called after the condition variable had been signaled making it wait until the next epoch, ultimately causing a crash.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


valeryz (approved):

Steve-White-UK (approved):
I have been running this build for a few minutes on the local cluster and one day's changes make it a completely different system, it is very noticeable.

I am very happy to see this closed, nice job @yaron-zilliqa

PR #3505, opened 2023-03-14
[master] ZIL-5131: remove unnecessary logging in serialization code.

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


chetan-zilliqa (commented):
Sometimes these logs are helpful to investigate time delays in serialize/deserialize operations.

yaron-zilliqa:
> Sometimes these logs are helpful to investigate time delays in serialize/deserialize operations.

Not really. This is just serialization and logs should (and would) be written when something goes wrong at the call site or even during serialization. Marking every method by BEG & END is a total waste. Especially for something like this.

Steve-White-UK (approved):

PR #3514, opened 2023-03-23
[master] Fixing condition variables - Lookup.cpp

valeryz:
## Description

The goal is to fix the ubiquitious error in using condition variables, where the condition is not being checked before waiting itself, which results into the recepient never aware that the message has arrived, in case it had arrived quickly.

This will be a sequence of PRs, so that the blast radius of each is limited.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libLookup/Lookup.cpp
+++ b/src/libLookup/Lookup.cpp
@@ -224,12 +224,15 @@ void Lookup::GetInitialBlocksAndShardingStructure() {
   }
   if (!m_rejoinInProgress) {
     // Ask for the sharding structure from lookup
+    {
+      std::unique_lock<std::mutex> cv_lk(m_mutexShardStruct);
+      m_shardStructSignal = false;
+    }
     ComposeAndSendGetShardingStructureFromSeed();
     std::unique_lock<std::mutex> cv_lk(m_mutexShardStruct);
-    // TODO: cv fix
-    if (cv_shardStruct.wait_for(
-            cv_lk, std::chrono::seconds(GETSHARD_TIMEOUT_IN_SECONDS)) ==
-        std::cv_status::timeout) {
+    if (!cv_shardStruct.wait_for(
+            cv_lk, std::chrono::seconds(GETSHARD_TIMEOUT_IN_SECONDS),
+            [this]() { return m_shardStructSignal; })) {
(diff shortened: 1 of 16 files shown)


src/libLookup/Lookup.cpp
```diff
@@ -204,26 +204,35 @@ void Lookup::GetInitialBlocksAndShardingStructure() {
     }
     LOG_GENERAL(INFO,
                 "TxBlockNum " << txBlockNum << " DSBlockNum: " << dsBlockNum);
+    {
+      std::unique_lock<std::mutex> cv_lk(m_mutexCvSetRejoinRecovery);
+      m_rejoinRecoverySignal = false;
```
yaron-zilliqa, line 209: Why set this here in a separate block and not right after line 211/215 below?

src/libLookup/Lookup.cpp
```diff
@@ -204,26 +204,35 @@ void Lookup::GetInitialBlocksAndShardingStructure() {
...
       if (m_rejoinInProgress) {
         break;
       }
     }
   }
   if (!m_rejoinInProgress) {
     // Ask for the sharding structure from lookup
+    {
+      std::unique_lock<std::mutex> cv_lk(m_mutexShardStruct);
+      m_shardStructSignal = false;
```
yaron-zilliqa, line 229: Same as above: why not just set after line 223/232 below?

yaron-zilliqa (changes requested):

src/libLookup/Lookup.cpp
valeryz: Because `ComposeAndSendGetDirectoryBlocksFromSeed` call below might result in the response being sent straight away, and we better have the mutex unlocked.  But even if not, this is a function that may take some time, and we shouldn't be held the mutex aquired during its execution.

src/libLookup/Lookup.cpp
valeryz: Same as above: we need to keep the mutex unlocked when the network communication request is initiated, because the response may come sooner than we expect (when this thread is interrupted for example, and only scheduled later).

Steve-White-UK (approved):

PR #3519, opened 2023-03-27
[master] Metrics and debugging

n-hutton:
Metrics and debugging from the recent debugging session


infra/k8s/base/configmap/init.py
```diff
@@ -639,6 +639,8 @@ def cmd_zilliqa_daemon(args, recovery=False, resume=False):
 
     start_sh = [
         '#!/bin/bash',
+        'ulimit -c unlimited',
+        'sysctl -w kernel.core_pattern=core.%p.%s.%c.%d.%P',
```
n-hutton, line 643: should this be here? The default location for core dumps seemed to be apport, the crash reporting manager. Only problem is, it doesn't exist in the container so dumps get lost.

JamesHinshelwood (approved):

PR #3521, opened 2023-03-28
[master] ZIL-5143: add tracing for epoch, consesnsus & blockstorage.

yaron-zilliqa:
## Backward Compatibility

- [] This is a breaking change

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libConsensus/ConsensusBackup.cpp
+++ b/src/libConsensus/ConsensusBackup.cpp
@@ -21,9 +21,13 @@
 #include "common/Constants.h"
 #include "common/Messages.h"
 #include "libMessage/Messenger.h"
+#include "libMetrics/Api.h"
+#include "libMetrics/TracedIds.h"
 #include "libNetwork/P2PComm.h"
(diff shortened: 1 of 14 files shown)


rrw-zilliqa (approved):
Looks good to me - you might want enums for the span names, and worth checking that this doesn't produce too much overhead when run.

Also, you'll want to decide whether to submit an equivalent PR to release/v8.9 ..

yaron-zilliqa:
> Looks good to me - you might want enums for the span names, and worth checking that this doesn't produce too much overhead when run.
> 
> Also, you'll want to decide whether to submit an equivalent PR to release/v8.9 ..

We're using enums for the filters and yeah - we might want to look into using enums a bit more here.
I've tested this locally and I think it's quite convenient since it shows things in terms of "Epochs" - easier to trace in which epoch something happened.

I will indeed get this into release/v8.9.. very much want it to in the release.

PR #3526, opened 2023-03-29
[master] Networking fix

n-hutton:
During testing of a rehearsal network, it would fail to generate microblocks for the shard of normals.

This was during a period of the consensus where normals would respond to the leader with a direct message (as opposed to the rumor system which is used to propagate the first step in consensus which is sharing the proposed microblock).

High rates of network timeout/failure were observed, and the peer queue would be ~200. The issue was determined to be that the backlog parameter in `evconnlistener_new_bind` defaults to 128 (the backlog of allowed pending connections), which is less than the 519 connections the leader would have to recieve at this point. This made it such that it could not be replicated on smaller networks as they wouldn't get to this network limit.


.github/workflows/ci-image-dev.yml
```diff
@@ -10,9 +10,6 @@ on:
   push:
     branches-ignore:
       - 'release/**'
-  pull_request:
```
JamesHinshelwood, line 13: Do we still want these changes? cc @frankmeds

JamesHinshelwood (approved):

PR #3530, opened 2023-03-29
[master] (fix) ZIL-5144: Don't send Scilla contract creation to ARCHIVAL_SEND_…

rrw-zilliqa:
…SHARD unless you are actually a lookup or archival node.

## Description

Fix ZIL-5144.

## Backward Compatibility

## Review Suggestion

Check that the logic is correct - please be careful; this is my first PR which touches actual logic and I'm far from sure I have it right - the only evidence I have is traces and the fact that my trivial deploy-a-Scilla-contract test works with this patch, and fails without it.

### DIFF
--- a/src/libServer/LookupServer.cpp
+++ b/src/libServer/LookupServer.cpp
@@ -2391,7 +2391,9 @@ std::pair<std::string, unsigned int> LookupServer::CheckContractTxnShards(
   if (!tx.IsEth() && scType == Transaction::CONTRACT_CREATION) {
     // Scilla smart CONTRACT_CREATION call should be executed in shard rather
     // than DS.
-    mapIndex = SEND_TYPE::ARCHIVAL_SEND_SHARD;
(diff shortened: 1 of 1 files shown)


valeryz:
The purpose of sending to shards rather than DS is to scale them.  contract creation, esp in Scilla should be so much more rare than contract calls that an advantage to scaling it through sharding almost doesn't exist.

At some point, we had unified this among EVM and Scilla, and had sent all creation to DS.  In the current codebase you are fixing, it seems it is no longer the case (don't know when it reverted back).

Anyhow, the change LGTM, but FWIW sending them to shards is not necessary at all.

valeryz (approved):

rrw-zilliqa:
Merged 3530 on the basis that we can then decide to merge 3532 over it - before I OK that one, I want to see it pass some tests. Before that, I want to be able to run some tests! Sorry @valeryz ...

PR #3533, opened 2023-03-30
[master] Add replay transaction script

JamesHinshelwood:
Allows you to replay transactions from one network on another network, assuming the validity properties are the same (e.g. the same network ID and state).

For example,
```
python replay-txn.py 0x2afa7221cbf2784b5ea631ac14bdb82968abbe1b592587bf9fd4e28611a8757c 'https://api.zilliqa.com' 'https://v880-rehearsal3-l2api.mainnet.aws.zilliqa.com'
```


rrw-zilliqa (approved):
Looks good to me :-)

PR #3535, opened 2023-03-31
[master] hexify get TX

n-hutton:


src/libServer/JSONConversion.cpp
```diff
@@ -632,13 +632,25 @@ const Json::Value JSONConversion::convertTxtoJson(
   _json["gasPrice"] = twr.GetTransaction().GetGasPriceQa().str();
   _json["gasLimit"] = to_string(twr.GetTransaction().GetGasLimitZil());
 
-  if (!twr.GetTransaction().GetCode().empty()) {
-    _json["code"] =
-        DataConversion::CharArrayToString(twr.GetTransaction().GetCode());
+  auto const code = twr.GetTransaction().GetCode();
+  auto const data = twr.GetTransaction().GetData();
+
+  // If we detect non ascii characters (i.e evm bytecode) we convert to hex
```
rrw-zilliqa, line 638: Why do we do this? It's worth providing an explanation in the comment or someone will have a bad experience, and, knowing no better, revert this patch ..

n-hutton: This will happen when the legacy API (GetTransaction) hits a eth style TX. It is not really a use case that should happen so often, but the old block explorers do this. In effect it is just cosmetic for a corner case.

PR #3536, opened 2023-03-31
[master] ZIL-5168: Misc fixes.

rrw-zilliqa:
## Description

Add localdev to hardhat.config.ts, and let localdev cope with non-utf-8 logs.

## Backward Compatibility

## Review Suggestion

Reading the code should do, but feel free to try it if you really want!


rrw-zilliqa:
Included in another PR - abandoning this one.

PR #3538, opened 2023-04-02
[master] ZIL-5165: eoa accounts ignore messages

rrw-zilliqa:
## Description

Fix ZIL-5165; unfortunately this PR brings quite a lot of mechanism along with it in the form of Saeed's tests for this situation and my additions to them that allow you to run them, so as to check that the fix is correct.

They also bring in a couple of minor localdev fixes that I thought were small enough not to warrant chopping them out of this PR and testing separately - I can do that, but it's a few hours' work and since we don't really bisect anyway I figured perhaps not worth delaying for.

## Backward Compatibility

## Review Suggestion

```sh
npx hardhat test test/scilla/FungibleToken.ts --network localdev
```

and check that it works. For localdev, try `snapshot-k8s` and see that it works now, and try reupping and observe that it now works (in that you don't need to restart the ingress manually).

Inspect the code and comments and check that you agree with my logic ..


rrw-zilliqa:
Review if you like, but @bzawisto has a different (probably better) fix for this; leaving this here so that we have an open PR that fixes this issue in the system, but I'll rework once Saeed has his test framework pushed and I can then rebase on top of that.

tests/EvmAcceptanceTests/test/scilla/FungibleToken.ts
```diff
@@ -0,0 +1,86 @@
...
+      console.log("Error: ", error)
+    }
+
+    expect(await getBalance(address)).to.be.eq(10)
+  });
+
+  it("Should not be possible to transfer to a contract which does not accept", async function() {
+    try {
+      let result = await contract.Transfer(null_contract.address, 10);
+      // console.log(`result ${JSON.stringify(result)}`);
```
its-saeed, line 77: Can we remove this commented code?

tests/EvmAcceptanceTests/test/scilla/FungibleToken.ts
```diff
@@ -0,0 +1,86 @@
...
+      expect(result.receipt.success).to.be.false;
+      expect(result.receipt.errors["1"].length).to.equal(1);
+      // 7 == CALL_CONTRACT_FAILED
```
its-saeed, line 80: We can define it as a `const` instead of having a comment.

tests/EvmAcceptanceTests/scripts/transfer.ts
its-saeed, line 1: It's not a clean piece of code to commit :) I just wrote it to test something.

its-saeed (approved):

PR #3548, opened 2023-04-05
[master] ZIL-5146: only allow lookup & seedprv node types to use telemetry 

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libMetrics/Common.h
+++ b/src/libMetrics/Common.h
@@ -18,6 +18,8 @@
 #ifndef ZILLIQA_SRC_LIBMETRICS_COMMON_H_
 #define ZILLIQA_SRC_LIBMETRICS_COMMON_H_
 
+#include <algorithm>
+#include <array>
 #include <string>
 
 namespace zil {
@@ -31,5 +33,16 @@ const std::string METRIC_SCHEMA{"https://opentelemetry.io/schemas/1.2.0"};
 }  // namespace zil
 
 constexpr double METRICS_VERSION{8.6};
+
+template <typename StringT>
+bool IsObservabilityAllowed(const StringT& identity) {
+  static std::array<std::string_view, 4> allowedIdentities = {
+      "seedprv", "seedpub", "dsguard", "lookup"};
+  return std::any_of(std::begin(allowedIdentities), std::end(allowedIdentities),
+                     [&identity](const auto& allowedIdentity) {
+                       return identity.find(allowedIdentity) !=
+                              std::string::npos;
+                     });
+}
 
 #endif  // ZILLIQA_SRC_LIBMETRICS_COMMON_H_
--- a/src/libMetrics/Logging.cpp
+++ b/src/libMetrics/Logging.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "Logging.h"
+#include "Common.h"
 #include "Tracing.h"
 
 #include "libUtils/Logger.h"
@@ -29,6 +30,8 @@
 #include "opentelemetry/sdk/common/global_log_handler.h"
 #include "opentelemetry/sdk/logs/logger_provider_factory.h"
 #include "opentelemetry/sdk/logs/simple_log_record_processor_factory.h"
+
+#include <boost/algorithm/string/case_conv.hpp>
 
 namespace otlp_exporter = opentelemetry::exporter::otlp;
 namespace logs_exporter = opentelemetry::exporter::logs;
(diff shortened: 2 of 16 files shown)


src/libMetrics/Metrics.cpp
```diff
@@ -39,31 +39,41 @@
...
+  // c-tor
+  zil::metrics::Filter::GetInstance().init();
+}
 
-void Metrics::Init() {
-  std::string cmp(METRIC_ZILLIQA_PROVIDER);
+void Metrics::Initialize(std::string_view identity /*= {}*/,
+                         std::string provider /* = METRIC_ZILLIQA_PROVIDER*/) {
+  static std::array<std::string_view, 2> allowedIdentities = {"seedprv",
+                                                              "lookup"};
```
JamesHinshelwood, line 57: Hmm, do we want metrics, tracing, etc. for normals, dsguards, seedpubs as well? I would assume so?

yaron-zilliqa: We do, but the identity I believe should be also lookup. Unless the init.py in testnet script is incorrect:

`parser.add_argument('--type', choices=['normal', 'lookup', 'new', 'seedprv'], help='node type (normal, lookup, new, seedprv)')
`

yaron-zilliqa: I think you're right and these aren't actually the only options. I'll add the others. Wouldn't hurt anyway.

JamesHinshelwood: I did a quick grep of `testnet`. I believe `--type` is only set for the join pod. For the others, it is inferred from the pod name by splitting on `-`. So if I understand correctly, the other nodes will have type "dsguard", "normal", etc.

JamesHinshelwood: I guess the important question is what is the `--type` of external miner nodes? Is it set in `launch_docker.sh`?

src/libMetrics/Logging.cpp
```diff
@@ -142,25 +142,36 @@ std::shared_ptr<logs::LoggerProvider> InitNoop() {
...
-Logging::Logging() {
-  std::string cmp(LOGGING_ZILLIQA_PROVIDER);
+void Logging::Initialize(std::string_view identity /*= {}*/,
+                         std::string provider /*= LOGGING_ZILLIQA_PROVIDER*/) {
+  static std::array<std::string_view, 4> allowedIdentities = {
+      "seedprv", "seedpub", "dsguard", "lookup"};
+  if (std::none_of(std::begin(allowedIdentities), std::end(allowedIdentities),
+                   [identity](const auto& allowedIdentity) {
+                     return identity.find(allowedIdentity) != std::string::npos;
+                   })) {
```
rrw-zilliqa, line 152: Works, but a bit convoluted? Might have been better to write an aux function (or at least a comment):

```
if (!std::any_of(std::begin(allowedIdentities), std::end(allowedIdentities), 
   /* Is this allowed? */
    [identity](const auto& allowedIdentity) { return identity.find(allowedIdentity) != std::string::npos } })) { 
     .. 
   }
```
?

src/libMetrics/Logging.cpp
```diff
@@ -142,25 +142,36 @@ std::shared_ptr<logs::LoggerProvider> InitNoop() {
...
+    provider = "NONE";
+  }
+
+  transform(provider.begin(), provider.end(), provider.begin(), ::tolower);
```
rrw-zilliqa, line 156: Worth a utility function? (or using the boost alg?) - this is really nitpicky!

src/libMetrics/Metrics.cpp
```diff
@@ -39,31 +39,41 @@
...
+  // c-tor
+  zil::metrics::Filter::GetInstance().init();
+}
 
-void Metrics::Init() {
-  std::string cmp(METRIC_ZILLIQA_PROVIDER);
+void Metrics::Initialize(std::string_view identity /*= {}*/,
+                         std::string provider /* = METRIC_ZILLIQA_PROVIDER*/) {
+  static std::array<std::string_view, 4> allowedIdentities = {
+      "seedprv", "seedpub", "dsguard", "lookup"};
```
rrw-zilliqa, line 57: Use a singleton rather than duplicating the list?

rrw-zilliqa (approved):
Looks good to me :-)

src/libMetrics/Logging.cpp
yaron-zilliqa: As a matter of fact that's how I initially wrote it and thought that was a bit less clear :D .... at any rate - I've moved this to an aux function and reused in Metrics.cpp & Tracing.cpp as well.

src/libMetrics/Logging.cpp
yaron-zilliqa: Yeah, changed to use the boost::algorithm::to_lower.

src/libMetrics/Metrics.cpp
yaron-zilliqa: Moved all this to a separate aux function so it's not repeated. (regardless, I hate singletons and don't really use them).

src/libMetrics/Common.h
```diff
@@ -32,4 +34,15 @@ const std::string METRIC_SCHEMA{"https://opentelemetry.io/schemas/1.2.0"};
 
 constexpr double METRICS_VERSION{8.6};
 
+template <typename StringT>
+bool IsObservabilityAllowed(const StringT& identity) {
+  static std::array<std::string_view, 4> allowedIdentities = {
+      "seedprv", "seedpub", "dsguard", "lookup"};
```
JamesHinshelwood, line 40: Should "normal" be here too?

PR #3551, opened 2023-04-05
[master] MVP-5165: Funds-bearing messages to EOAs should cause a transfer to occur

rrw-zilliqa:
## Description

(fix) MVP-5165: Transfers to EOAs do not count towards the max edge limit. (feat) Update package-lock
(feat) Re-enable SendZil tests (which now pass)
(feat) MVP-5165: Test funds transfers to EOAs.

## Backward Compatibility

## Review Suggestion

Check that tests test messages to EOAs correctly.
Look over code and check that it looks reasonable.
Run code and check that tests pass with this patch, but do not without it.


JamesHinshelwood (commented):
LGTM, but was `package-lock.json` meant to have a 16000 line diff? If yes, can we do it in a separate PR?

JamesHinshelwood (approved):
I removed the `package-lock.json` changes, hope you don't mind!

rrw-zilliqa:
(PS in general happy to commit tests in a separate PR if you want to do that - just feel bad separating the code from the tests that test it because it temporarily leaves master in a state of containing code, but no tests for it and thus being subject to breakage .. )

PR #3557, opened 2023-04-10
[release/v8.9] Zil 5176 : Allow node to sync if not in shard

chetan-zilliqa:
## Description

 If a node is not a part of the shard, we should allow it to sync after reading complete persistence. So at a point where the node determines that it is not part of the shard, we called the `StartSynchronization` function. The `StartSynchronization` function allows the node to perform normal sync with the network i.e fetch block one by one. I have to introduce `m_allowRecoveryAllSync` to make sure that the code does not change the node's state anytime. When the POW window is reached, the node attempts to perform POW and if it joins the network successfully then it gets ProcessVCDSBlocksMessage  where I reset the variable m_allowRecoveryAllSync

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test

### DIFF
--- a/src/libNode/Node.cpp
+++ b/src/libNode/Node.cpp
@@ -1112,23 +1112,15 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
         m_mediator.m_ds->m_shards, m_mediator.m_ds->m_publicKeyToshardIdMap,
         m_mediator.m_ds->m_mapNodeReputation);
   }
+  bool rejoinCondition = REJOIN_NODE_NOT_IN_NETWORK && !LOOKUP_NODE_MODE && !bDS;
+  LOG_GENERAL(INFO, "rejoinCondition = "<< rejoinCondition); //TODO: remove this log before merge
 
-  if (REJOIN_NODE_NOT_IN_NETWORK && !LOOKUP_NODE_MODE && !bDS) {
-    if (!bInShardStructure) {
-      LOG_GENERAL(
-          WARNING,
-          "Node " << m_mediator.m_selfKey.second
-                  << " is not in network, apply re-join process instead");
+  if (rejoinCondition && bIpChanged) {
+    LOG_GENERAL(
+        INFO, "My IP has been changed. So will broadcast my new IP to network");
+    if (!UpdateShardNodeIdentity()) {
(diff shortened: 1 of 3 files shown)


src/libNode/Node.cpp
```diff
@@ -1112,23 +1112,15 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
...
-  if (REJOIN_NODE_NOT_IN_NETWORK && !LOOKUP_NODE_MODE && !bDS) {
-    if (!bInShardStructure) {
-      LOG_GENERAL(
-          WARNING,
-          "Node " << m_mediator.m_selfKey.second
-                  << " is not in network, apply re-join process instead");
+  if (rejoinCondition && bIpChanged) {
+    LOG_GENERAL(
+        INFO, "My IP has been changed. So will broadcast my new IP to network");
+    if (!UpdateShardNodeIdentity()) {
```
JamesHinshelwood, line 1121: In theory it should be possible for a node's identity to change, but the IP address not to change. e.g. Kubernetes could just shuffle the same pods around different k8s nodes. Should we try to handle this case?

src/libNode/Node.cpp
```diff
@@ -1112,23 +1112,15 @@ bool Node::StartRetrieveHistory(const SyncType syncType,
         m_mediator.m_ds->m_shards, m_mediator.m_ds->m_publicKeyToshardIdMap,
         m_mediator.m_ds->m_mapNodeReputation);
   }
+  bool rejoinCondition = REJOIN_NODE_NOT_IN_NETWORK && !LOOKUP_NODE_MODE && !bDS;
+  LOG_GENERAL(INFO, "rejoinCondition = "<< rejoinCondition); //TODO: remove this log before merge
```
JamesHinshelwood, line 1116: Reminder to remove :)

JamesHinshelwood (approved):
LGTM, please merge to master at the same time :)

src/libNode/Node.cpp
chetan-zilliqa: If identity changed because of pods reshuffling, this would still be considered IP change from node per see.
So I think this should still work without any change.

PR #3571, opened 2023-04-13
[master] ZIL-5189: fix undefined behaviour issues

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/depends/common/FixedHash.h
+++ b/src/depends/common/FixedHash.h
@@ -23,6 +23,7 @@
 
 #include <algorithm>
 #include <array>
+#include <bit>
 #include <cstdint>
 #include <random>
 #include <boost/functional/hash.hpp>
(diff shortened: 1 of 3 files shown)


src/depends/common/FixedHash.h
```diff
@@ -338,19 +339,11 @@ namespace dev
...
-        return (hash1[0] == hash2[0]) && (hash1[1] == hash2[1]) && (hash1[2] == hash2[2]) && (hash1[3] == hash2[3]);
-    }
-
     /// Fast std::hash compatible hash function object for h256.
     template<> inline size_t FixedHash<32>::hash::operator()(FixedHash<32> const& value) const
     {
-        uint64_t const* data = reinterpret_cast<uint64_t const*>(value.data());
-        return boost::hash_range(data, data + 4);
+        auto data = std::bit_cast<std::array<uint64_t, 4>>(value.m_data);
+        return boost::hash_range(data.begin(), data.end());
```
rrw-zilliqa, line 346: Why are we hashing a hash? You're not going to make it any more random, you know! Perhaps worth a ticket to investigate deleting?

rrw-zilliqa (approved):
Looks good to me :-)

src/depends/common/FixedHash.h
yaron-zilliqa: My guess is that whoever wrote this tried to make it more "efficient" somehow by hashing the uint64_t instead of the actual characters.
It's possible to delete this specialization as well, but I thought it might be a bit safer to do the same thing in case there's some code that might break if this is changed.
So this will produce the same hash as it always has. After the release we can probably investigate if it's worth changing.

PR #3572, opened 2023-04-13
[master] Fix gas consistency test against local network

n-hutton:
In this test, the value is too small when testing against a local network with realistic gas charges, to actually pay for the transaction. Also, it needs to block to wait for the funding transaction to clear.


tests/EvmAcceptanceTests/test/Transfer.ts
```diff
@@ -153,20 +152,25 @@ describe("Transfer ethers", function () {
   it("should return check gas and funds consistency", async function () {
     let rndAccount = ethers.Wallet.createRandom();
 
-    const FUND = BigNumber.from(100_000_000_000);
+    const FUND = BigNumber.from(200_000_000_000_000_000n);
```
its-saeed, line 155: Maybe using something like `ethers.utils.parseUnits("10", "ether")` is more readable.

PR #3577, opened 2023-04-13
Adding erc20/zrc2 interop example

bzawisto:
## Description
PR contains an example of how erc20/zrc2 can be implemented with additional bridging contract. It was assumed that ERC20 functionality (albeit a bit simplified) is embedded into Bridge. Separate ERC20 contract is still doable though it may require some non-standard methods (i.e. burn and mint). 

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)

PR #3591, opened 2023-04-20
[master] Rework zil mint check on selfdestruct

n-hutton:
This checks that the refunded amount from contracts is EXACT. Before it was incorrectly understanding 'fund' to mean how much to transfer, when it was what the (non-contract) account value should become.

The logic should be the same.

### DIFF
--- a/src/libCps/Amount.h
+++ b/src/libCps/Amount.h
@@ -49,6 +49,17 @@ class Amount final {
   constexpr auto operator>(const Amount& other) const {
     return !(*this <= other);
   }
+  constexpr auto operator+(const Amount& rhs) const {
+    // return the sum of two numbers as biggest type to avoid overflow
+    return Amount{this->toWei() + rhs.toWei()};
+  }
+  constexpr auto operator-(const Amount& rhs) const {
+    // return the sum of two numbers as biggest type to avoid overflow
+    return Amount{this->toWei() - rhs.toWei()};
+  }
+  constexpr auto operator==(const Amount& rhs) const {
+    return this->toWei() == rhs.toWei();
+  }
 
  private:
   constexpr Amount(const uint256_t& wei) : m_value(wei){};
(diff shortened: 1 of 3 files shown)


src/libCps/Amount.h
```diff
@@ -49,6 +49,17 @@ class Amount final {
   constexpr auto operator>(const Amount& other) const {
     return !(*this <= other);
   }
+  constexpr auto operator+(const Amount& rhs) const {
+    // return the sum of two numbers as biggest type to avoid overflow
```
JamesHinshelwood, line 53: Maybe I'm misunderstanding the comment - How are you avoiding overflow?

src/libCps/Amount.h
```diff
@@ -49,6 +49,17 @@ class Amount final {
...
+    return Amount{this->toWei() + rhs.toWei()};
+  }
+  constexpr auto operator-(const Amount& rhs) const {
+    // return the sum of two numbers as biggest type to avoid overflow
```
JamesHinshelwood, line 57: nit: returns the sum?

src/libCps/CpsRunEvm.cpp
```diff
@@ -648,24 +648,30 @@ void CpsRunEvm::HandleApply(const evm::EvmResult& result,
...
-          ", requested: " + funds.toWei().convert_to<std::string>();
-
-      LOG_GENERAL(WARNING, "possible zil mint! " << error);
-      mAccountStore.TransferBalanceAtomic(accountToRemove, fundsRecipient,
-                                          currentFunds);
-      span.SetError(error);
+    // Funds is what we want our contract to become/be modified to.
+    // Check that the contract funds plus the current funds in our account
+    // is equal to this value
+    if(!(funds == recipientPreFunds + currentContractFunds)) {
```
JamesHinshelwood, line 663: nit
```suggestion
    if(funds != recipientPreFunds + currentContractFunds) {
```

JamesHinshelwood (approved):

src/libCps/Amount.h
n-hutton: I thought something like this: if you use toQa you downcast 256->128 so you could concevably overflow the sum. However, now that I look again, there is only a constructor for 256 (128 in variant unused?)... so it all seems moot.

PR #3593, opened 2023-04-21
[master] ZIL-5001: Enable and fix out of gas test

JamesHinshelwood:
I've tested on Sepolia and the behaviour is consistent there - the contract does not revert, it fails with an out of gas error.

### DIFF
--- a/tests/EvmAcceptanceTests/test/ContractRevert.ts
+++ b/tests/EvmAcceptanceTests/test/ContractRevert.ts
@@ -36,8 +36,14 @@ describe("Revert Contract Call", function () {
     await expect(this.contract.callChainOk()).not.to.be.reverted;
   });
 
-  // FIXME: https://zilliqa-jira.atlassian.net/browse/ZIL-5001
-  xit("Should be reverted without any reason if specified gasLimit is not enough to complete txn", async function () {
-    await expect(this.contract.outOfGas({gasLimit: 100000})).to.be.revertedWithoutReason();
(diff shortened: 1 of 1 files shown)


tests/EvmAcceptanceTests/test/ContractRevert.ts
```diff
@@ -36,8 +36,14 @@ describe("Revert Contract Call", function () {
     await expect(this.contract.callChainOk()).not.to.be.reverted;
   });
 
-  // FIXME: https://zilliqa-jira.atlassian.net/browse/ZIL-5001
-  xit("Should be reverted without any reason if specified gasLimit is not enough to complete txn", async function () {
-    await expect(this.contract.outOfGas({gasLimit: 100000})).to.be.revertedWithoutReason();
+  it("Should be reverted without any reason if specified gasLimit is not enough to complete txn", async function () {
+    const txn = await this.contract.outOfGas({gasLimit: 100000});
+    expect(txn).not.to.be.reverted;
+    try {
```
its-saeeed, line 42: I think we can use something like `expect(...).to.throw(...)`

its-saeeed (approved):

tests/EvmAcceptanceTests/test/ContractRevert.ts
its-saeeed: https://www.chaijs.com/api/bdd/#method_throw

PR #3594, opened 2023-04-21
Simplify `Amount`

JamesHinshelwood:
As far as I can see, we never use the `uint128_t` half of the `m_value` variant. When passed an amount in Qa, we rescale it to Wei by multiplying it by `EVM_ZIL_SCALING_FACTOR` anyway.

Therefore, we simplify the type to just `uint256_t` and remove the unused branches.


rrw-zilliqa (approved):
Looks good to me :-)

PR #3595, opened 2023-04-21
[master] Encode EVM init data as hexadecimal

JamesHinshelwood:
Previously, devex would hang when trying to load any screen which contained an EVM transaction, because the response contained an error. Now, devex copes slightly better (but still crashes and displays a blank screen if you click on the transaction).


rrw-zilliqa (approved):
Looks good to me :-)

PR #3596, opened 2023-04-21
Add some additional logging for EVM execution

JamesHinshelwood:
This makes it easier to find the error when an EVM call failed. We don't expose the error to the user, but this does at least let us tell by looking at the logs.

There are a few other miscellaneous logs in this commit too, sorry.


rrw-zilliqa (approved):
Looks good to me :-)

PR #3600, opened 2023-04-23
[master] Delegate call fix

n-hutton:
There was an issue with delegate call such that if you set a value and then called delegate call, the delegated call requests that value from the chain, which is incorrect.

I found another issue in fixing this one which is that the originally calling contract also doesn't see the value of the delegated call once it has completed.

The changes to Delegatecall.sol did not pass the `require` on master, now it does.

In order to fix this there was two changes.

- When there is a trap call to make the delegate, the current state changes so far are applied CPP side so that subsequent continuations can query the correct values via IPC.

- When a call completes, it will presumably traverse back down the stack. These pending continuations are held in a map in the EVM, and the state changes of this completed call are searched for and applied in all other continuations (these state changes are not to account/balance values).

Refer to inline comments. There is one potential way that this PR badly breaks things and that is if a static call on completion is able to modify state of other continuations via this second fix. There is a test to guard this but I will double check. The fix also relies on the fact that all active continuations are from the same TX/call which I think is fine.

### DIFF
--- a/evm-ds/src/evm_server_run.rs
+++ b/evm-ds/src/evm_server_run.rs
@@ -23,7 +23,7 @@ use crate::precompiles::get_precompiles;
 use crate::pretty_printer::log_evm_result;
 use crate::protos::Evm as EvmProto;
 use crate::{scillabackend, LoggingEventListener};
-use protobuf::Message;
+use protobuf::{Message};
 
 #[allow(clippy::too_many_arguments)]
 pub async fn run_evm_impl(
@@ -71,7 +71,15 @@ pub async fn run_evm_impl(
                 let result = handle_panic(tx_trace, gas_limit, "Continuation not found!");
                 return Ok(base64::encode(result.write_to_bytes().unwrap()));
             }
+
             let recorded_cont = recorded_cont.unwrap();
+
+            // print the continuation storage
+            eprintln!("Continuation memory: {:?}", recorded_cont.memory);
+            eprintln!("Continuation stack: {:?}", recorded_cont.stack);
+            eprintln!("Continuation storages: {:?}", recorded_cont.storages);
+            eprintln!("Continuation id: {:?}", continuation.get_id());
+
             let machine = Machine::create_from_state(Rc::new(recorded_cont.code), Rc::new(recorded_cont.data),
                                                               recorded_cont.position, recorded_cont.return_range, recorded_cont.valids,
                                                               recorded_cont.memory, recorded_cont.stack);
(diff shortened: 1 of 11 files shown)


evm-ds/src/evm_server_run.rs
```diff
@@ -225,8 +228,9 @@ fn build_exit_result(
                         modify.set_reset_storage(reset_storage);
                         let storage_proto = storage
                             .into_iter()
-                            .map(|(k, v)| backend.encode_storage(k, v).into())
+                            .map(|(k, v)| { continuations.lock().unwrap().update_states(address, k, v); backend.encode_storage(k, v).into()})
```
n-hutton, line 231: applying state changes to other pending continuations - seems a little hacky to do it this way

evm-ds/src/evm_server_run.rs
```diff
@@ -253,6 +259,51 @@ fn build_call_result(
     let mut trap_reason = EvmProto::ExitReason_Trap::new();
     trap_reason.set_kind(EvmProto::ExitReason_Trap_Kind::CALL);
     let mut exit_reason = EvmProto::ExitReason::new();
+
+    let (state_apply, _) = executor.into_state().deconstruct();
+
+    // We need to apply the changes made to the state so subsequent calls can
+    // see the changes.
+    result.set_apply(
```
n-hutton, line 267: applying pending changes

src/libCps/CpsRunEvm.cpp
```diff
@@ -204,6 +204,65 @@ CpsExecuteResult CpsRunEvm::HandleCallTrap(const evm::EvmResult& result) {
       ProtoToUint(callData.transfer().value()).convert_to<std::string>());
 
   uint64_t remainingGas = result.remaining_gas();
+  Address thisContractAddress = ProtoToAddress(mProtoArgs.address());
+  Address fundsRecipient;
+  Amount funds;
+
+  // Apply the evm state changes made so far so subsequent contract calls
+  // can see the changes (delegatecall)
+  for (const auto& it : result.apply()) {
```
n-hutton, line 213: this block is basically a copy paste of applying state changes

src/libCps/CpsRunEvm.cpp
```diff
@@ -204,6 +204,65 @@ CpsExecuteResult CpsRunEvm::HandleCallTrap(const evm::EvmResult& result) {
...
+          for (const auto& x : states) {
+            toDeletes.emplace_back(x.first);
+          }
+
+          mAccountStore.UpdateStates(thisContractAddress, {}, toDeletes, true);
+        }
+        // Actually Update the state for the contract (only allowed for
+        // thisContractAddress in non-static context!)
+        for (const auto& sit : it.modify().storage()) {
+          if (iterAddress != thisContractAddress ||
```
rrw-zilliqa, line 241: Is there any magic reason why this is inside the loop? No harm to having it there (the compiler will hoik it anyway, if it has any sense), just wondering if there is something I'm missing?

rrw-zilliqa (approved):
Looks good to me, though I'll try to write some negative examples, too, and I too worry about static calls  ...

PR #3599, opened 2023-04-22
[master] Fix the gas limit used for gas charging

valeryz:
## Backward Compatibility

- [ ] This is not a breaking change
- [ ] This is a breaking change

### Implementation

- [ ] **ready for review**

### Integration Test (Core Team)

- [ ] local machine test

- [ ] small-scale cloud test

PR #3601, opened 2023-04-24
[master] ZIL-5208: fix evm gas rrw

rrw-zilliqa:
## Description

* Scale EVM gas so we don't undercharge (equiv to #3599 )
* Round gas used so that deployments are zero-sum
* Trivial test case for the same.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


JamesHinshelwood (approved):
Looks good to me!

PR #3602, opened 2023-04-24
[master] Static call test

n-hutton:
As a result of the fix of #3600 on completion of a call the states that call changed would be reflected to lower level continuations. If it happened such that this call was a static call that somehow modified an address matching another continuation this could modify contract state in a static context, which would be incorrect.

To fix this, we note whether the call we are making is static and do not apply state changes after the call completes. This causes the test to pass.

There is another potential edge case to look at: if the static call gets trapped, when it returns, does it still know it is in a static context? The EVM continuation does not record that it is static so we are relying on the CPP side to do so correctly.


rrw-zilliqa (approved):
Looks good to me :-)

PR #3605, opened 2023-04-24
[master] Static to non static switch allowed erronerously

n-hutton:
Final bug in the money market tests. I mentioned in a previous PR that static calls might not be working correctly and it turned out to be the case.

There is a guard in the CPP side that triggers when a call trap is generated that has a non static context, when its parent was static: "Attempt to change context from static to non-static in precompile-trap". This could happen if a static call then made a delegate call.

This caused healthy TX to get rejected at the estimate gas stage.

The fix to this is to make sure the call trap inherits its parents' static-ness.


rrw-zilliqa (approved):
Looks good to me :-)

PR #3610, opened 2023-04-25
Implement some APIs for Otterscan

JamesHinshelwood:
This adds the following JSON-RPC APIs:

- `erigon_getHeaderByNumber`
- `ots_getApiLevel`
- `ots_hasCode`
- `ots_getBlockDetails`
- `ots_getBlockTransactions`
- `ots_getContractCreator` (requires `ARCHIVAL_LOOKUP_WITH_TX_TRACES`)

The `ots_*` APIs are supposedly documented at https://github.com/otterscan/otterscan/blob/develop/docs/custom-jsonrpc.md, but I found a few discrepancies that meant it was easier to reverse-engineer the API from the (quite readable) Otterscan source code instead.

This is enough to get Otterscan running with the ability to view blocks, accounts and transactions. Many features rely on further APIs which we have not yet implemented though.


src/libPersistence/BlockStorage.cpp
```diff
@@ -1648,6 +1649,24 @@ bool BlockStorage::GetMinerInfoShards(const uint64_t& dsBlockNum,
   return found;
 }
 
+bool BlockStorage::PutContractCreator(const dev::h160 address, const dev::h256 txnHash) {
+  if (!m_contractCreatorDB) {
+    return true;
```
n-hutton, line 1654: shouldn't this be false?

src/libPersistence/BlockStorage.cpp
```diff
@@ -1648,6 +1649,24 @@ bool BlockStorage::GetMinerInfoShards(const uint64_t& dsBlockNum,
...
+  return m_contractCreatorDB->Insert(address.asBytes(), txnHash.asBytes());
+}
+
+dev::h256 BlockStorage::GetContractCreator(const dev::h160 address) {
+  if (!m_contractCreatorDB) {
+    return dev::h256();
+  }
+
+  lock_guard<mutex> g(m_contractCreatorMutex);
+  return dev::h256(reinterpret_cast<const unsigned char*>(m_contractCreatorDB->Lookup(address.asBytes()).c_str()), dev::h256::ConstructFromPointerType::ConstructFromPointer);
```
n-hutton, line 1667: could this potentially be a memory access violation if Lookup returns empty string => h256 constructor is memcpying fixed width memory?

n-hutton (changes requested):
minor comments

src/libPersistence/BlockStorage.cpp
JamesHinshelwood: So `m_contractCreatorDB` won't be initialised when `!LOOKUP_NODE_MODE`. But I decided to still call `PutContractCreator` unconditionally and just silently return if we weren't a lookup node. I can change the setup if you think its unclear?

PR #3617, opened 2023-05-02
[master] Fix bug 5217

Steve-White-UK:
Fix the ZIL-5217 and use access control correctly
Add the access control to the error response as well as normal response.
add access control to error response.

## Backward Compatibility

### Implementation

- [x ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)

PR #3618, opened 2023-05-02
[master] ZIL-5214: Make request queue size and threads configurable

JamesHinshelwood:
We also adjust the defaults to 65536 and 64, which is closer to the values from v8.1.0.


rrw-zilliqa (approved):
Looks good to me :-)

chetan-zilliqa:
@JamesHinshelwood , maybe make it a part of the constants.xml file so that we can even tweak it at a later point if required.

JamesHinshelwood:
> @JamesHinshelwood , maybe make it a part of the constants.xml file so that we can even tweak it at a later point if required.

I added them to constants, but just set the default values to what we agreed. So we can change them later if needed.

PR #3619, opened 2023-05-02
[master] ZIL-5214: Revert priority mutex implementation for AccountStore

JamesHinshelwood:


rrw-zilliqa (commented):
What now stops repeated contract read requests from starving the sync thread so badly that the seedpub drops out of consensus?

chetan-zilliqa (approved):

chetan-zilliqa:
I think let's wait for the PR to merge.

PR #3623, opened 2023-05-07
[master] Adding UUPS test cases

bzawisto:
## Description
Added some test cases for UUPS. All seems to be working as expected. 

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


tests/EvmAcceptanceTests/test/gas-transfer/GasTransferTest.ts
```diff
@@ -27,48 +33,56 @@ class TestCase {
...
+    new TestCase(
+      "0",
+      BigNumber.from(ONE_GWEI),
+      BigNumber.from(0),
+      BigNumber.from(0),
+      BigNumber.from(ONE_GWEI),
+      BigNumber.from(ONE_GWEI),
+      BigNumber.from(ONE_GWEI)
+    )
+    //   new TestCase( ONE_GWEI, BigNumber.from(0), BigNumber.from(0), BigNumber.from(0)),
```
n-hutton, line 53: dead code here

n-hutton (changes requested):
minor code cleanup

tests/EvmAcceptanceTests/contracts/proxy/ImplementationV1.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.8.9;
+
+import "./ProxyStorageV1.sol";
+import "./interfaces/InterfaceV1.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+contract ImplementationV1 is ProxyStorage, Initializable, OwnableUpgradeable, UUPSUpgradeable, InterfaceV1
```
rrw-zilliqa, line 10: Contract names in hardhat are global, so probably worth giving this a more self-explanatory name (though I admit mine are just as bad!)

tests/EvmAcceptanceTests/test/gas-transfer/GasTransferTest.ts
rrw-zilliqa: I think this patch removes the dead code .. ?

tests/EvmAcceptanceTests/test/proxy/UUPStest.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  it("Should not be able to initialize the same proxy contract twice", async function () {
+    await expect(proxyContract.initialize(123)).to.be.rejected;
+  });
+
+  it("Should not be possible to initialize the same proxy contract twice with different signer", async function () {
+    const newSigner = await parallelizer.takeSigner();
+    await expect(proxyContract.connect(newSigner).initialize(123)).to.be.rejected;
+  });
+});
```
rrw-zilliqa, line 73: I think this works fine as a test of UUPS proxies, but not as a test of UUPS proxies calling other UUPS proxies, which is the bug the OP reported. Not that we shouldn't add it, but .. (I have a test that does do this - will PR today - sorry!)

rrw-zilliqa (approved):
LGTM, but we should (maybe later) rename the contracts so we don't get confused about names later.

rrw-zilliqa:
Closing at @bzawisto 's request.

PR #3628, opened 2023-05-15
[master] Change Open Telemetry Tracing to use a batch processor so that we do not use synchronous writing. see ticket 5247

Steve-White-UK:
## Description
Enable batch processing for non stdout tracers.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK:
This PR is trashed and superceed by https://github.com/Zilliqa/Zilliqa/pull/3653 which has the actual ten lines needed.

PR #3631, opened 2023-05-23
[master] Interop fixes

bzawisto:
## Description
Scilla contracts were not exported at some places - which wasn't expected and - in result it returned empty json. 
Also, we should wait for receipt otherwise accounts are not updated at the time the assertion is made.

## Backward Compatibility

### Implementation

- [X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libEth/Eth.cpp
+++ b/src/libEth/Eth.cpp
@@ -257,15 +257,6 @@ void DecorateReceiptLogs(Json::Value &logsArrayFromEvm,
     logEntry["blockHash"] = blockHash;
     logEntry["blockNumber"] = blockNum;
     logEntry["logIndex"] = (boost::format("0x%x") % logIndex).str();
-    // remove unsupported/add missing keys in receipt objects
-    logEntry.removeMember("params");
-    logEntry.removeMember("_eventname");
-    if (!logEntry.isMember("topics")) {
-      logEntry["topics"] = Json::arrayValue;
-    }
-    if (!logEntry.isMember("data")) {
-      logEntry["data"] = "0x";
-    }
     ++logIndex;
   }
 }
--- a/src/libEth/Eth.cpp
+++ b/src/libEth/Eth.cpp
@@ -257,6 +257,15 @@ void DecorateReceiptLogs(Json::Value &logsArrayFromEvm,
     logEntry["blockHash"] = blockHash;
     logEntry["blockNumber"] = blockNum;
     logEntry["logIndex"] = (boost::format("0x%x") % logIndex).str();
+    // remove unsupported keys in receipt objects
+    logEntry.removeMember("params");
+    logEntry.removeMember("_eventname");
+    if (!logEntry.isMember("topics")) {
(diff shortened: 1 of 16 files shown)


src/libEth/Eth.cpp
```diff
@@ -257,6 +257,15 @@ void DecorateReceiptLogs(Json::Value &logsArrayFromEvm,
     logEntry["blockHash"] = blockHash;
     logEntry["blockNumber"] = blockNum;
     logEntry["logIndex"] = (boost::format("0x%x") % logIndex).str();
+    // remove unsupported/add missing keys in receipt objects
+    logEntry.removeMember("params");
+    logEntry.removeMember("_eventname");
+    if (!logEntry.isMember("topics")) {
```
rrw-zilliqa, line 263: I need to check, but does this not effectively suppress all events from Scilla? We don't want to do that, because if we do, there is no way at all for indexers to eg. track ZRC-2 balances. Changing the format is (probably) fine, but removing the data isn't.

rrw-zilliqa (changes requested):
Changes to the event structure I think need looking at ?
Other than that, looks broadly fine - will check again when I'm back home this evening - sorry!

src/libEth/Eth.cpp
bzawisto: I removed unsupported fields (leaving parts of scilla receipts present). How should we tackle that - I can rework it a bit to transform maps of events to plain arrays supported by web3 libraries, would that make sense?

rrw-zilliqa: I'm afraid I'm going to need to look at Scilla receipts to see what they do, but the choices are laid out in RFC075 - either we implement a new way of building Scilla receipts into EVM transactions that is compatible with web3 parsing libraries, in which case yes - you need to somehow format them in a way compatible with ethereum receipts _and_ write it down in the RFC, or we leave it the way it was.

rrw-zilliqa: Removing data isn't going to work for anyone.

bzawisto: Well, the only possibility in this case is to rework scilla events. There are some nuances on how we should treat event name (whether it should be signature or just a name), how to handle indexed arguments, how should that affect bloom filters, etc...

rrw-zilliqa: I would argue that leaving well enough alone was the right thing to do, but if you're passionate about dong this, how about you have a stab at fixing RFC075 and I'll race you to writing a spec that looks like it might work?

Note that we probably want to try to reduce the load on indexers - apart from anything else, viewblock is going to have to be fixed to work with our new event format whatever happens, and the less work that is, the more likely indexers won't be broken in the future.

rrw-zilliqa: (erm, sorry, that was poorly explained - the less work it is to write a Zilliqa indexer, the more likely they won't in future be broken)

chetan-zilliqa (approved):

PR #3634, opened 2023-05-23
Otterscan final apis

n-hutton:


evm-ds/src/cps_executor.rs
```diff
@@ -345,6 +345,14 @@ impl<'a> Handler for CpsExecutor<'a> {
         }
     }
 
+    fn get_create_address(
```
n-hutton, line 348: now a requirement of handler via the evm PR

evm-ds/src/evm_server_run.rs
```diff
@@ -157,25 +154,31 @@ pub async fn run_evm_impl(
                 listener.finished_call();
 
                 match exit_reason {
-                    evm::ExitReason::Succeed(_) => {}
+                    evm::ExitReason::Revert(_) => {
```
n-hutton, line 157: when tx fails, capture error blob

scripts/integration_test_js.sh
```diff
@@ -118,9 +118,12 @@ else
 
     cd tests/EvmAcceptanceTests/
     npm install
-    DEBUG=true MOCHA_TIMEOUT=20000 npx hardhat test
```
n-hutton, line 121: this has effect of not interleaving the hardhat output with the iso server output, it just dumps it after

JamesHinshelwood (commented):
Overall, LGTM.

Probably makes sense for someone more qualified in C++/ZQ1 to have a review too. @Steve-White-UK ?

JamesHinshelwood (approved):

Steve-White-UK (approved):
I so no reason not to merge to master if James is happy with that.

src/libData/AccountStore/AccountStoreSCEvm.cpp
```diff
@@ -283,6 +284,38 @@ bool AccountStoreSC::EvmProcessMessage(EvmProcessContext &params,
...
+  return trace_json.toStyledString();
+}
+
+void getAddressesFromTrace(const std::string &trace, std::set<std::string> &addresses) {
+  Json::Value trace_json;
+  JSONUtils::GetInstance().convertStrtoJson(trace, trace_json);
+
+  Json::Value parsed;
+
+  auto const item = trace_json["otter_addresses_called"];
```
bzawisto, line 305: Could it be just `auto const parsed = ...`?

bzawisto (approved):

PR #3643, opened 2023-06-06
[master] Probe erc165 interface

bzawisto:
## Description
Detect for ScillaReceiver presence in evm contract and if it supports ScillaReceiver iface - revert.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libCps/CpsRunEvm.cpp
```diff
@@ -741,6 +741,95 @@ void CpsRunEvm::HandleApply(const evm::EvmResult& result,
...
+  evm::EvmArgs args;
+  *args.mutable_address() = AddressToProto(destinationAddress);
+  const auto code = accStore.GetContractCode(destinationAddress);
+  *args.mutable_code() = DataConversion::CharArrayToString(StripEVM(code));
+
+  *args.mutable_data() = DataConversion::CharArrayToString(
+      DataConversion::HexStrToUint8VecRet(ERC165METHOD));
+  *args.mutable_caller() = AddressToProto(caller);
+  *args.mutable_origin() = AddressToProto(ctx.origSender);
+  args.set_gas_limit(30000);
```
rrw-zilliqa, line 762: Worth a comment here to say that this 30k limit comes from ERC165.

src/libCps/CpsRunEvm.cpp
```diff
@@ -741,6 +741,95 @@ void CpsRunEvm::HandleApply(const evm::EvmResult& result,
...
+    if (!std::empty(evmResult.return_value()) &&
+        static_cast<uint8_t>(evmResult.return_value().back()) == 0x01) {
+      return false;
+    }
+  }
+  // Eventually check support for scilla interface in evm contract
+  {
+    // Check if destination supports 'function
+    // handle_scilla_message(string,bytes)'
+    constexpr auto SUPPORT_SCILLA_IFACE =
```
rrw-zilliqa, line 814: Comment to say how you got the value SUPPORT_SCILLA_IFACE?

src/libCps/CpsRunEvm.cpp
```diff
@@ -741,6 +741,95 @@ void CpsRunEvm::HandleApply(const evm::EvmResult& result,
   }
 }
 
+bool CpsRunEvm::ProbeERC165Interface(CpsAccountStoreInterface& accStore,
```
rrw-zilliqa, line 744: You probably want to split this out into a ProbeERC165Interface(ifaceId), and a call to ProbeERC165Interface(keccak256("scilla_handle_event", .. )) so that future checks for interface support don't have to replicate the generic ERC165 support code, but we can do that later..

rrw-zilliqa (approved):
Some v minor nitpicks, otherwise looks good to me :-)

PR #3644, opened 2023-06-06
Zil 5250 scilla server should be restarted if it crashes or is killed

Steve-White-UK:
## Description

To detect RPC error when a call is made with no server

Align the behavior with that of evm-ds

Recoverability and Zilliqa 1.5 release
## Backward Compatibility

## Review Suggestion

launch a system with this fix and log into a pod and kill scilla_server, the process should be restarted on the next scilla request.
if scilla-server were to crash then it would be silently repaired by this fix, a log message would be generated to inform the log that scilla was determined to be not running.

### Implementation

- [ X] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libScilla/ScillaClient.cpp
```diff
@@ -78,6 +78,29 @@ void ScillaClient::Init() {
   }
 }
 
+bool ScillaClient::isScillaRuning() {
+  std::string cmdStr = "ps aux | awk '{print $2\"\\t\"$11}' | grep \"" +
+                       SCILLA_SERVER_BINARY + "\" | awk '{print $1}'";
```
rrw-zilliqa, line 83: Yuck! but I guess this is idiomatic in our horrible, horrible world :-(

src/libScilla/ScillaClient.cpp
```diff
@@ -248,7 +287,15 @@ bool ScillaClient::CallDisambiguate(uint32_t version, const Json::Value& _json,
...
     } else {
-      result = e.what();
+      if (e.GetCode() == jsonrpc::Errors::ERROR_RPC_JSON_PARSE_ERROR ||
+          e.GetCode() == jsonrpc::Errors::ERROR_CLIENT_CONNECTOR){
+        LOG_GENERAL(WARNING, "Looks like connection problem");
+        if (!isScillaRuning()) {
+          LOG_GENERAL(WARNING, "Scilla is not running");
+          CheckClient(version,true);
+        }
+        return CallChecker(version, _json, result, counter - 1);
```
rrw-zilliqa, line 297: This is cut&pasted three times - maybe think about putting it in a function?

rrw-zilliqa (approved):
Basically looks good to me :-)

src/libScilla/ScillaClient.cpp
yaron-zilliqa: Agreed; an alternative (a bit more effort though) would be to use Boost.Process to launch Scilla and use the launcher/handle as needed. Integrates nicely with Boost.Asio as well.

Steve-White-UK: I left it as humour, it was generated by github auto pilot, I personally would have used pidof, but copilot did a perfect job of emulating the programming style of the original author.

PR #3647, opened 2023-06-07
[master] Adding support for reading scilla ctor immutables

bzawisto:
## Description
Read plain ctor fields from scilla contract.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)
(I'd say we should have tests for the priority of init data vs fields, but IIRR Scilla won't let you overlap them, so it's a moot point, I think?)

PR #3650, opened 2023-06-07
[master] Return response from `eth_getCode` in lowercase

JamesHinshelwood:
The specification doesn't explicitly document this, but its implied by the fact that most implementations return all hex data in lowercase.

Without this, some contract verification tools break, because they detect a mismatch in the compiled and deployed bytecode.


Steve-White-UK (approved):
I am assured this passes our tests so appproved.

PR #3670, opened 2023-06-15
[master] ZIL-5278: Retry sending outgoing messages a fixed number of times

JamesHinshelwood:
We keep the existing time-based retries, but increase the default value from 5 seconds to 300 seconds. My intention here is to effectively disable the time-based timeouts, but keep the option to re-enable it with a configuration change.

We also start respecting the `MAXRETRYCONN` configuration again which limits the number of times we will retry the sending of a message.


Steve-White-UK (commented):
The implementation is fine, not going to get nit-picky over the C++, that is just not me.

I am going to dwell on it overnight, something doesn't sit right with me on this, I am unsure of the impact of the message hanging around in here for any significant time. 

I have merged it into my local branch of latest 9.2 and let it rip and its fine , I would like to put it onto some kind of harness that simulates bad connections and highlight some vars in metrics to see what behavior looks like in odd conditions, also I would love to study previous,current and the artem version of this to understand the migrations.

in the morning though.

PR #3672, opened 2023-06-16
[release/v9.2] ZIL-5281: client side for hot upgrades

yaron-zilliqa:
## Description

Changes to zilliqad:
* Download a manifest from S3
* If changed, execute the manifest based on the action specified.
* Communicate with the zilliqa process through a pipe (actually two pipes: read & write)
* Upon acknowledgement, wait for zilliqa to stop, perform the ugprade & restart with no recovery flag.

Changes to zilliqa:
* Listens to notifications from zilliqad (through a pipe).
* Upon an upgrade request, start rejecting transactions at the quiesce-at DS block and shutdown at upgrade-at DS block.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK (commented):
Can we do a code walk-through on Monday?

rrw-zilliqa (changes requested):
Some notes (which I'm afraid I put in a text file, so they're not inline)

* Before we ship this, we need to provide some guardrails to ensure we can't rugpull miner nodes (this is for their protection and ours).
   : Add a mode which refuses to upgrade unless the SHA256 of the downloaded image matches that of a value in an updater.conf.js file.
   (this doesn't need to be default, but there does need to be a way to do it).

* Decide what to do about making config file updates. Easiest would be
  to just have them be manual, but then we need to check that we're
  not upgrading zilliqa until we have a config file that is compatible
  with the version we are upgrading to (corollary - we need to make
  sure we don't start an older version with a newer configuration
  file).

* When passing data between programs, it's best to format as JSON (or
  some universally agreed format) rather than parsing strings - cuts
  down on bugs but also attack surface.

* I'm reasonably sure the comms process only works because of atomicity of read()
  and write(), which in turn only works because asio does a single
  read() or write() per readSome()/writeSome() - might want to check
  that this is actually a guarantee.

* Your validation check (DaemonListener.cpp:56) is probably fragile - can you be sure that if I set <blk>+1 at an arbitrary point in the cycle, we'll quiesce correctly at <blk>+1 ? (and you probably want updateDSBlock > quiesceDSBlock)

* Quite a bit of cut & paste (eg. of S3 downloads) in ZilliqaUpdater.cpp - prob best refactored
* It's often easier to do things like SHA256 in-process; shelling out (particularly to absolute paths) can cause obscure breakage later.
* Worth adding docs for what the manifest file format actually is - reverse-engineering it from the parsing code every time is going to be tedious
* Decode the manifest once and then use the (type-safe) values - don't keep looking up manifest["upgrade-at-dsblock"] - you'll get it wrong and cause horror.
* ZilliqaUpdater::Update() doesn't do atomic file copies (AFAICT), so will corrupt zilliqa if it's interrupted.
* .. and will also fail if the backup already exists (so going around again will fail to upgrade), because it reports an error if the source and destination are the same.
* Name files 'zilliqa-<version>.tar.bz' not just 'zilliqa.tar.bz2' or you will get very confused about which version is which

yaron-zilliqa:
>     * Before we ship this, we need to provide some guardrails to ensure we can't rugpull miner nodes (this is for their protection and ours).
>       : Add a mode which refuses to upgrade unless the SHA256 of the downloaded image matches that of a value in an updater.conf.js file.
>       (this doesn't need to be default, but there does need to be a way to do it).
There's already a check that refuses to updated unless the SHA256 is identical to what we send in the manifest. Perhaps I'm missing what you meant about a separate mode or having it in some file?

>     * Decide what to do about making config file updates. Easiest would be
>       to just have them be manual, but then we need to check that we're
>       not upgrading zilliqa until we have a config file that is compatible
>       with the version we are upgrading to (corollary - we need to make
>       sure we don't start an older version with a newer configuration
>       file).
That's true; this is meant as a preliminary, simple version. If needed this can be extended through adding fields/actions to the manifest in the future.

>     * When passing data between programs, it's best to format as JSON (or
>       some universally agreed format) rather than parsing strings - cuts
>       down on bugs but also attack surface.
True, but this is so simple that I didn't this it even merited JSON, honestly. I was also trying ti refrain from introducing other dependencies (but by now jsoncpp_lib is one of them, actually...). I can change this.

>     * I'm reasonably sure the comms process only works because of atomicity of read()
>       and write(), which in turn only works because asio does a single
>       read() or write() per readSome()/writeSome() - might want to check
>       that this is actually a guarantee.
At which point will you have more than one? The read/write are done to different pipes (doing this in a single one can cause issues such as dead-locks even), but this is a fairly standard way of writing/reading using Asio.

>     * Your validation check (DaemonListener.cpp:56) is probably fragile - can you be sure that if I set +1 at an arbitrary point in the cycle, we'll quiesce correctly at +1 ? (and you probably want updateDSBlock > quiesceDSBlock)
Testing this I'm also a bit confused about whether things start from 0 or 1 so not entirely sure if this comparison is just right.
Not sure I understand your point about the +1 at an arbitrary point in the cycle?
There's already a check for updateDSBlock > quiesceDSBlock; the update is rejected in this case (in DaemonListener.cpp).

>     * Quite a bit of cut & paste (eg. of S3 downloads) in ZilliqaUpdater.cpp - prob best refactored
Twice, but yeah - I'll refactor this.

>     * It's often easier to do things like SHA256 in-process; shelling out (particularly to absolute paths) can cause obscure breakage later.
Agreed, but as mentioned above - I didn't initially want to introduce more dependencies and we're already doing tons of things in the daemon by invoking processes, so thought that was reasonable in this case.

>     * Worth adding docs for what the manifest file format actually is - reverse-engineering it from the parsing code every time is going to be tedious
Yes - I'll this in the JIRA ticket later.

>     * Decode the manifest once and then use the (type-safe) values - don't keep looking up manifest["upgrade-at-dsblock"] - you'll get it wrong and cause horror.
> 
>     * ZilliqaUpdater::Update() doesn't do atomic file copies (AFAICT), so will corrupt zilliqa if it's interrupted.
Not sure what you mean by atomic file copies here. Using rename? or using a (hard) link? Given that there are multiple operations on the file system not sure how this can be done more atomically. 

>     * .. and will also fail if the backup already exists (so going around again will fail to upgrade), because it reports an error if the source and destination are the same.
It wouldn't fail in this case because due to the copy they wouldn't be the same entity on the file system (unlike if one was a sym-/hard link of the other). Regardless, probably a good idea to clean the backup file upon failure as well.

>     * Name files 'zilliqa-.tar.bz' not just 'zilliqa.tar.bz2' or you will get very confused about which version is which
Fair point; this actually just the local file name to make it easier a bit (everything resides in the uuid of the update). We can add another field to the manifest so that the updater knows what prefix to use.

Steve-White-UK (approved):
This code should be passive and will be used in the future when it deemed stable to perform a hotfix.

PR #3673, opened 2023-06-18
Scilla call tests

bzawisto:
## Description
Luckily, the issue with invisible scilla updates to evm contract withing single frame turned out to be rather straightforward to fix. (See ScillaCallComplex.sol).
There are also many new tests added.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK (commented):
Looking Good - Executing tests as we speak, expect to approve later if passes successfully.

Steve-White-UK (approved):
Passes the tests nicely, should probably get an rrw stamp but LGTM

tests/EvmAcceptanceTests/contracts/rfc75/ScillaCallComplex.sol
```diff
@@ -0,0 +1,77 @@
...
+        uint256 argsLength = encodedArgs.length;
+        bool success;
+        bytes memory output = new bytes(36);
+        assembly {
+            success := staticcall(21000, 0x5a494c92, add(encodedArgs, 0x20), argsLength, add(output, 0x20), 32)
+        }
+        require(success);
+        (funds) = abi.decode(output, (uint128));
+        return funds;
+
```
rrw-zilliqa, line 74: Add some negative tests here - what happens if success == false? If the txn can succeed when success=false, test this.

Test booleans, ints, ByStr, Address, List (and update RFC075 to indicate that maps are implemented)

tests/EvmAcceptanceTests/contracts/scilla/rfc75/ScillaCallComplex.scilla
```diff
@@ -0,0 +1,30 @@
...
+  = Emp ByStr20 Uint128
+
+ field nestedMap: Map ByStr20 (Map ByStr20 Uint128)
+    = Emp ByStr20 (Map ByStr20 Uint128)
+
+transition setValue(to: ByStr20, val: Uint128)
+    simpleMap[to] := val;
+    nestedMap[to][to] := val;
+    value := val;
+    msg = {_tag : "SomeMessage"; _recipient : to; sender : _sender; recipient : to; _amount: Uint128 0};
```
rrw-zilliqa, line 27: use different values for each of simpleMap, nestedMap, and value so that if you read the wrong one in the chain code, the test fails.

rrw-zilliqa (approved):
LGTM on a first look, but worth addressing the comments in another PR - thanks!

PR #3676, opened 2023-06-19
[master] API enable otterscan extended tracing

n-hutton:
In the tests, it is possible in a devnet it is testing the otter and trace tx tests against a network that doesn't have the trace flag on.

This adds an endpoint so that the flag can be externally set. This lets the test make sure it is set before it runs (it will still fail if you are not running against a seed node though, ie one with ARCHIVAL_LOOKUP set to true)

### DIFF
--- a/src/cmd/isolated_server.cpp
+++ b/src/cmd/isolated_server.cpp
@@ -107,8 +107,6 @@ int main(int argc, const char* argv[]) {
   ENABLE_EVM = true;
   LOOKUP_NODE_MODE = true;
   SCILLA_SERVER_LOOP_WAIT_MICROSECONDS = 100000;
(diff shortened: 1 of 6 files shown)


src/cmd/isolated_server.cpp
```diff
@@ -107,8 +107,6 @@ int main(int argc, const char* argv[]) {
   ENABLE_EVM = true;
   LOOKUP_NODE_MODE = true;
   SCILLA_SERVER_LOOP_WAIT_MICROSECONDS = 100000;
-  ARCHIVAL_LOOKUP_WITH_TX_TRACES =
```
n-hutton, line 110: don't need this since test can turn it on now

PR #3679, opened 2023-06-20
[master] Unified gas tracking

bzawisto:
## Description
Track gas left in single place to avoid rare bugs when gas consumption is not calculated properly.
It could be observed when there were multiple scilla messages sent and gas limit was set at the time of their creation, not execution.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/libCps/CpsRunEvm.cpp
```diff
@@ -873,11 +873,11 @@ void CpsRunEvm::ProvideFeedback(const CpsRun& previousRun,
     }
   } else {
     const auto& scillaResult = std::get<ScillaResult>(results.result);
-    const auto remainingGas =
-        GasConv::GasUnitsFromCoreToEth(scillaResult.gasRemained);
+    // const auto remainingGas =
```
n-hutton, line 876: delete commentd out code?

n-hutton (changes requested):
minor

PR #3683, opened 2023-06-20
[master] ZIL-5278: Ignore blacklist when sending consensus challenges

JamesHinshelwood:


src/libNetwork/P2PComm.h
```diff
@@ -139,7 +139,8 @@ class P2PComm {
   void SendMessage(const std::deque<Peer>& peers, const zbytes& message,
                    unsigned char startByteType = zil::p2p::START_BYTE_NORMAL,
                    bool inject_trace_context = false,
-                   bool bAllowSendToRelaxedBlacklist = false);
+                   bool bAllowSendToRelaxedBlacklist = false,
```
JamesHinshelwood, line 142: Is this a bug? The argument order is swapped between the header and source files. Compare this to
https://github.com/Zilliqa/Zilliqa/blob/6758a8c7f97c36b0a3eca3e3c80b54aff8c213bd/src/libNetwork/P2PComm.cpp#L1255-L1256

Steve-White-UK: Yes that is clearly a bug

JamesHinshelwood: Part of me is scared to change it, in case we're relying on the backwards ordering now that it is in production? Shall we fix it in a separate PR, outside of this release?

Steve-White-UK: I have it out for test now, I was going to run a test with the PR as it is and then compare it with a run with the bug corrected, the sad part is there is a clang warning for this that has possibly been disabled otherwise it would have been picked up by the latest compiler.

src/libNetwork/SendJobs.cpp
```diff
@@ -291,6 +293,11 @@ class PeerSendQueue : public std::enable_shared_from_this<PeerSendQueue> {
   bool CheckAgainstBlacklist() {
     auto sz = m_queue.size();
     if (sz > 0 && IsBlacklisted(m_peer, false)) {
+      auto& item = m_queue.front();
```
yaron-zilliqa, line 296: I would probably add a comment explaining why this is necessary.

yaron-zilliqa (approved):

Steve-White-UK:
Please do not merge as it is, Chetan and Steve have tested this and do not believe this is sufficient, we are still investigating and will be back on it 25/6/2023

chetan-zilliqa (changes requested):
We would need to set the `ignoreWhteilist` to `true` as the last params in the following places so that the messages are sent on retry.
https://github.com/Zilliqa/Zilliqa/blob/69dd66b01399290421ed16f927fb2caba23caafb/src/libConsensus/ConsensusLeader.cpp#L327
https://github.com/Zilliqa/Zilliqa/blob/69dd66b01399290421ed16f927fb2caba23caafb/src/libConsensus/ConsensusLeader.cpp#L572
https://github.com/Zilliqa/Zilliqa/blob/69dd66b01399290421ed16f927fb2caba23caafb/src/libConsensus/ConsensusLeader.cpp#L837
Frankly speaking, we could just reuse the `bAllowSendToRelaxedBlacklist`, but this is also set in the job. however, this will again create confusion because the same flag is used for strict checking.

PR #3684, opened 2023-06-21
Mirror evm ds

n-hutton:
These are a series of changes made to integrate evm into z2. The behavour should be identical to before. The changes are commented inline, and in short are:

- main body of run_evm moved into run_evm_direct which can be called by either z1 or z2
- the use of the backend is now generic so that either backend can be passed in
- in order to support this, backend functions that were ScillaBackend specific have been moved to free functions that are only used in the z1 case


evm-ds/src/precompiles/modexp.rs
```diff
@@ -133,7 +133,8 @@ fn required_gas(input: &[u8]) -> Result<u64, ExitError> {
     // Multiply the cost by 20 if the modulus is even, to account for the slow implementation of `BigUint::modpow`.
     // Based on https://github.com/paritytech/frontier/blob/master/frame/evm/precompile/modexp/src/lib.rs
     // Licensed under the Apache License, Version 2.0.
-    let cost = core::cmp::max(200, saturating_round(gas)).saturating_mul(if (modulus % 2).is_zero() { 20 } else { 0 });
+    let cost = core::cmp::max(200, saturating_round(gas))
```
n-hutton, line 136: cargo fmt change

PR #3687, opened 2023-06-22
[master] ZIL-5288: Add a --native option to build-scilla to allow building it natively.

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/localdev.py
+++ b/scripts/localdev.py
@@ -970,21 +970,85 @@ def build_scilla(config, driver, tag):
                capture_output = False)
     return image_name
 
+def build_native_scilla(config):
+    build_env = os.environ.copy()
+    build_env.update(config.default_env)
+
+    if not build_env.get("VCPKG_ROOT"):
+        raise GiveUp("Environment variable VCPKG_ROOT must be defined to point to vcpkg")
+
+    vcpkg_triplet = run_or_die(config, ["scripts/vcpkg_triplet.sh"], in_dir = SCILLA_DIR, env = build_env,
+               capture_output = True).decode('utf-8')
+
+    build_env["SCILLA_REPO_ROOT"] = SCILLA_DIR
+    build_env["PKG_CONFIG_PATH"] = f"{os.path.join(SCILLA_DIR, 'vcpkg_installed', vcpkg_triplet, 'lib', 'pkg-config')}" + f":{build_env['PKG_CONFIG_PATH']}" if build_env.get('PKG_CONFIG_PATH') else ''
+
+    # Cleanup
+    shutil.rmtree(os.path.join(SCILLA_DIR, "vcpkg_installed"), ignore_errors = True)
+    shutil.rmtree(os.path.join(SCILLA_DIR, "_build"), ignore_errors = True)
+
+    output = run_or_die(config, ["grep", "OCAML_VERSION_RECOMMENDED=", "Makefile"], in_dir = SCILLA_DIR, env = build_env,
+               capture_output = True).decode('utf-8')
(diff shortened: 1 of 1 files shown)


scripts/localdev.py
```diff
@@ -970,21 +970,81 @@ def build_scilla(config, driver, tag):
...
+    build_env.update(config.default_env)
+
+    if not build_env.get("VCPKG_ROOT"):
+        raise GiveUp("Environment variable VCPKG_ROOT must be defined to point to vcpkg")
+
+    build_env["SCILLA_REPO_ROOT"] = SCILLA_DIR
+
+    # Cleanup
+    shutil.rmtree(os.path.join(SCILLA_DIR, "vcpkg_installed"), ignore_errors = True)
+    shutil.rmtree(os.path.join(SCILLA_DIR, "_build"), ignore_errors = True)
```
rrw-zilliqa, line 984: This is going to lead to extremely long build times (esp on OS X) if we don't check if it's been done first - you probably want to introduce a sentinel here, because otherwise every time you invoke build_native_scilla() you will need to wait for deps to be built (and this includes every time you want to test a change to scilla)

scripts/localdev.py
```diff
@@ -970,21 +970,81 @@ def build_scilla(config, driver, tag):
...
+        if len(p) > 1:
+            build_env[p[0]] = p[1][1:-1]
+
+    # Set PWD explicitly to get around usage in Scilla's Makefile; in sub-shells
+    # this will resolve to the wrong directly which could mess up rpath.
+    build_env["PWD"] = SCILLA_DIR
+    run_or_die(config, ["make", "opamdep"], in_dir = SCILLA_DIR, env = build_env,
+               capture_output = False)
+    run_or_die(config, ["make"], in_dir = SCILLA_DIR, env = build_env,
+               capture_output = False)
```
rrw-zilliqa, line 1014: OK, but why is this here and not in the makefile? I think I'm missing something .. ?

scripts/localdev.py
yaron-zilliqa: If vcpkg has already been installed (either from a previous compilation or Zilliqa), it would simply be installed. Anything new or different will be compiled (and be used by vcpkg in subsequent installs).
The scripts/build_deps.sh in scilla does the same actually and checks if it exists or not so this doesn't really make much of a difference.

The reason for deleting both (predominantly _build) is because there have been strange issues related to ocaml/dune cache.

scripts/localdev.py
yaron-zilliqa: These are in the Makefile but they're separate target and apart from the Makefile being fairly old (and at least partially out-of-date?), I find that something there & in the instructions is very sensitive to the order in which this is done.
So I was hoping to do this independently for now without having this depend on other PRs in Scilla. Once stable, we will fix Scilla & the instructions.

scripts/localdev.py
rrw-zilliqa: I don't think it does - it calls `scripts/vcpkg.sh` which does:

```
# If already installed, exit early.
if [[ -d vcpkg_installed ]]
then
    echo "Found vcpkg_installed, not installing again"
    exit 0
fi
```

but good point - sorry - I'd not twigged that vcpkg will cache separately from vcpkg_installed - in that case, np (though if the problem is in the build cache, I suspect you will simply regenerate it here .. )

rrw-zilliqa (approved):
You'll also want to fix build_native_to_workspace() to use build_scilla(), presumably, or isolated server won't work?

(mind out for #3688 though)

PR #3688, opened 2023-06-22
[master] Zil 5287 isolated server scilla fix

rrw-zilliqa:
## Description

Make scilla-server run again on x64.

## Backward Compatibility

## Review Suggestion

Check code. Run isolated on OS X and x64 (works on x86, OS X/x86 for me - can't test aarch64)

### Implementation

- [X ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/localdev.py
+++ b/scripts/localdev.py
@@ -613,8 +613,11 @@ def isolated(config, enable_evm = True, block_time_ms = None):
     with open(os.path.join(target_workspace, 'constants.xml'), 'w') as f:
         f.write(output_config)
     new_env = os.environ.copy()
-    #old_path = new_env.get("DYLD_LIBRARY_PATH", "")
-    #new_env["DYLD_LIBRARY_PATH"] = f"{target_workspace}/lib:{old_path}"
+    # This used to contain DYLD_LIBRARY_PATH - and it probably still should,
+    # but I don't have a Mac to hand - rrw 2023-06-22
+    for var in [ "LD_LIBRARY_PATH" ]:
+        old_path = new_env.get(var, "")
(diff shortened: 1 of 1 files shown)


scripts/localdev.py
```diff
@@ -613,8 +613,11 @@ def isolated(config, enable_evm = True, block_time_ms = None):
     with open(os.path.join(target_workspace, 'constants.xml'), 'w') as f:
         f.write(output_config)
     new_env = os.environ.copy()
-    #old_path = new_env.get("DYLD_LIBRARY_PATH", "")
-    #new_env["DYLD_LIBRARY_PATH"] = f"{target_workspace}/lib:{old_path}"
+    # This used to contain DYLD_LIBRARY_PATH - and it probably still should,
+    # but I don't have a Mac to hand - rrw 2023-06-22
+    for var in [ "LD_LIBRARY_PATH" ]:
```
yaron-zilliqa, line 618: Is this really necessary? The paths should really be embedded in the binary so I guess for some they're not? Perhaps might be worthwhile pointing out so we can fix.

rrw-zilliqa: If it wasn't, my isolated server would've worked! To reproduce (at least on my machine), you can:

cd ../scilla
git clean -dxf
cd ../zilliqa
./scripts/localdev.py isolated

isolated server comes up, but scilla-server crashes because it can't find libSchnorr.so

yaron-zilliqa: Fair enough; I just merged my PR for the --native option. I believe it should fix this issue which would make this change redundant. Give it a try.

yaron-zilliqa (approved):

PR #3690, opened 2023-06-27
[master] make spurious otter messages go away

n-hutton:
when estimating gas, you don't want to save the otter output. This made it print a warning but it was spurious.


Steve-White-UK (approved):
LGTM

PR #3695, opened 2023-06-28
[master] Cleanup of fixes for chainlink

n-hutton:
Fixes for chainlink, comments inline:
- Subscription bug
- Prev hash not set
- Block JSON fields not parseable

There are not CI tests for these, it would be good if there were the following tests:
- Blocks have prevHash set correctly
- Subscription to newHeads works, and it works if you subscribe multiple times (all subscribers recieve the block)


src/libServer/IsolatedServer.cpp
```diff
@@ -1181,9 +1181,13 @@ TxBlock IsolatedServer::GenerateTxBlock() {
...
   }
 
+  auto const prevTxBlock = m_sharedMediator.m_txBlockChain.GetLastBlock();
+
+  auto const prevHash = m_blocknum == 0 ? BlockHash(): prevTxBlock.GetBlockHash();
+
   TxBlockHeader txblockheader(0, m_currEpochGas, 0, m_blocknum,
                               TxBlockHashSet(), numtxns, m_key.first,
-                              TXBLOCK_VERSION);
+                              1, TXBLOCK_VERSION, CommitteeHash(), prevHash);
```
n-hutton, line 1190: this was actually incedentally incorrectly setting the version as the ds block number here

src/libServer/IsolatedServer.cpp
```diff
@@ -1181,9 +1181,13 @@ TxBlock IsolatedServer::GenerateTxBlock() {
     m_txnBlockNumMap[m_blocknum].clear();
   }
 
+  auto const prevTxBlock = m_sharedMediator.m_txBlockChain.GetLastBlock();
+
+  auto const prevHash = m_blocknum == 0 ? BlockHash(): prevTxBlock.GetBlockHash();
```
n-hutton, line 1186: Set the prev hash in blocks (iso server only)

src/libServer/JSONConversion.cpp
```diff
@@ -160,8 +160,8 @@ const Json::Value JSONConversion::convertTxBlocktoEthJson(
   // Required by ethers
   retJson["extraData"] = "0x";
   retJson["nonce"] = "0x0";
-  retJson["receiptsRoot"] = "0x";
-  retJson["transactionsRoot"] = "0x0";
+  retJson["receiptsRoot"] = "0x0000000000000000000000000000000000000000000000000000000000000000";
```
n-hutton, line 163: chainlink unable to parse non hash-length strings, so make them full length. I think this should be fine, no client actually uses these fields I believe

tests/EvmAcceptanceTests/hardhat.config.ts
```diff
@@ -113,13 +113,15 @@ const config: HardhatUserConfig = {
       miningState: false
     },
     testnet: {
-      url: "https://devnetnh-l2api.dev.z7a.xyz",
-      websocketUrl: "wss://devnetnh-l2api.dev.z7a.xyz",
+      url: "https://testnet-receipts-l2api.dev.z7a.xyz",
```
n-hutton, line 116: convenience

src/libEth/filters/SubscriptionsImpl.h
```diff
@@ -70,14 +71,32 @@ class SubscriptionsImpl {
...
     Id id = 0;
 
-    /// True if this conn subscribed to pending txns
-    bool subscribedToPendingTxns = false;
+    // populated if this conn subscribed to pending txns.
+    // since a single connection could have multiple subscriptions, we need to
+    // keep a set of subscription IDs here. They are non-overlapping
+    // with subscribedToNewHeads
+    //bool subscribedToPendingTxns = false;
+    std::unordered_set<uint64_t> subscribedToPendingTxn;
```
n-hutton, line 79: So, the same subscription ID was being used for every connection to newHeads. This is completely fine until there are multiple newHeads subscriptions in the same single connection. This is a bit of a corner case.

Anyway, to solve this, connectinos now keep a set of subscription IDs which are unique and not re-used for subscription types (hewHeads, pendingTxn). This has been confirmed to work with the chainlink software.

Steve-White-UK (approved):
OK

PR #3696, opened 2023-06-29
[master] ZIL-5298: add connection timer and make reconnection time configurable

yaron-zilliqa:
## Description

Added the following (optional) constants:
* CONNECTION_TIMEOUT_IN_MS
* RECONNECT_INTERVAL_IN_MS

The reconnection timer is initially used to enforce a timeout before invoking async_connect.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/common/Constants.cpp
+++ b/src/common/Constants.cpp
@@ -494,9 +494,9 @@ const unsigned int MAX_WHITELISTREQ_LIMIT{
 const unsigned int SENDJOBPEERS_TIMEOUT{
     ReadConstantNumeric("SENDJOBPEERS_TIMEOUT", "node.p2pcomm.")};
 const unsigned int CONNECTION_TIMEOUT_IN_MS{
-    ReadConstantNumeric("CONNECTION_TIMEOUT_IN_MS", "node.p2pcomm.", 60000)};
+    ReadConstantNumeric("CONNECTION_TIMEOUT_IN_MS", "node.p2pcomm.", 30000)};
 const unsigned int RECONNECT_INTERVAL_IN_MS{
-    ReadConstantNumeric("RECONNECT_INTERVAL_IN_MS", "node.p2pcomm.", 5000)};
+    ReadConstantNumeric("RECONNECT_INTERVAL_IN_MS", "node.p2pcomm.", 2500)};
 
 // PoW constants
 const bool FULL_DATASET_MINE{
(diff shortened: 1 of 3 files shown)


chetan-zilliqa (commented):
@yaron-zilliqa , thanks for the PR.
I understand that we set `wait_timer` before connecting so that we can timeout and reconnect after connecting the timer expiry.

can you help to verify my understanding and let me know the behaviour in both scenarios ?

**Scenario 1) The connect takes ~2 minutes(Socket unreachable) before returning connection timeout:**  If the node returns after 60 seconds connect time as it tries to reconnect, however, there are two conditions -
`    if (!CheckAgainstBlacklist()  || ExpiredOrDone()) { return;}`

`CheckAgainstBlacklist` would return true as the node is not blacklisted, but `ExpiredOrDone `would be true(> 5) so it won't perform reconnect attempt. 
**Scenario 2) The connect takes ~1 second (Process down) and returns with ec(connection timeout immediately) :** If the node returns after 1 second, `OnConnected` would return ec code(connection timeout) which means the node is relaxed blacklisted. do the node retry in this case ? If not, I think we would need to check allow_relaxed_blacklist  in `CheckAgainstBlacklist`  function and return such that we allow the node to reconnect.

yaron-zilliqa:
> **Scenario 1) The connect takes ~2 minutes(Socket unreachable) before returning connection timeout:** If the node returns after 60 seconds connect time as it tries to reconnect, however, there are two conditions - ` if (!CheckAgainstBlacklist() || ExpiredOrDone()) { return;}`
> 
> `CheckAgainstBlacklist` would return true as the node is not blacklisted, but `ExpiredOrDone `would be true(> 5) so it won't perform reconnect attempt.
Why would ExpiredOrDone return true simply after 5 seconds? It will return true for a timeout after std::max(5000u, TX_DISTRIBUTE_TIME_IN_MS * 3 / 4). So a minimum of 5 seconds but the expiry would be TX_DISTRIBUTE_TIME_IN_MS * 3 / 4 which I believe is quite a few minutes.

>**Scenario 2) The connect takes ~1 second (Process down) and returns with ec(connection timeout immediately) :** If the node returns after 1 second, `OnConnected` will return ec code(connection timeout) and it will try one more reconnection attempt before marking the node as relaxed blacklisted. do the node retry in this case? If not, we would need to check allow_relaxed_blacklist in `CheckAgainstBlacklist` function and return such that we allow the node to reconnect.
> 
> I think we would need to set m_expiry to a larger duration say (3 minutes).
If async_connect returns immediately, the the code works as it always has. The change isn't meant to address this, but only a timer for the async_connect phase in case it takes a very long time.

Steve-White-UK (approved):
LGTM

PR #3706, opened 2023-07-04
[master] Preserve msg_fallback_handler in the chain of scilla calls

bzawisto:
## Description
Msg fallback handler should be kept as set in scillaCall2 no matter how many indirect scilla invocations we have before a msg is sent to evm contract.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)

PR #3715, opened 2023-07-06
Fixed firstPage and lastPage assignment

lucac-zilliqa:
Fixed the assignment of these fields so that they behaved according to the Otterscan API specification

### DIFF
--- a/src/libServer/EthRpcMethods.cpp
+++ b/src/libServer/EthRpcMethods.cpp
@@ -2126,11 +2126,10 @@ Json::Value EthRpcMethods::OtterscanSearchTransactions(const std::string& addres
     response["txs"] = txs;
     response["receipts"] = receipts;
 
-    // Otterscan docs. If results are less than pagesize, results returned as-is.
-    if (!(res.size() < pageSize)) {
-      response["firstPage"] = before || !wasMore;
-      response["lastPage"] = !before || !wasMore;
-    }
+    // Otterscan docs:
+    // These are the conditions for which these variables are set to true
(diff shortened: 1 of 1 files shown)


src/libServer/EthRpcMethods.cpp
```diff
@@ -2126,11 +2126,10 @@ Json::Value EthRpcMethods::OtterscanSearchTransactions(const std::string& addres
...
 
-    // Otterscan docs. If results are less than pagesize, results returned as-is.
-    if (!(res.size() < pageSize)) {
-      response["firstPage"] = before || !wasMore;
-      response["lastPage"] = !before || !wasMore;
-    }
+    // Otterscan docs:
+    // These are the conditions for which these variables are set to true
+    response["firstPage"] = (before && !blockNumber) || (!before && !wasMore);
+    response["lastPage"] = (!before && !blockNumber) || (before && !wasMore);
```
JamesHinshelwood, line 2132: I wonder if it would make sense to break up these booleans to make them more readable. Currently I struggle to work out the logic.
Could we do something like
```
if (before) {
    response["firstPage"] = ...;
    response["lastPage"] = ...;
} else {
    response["firstPage"] = ...;
    response["lastPage"] = ...;
}
?

lucac-zilliqa: In my head I saw it as firstPage is true if the API call was ots_searchTransactionsBefore and block number 0 was passed into the call or if the API call was ots_searchTransactionsAfter and the entire chain has been explored. Same for lastPage. I thought that would be easier to read (maybe I should make the blockNumber more explicit?)

n-hutton (approved):

PR #3721, opened 2023-07-12
[master] Added check for correct files in home directory

lucac-zilliqa:
installme.sh requires that the contents of docs/setup/ are copied into the home directory before being executed.

This change adds a check for the files in the directory and will print a message if not and exit.

The first condition checks whether the executing script is located in the home directory. If the script can be executed elsewhere as long as the other files are in the home directory, then we can remove this check.

The second and third check for the presence of setenv.sh and runlocaldev.sh in the home directory.


rrw-zilliqa (approved):
Clunky, but LGTM :-) - thank you!

PR #3727, opened 2023-07-18
Added additional tests to check edge cases

lucac-zilliqa:
Added additional tests for the Otterscan APIs to test for the edge cases brought up in ZIL-5306 and ZIL-5307. 

Additional questions: When should I use the `Parrelizer` class instead of the basic `ethers.getContractFactory`? I also created a contract with `signed` being undefined. What is the default `Signer` value here?


rrw-zilliqa (approved):
Looks good to me :-)

PR #3730, opened 2023-07-26
[master] Added transactionCount field to fullblock object

lucac-zilliqa:
Added a simple line that adds this new field.

Question: I see that in the above function `GetBlockDetails`, it gets the transaction count using `txBlock.GetHeader().GetNumTxs()`. Does this have any performance benefits over `transactions.size()`?


JamesHinshelwood (approved):

JamesHinshelwood:
> Question: I see that in the above function GetBlockDetails, it gets the transaction count using txBlock.GetHeader().GetNumTxs(). Does this have any performance benefits over transactions.size()?

In theory they're different numbers, but you would hope they are always consistent. In both cases all you're doing is reading an integer from a known position on the stack, so the performance impact of both will be minimal (and might even be zero after compilation and optimisation).

n-hutton (approved):

PR #3731, opened 2023-07-28
[master] Fixed function so correctly returns GasLimit

lucac-zilliqa:
I removed the line that caused this API call to constantly return "0x1" as the GasLimit of a block.

### DIFF
--- a/src/libServer/EthRpcMethods.cpp
+++ b/src/libServer/EthRpcMethods.cpp
@@ -2240,8 +2240,6 @@ Json::Value EthRpcMethods::GetBlockDetails(const uint64_t blockNumber) {
   uint128_t fees = (isVacuous ? 0 : txBlock.GetHeader().GetRewards() * EVM_ZIL_SCALING_FACTOR);
   auto jsonBlock = GetEthBlockCommon(txBlock, false);
 
-  jsonBlock["gasLimit"] = "0x1";
-
   jsonBlock.removeMember("transactions");
   jsonBlock["transactionCount"] = txBlock.GetHeader().GetNumTxs();
   jsonBlock["logsBloom"] = Json::nullValue;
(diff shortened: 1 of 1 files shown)


src/libServer/EthRpcMethods.cpp
```diff
@@ -2240,8 +2240,6 @@ Json::Value EthRpcMethods::GetBlockDetails(const uint64_t blockNumber) {
   uint128_t fees = (isVacuous ? 0 : txBlock.GetHeader().GetRewards() * EVM_ZIL_SCALING_FACTOR);
   auto jsonBlock = GetEthBlockCommon(txBlock, false);
 
-  jsonBlock["gasLimit"] = "0x1";
```
JamesHinshelwood, line 2243: @n-hutton Why was this here? It looks suspiciously intentional.

lucac-zilliqa: I was wondering that as well. I sent Nathan a message on Slack asking about this. `GetEthBlockCommon` is called in this top level function in which it then calls `JSONConversion::convertTxBlocktoEthJson` which contains this line `retJson["gasLimit"] = (boost::format("0x%x") % txheader.GetGasLimit()).str();`. It seems then that it should normally return a JSON object containing a "gasLimit" field without the line I removed.

lucac-zilliqa: @n-hutton responded and said that it was to prevent a division by zero error as running it in isolated server mode it would default to 0 and UI would break due to a division by 0 error

n-hutton: Yup, the otterscan UI wouldn't even render which took some digging to find out why. The fix looks good though!

n-hutton (approved):

PR #3737, opened 2023-08-04
[master] Fix waiting for pods indefinitely in localdev.py

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/scripts/localdev.py
+++ b/scripts/localdev.py
@@ -393,7 +393,7 @@ def grafana_down(config):
     """ Let helm undeploy grafana """
     helm_remove_repository(config, 'grafana')
 
-def prometheus_up(config, testnet_name, count = 23):
+def prometheus_up(config, testnet_name, count = 24):
     """ Let helm deploy prometheus """
     ips = []
     while True:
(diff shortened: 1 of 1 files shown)


Steve-White-UK (changes requested):
The code should not have a hard-coded constant, the actual number can be calculated from the values in the parameters to the bootstrap command.

WE have to adjust it every time you run a slightly different bootstrap, very annoying.

yaron-zilliqa:
> 

Yes, but that value depends on the fanout value which has some strange calculation in the testnet repo.
I'd rather have that than copying all that code.

JamesHinshelwood:
Counting the number of lines from `kubectl get pods` seems super fragile. You can probably do this with a `kubectl wait --for=condition=ready pod -l app=zilliqa` (might be the wrong label) or similar.

yaron-zilliqa:
> 

Yes it is super fragile, except it's unfortunately needed for Prometheus' configuration and the configuration is hard coded anyway, and the script was already doing something similar. I'll see if waiting for 'ready' works.

Steve-White-UK (approved):

Steve-White-UK (changes requested):
Needs a bit of adjustment for Q4 dev I will take a look in the morning.

Steve-White-UK (approved):

rrw-zilliqa:
Closed at @yaron-zilliqa 's request.

PR #3739, opened 2023-08-10
[q4-working-branch] Fixing test

bzawisto:
This fixes sporadic transaction-replaced issues. There could be more in other test suites and we should fix them in the same way.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/tests/EvmAcceptanceTests/test/ERC20isZRC2.ts
+++ b/tests/EvmAcceptanceTests/test/ERC20isZRC2.ts
@@ -34,18 +34,23 @@ describe("ERC20 Is ZRC2", function () {
       this.skip();
     }
 
-    contractOwner = await parallelizer.takeSigner();
+    const signers = await ethers.getSigners();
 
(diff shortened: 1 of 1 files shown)


tests/EvmAcceptanceTests/test/ERC20isZRC2.ts
```diff
@@ -34,18 +34,23 @@ describe("ERC20 Is ZRC2", function () {
...
-      "FungibleToken",
-      await contractOwner.getAddress(),
-      "ERC20isZRC2 Token",
-      "SDT",
-      2,
-      1_000
+    const signers = await ethers.getSigners();
+
+    if(signers.length < 3) {
+      this.skip();
```
its-saeed, line 40: Why not `expect` this value? To fail if conditions are not satisfied, instead of ignoring it silently.

its-saeed (approved):

Steve-White-UK (approved):

PR #3760, opened 2023-08-22
ZIL-5332-debug_accountRange

lucac-zilliqa:
This pull request will add the functionality of debug_accountRange which will return a pagination style response of addresses in the network.

The idea is to implement a cache (an array) which will contain address hexes (still to be determined) which can then be indexed into a slice of the array returned (or also include the corresponding account information).


530
lucac-zilliqa, line 1: Ignore this file. I will remove this from the commit

src/libData/AccountStore/AccountStore.cpp
```diff
@@ -843,3 +854,26 @@ void AccountStore::PrintAccountState() {
...
+  for(auto it = m_state.begin(); it != m_state.end(); ++it){
+    std::pair<zbytesConstRef , zbytesConstRef >item = it.at();
+    std::array<unsigned char, 40> arr;
+    std::copy(item.first.begin(), item.first.end(), arr.begin());
+    m_cache.push_back(arr);
+  }
+}
+
+void AccountStore::PrintAddressCache(){
+  for (const std::array<unsigned char, 40>& entry : m_cache) {
```
Steve-White-UK, line 875: PrintAddressCache should probably lock the mutex as it may be possible for another thread to change the cache and invalidate the iterator, if there is only one thread then it possibly does not need a mutex.

Steve-White-UK (changes requested):
If you can add a test to the EvmAcceptanceTests we can take it on board and this test will be executed via ci.

Need to make sure it only looks for short durations and that it doesn't use excessive memory, apart from that it should be OK. 


If you provide a test we will test it out as part of our current schedule.

PR #3774, opened 2023-09-11
[q4-working-branch] Q4 working branch multiplier

Steve-White-UK:
## Description

This is the new CPP Multiplier which replaces the go duplicator.

It should be compatible with go duplicator in that it is invoked with the same parameters and it forwards messages from the inbound port to the nodes specified by getting the contents of the URL supplied.

This version does no caching or discarding of data, what goes in goes out.

Replace goduplicator if we need to.

## Backward Compatibility

## Status

Very new should be used in conjunction with dev-minikube-local which will need to be updated to invoke this executable.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK (commented):
LGTM

its-saeed (approved):

PR #3795, opened 2023-09-29
[master] GetTransactionStatus Should respect txn query with or without 0x.

its-saeed:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libRemoteStorageDB/RemoteStorageDB.cpp
+++ b/src/libRemoteStorageDB/RemoteStorageDB.cpp
@@ -246,7 +246,7 @@ bool RemoteStorageDB::UpdateTxn(const string& txnhash, const TxnStatus status,
   return true;
 }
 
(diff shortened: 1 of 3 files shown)


rrw-zilliqa (approved):
Looks good to me - but is there a test for this that this fixes? going to assume there is ..

chetan-zilliqa (approved):

its-saeed:
> Looks good to me - but is there a test for this that this fixes? going to assume there is ..

Chetan and I tried to cross-check it and it worked fine. The rest of the API uses the same approach. However because these kinds of APIs are not supported on the isolated server and devnet, I wasn't able to test it using an acceptance test. 
If we can have some kind of in-memory database for iso server/devent that would be great for writing acceptance tests.

@chetan-zilliqa will test it specifically after we have a new testnet and I'll try to see if we can have an in memory/in file database to use for test purposes.

PR #3797, opened 2023-10-04
[master] Kill evm and scilla client by PID not by name.

its-saeed:
## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


chetan-zilliqa:
Some test cases to check. Feel free to ignore if already verified those.
1) kill zilliqa process and check what happens. should see one scilla and evm process running.
2) The scilla process should restart itself after every 99th block or based on what value you set in `NUM_FINAL_BLOCK_PER_POW`.

chetan-zilliqa (approved):

PR #3805, opened 2023-10-10
[master] ZIL-5403: check if a LevelDB object is null or not

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


src/depends/libDatabase/LevelDB.cpp
```diff
@@ -15,610 +15,493 @@
...
-    leveldb::Status s = leveldb::DestroyDB(STORAGE_PATH + PERSISTENCE_PATH +
-        (this->m_subdirectory.size() ? "/" + this->m_subdirectory : "") + "/" + this->m_dbName,
-        leveldb::Options());
-    if (!s.ok())
-    {
-        LOG_GENERAL(WARNING, "[DeleteDB] Status: " << s.ToString());
-        return -1;
-    }
+int LevelDB::DeleteDBForNormalNode() {
+  m_db.reset();
```
rrw-zilliqa, line 425: Should this not handle the case where m_db is nullptr?

rrw-zilliqa: (there are other instances below)

rrw-zilliqa (approved):
Check comments, but I don't think this makes things worse .. worth testing quite thoroughly, though, because bugs here might be quite bad.

src/depends/libDatabase/LevelDB.cpp
rrw-zilliqa: :+1:

src/depends/libDatabase/LevelDB.cpp
```diff
@@ -15,610 +15,493 @@
...
+template <typename ArgT>
+std::string LookupImpl(const std::shared_ptr<leveldb::DB> db,
+                       const std::string& dbName, ArgT&& arg,
+                       bool* found = nullptr) {
+  std::string value;
+  if (!db) {
+    LOG_GENERAL(WARNING, "LevelDB " << dbName << " isn't available");
+    if (found) {
+      *found = false;
+    }
```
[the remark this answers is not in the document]
JamesHinshelwood: @yaron-zilliqa Sorry, I'm not understanding this properly, but the new code does jump out to me as a bit scary. I agree that its dangerous to assume 'DB doesn't exist' is the same as 'DB exists, but is empty'.

src/depends/libDatabase/LevelDB.cpp
```diff
@@ -15,610 +15,493 @@
...
-    if(!(std::filesystem::exists(path)))
-    {
-        LOG_GENERAL(WARNING, path + " does not exist");
-        return;
+  auto s = db->Get(leveldb::ReadOptions(), std::forward<ArgT>(arg), &value);
+  if (!s.ok()) {
+    if (!s.IsNotFound()) {
+      LOG_GENERAL(WARNING, "LevelDB " << dbName << " status is not OK - "
+                                      << s.ToString());
+    }
```
JamesHinshelwood, line 45: I'm not sure this makes sense. Would `s.IsNotFound()` be `true` if the DB exists, but the key (`arg`) doesn't exist in it?

yaron-zilliqa: It would; this is backward compatible with the current implementation. Haven't changed anything there.

src/depends/libDatabase/LevelDB.cpp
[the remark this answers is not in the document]
yaron-zilliqa: The methods haven't been reordered. The only change in logic here is checking for a nullptr before accessing it.
The previous methods were really a copy of one another with a different construction of leveldb::Slice, so the not simply invoke XXXXXImpl(...) instead of repeating the code.
The problem it solves is that the process won't crash (and then redownload persistence upon restart) if the required LevelDB object was NULL. I came across this scenario this time round when setting ARCHIVAL_LOOKUP_WITH_TX_TRACES to true and calling ots_getTransactionError.

PR #3807, opened 2023-10-10
[master] (fix) ZIL-5407 Specifying --isolated-server-accounts in localdev shou…

rrw-zilliqa:
…ld not stop it including isolated server accounts.

## Description

Don't have counterintuitive behaviour in localdev (also, this makes z blockchain fail to include these accounts in its builds properly)

## Backward Compatibility

## Review Suggestion

Read code and test.

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


scripts/localdev.py.orig
JamesHinshelwood, line 1: Why does this exist?

scripts/localdev.py
```diff
@@ -502,8 +502,9 @@ def tempo_down(config):
               default='localdev',
               show_default=True,
               help="The test network's name")
-@click.option("--isolated-server-accounts",
-              is_flag=True,
+# Need to specify like this, or passing --isolated-server-accounts will cause isolated server accounts to be disabled
+# - rrw 2023-10-10
+@click.option("--isolated-server-accounts/--no-isolated-server-accounts",
```
yaron-zilliqa, line 507: Not sure I quite understand this change.

scripts/localdev.py.orig
rrw-zilliqa: It shouldn't - sorry! Will zap.

scripts/localdev.py
rrw-zilliqa: If you set default=True and then pass '--isolated-server-accounts', the effect is to turn isolated server accounts off. If you make the option explicit like this , that behaviour doesn't occur. I'll update the comment ..

JamesHinshelwood (dismissed):

PR #3811, opened 2023-10-13
[master] Remove --skip-non-guard-ds config for desharded mode

chetan-zilliqa:
## Description

`--skip-non-guard-ds` was there for some legacy reasons and we do not need the flag in current desharded mode.
Also update some config to make network bit decentralised with both mining nodes and DS guard nodes.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (dismissed):
Looks good to me :-)

rrw-zilliqa (approved):
Still looks good to me :-)

PR #3822, opened 2023-10-16
[master] ZIL-5354 sendTransaction doesn't work - response handled differently in ethers 6, looking for 0x0 in data not empty string.

Steve-White-UK:
ZIL-5354 blank response in data not handled by ethers 6, switched to 0x0

## Description

Fix bug outlined in ZIL-5354 

## Backward Compatibility

## Status

In testing, tests passed locally

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test

### DIFF
--- a/src/libServer/JSONConversion.cpp
+++ b/src/libServer/JSONConversion.cpp
@@ -698,7 +698,13 @@ const Json::Value JSONConversion::convertTxtoEthJson(
    } else {
      inputField += callData;
    }
+ } else {
+   // Append extra '0x' prefix iff GetCode() gave empty string
+   if (inputField.empty()) {
+     inputField += "0x";
+   }
  }
+
  retJson["input"] = inputField;
  // ethers also expects 'data' field
  retJson["data"] = inputField;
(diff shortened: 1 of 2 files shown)


its-saeed (approved):

src/libServer/JSONConversion.cpp
```diff
@@ -698,7 +698,13 @@ const Json::Value JSONConversion::convertTxtoEthJson(
    } else {
      inputField += callData;
    }
+ } else {
+   // Append extra '0x' prefix iff GetCode() gave empty string
+   if (inputField.empty()) {
+     inputField += "0x";
```
JamesHinshelwood, line 704: Would it be neater to write `std::string inputField = "0x";` on line 686, then remove all the extra `0x`es below?

k5ssv: OK, I saved another variable too, I have been causing problems by making things go too fast, but it causes a smile at least.

its-saeed (approved):

src/libServer/JSONConversion.cpp
JamesHinshelwood: I meant that if we set `inputField` to `0x` when it is first defined above, then we can remove all the conditional adding of `0x`es later.

PR #3827, opened 2023-10-17
[master] ZIL-5376: Support type 1 and type 2 Ethereum transactions

JamesHinshelwood:
A brief description of how these transaction types differ and the potentially controversial choices I made while supporting them...

Type 1 transactions are defined in EIP-2930 and allow access lists to be provided with transactions. In Zilliqa, access lists are completely ignored and do not have any affect on gas accounting, but we do keep track of them so the transaction can be reconstructed later.

Type 2 transactions are defined in EIP-1559 and accept access lists, plus they allow for a different model for gas accounting. Instead of setting a single `gasPrice` for a transaction, you set a 'priority fee' which is paid in addition to the block's base fee. A maximum fee is also sent in transactions which limits the total fee the user is willing to pay (base fee + priority fee). In Zilliqa, we ignore the priority fee and just use the maximum fee as the entire fee for the transaction.

Its also worth noting that I removed the ability to configurability of transaction versions and just hardcoded the values that we use everywhere. We were already hardcoding these in many places in the source code, so there was only one set of configured values which would actually result in a working network anyway.

### DIFF
--- a/src/libEth/Eth.cpp
+++ b/src/libEth/Eth.cpp
@@ -73,28 +73,42 @@ Json::Value populateReceiptHelper(
   ret["r"] = GetR(sig);
   ret["s"] = GetS(sig);
 
+  switch (tx.GetVersionIdentifier()) {
+    case TRANSACTION_VERSION:
+      ret["type"] = "0x0"; // FIXME?
+      break;
+    case TRANSACTION_VERSION_ETH_LEGACY:
+      ret["type"] = "0x0";
+      break;
+    case TRANSACTION_VERSION_ETH_EIP_2930:
+      ret["type"] = "0x1";
+      break;
+    case TRANSACTION_VERSION_ETH_EIP_1559:
+      ret["type"] = "0x2";
+      break;
+    default:
+      LOG_GENERAL(WARNING, "Invalid transaction type");
+      break;
+  }
+
   return ret;
 }
 
-// Given a RLP message, parse out the fields and return a EthFields object
-EthFields parseRawTxFields(std::string const &message) {
+EthFields parseLegacyTransaction(zbytes asBytes) {
   EthFields ret;
-
-  zbytes asBytes;
-  DataConversion::HexStrToUint8Vec(message, asBytes);
 
   dev::RLP rlpStream1(asBytes,
                       dev::RLP::FailIfTooBig | dev::RLP::FailIfTooSmall);
 
   if (rlpStream1.isNull()) {
-    LOG_GENERAL(WARNING, "Failed to parse RLP stream in raw TX! " << message);
+    LOG_GENERAL(WARNING, "Failed to parse RLP stream in raw TX!");
     return {};
   }
 
   int i = 0;
   // todo: checks on size of rlp stream etc.
 
-  ret.version = DataConversion::Pack(CHAIN_ID, 2);
+  ret.version = DataConversion::Pack(CHAIN_ID, TRANSACTION_VERSION_ETH_LEGACY);
 
   // RLP TX contains: nonce, gasPrice, gasLimit, to, value, data, v,r,s
   for (auto it = rlpStream1.begin(); it != rlpStream1.end();) {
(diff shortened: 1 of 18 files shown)


rrw-zilliqa (approved):
Looks good to me, but maybe some comment fixes would be good (perhaps in another PR?)

src/libEth/Eth.cpp
```diff
@@ -73,28 +73,42 @@ Json::Value populateReceiptHelper(
   ret["r"] = GetR(sig);
   ret["s"] = GetS(sig);
 
+  switch (tx.GetVersionIdentifier()) {
+    case TRANSACTION_VERSION:
+      ret["type"] = "0x0"; // FIXME?
```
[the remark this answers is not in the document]
JamesHinshelwood: I wasn't sure if it made sense to return `"type": "0x0"` for Zilliqa transactions when you call `eth_getTransactionByHash` on them. It is the current behaviour, but I figured it might make sense to use the type to differentiate non-Eth transactions too?

We'd want to pick a value which is sufficiently unlikely to ever be used by an Ethereum transaction type.

Do you think this is worth changing now or should I just stick with the existing behaviour of returning `"type": "0x0"` for both Ethereum legacy transactions and Zilliqa transactions?

src/libEth/Eth.cpp
```diff
@@ -145,6 +159,178 @@ EthFields parseRawTxFields(std::string const &message) {
...
+        ret.signature.insert(ret.signature.end(), b.begin(), b.end());
+      } break;
+      case 10:  // signatureS
+      {
+        zbytes b = dev::toBigEndian(dev::u256(*it));
+        ret.signature.insert(ret.signature.end(), b.begin(), b.end());
+      } break;
+      default:
+        LOG_GENERAL(WARNING, "too many fields received in rlp!");
+    }
```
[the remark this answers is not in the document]
JamesHinshelwood: Not really - We should fail with an error if the RLP doesn't contain exactly the right number of items. However the only way I could find to propagate errors in this bit of code was to log a warning and just return an empty byte array :sob: 

Is this what you meant?

src/libEth/Eth.cpp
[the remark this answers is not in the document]
JamesHinshelwood: OTOH maybe if we return a `type` outside of `{0,1,2}`, clients like Ethers will just explode when trying to parse such transactions...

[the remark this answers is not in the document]
JamesHinshelwood: I had a look and other Ethereum-compatible chains are using other `type`s to represent their own things (e.g. [Optimism](https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#the-deposited-transaction-type) uses a `type` of `0x7E` for its own transactions).

JamesHinshelwood:
Something that just crossed my mind - We probably should charge gas at least for the extra access list in transaction requests to prevent people sending stupidly long access lists without any cost.

JamesHinshelwood:
Thanks for spotting @bzawisto, I'll update the transaction roots too.

src/libEth/Eth.cpp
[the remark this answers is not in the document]
rrw-zilliqa: Those ^^^ are good things to write in a comment, so when someone sees // FIXME, they don't have to search github to work out what you mean :-)

src/libEth/Eth.cpp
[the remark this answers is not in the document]
rrw-zilliqa: Works for me :-) - just checking it was intentional that you break out of the loop early if there aren't enough RLP items.

src/libEth/Eth.cpp
```diff
@@ -73,28 +73,42 @@ Json::Value populateReceiptHelper(
...
+      LOG_GENERAL(WARNING, "Invalid transaction type");
+      break;
+  }
+
   return ret;
 }
 
-// Given a RLP message, parse out the fields and return a EthFields object
-EthFields parseRawTxFields(std::string const &message) {
+EthFields parseLegacyTransaction(zbytes asBytes) {
```
yaron-zilliqa, line 97: Why is this being copied? unless you're planning to move this below (into RLP), better to pass a const reference.

src/libEth/Eth.cpp
```diff
@@ -145,6 +159,178 @@ EthFields parseRawTxFields(std::string const &message) {
   return ret;
 }
 
+EthFields parseEip2930Transaction(zbytes asBytes) {
```
yaron-zilliqa, line 162: As above.

src/libEth/Eth.cpp
```diff
@@ -145,6 +159,178 @@ EthFields parseRawTxFields(std::string const &message) {
...
+    return {};
+  }
+
+  int i = 0;
+
+  ret.version = DataConversion::Pack(CHAIN_ID, TRANSACTION_VERSION_ETH_EIP_2930);
+
+  // RLP TX contains: chainId, nonce, gasPrice, gasLimit, to, value, data, accessList, signatureYParity, signatureR, signatureS
+  for (auto it = rlpStream1.begin(); it != rlpStream1.end();) {
+    auto byteIt = (*it).operator zbytes();
```
yaron-zilliqa, line 179: Why is this here? looks like it's only used in case 6; You can just use:
`ret.code = static_cast<...>(*it)` there which will invoke the operator.

In addition, why not just have: `for (const auto& value = rlpStream1) { ... }`? You don't seem to be manipulating the iterator other than incrementing it.

src/libMessage/Messenger.cpp
```diff
@@ -802,6 +819,22 @@ bool ProtobufToTransactionCoreInfo(
     copy(protoTxnCoreInfo.data().begin(), protoTxnCoreInfo.data().end(),
          txnCoreInfo.data.begin());
   }
+  txnCoreInfo.accessList.reserve(protoTxnCoreInfo.accesslist_size());
+  for (const auto& item : protoTxnCoreInfo.accesslist()) {
+    dev::h160 address = dev::h160(item.address(), dev::h160::ConstructFromStringType::FromBinary);
```
yaron-zilliqa, line 824: ` = dev::h160` isn't required. Can just initialize directly.

yaron-zilliqa (changes requested):

src/libEth/Eth.cpp
JamesHinshelwood: Copy pasted from the other parsing code. I will simplify :+1:

src/libEth/Eth.cpp
```diff
@@ -145,6 +160,176 @@ EthFields parseRawTxFields(std::string const &message) {
...
+    return {};
+  }
+
+  int i = 0;
+
+  ret.version = DataConversion::Pack(CHAIN_ID, TRANSACTION_VERSION_ETH_EIP_2930);
+
+  // RLP TX contains: chainId, nonce, gasPrice, gasLimit, to, value, data, accessList, signatureYParity, signatureR, signatureS
+  for (const auto& it : rlpStream1) {
+    auto byteIt = it.operator zbytes();
```
yaron-zilliqa, line 180: again, why are you using this syntax?

src/libMessage/Messenger.cpp
```diff
@@ -802,6 +819,22 @@ bool ProtobufToTransactionCoreInfo(
...
+  for (const auto& item : protoTxnCoreInfo.accesslist()) {
+    dev::h160 address(item.address(), dev::h160::ConstructFromStringType::FromBinary);
+    std::vector<dev::h256> storageKeys;
+    storageKeys.reserve(item.storagekeys_size());
+    for (const auto& key : item.storagekeys()) {
+      dev::h256 storageKey(key, dev::h256::ConstructFromStringType::FromBinary);
+      storageKeys.push_back(storageKey);
+    }
+    auto accessListItem = std::make_pair(address, storageKeys);
+    txnCoreInfo.accessList.push_back(accessListItem);
```
yaron-zilliqa, line 832: std::move(accessListItem) ?

yaron-zilliqa (dismissed):

src/libEth/Eth.cpp
```diff
@@ -145,6 +160,172 @@ EthFields parseRawTxFields(std::string const &message) {
...
+        ret.maxPriorityFeePerGas = uint128_t{it};
+        break;
+      case 3:
+        ret.maxFeePerGas = uint128_t{it};
+        break;
+      case 4:
+        ret.gasLimit = uint64_t{it};
+        break;
+      case 5:
+        ret.toAddr = zbytes(it);;
```
yaron-zilliqa, line 262: Two ;;
Can you please check & confirm that this will indeed invoke the operator() ? I'm not entirely sure in this case. You can instead try static_cast<zbytes>(it) if it doesn't.

src/libEth/Eth.cpp
```diff
@@ -145,6 +160,172 @@ EthFields parseRawTxFields(std::string const &message) {
...
+        break;
+      case 6:
+        ret.amount = uint128_t(it);
+        break;
+      case 7:
+        ret.code = zbytes(it);
```
yaron-zilliqa, line 268: As above.

src/libEth/Eth.cpp
```diff
@@ -145,6 +160,172 @@ EthFields parseRawTxFields(std::string const &message) {
...
+        ret.nonce = uint32_t(it);
+        break;
+      case 2:
+        ret.gasPrice = uint128_t{it};
+        break;
+      case 3:
+        ret.gasLimit = uint64_t{it};
+        break;
+      case 4:
+        ret.toAddr = zbytes(it);
```
yaron-zilliqa, line 193: See comment about operator() below.

src/libEth/Eth.cpp
```diff
@@ -145,6 +160,172 @@ EthFields parseRawTxFields(std::string const &message) {
...
+        break;
+      case 5:
+        ret.amount = uint128_t(it);
+        break;
+      case 6:
+        ret.code = zbytes(it);
```
yaron-zilliqa, line 199: See comment about operator() below.

yaron-zilliqa (approved):

src/libEth/Eth.cpp
JamesHinshelwood: Sorry, I don't understand what you mean by "confirm that this will indeed invoke the operator()"? (I'm probably missing some C++ knowledge here)

It is definitely reading the RLP bytes and converting them to a  `zbytes` though, because the tests still parse. If that wasn't happening we would get signature verification failures.

yaron-zilliqa: The old code invoked the conversion operator() explicitly, so to make it equivalent I suggested you check that this change does the same instead of invoking the constructor and then copying or converting.
If you're saying that this is doing what it's supposed to do then I'd say it's fine.

yaron-zilliqa (approved):

PR #3833, opened 2023-10-20
[master] BugFix ZIL-5424 : Fix for mismatch in txn pool data

chetan-zilliqa:
## Description

The public devnet was stalled many times when we fire some transactions.
The reason for this is the consensus issue with txn pool data.

The consensus stalls because every node has different view of transactions and eventually that couldn’t reach consensus.

Root cause : There are some timing delays in nodes when they finish consensus for a round. A faulty backup was clearing the txns pool of new transaction thinking that this transactions belong to old epoch. The lookup dispatches the txns when it receives last final blocks. However, one of the backup still in the  DONE  state (some few milliseconds gap) and it receives the FORWARD_TRANSACTION message. At the end of consensus we clear the txn pool. So it clear new set of transactions.

Fix : Introduced sufficient delays(while maintaining blocktime) before transactions are send to nodes.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)

Fixing the root cause would be better, but little chance of that at this point ..

PR #3836, opened 2023-10-27
Zil 5450 improvements to network and upgrade of tests for native code test node test zilliqa

Steve-White-UK:
## Description
Upgrade of the test components to work using the native build of Zilliqa

a new test directory native exists which contains scripts to start zilliqa in native mode according to the definition of these scripts.

This environment is provided using the test directories 

Node
Zilliqa
native

This PR also contains some small improvements made to the network layer, to gracefully close sockets and some better error messages indicating which ports are in use when a socket connection fails.

These tests currently work from unaltered constants_local,XML in the Zilliqa trunk.

scripts are as documented on the Zilliqa Read.me

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


Steve-White-UK (commented):
Remember to review UnixDomainSocket class, this one kinda slipped through without notice.

PR #3848, opened 2023-10-30
[master] ZIL-5406: add support for a cdn

yaron-zilliqa:
## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


mauromedda (approved):

scripts/localdev.py
```diff
@@ -731,14 +734,19 @@ def stop_testnet(config, testnet_name):
...
 
     if os.path.exists(instance_dir):
         print(f"Removing old testnet configuration ..")
         shutil.rmtree(instance_dir)
     print(f"Generating testnet configuration .. ")
+
+    # IMPORTANT: if you change any of the flags make sure to update the count returned
+    #            by this function so that prometheus_up knows how many pods to wait for.
+    # 20 normal pods (i.e. 5 ds + 15 non-ds), 1 lookup, 2 seedpubs, 1 seedprv.
+    count = 20 + 1 + 2 + 1
```
rrw-zilliqa, line 749: This seems dodgy - what happens if this count isn't right? Will the user get a reasonable error message, or will they have to go hunting for why everything's bust .. ?

rrw-zilliqa (dismissed):
not sure the native stuff is all that useful, but it doesn't look like it does any harm.
The rest seems reasonable. Looks good to me.

scripts/localdev.py
yaron-zilliqa: Accidental.

yaron-zilliqa:
> not sure the native stuff is all that useful, but it doesn't look like it does any harm. The rest seems reasonable. Looks good to me.

That wasn't meant to be committed. I've reverted it.

mauromedda (approved):

PR #3871, opened 2023-11-10
[master] US-322: ots_hasCode() should take an integer as its second argument.

rrw-zilliqa:
## Description

Fix ots_hasCode so that otterscan doesn't choke when calling it.

## Backward Compatibility

- [] This is not a breaking change

## Review Suggestion
Check that otterscan now works with ZQ1 (in particular, txn details and logs)

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


JamesHinshelwood (commented):
I think this can in fact take either a string or a number :(
https://github.com/Zilliqa/otterscan/blob/develop/src/useErigonHooks.ts#L690
(`type BlockTag = BigNumberish | string`)

JamesHinshelwood (approved):
Approving in case the frontend doesn't actually send a string in practice? (though on my brief glance it looks like it might, since the default parameter is a string)

PR #3876, opened 2023-11-23
[master] Send txns from lookups instead of seeds when final block arrives

bzawisto:
## Description
We now forward from seeds to lookups and - when new block comes - send the entire mempool from Lookup to ds nodes.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)

PR #3884, opened 2023-11-29
[release/v9.3] Check null condition before accessing dsblock consensus object

chetan-zilliqa:
## Description

A newly joined node receives an invalid gossip message for `DS_FINALBLOCKCONSENSUS`. As the newly joined node wasn't part of the ds block consensus, it has no consensus object created. But because of timing issues, it receives late gossip messages for DS block consensus which we should ignore as the consensus object is not initialized.
If the consensus object is created by other message, the state check would automatically handle the message or reject it.

## Backward Compatibility

### Implementation

- [x] **ready for review**

- [ ] local machine test

- [x] small-scale cloud test


chetan-zilliqa:
Created https://github.com/Zilliqa/Zilliqa/pull/3885 against master

PR #3888, opened 2023-11-30
[master] After view-change fixes

bzawisto:
## Description
Some backup nodes were keeping microblocks with different gas limits which are no longer valid for re-run of consensus for the same epoch.

## Backward Compatibility

### Implementation

- [ ] **ready for review**

- [ ] local machine test

- [ ] small-scale cloud test


rrw-zilliqa (approved):
Looks good to me :-)

